Issue #280, opened 2024-01-29
beagle.conf should be removed.  chan_beagle has not been ported to ASL3


KB4MDD:
The configuration file beagle.conf should be removed from configs/rpt.  This configuration file is associated with chan_beagle which is not included in ASL3.

okcrum:
Which beagleboard is it that we are supposed to support?

https://www.beagleboard.org/boards

Allan-N:
FWIW, what I'd like to do here (and it's kinda on my TODO list) is to commit whatever changes we have, commented out as necessary, and merge them into master.  Then, remove all of the BeagleBone support.  The idea here is to clean out the cruft ... but still be able to look into the repo history for whatever support we had.  Oh, and I don't want to have to worry about a WIP PR/branch being deleted.  That's why I'm thinking that we should get the changes committed to "master"

okcrum:
There's no chan_beagle.c ATM, so remove the /etc/asterisk/beagle.conf file, modify modules.conf, and comment out what's left in the code?

Allan-N:
>There's no chan_beagle.c ATM

See https://github.com/AllStarLink/app_rpt/pull/319 :)

okcrum:
Should have written "no current chan_beagle.c in the current master branch".

okcrum:
Reopened to complete archiving chan_beagle

okcrum:
Marking this closed; the aim of the exercise was to merge the chan_beagle code cleanly into master, then remove it so there would be a "safety" copy should the branch issue-280 get deleted. This is now complete.

PR #319, opened 2024-04-29
chan_beagle: Initial port to compile on current Asterisk.

InterLinked1:
There are still a few unresolved references to SND_PCCV stuff on my machine, but otherwise, this updates the code for current Asterisk APIs. The sample config
has been moved out of the module itself into its own sample file.

Resolves: #280


okcrum:
Reason for this commit is to preserve this code in the master branch, so it won't get lost in a deleted branch.

Allan-N (approved):

Allan-N:
>The clang-format file isn't new, is it? I wonder if this branch needs a rebase

@okcrum did a rebase the branch.  Not sure why the clang-format file didn't make it.

PR #346, opened 2024-07-04
chan_simpleusb/chan_usbradio: Use monotonic clock

### LINKED ISSUE #344: chan_simpleusb: Replace time() with clock_gettime()
Based on https://github.com/AllStarLink/ASL3/issues/76, the boot-time clock skew seems to cause chan_simpleusb to lock the PTT open until a restart of asterisk. Delaying asterisk start on boot until NTP has synchronized seems to resolve the issue but that is not a reliable fix. Within the code, comparisons are being made using `time(2)` when the likely resolution is to use `clock_gettime(2)` with a `CLOCK_MONOTONIC` type. That way, "wallclock" time changes in the OS don't affect the USB channel driver.

KB4MDD:
This changes chan_simpleusb and chan_usbradio to use the monotonic clock.  This will address timing issues when NTP updates the internal clock.

This closes #344

### DIFF
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -397,6 +397,38 @@ static struct ast_channel_tech simpleusb_tech = {
 };
 
 /*!
+ * \brief Get system monotonic 
+ * This returns the CLOCK_MONOTONIC time
+ * \param second	Pointer to time_t to receive the time.
+ */
+static void rpt_time(time_t *second)
+{
+	struct timespec ts;
+	
+	clock_gettime(CLOCK_MONOTONIC, &ts);
+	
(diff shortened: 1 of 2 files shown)


Allan-N (approved):
As an alternative to adding rpt_time() and rpt_tvnow() to both chan_simpleusb.c and chan_usbradio.c "I" might have placed both of these functions along side of rpt_starttime().  

But, as-is, the changes look good.

tsawyer:
@KB4MDD Naveen mentioned that this change could have been made to res_usbradio. He didn't think it needs to be changed right away. Perhaps you could to another PR later on this.

PR #355, opened 2024-07-11
Fix issue where bad registration info leads to NULL dereference crash

Allan-N:
When you start off with a "rpt_http_registrations.conf" containing an error and attempt to register then you may get back an unexpected response.  If you don't check for the unexpected then you can end up with a crash.  This change adds a missing check for the unexpected (lack of "data").


KB4MDD (approved):
This looks good.

PR #356, opened 2024-07-12
chan_simpleusb / chan_usbradio: Modify cos operation

### LINKED ISSUE #329: Simpleusb-tune-menu COS Output Logic
From the simple-tune-menu when both COR Input and CTCSS Input are set to Off COS Output is Keyed. If an input is off it should have no role determining COS Output which should be Clear by default. A USB device must be connected to duplicate this issue. Both a URI and a RA42M exhibit this behavior. 

```
Please enter your selection now: v
Receiver/Transmitter Status Display:
  COS   | CTCSS  | COS   | PTT
 Input  | Input  | Out   | Out
 Off    | Off    | Keyed | Keyed
```

KB4MDD:
chan_simpleusb and chan_usbradio should not activate the receiver 
when the carrierfrom and ctcssfrom are set to no.

This closes #329.

### DIFF
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -2260,6 +2260,13 @@ static struct ast_frame *simpleusb_read(struct ast_channel *c)
 		sd = 1;
 	}
 	o->rx_ctcss_active = sd;
+	
+	/* Special case where cd and sd have been configured for no */
+	if (o->rxcdtype == CD_IGNORE && o->rxsdtype == SD_IGNORE) {
+		cd = 0;
+		sd = 0;
+	}
+	
 	/* Timer for how long TX has been unkeyed - used with txoffdelay */
 	if (o->txoffdelay) {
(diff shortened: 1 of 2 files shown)


Allan-N (approved):
OK w/me.

channels/chan_simpleusb.c
```diff
@@ -2260,6 +2260,13 @@ static struct ast_frame *simpleusb_read(struct ast_channel *c)
 		sd = 1;
 	}
 	o->rx_ctcss_active = sd;
```
mkmer, line 2262: Shouldn't this assignment follow the new code?

KB4MDD: I left that so that the live display would show the actual signal being received.  We may have a cos activation and that would be displayed, however, if the options are set to 'no', the ptt would not be keyed.

Setting both options to 'no' is a real unique situation.  It would be a case where the device can only transmit.

PR #357, opened 2024-07-12
Update configs for missing options

KB4MDD:
This updates simpleusb and usbradio config files for missing parallel port options.

### DIFF
--- a/configs/rpt/simpleusb.conf
+++ b/configs/rpt/simpleusb.conf
@@ -67,11 +67,15 @@ carrierfrom = usbinvert             ; no,usb,usbinvert
                                     ; no - no carrier detection at all
                                     ; usb - from the COR line on the USB sound fob (Active high)
                                     ; usbinvert - from the inverted COR line on the USB sound fob (Active low)
+                                    ; pp - from parallel port (Active high)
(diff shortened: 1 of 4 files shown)


tsawyer (approved):

tsawyer:
Have we sorted out how to get config file changes out to users?

Allan-N:
> Have we sorted out how to get config file changes out to users?

AFAIK, we have not yet figured this out.  @jxmx, any thoughts?

Allan-N (approved):
The (doc) changes are fine.

jxmx:
I think we need to have a larger, strategic discussion on what problem(s) we're trying to architect for when we're talking about "getting config file changes out to users". For this fix, I think it should just be merged and appear in the next release when needed. I would assume that the user base of needed an actual parallel port in 2024 is vanishingly small AND that user base that is nursing equipment along that still has one should be able to edit a text file. 

In general, we should likely talk about breaking up some of these monolithic configuration files and make good use of the include ability inside the configuration files. And since Asterisk seems to properly obey "last directive wins" we could create a set of includes that are marked "DO NOT EDIT THIS" and then forcefully blow them away on upgrades with a postinst script when we need to.

InterLinked1 (approved):

PR #370, opened 2024-07-24
rpt.conf: Update for default events on GPIO4

KB4MDD:
This updates rpt.conf to enable the events section and add commented out defaults for GPIO4.

This was requested by KE6PCV, however, I find this useful also.  I use sound card devices from two manufactures that support the GPIO4 LED.

### DIFF
--- a/configs/rpt/rpt.conf
+++ b/configs/rpt/rpt.conf
@@ -134,7 +134,7 @@ linkunkeyct = ct8                   ; sent when a transmission received over the
                                     ; 1. node number in this stanza (us)
                                     ; 2. node number being disconnected from us (them)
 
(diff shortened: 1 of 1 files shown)


tsawyer (commented):
We have to find a better way to implement this rather than changing the default configs. This needs to be a menu item, perhaps part of the hotspot "shortcut".

tsawyer:
@Allan-N, please note about adding a menu shortcut.

KB4MDD:
Is there a reason that it cannot go out with new installs?  

I understand the need for menu options, but there are those that don't use the menu and having it in the file by default is great.  

I believe that the installer does not overwrite the settings in /etc/asterisk.

KB4MDD:
If this is the case, we can never update the default configs to add new features or to correct comments in the config files.

PR #372, opened 2024-07-26
app_rpt: Correct 5 digit node lookup from 4 digit node

KB4MDD:
This updates app_rpt to address a problem with looking up a 5 or 6 digit node from a 4 digit node. The 4 digit node could not lookup 5 or 6 digit nodes when it did not use an external nodes file. This would happen if you are not using the node updater and depending on DNS only.

The problem was associated with the internal 'longestnode' variable. We did not have an external file to determine the longest node number. In this case, the longest node was set to 4.

A new rpt.conf [general] item, max_dns_node_length, has been added with a default value of 6. The routines that determine the longest nodes will now MAX the value it determines with rpt_max_dns_node_length. The configurable item cannot be less than 4 or more then 63.

By default, this configurable item will not be in rpt_conf. It can be used in the future if we need to support longer node numbers.

Additional level 4 debugging was added to the DNS lookup routine.

Without this change, the problem can be remediated with existing installs, that don't use the node updater, by adding the following to rpt.conf [nodes].

999999 = radio@127.0.0.1:4569/999999,NONE

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -424,6 +424,7 @@ static int nrpts = 0;
 
 /* general settings */
 enum rpt_dns_method rpt_node_lookup_method = DEFAULT_NODE_LOOKUP_METHOD;
+int rpt_maximum_node_length = 6;
 
 int max_chan_stat[] = { 22000, 1000, 22000, 100, 22000, 2000, 22000 };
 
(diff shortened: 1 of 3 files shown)


Allan-N (approved):
Looks good.

PR #373, opened 2024-07-26
Increase HTTP registration timeout

Allan-N:
The current HTTP registration timeout (5s) has proven to be too slow for some nodes.  Bump the timeout to 15s.


KB4MDD (approved):
This looks good.

PR #379, opened 2024-08-01
Update the permissions of the "archivedir" .WAV files.

Allan-N:
When using "archivedir = /var/spool/asterisk/monitor" we create files in the [spool] directory. We are setting the filesystem permissions on the ".txt" files to be 0644.  We should do the same for the ".WAV" files.


KB4MDD (approved):
Looks good.

InterLinked1 (approved):

PR #394, opened 2024-08-28
Support URIx eeprom Serial Numbers

coolacid:
When using a URIx with an eeprom, the eeprom supports setting a serial number for the device. The serial number is exposed as in the device descriptor as iSerialNumber. 

This PR allows the use of a serial number to identify which device to load in chan_usbradio. 

On channel hidthread startup, the PR will loop through the AST radio device strings looking for a matching configured serial number. If a match is found, the devstr is updated to match the one found in the search loop. 

Should a usb device match not be found, chan_usbradio continues as normal by using the configured devstr, and finally searching for any available device. 

At this time, I only have very terrible python code to modify device serial numbers, and it is not clean enough to include in this PR. 

I've been using a version of this for the old apt_rpt for a while, and it's worked great. This is an attempt to port to the new V3 version.

```
[2024-08-28 14:37:10.221] NOTICE[11380] chan_usbradio.c: Checking for USB device with serial 1000
[2024-08-28 14:37:10.310] NOTICE[11380] chan_usbradio.c: Device Serial 1234 vs 1000
[2024-08-28 14:37:10.398] NOTICE[11380] chan_usbradio.c: Device Serial 1000 vs 1000
[2024-08-28 14:37:10.398] NOTICE[11380] chan_usbradio.c: Found device serial 1000 at 2-1:1.0 for 1999

```

### DIFF
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -677,7 +677,7 @@ static int get_usb_serial(char *devstr, char *serial)
     struct usb_dev_handle *usb_handle;
     int length;
 
-    usb_dev = hid_device_init(devstr);
+    usb_dev = ast_radio_hid_device_init(devstr);
     usb_handle = usb_open(usb_dev);
     length = usb_get_string_simple(usb_handle, usb_dev->descriptor.iSerialNumber, serial, sizeof(serial));
     usb_close(usb_handle);
@@ -890,18 +890,25 @@ static void *hidthread(void *arg)
 		if (strlen(o->serial) > 0)
 		{
 			ast_log(LOG_NOTICE, "Checking for USB device with serial %s\n", o->serial);
-			for(s = usb_device_list; *s; s += strlen(s) + 1)
+			int index = 0;
+			char *index_devstr = NULL;
+			for(;;)
 			{
+				index_devstr = ast_radio_usb_get_devstr(index);
+				if (ast_strlen_zero(index_devstr)) {
+					/* No more devices, so break out of the loop */
+					break;
+				}
 				/* Go through the list of usb devices, and get the serial numbers */
-				if (get_usb_serial(s, serial) == 0) continue;
+				if (get_usb_serial(index_devstr, serial) == 0) continue;
 				ast_log(LOG_NOTICE, "Device Serial %s vs %s\n", o->serial, serial);
 				if (strcmp(o->serial, serial) == 0)
 				{
 					/* We found a device with the matching serial number
 					 * Set the devstr to the matching device
 					 */
-					ast_log(LOG_NOTICE, "Found device serial %s at %s for %s\n",o->serial, s, o->name);
-					ast_copy_strin(o->devstr, s, sizeof(o->devstr));
+					ast_log(LOG_NOTICE, "Found device serial %s at %s for %s\n",o->serial, index_devstr, o->name);
+					ast_copy_string(o->devstr, index_devstr, sizeof(o->devstr));
 					break;
 				}
 			}
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -897,6 +897,7 @@ static void *hidthread(void *arg)
 				index_devstr = ast_radio_usb_get_devstr(index);
 				if (ast_strlen_zero(index_devstr)) {
 					/* No more devices, so break out of the loop */
+					ast_log(LOG_NOTICE, "USB Device with serial %s not found.\n", o->serial);
 					break;
 				}
 				/* Go through the list of usb devices, and get the serial numbers */
@@ -911,6 +912,7 @@ static void *hidthread(void *arg)
 					ast_copy_string(o->devstr, index_devstr, sizeof(o->devstr));
 					break;
 				}
+				index++;
 			}
 		}
 		/* Automatically assign a devstr if one was not specified in the configuration. */
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -202,6 +202,7 @@ struct chan_usbradio_pvt {
 	int boost;					/* input boost, scaled by BOOST_SCALE */
 	char devicenum;
 	char devstr[128];
+	char serial[14];
 	int spkrmax;
 	int micmax;
 	int micplaymax;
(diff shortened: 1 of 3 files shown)


KB4MDD:
This support will also need to be added to chan_simpleusb.  We are attempting to keep the same features in both drivers.  

Common routines, like get_usb_serial, will need to be in res_usbradio.

Please review https://docs.asterisk.org/Historical-Documentation/Development/Historical-Policies-and-Procedures/Code-Review/Coding-Guidelines/

We will run through a code review soon.

coolacid:
Thanks for the feedback. I'll see about moving the get_usb_serial to the resource. I haven't looked at simple_usb code at all, so probably need a little direction on that.

KB4MDD:
You will find the routines that you changed are almost the same in simple_usb

coolacid:
Thanks, a quick look seemed to say that. I'll work on implementing and testing in simple_usb a little later.

coolacid:
Have to put this down for a bit. SimpleUSB works except for one bug. It's suppose to collect the serial number when an auto-assign happens. With adding some debug I see it enter my if statement just fine, but `simpleusb-tune-menu` won't show the serial number. If I set the serial number in the config, it all works as expected.

res/res_usbradio.c
```diff
@@ -719,6 +719,33 @@ void ast_radio_time(time_t *second)
...
+ * \brief Get serial number from device if available
+ *	This function will attempt to get the serial number from a media device
+ *
+ * \param devstr	The USB device string
+ * \param serial	Buffer to return the serial number
+ *
+ * \retval		Length of found serial number
+ */
+
+int ast_radio_get_usb_serial(char *devstr, char *serial)
```
Allan-N, line 732: This function really needs to have the size of the "char *serial" buffer passed in as an argument.

res/res_usbradio.c
```diff
@@ -719,6 +719,33 @@ void ast_radio_time(time_t *second)
...
+{
+    struct usb_device *usb_dev;
+    struct usb_dev_handle *usb_handle;
+    int length;
+
+    usb_dev = ast_radio_hid_device_init(devstr);
+    usb_handle = usb_open(usb_dev);
+    length = usb_get_string_simple(usb_handle, usb_dev->descriptor.iSerialNumber, serial, sizeof(serial));
```
Allan-N, line 740: sizeof(serial) is NOT what you want here.  You want the caller to pass in the size of the provided buffer and pass that along to usb_get_string_simple()

channels/chan_simpleusb.c
```diff
@@ -347,6 +348,7 @@ static struct chan_simpleusb_pvt simpleusb_default = {
 	.rxondelay = 0,
 	.txoffdelay = 0,
 	.pager = PAGER_NONE,
+	.serial = "",
```
Allan-N, line 351: I'm not sure about this since the variable is "char serial[14]".  Also, with the default struct being zero'd out then is this needed (kinda the same way that .devstr is not being initialized)?

channels/chan_simpleusb.c
```diff
@@ -906,6 +908,35 @@ static void *hidthread(void *arg)
 		 * found device.
 		 */
 		ast_radio_time(&o->lasthidtime);
+		/* If configuration has a serial number defined, find the device */
+		if (strlen(o->serial) > 0)
+		{
+			ast_log(LOG_NOTICE, "Checking for USB device with serial %s\n", o->serial);
```
Allan-N, line 914: This ast_log() ... and the others below ... really feel like debug messages.  Maybe switch them to ast_debug() ?

channels/chan_simpleusb.c
```diff
@@ -906,6 +908,35 @@ static void *hidthread(void *arg)
...
+				index_devstr = ast_radio_usb_get_devstr(index);
+				if (ast_strlen_zero(index_devstr)) {
+					/* No more devices, so break out of the loop */
+					ast_log(LOG_NOTICE, "USB Device with serial %s not found.\n", o->serial);
+					break;
+				}
+				/* Go through the list of usb devices, and get the serial numbers */
+				if (ast_radio_get_usb_serial(index_devstr, serial) == 0) continue;
+				ast_log(LOG_NOTICE, "Device Serial %s vs %s\n", o->serial, serial);
+				if (strcmp(o->serial, serial) == 0)
```
Allan-N, line 928: Pull the brace ('{') up to this line.

channels/chan_simpleusb.c
```diff
@@ -906,6 +908,35 @@ static void *hidthread(void *arg)
...
+			for(;;)
+			{
+				index_devstr = ast_radio_usb_get_devstr(index);
+				if (ast_strlen_zero(index_devstr)) {
+					/* No more devices, so break out of the loop */
+					ast_log(LOG_NOTICE, "USB Device with serial %s not found.\n", o->serial);
+					break;
+				}
+				/* Go through the list of usb devices, and get the serial numbers */
+				if (ast_radio_get_usb_serial(index_devstr, serial) == 0) continue;
```
Allan-N, line 926: The app_rpt project follows the asterisk coding standards where single/standalone statements are not allowed.   This should be :
```
if (ast_radio_get_usb_serial(index_devstr, serial) == 0) {
    continue;
}
```

channels/chan_simpleusb.c
```diff
@@ -3087,6 +3123,7 @@ static void _menu_print(int fd, struct chan_simpleusb_pvt *o)
 	ast_cli(fd, "Active radio interface is [%s]\n", simpleusb_active);
 	ast_mutex_lock(&usb_dev_lock);
 	ast_cli(fd, "Device String is %s\n", o->devstr);
+	ast_cli(fd, "Device Serial is %s\n", o->serial);
```
Allan-N, line 3126: I have mixed thoughts about reporting the device serial if it's not set.

Allan-N (changes requested):

coolacid:
Thanks @Allan-N for the review. C isn't my primary language, so I appreciate the hand holding.

Allan-N:
FYI : Asterisk [Coding Guidelines]( https://docs.asterisk.org/Development/Policies-and-Procedures/Coding-Guidelines/)

Allan-N (changes requested):
One other (set of) changes to consider.  The tune menus allow you to switch the device associated with a node (handy/needed if you have multiple dongles connected).  This is handled in chan_simpleusb.c by the usb_device_swap() function.  Seems like that function also needs to be aware of the serial #.

PR #396, opened 2024-09-06
Add RX Audio Statistics and ADC Clip Detection feature. This adds a n…

davidgsd:
Add RX Audio Statistics and ADC Clip Detection feature. This adds a new rxaudiostats parameter to simpleusb.conf, which enables a new function check_rx_audio() in chan_simpleusb.c to be called during processing of received USB audio frames. If ADC clipping is then detected GPIO4 is set for 500mS to support illumination of a Clip LED on the node / audio interface. Statistics collected also include peak and average RMS audio levels, averaged over the previous 1 second, which can be displayed from a new simpleusb-tune-menu 'R' option or AMI 'susb tune menu-support a' function.

The background behind this feature is discussed in this ASL forum post: https://community.allstarlink.org/t/audio-clip-led-feature-addition-to-asl3/21651

I have reviewed the schematics for a wide variety of USB radio interfaces used with ASL including models from Repeater-Builder, Masters Communications, DMK Engineering, Kits-for-Hams, BH7NOR, Digirig, as well as C-Media CM108AH/B sound fobs, and found that GPIO4 is not currently used in any of these interfaces, and thus is ideal for use as an LED clip indicator, to allow future URIs - or in the case of radio-less nodes a USB Communications Interface (UCI) - to provide a simple, low cost means to notify users of audio level issues. As more users get on AllStar, improperly adjusted audio levels are an increasingly important issue, and this addition will make it easier for users to confirm audio levels with a clip LED present on future products as well as with an added simpleusb-tune-menu option that shows more precise statistics than the existing "Set Rx Voice Level" utility.

I have tested this feature on several nodes using various URIs/UCIs with C-Media CM108AH, CM119A, and CM108B ICs. On each of these I added an LED output on GPIO4, and confirmed that the LED only illuminates when actual clipping occurred. See the function header documentation in chan_simpleusb.c for more details. The statistics collected are simple and require essentially zero CPU use, and under 1KB of additional RAM use.  Tests with the feature enabled and disabled resulted in no noticeable difference in asterisk's CPU use as shown by top. In both cases CPU use averages 10% +/- 1% on a Dell Wyse thin-client PC with 1.4GHz Atom CPU, with both the Allmon3 and AllScan dashboard apps running.

Below screenshot shows the operation of the new simpleusb-tune-menu "R" option, which prints once per second the peak and average rx audio stats along with number of clip events detected, until the Enter key is pressed - similar to how other menu options work such as the "2) Set Rx Voice Level" or "V) View COS, CTCSS and PTT Status".

susb-tune-menu-R

To those with experience in audio production, recording, engineering, etc., these basic audio statistics and Clip LED support should be a welcome addition. Audio equipment products such as mixers, mic preamps, dynamics processors, etc. almost always have clip LEDs on the inputs thereby allowing levels to be quickly and easily optimized (with no need to SSH into anything).

For those who use AllStar mainly for repeaters and have high-end communications test equipment, this feature may not be as interesting, but most AllStar users have small inexpensive nodes and limited abilities to optimize audio levels, or may not be able to easily SSH into their node and figure out how to use the tune-menu utility properly. Unfortunately it takes only 1 or 2 users on an AllStar-linked repeater, hub, or net who have audio levels that are too high or too low, and a significant inconvenience then results for everyone else as they have to ride their volume controls up and down as various people key up. Distorted audio is the bane of digital/VOIP modes and with the expanding number of ways that people can get on AllStar I feel it's important as an AllStar user and vendor to make it as easy as possible for users to get their audio levels right.  (This feature could also be extended in the future to support AGC/leveling functionality that could for example smoothly and gradually reduce or increase the rxmix setting when clipping is detected or if signal levels are excessively low.)

I was careful to really thoroughly look at how chan_simpleusb and the tune-menu utilities are set up, how configuration parameters and other variables and structures are managed, and how all this works in various use cases, including in the context of simpleusb supporting multiple nodes and interfaces on one server. I believe my implementation of this feature has closely followed the style and implementation of similar functions in the app_rpt code, while being simple and cleanly partitioned. I believe there should be essentially zero risk of bugs or issues with this feature addition. If I can provide any additional test info let me know. The feature is pretty easy to test since all it's doing is looking at incoming audio and then showing a few basic statistics in a tune-menu command, and setting a CM1xxx GPIO when clipping is detected. (Clipping can easily be induced just by increasing the rxmix setting in the tune-menu.)

I have not added this to the usbradio driver since that tends to be used more in unattended applications such as repeaters rather than in small personal nodes. But it would be very easy to also add to usbradio by moving the 2 new functions to a common file and copying the handful of other lines of code into the corresponding usbradio files.

### DIFF
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -598,10 +598,17 @@ static int hidhdwconfig(struct chan_simpleusb_pvt *o)
 		o->valid_gpios = 1;			/* for GPIO 1 */
 	}
 	/* validate checkrxaudio setting (Clip LED GPIO#) */
-	if (o->checkrxaudio && (o->checkrxaudio >= GPIO_PINCOUNT || !(o->valid_gpios & (1 << (o->checkrxaudio - 1)))))
+	if (o->checkrxaudio)
 	{
-		ast_log(LOG_ERROR, "Channel %s: checkrxaudio = GPIO%d not supported\n", o->name, o->checkrxaudio);
-		o->checkrxaudio = 0;
+		if (o->checkrxaudio >= GPIO_PINCOUNT || !(o->valid_gpios & (1 << (o->checkrxaudio - 1))))
+		{
+			ast_log(LOG_ERROR, "Channel %s: checkrxaudio = GPIO%d not supported\n", o->name, o->checkrxaudio);
+			o->checkrxaudio = 0;
+		}
+		else
+		{
+			o->hid_gpio_ctl |= 1 << (o->checkrxaudio - 1); /* confirm Clip LED GPIO set to output mode */
+		}
 	}
 	o->hid_gpio_val = 0;
 	for (i = 0; i < GPIO_PINCOUNT; i++) {
@@ -2101,8 +2108,14 @@ static struct ast_frame *simpleusb_read(struct ast_channel *c)
 				if (f1->datalen - src >= l) {	
 					/* enough to fill a frame */
 					memcpy(o->simpleusb_write_buf + o->simpleusb_write_dst, (char *) f1->data.ptr + src, l);
+
+					/* TBR - below appears to be an attempt to match levels to the original CM108
+					 * IC which has been out of production for over 10 years. Scaling audio to 
+					 * 109.375% will result in clipping! Any adjustments for CM1xxx gain differences
+					 * should be made in the mixer settings, not in the audio stream itself.
+					 */
 					/* Adjust the audio level for CM119 A/B devices */
-					if (o->devtype != C108_PRODUCT_ID) {
+					if (o->devtype != C108_PRODUCT_ID && !o->checkrxaudio) {
 						register int v;
 
 						sp = (short *) o->simpleusb_write_buf;
@@ -2118,6 +2131,7 @@ static struct ast_frame *simpleusb_read(struct ast_channel *c)
 							*sp++ = v;
 						}
 					}
+
 					sp = (short *) o->simpleusb_write_buf;
 					sp1 = outbuf;
 					doright = 1;
@@ -3285,7 +3299,6 @@ static void tune_write(struct chan_simpleusb_pvt *o)
  *		1 - get node names that are configured in simpleusb.conf
  *		2 - print parameters
  *		3 - get node names that are configured in simpleusb.conf, except current device
- *		a - receive audio statistics display
  *		b - receiver tune display
  *		c - receive level
  *		f - txa level
@@ -3303,6 +3316,7 @@ static void tune_write(struct chan_simpleusb_pvt *o)
  *		t - change rx on delay
  *		u - change tx off delay
  *		v - view cos, ctcss and ptt status
+ *		y - receive audio statistics display
  *
  * \param fd			Asterisk CLI fd
  * \param o				Private struct.
@@ -3346,24 +3360,6 @@ static void tune_menusupport(int fd, struct chan_simpleusb_pvt *o, const char *c
 			x++;
 		}
 		ast_cli(fd, "\n");
-		break;
-	case 'a':					/* display receive audio statistics (interactive) */
-	case 'A':					/* display receive audio statistics (once only) */
-		if (!o->hasusb) {
-			ast_cli(fd, USB_UNASSIGNED_FMT, o->name, o->devstr);
-			break;
-		}
-		if (!o->checkrxaudio) {
-			ast_cli(fd, "checkrxaudio is currently Disabled in simpleusb.conf\n");
-			break;
-		}
-		for (;;) {
-			ast_radio_print_rx_audio_stats(fd, &o->rxaudiostats);
-			if (cmd[0] == 'A')
-				break;
-			if (ast_radio_poll_input(fd, 1000))
-				break;
-		}
 		break;
 	case 'b':					/* receiver tune display */
 		if (!o->hasusb) {
@@ -3514,6 +3510,24 @@ static void tune_menusupport(int fd, struct chan_simpleusb_pvt *o, const char *c
 			break;
 		}
 		tune_rxtx_status(fd, o);
+		break;
+	case 'y':					/* display receive audio statistics (interactive) */
+	case 'Y':					/* display receive audio statistics (once only) */
+		if (!o->hasusb) {
+			ast_cli(fd, USB_UNASSIGNED_FMT, o->name, o->devstr);
+			break;
+		}
+		if (!o->checkrxaudio) {
+			ast_cli(fd, "checkrxaudio is currently Disabled in simpleusb.conf\n");
+			break;
+		}
+		for (;;) {
+			ast_radio_print_rx_audio_stats(fd, &o->rxaudiostats);
+			if (cmd[0] == 'Y')
+				break;
+			if (ast_radio_poll_input(fd, 1000))
+				break;
+		}
 		break;
 	default:
 		ast_cli(fd, "Invalid Command\n");
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -272,7 +272,6 @@ struct chan_simpleusb_pvt {
 	int hid_io_ctcss;
 	int hid_io_ctcss_loc;
 	int hid_io_ptt;
-	int hid_io_clip_led;			/* indicator to alert user of ADC clipping */
 	int hid_gpio_loc;
 	int32_t hid_gpio_val;
 	int32_t valid_gpios;
@@ -315,7 +314,7 @@ struct chan_simpleusb_pvt {
 	struct timeval tonetime;
 	int toneflag;
 	int duplex3;
-	int rxaudiostats;
+	int checkrxaudio;
 	
 	int32_t discfactor;
 	int32_t discounterl;
@@ -328,14 +327,7 @@ struct chan_simpleusb_pvt {
 	char *gpios[GPIO_PINCOUNT];
 	char *pps[32];
 
-	/* Rx audio (ADC) statistics variables. susb tune-menu "R" command displays
-	 * stats data (peak, average, min, max levels and clipped sample count).
-	 */
-#define AUDIO_STATS_LEN 50 			/* number of 20mS frames. 50 => 1 second buf len */
-	unsigned short maxbuf[AUDIO_STATS_LEN];		/* peak sample value per frame */
-	unsigned short clipbuf[AUDIO_STATS_LEN];	/* number of clipped samples per frame */
-	unsigned int pwrbuf[AUDIO_STATS_LEN];		/* total RMS power per frame */
-	short rxaudiostats_index;					/* Index within buffers, updated as frames received */
+	struct rxaudiostatistics rxaudiostats;
 	
 	ast_mutex_t usblock;
 };
@@ -355,8 +347,8 @@ static struct chan_simpleusb_pvt simpleusb_default = {
 	.rxondelay = 0,
 	.txoffdelay = 0,
 	.pager = PAGER_NONE,
-	.rxaudiostats = 1,
-	.rxaudiostats_index = 0
+	.checkrxaudio = 1,
+	.rxaudiostats.index = 0
 };
 
 /*	DECLARE FUNCTION PROTOTYPES	*/
@@ -380,8 +372,6 @@ static int simpleusb_setoption(struct ast_channel *chan, int option, void *data,
 static void tune_menusupport(int fd, struct chan_simpleusb_pvt *o, const char *cmd);
 static void tune_write(struct chan_simpleusb_pvt *o);
 static int _send_tx_test_tone(int fd, struct chan_simpleusb_pvt *o, int ms, int intflag);
-static void check_rx_audio(struct chan_simpleusb_pvt *o, short len);
-static void print_rx_audio_stats(int fd, struct chan_simpleusb_pvt *o);
 
 static char *simpleusb_active;	/* the active device */
 
@@ -574,7 +564,6 @@ static int hidhdwconfig(struct chan_simpleusb_pvt *o)
 		o->hid_io_ctcss = 2;		/* GPIO 2 is External CTCSS */
 		o->hid_io_ctcss_loc = 1;	/* is GPIO 2 */
 		o->hid_io_ptt = 8;			/* GPIO 4 is PTT */
-		o->hid_io_clip_led = 0;		/* No Clip LED on this HW */
 		o->hid_gpio_loc = 1;		/* For ALL GPIO */
 		o->valid_gpios = 1;			/* for GPIO 1 */
 	} else if (o->hdwtype == 0) {	//dudeusb
@@ -585,7 +574,6 @@ static int hidhdwconfig(struct chan_simpleusb_pvt *o)
 		o->hid_io_ctcss = 1;		/* VOL UP External CTCSS */
 		o->hid_io_ctcss_loc = 0;	/* VOL UP External CTCSS */
 		o->hid_io_ptt = 4;			/* GPIO 3 is PTT */
-		o->hid_io_clip_led = 3;		/* GPIO 4 is Clip LED */
 		o->hid_gpio_loc = 1;		/* For ALL GPIO */
 		o->valid_gpios = 0xfb;		/* for GPIO 1,2,4,5,6,7,8 (5,6,7,8 for CM-119 only) */
 	} else if (o->hdwtype == 2) {	//NHRC (N1KDO) (dudeusb w/o user GPIO)
@@ -596,7 +584,6 @@ static int hidhdwconfig(struct chan_simpleusb_pvt *o)
 		o->hid_io_ctcss = 1;		/* VOL UP is External CTCSS */
 		o->hid_io_ctcss_loc = 0;	/* VOL UP CTCSS */
 		o->hid_io_ptt = 4;			/* GPIO 3 is PTT */
-		o->hid_io_clip_led = 3;		/* GPIO 4 is Clip LED */
 		o->hid_gpio_loc = 1;		/* For ALL GPIO */
 		o->valid_gpios = 0;			/* for GPIO 1,2,4 */
 	} else if (o->hdwtype == 3) {	// custom version
@@ -607,9 +594,14 @@ static int hidhdwconfig(struct chan_simpleusb_pvt *o)
 		o->hid_io_ctcss = 2;		/* GPIO 2 is External CTCSS */
 		o->hid_io_ctcss_loc = 1;	/* is GPIO 2 */
 		o->hid_io_ptt = 4;			/* GPIO 3 is PTT */
-		o->hid_io_clip_led = 3;		/* GPIO 4 is Clip LED */
 		o->hid_gpio_loc = 1;		/* For ALL GPIO */
 		o->valid_gpios = 1;			/* for GPIO 1 */
+	}
+	/* validate checkrxaudio setting (Clip LED GPIO#) */
+	if (o->checkrxaudio && (o->checkrxaudio >= GPIO_PINCOUNT || !(o->valid_gpios & (1 << (o->checkrxaudio - 1)))))
+	{
+		ast_log(LOG_ERROR, "Channel %s: checkrxaudio = GPIO%d not supported\n", o->name, o->checkrxaudio);
+		o->checkrxaudio = 0;
 	}
 	o->hid_gpio_val = 0;
 	for (i = 0; i < GPIO_PINCOUNT; i++) {
@@ -2288,12 +2280,20 @@ static struct ast_frame *simpleusb_read(struct ast_channel *c)
 
 	/* Check for ADC clipping and input audio statistics before any filtering is done.
 	 * FRAME_SIZE define refers to 8Ksps mono which is 160 samples per 20mS USB frame.
-	 * check_rx_audio() takes the read buffer as received (48K stereo), extracts the
-	 * mono 48K channel and checks amplitude and distortion characteristics.
+	 * ast_radio_check_rx_audio() takes the read buffer as received (48K stereo),
+	 * extracts the mono 48K channel, checks amplitude and distortion characteristics,
+	 * and returns true if clipping was detected.
 	 */
-	if(o->rxaudiostats)
+	if(o->checkrxaudio)
 	{
-		check_rx_audio(o, 12 * FRAME_SIZE);
+		if (ast_radio_check_rx_audio((short *) o->simpleusb_read_buf, &o->rxaudiostats, 12 * FRAME_SIZE))
+		{
+			/* Set Clip LED GPIO pulsetimer if not already set */
+			if (!o->hid_gpio_pulsetimer[o->checkrxaudio - 1])
+			{
+				o->hid_gpio_pulsetimer[o->checkrxaudio - 1] = CLIP_LED_HOLD_TIME_MS;
+			}
+		}
 	}
 
 	/* Downsample received audio from 48000 stereo to 8000 mono */
@@ -3353,12 +3353,12 @@ static void tune_menusupport(int fd, struct chan_simpleusb_pvt *o, const char *c
 			ast_cli(fd, USB_UNASSIGNED_FMT, o->name, o->devstr);
 			break;
 		}
-		if (!o->rxaudiostats) {
-			ast_cli(fd, "rxaudiostats is currently Disabled in simpleusb.conf\n");
+		if (!o->checkrxaudio) {
+			ast_cli(fd, "checkrxaudio is currently Disabled in simpleusb.conf\n");
 			break;
 		}
 		for (;;) {
-			print_rx_audio_stats(fd, o);
+			ast_radio_print_rx_audio_stats(fd, &o->rxaudiostats);
 			if (cmd[0] == 'A')
 				break;
 			if (ast_radio_poll_input(fd, 1000))
@@ -3698,7 +3698,7 @@ static struct chan_simpleusb_pvt *store_config(const struct ast_config *cfg, con
 		CV_BOOL("deemphasis", o->deemphasis);
 		CV_BOOL("preemphasis", o->preemphasis);
 		CV_UINT("duplex3", o->duplex3);
-		CV_BOOL("rxaudiostats", o->rxaudiostats);
+		CV_UINT("checkrxaudio", o->checkrxaudio);
 		CV_END;
 		
 		for (i = 0; i < GPIO_PINCOUNT; i++) {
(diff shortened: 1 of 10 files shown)


jxmx:
Marked for review and inclusion in the next app_rpt release

KB4MDD:
Hotspot radio and SHARI devices presently use GPIO 4 for COS indication.  How would the default option of enabled affect these devices?

Hardware that implements this feature may use the Simple USB or USB Radio channel driver.  Is it possible to implement this feature in chan_usbradio?  We have been working to keep the feature set the same in both drivers.

davidgsd:
> Hotspot radio and SHARI devices presently use GPIO 4 for COS indication. How would the default option of enabled affect these devices?
> 
> Hardware that implements this feature may use the Simple USB or USB Radio channel driver. Is it possible to implement this feature in chan_usbradio? We have been working to keep the feature set the same in both drivers. The common code is in res_usbradio.

I had not seen the SHARI schematic before but managed to find one online and see that they do indeed have an LED on GPIO4, and after some further investigation also found that the Masters Comms RL-20 also supports a COS LED on GPIO 4 or 8.  After having reviewed those and the other URI schematics I have and the App_rpt code it looks like GPIO1 is the better option. Do you know of any products that might be using GPIO1? If not I can update things to use that.  I could also update the rxaudiostats config parameter to be a numeric instead of boolean setting where 0=feature off, 1=use GPIO1, ... etc., with a default value of GPIO1.  This would allow it to be easily changed for users who might happen to be using a GPIO for custom event logic.

And yes it would be easy to move the 2 main functions to a common file and implement this feature in usbradio also. To do this should I move the 2 functions to ../res/res_usbradio.c (which is used by both simpleusb and usbradio), or might it be preferable to create a new .c and .h file (e.g. maybe ../res/res_audio_utils.c & .h) ?

Thanks,
David

InterLinked1:
> And yes it would be easy to move the 2 main functions to a common file and implement this feature in usbradio also. To do this should I move the 2 functions to ../res/res_usbradio.c (which is used by both simpleusb and usbradio), or might it be preferable to create a new .c and .h file (e.g. maybe ../res/res_audio_utils.c & .h) ?

They should just go in res_usbradio; given the size of this patch and the integration required, I don't think a separate file would make sense. It wouldn't just be a sibling file either, a new directory needs to be created whenever a module consists of multiple files and the Makefiles need to be modified too. Not worth it for a small change like this.

I haven't looked at your patch in detail, but it should also follow the Asterisk coding guidelines - e.g. no newline after if, whitespacing, braces even for single statements, etc: https://docs.asterisk.org/Development/Policies-and-Procedures/Coding-Guidelines/

davidgsd:
I'm seeing a strange issue, hopefully one of you knows how to fix it.  If I have a function in res_usbradio.c, and put a call to that function in chan_simpleusb.c, something gets corrupted and the simpleusb channel driver will not load. No error messages seem to show up anywhere. I can log into AMI and if I type 'core show help susb' it says susb doesn't exist.  If I comment out the one call then everything works fine. It's as if the linker is not able to correctly link res_usbradio to chan_simpleusb. It's not a run-time issue, the call to ast_res_xxx() can be in a part of code that never executes.

If this does not sound familar, try opening res_usbradio.c, add the following line,
int ast_test123(void) { return 1; }

then in res_usbradio.h, add the following line,
int ast_test123(void);

then in chan_simpleusb.c add a call to ast_test123();

It will compile and make install fine but simpleusb channel driver will then not load. Does anyone have any ideas?

InterLinked1:
> I'm seeing a strange issue, hopefully one of you knows how to fix it. If I have a function in res_usbradio.c, and put a call to that function in chan_simpleusb.c, something gets corrupted and the simpleusb channel driver will not load. No error messages seem to show up anywhere. I can log into AMI and if I type 'core show help susb' it says susb doesn't exist. If I comment out the one call then everything works fine. It's as if the linker is not able to correctly link res_usbradio to chan_simpleusb. It's not a run-time issue, the call to ast_res_xxx() can be in a part of code that never executes.
> 
> If this does not sound familar, try opening res_usbradio.c, add the following line, int ast_test123(void) { return 1; }
> 
> then in res_usbradio.h, add the following line, int ast_test123(void);
> 
> then in chan_simpleusb.c add a call to ast_test123();
> 
> It will compile and make install fine but simpleusb channel driver will then not load. Does anyone have any ideas?

What is the error message when you run "module load chan_simpleusb"?

davidgsd:
ERROR[1053]: loader.c:283 module_load_error: Error loading module 'chan_simpleusb': /usr/lib/x86_64-linux-gnu/asterisk/modules/chan_simpleusb.so: undefined symbol: ast_test123

InterLinked1:
> ERROR[1053]: loader.c:283 module_load_error: Error loading module 'chan_simpleusb': /usr/lib/x86_64-linux-gnu/asterisk/modules/chan_simpleusb.so: undefined symbol: ast_test123

Try naming the function starting with `ast_radio`. Only those functions are exported.

davidgsd:
Hi, yes using ast_radio_ prefix fixes it.  Thanks

davidgsd:
I have completed the above changes and tested the feature pretty thoroughly with usbradio and also again with simpleusb. Also reviewed the coding standards a little more closely, added some spaces in front of keywords, etc. LMK if any questions.

channels/chan_usbradio.c
```diff
@@ -549,6 +554,19 @@ static int hidhdwconfig(struct chan_usbradio_pvt *o)
 		o->hid_gpio_loc = 1;		/* For ALL GPIO */
 		o->valid_gpios = 1;			/* for GPIO 1 */
 	}
+	/* validate checkrxaudio setting (Clip LED GPIO#) */
+	if (o->checkrxaudio)
```
InterLinked1, line 558: opening brace must be on same line

InterLinked1 (changes requested):

channels/chan_simpleusb.c
```diff
@@ -344,6 +347,8 @@ static struct chan_simpleusb_pvt simpleusb_default = {
 	.rxondelay = 0,
 	.txoffdelay = 0,
 	.pager = PAGER_NONE,
+	.checkrxaudio = 1,
```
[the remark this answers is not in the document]
KB4MDD: I am wondering if this needs to be off until the hardware vendors begin including this in their hardware designs.

I understand that there is a plan to provide sample configurations for different vendor hardware.  This could be enabled in those samples for the hardware that supports this feature.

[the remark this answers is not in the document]
davidgsd: There are a few reasons why I left the default as off initially for usbradio.
* I believe the majority of low-cost / small / radio-less nodes use simpleusb, whereas usbradio tends to be used more in repeaters or higher-end nodes ie. with commercial radios or mobile radios supporting flat audio. As a result, I suspect that people who are using usbradio are probably more likely on average to have properly calibrated their audio levels, and significantly less likely to be clipping their URI ADCs. So there should not be as much urgency to get this feature out to usbradio users, and, they are more probably likely to be more savvy about the config settings and can enable it easily if they want to. Also, usbradio users may be more likely to be using GPIOs for specialized purposes such as a repeater / remote base control line.
* There is also a serious issue in the existing code where there are some old "hacks" to match I/O levels to the original CM108 IC, which has been out of production for over 10 years (see "TBR" comments I added about this in the code), and thus in both simpleusb and usbradio these hacks are scaling the raw Tx audio by 110%, which **will** cause clipping in tx audio for all URIs that do not use the very old original CM108.  Those 3 TBR sections really should be removed now, but doing so will result in tx audio levels then being reduced by ~10% in both simpleusb and usbradio (for all URIs except those using the original CM108), and in rx levels being increased by ~2dB in usbradio (for all non-CM108 URIs). So for this reason I left the default as off for usbradio, with the idea that once these level matching hacks are removed and users made aware that they may need to adjust levels after upgrading, then at that point we could have checkrxaudio default to enabled.  And I would suggest that these hacks be removed ASAP, and could open a separate issue on that now if it would be helpful, and was going to do so either now or after this feature was added. (Seemed easier to add this first, then remove the old level scaling hacks as a separate issue and PR.)
* For simpleusb this feature should default to enabled because the users who are the most likely to be introducing clipped audio into the AllStarLink network are probably in general also those who are the least likely to later go in and enable the feature. In reviewing about a dozen URI schematics, none use GPIO1, and the very small number of simpleusb users who might use it as an output eg. for some sort of Event logic output, probably are familiar with the configuration settings and it should be easy for them to set the checkrxaudio parameter to off or to a different GPIO#.

channels/chan_simpleusb.c
```diff
@@ -314,6 +314,7 @@ struct chan_simpleusb_pvt {
 	struct timeval tonetime;
 	int toneflag;
 	int duplex3;
+	int checkrxaudio;
```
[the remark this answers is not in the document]
davidgsd: Now adding "enables RxAudioStats feature & Clip LED output on specified GPIO#" comment here and in chan_usbradio.c.

channels/chan_usbradio.c
```diff
@@ -549,6 +554,19 @@ static int hidhdwconfig(struct chan_usbradio_pvt *o)
...
+	{
+		if (o->checkrxaudio >= GPIO_PINCOUNT || !(o->valid_gpios & (1 << (o->checkrxaudio - 1))))
+		{
```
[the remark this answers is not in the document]
davidgsd: Updating these now.

configs/samples/usbradio.conf.sample
```diff
@@ -154,6 +154,11 @@
                             ; 1 - full duplex
 ;duplex3 = 0                ; duplex 3 gain setting (0 to disable) ???
 
+;checkrxaudio = 0           ; Enable receive audio statistics and ADC clip detection (0 to disable).
```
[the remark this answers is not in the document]
davidgsd: See earlier comments.

include/asterisk/res_usbradio.h
```diff
@@ -433,3 +445,67 @@ void ast_radio_time(time_t *second);
...
+ * - Min and max signal power levels averaged within each USB frame (dBFS)
+ *   These define average dynamic range (dB)
+ * - Min and max signal power averaged over a longer time period (dBFS)
+ *   These define total signal power and peak-to-average power ratio
+ *
+ * \author      	NR9V
+ * \param sbuf  	Rx audio sample buffer
+ * \param o	    	Rx Audio Stats data structure
+ * \param len   	Length of data in sbuf
+ * \return 	    	None
```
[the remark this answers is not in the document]
davidgsd: I have corrected the function header documentation to indicate the return is "1 if clipping detected, 0 otherwise".

res/res_usbradio.c
```diff
@@ -732,6 +732,92 @@ struct timeval ast_radio_tvnow(void)
...
+#define CLIP_SAMP_THRESH       0x7eb0
+#define CLIP_EVENT_MIN_SAMPLES 3
+int ast_radio_check_rx_audio(short *sbuf, struct rxaudiostatistics *o, short len)
+{
+	unsigned short i, j, val, max = 0, seq_clips = 0;
+	double pwr = 0.0;
+	short buf[FRAME_SIZE], last_clip = -1;
+
+	/* validate len and index */
+	if (len > 12 * FRAME_SIZE)
```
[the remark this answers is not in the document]
davidgsd: This has been updated.

res/res_usbradio.c
```diff
@@ -732,6 +732,92 @@ struct timeval ast_radio_tvnow(void)
...
+	}
+	tpwr /= AUDIO_STATS_LEN;
+	/* Convert to dBFS / dB */
+	scale = 1.0 / (double) (1 << 30);
+	dpk = (pk > 0.0) ? 10 * log10(pk * pk * scale) : -96.0;
+	tpwr = (tpwr > 0.0) ? 10 * log10(tpwr * scale) : -96.0;
+	dmin = minpwr ? 10 * log10(minpwr * scale) : -96.0;
+	dmax = maxpwr ? 10 * log10(maxpwr * scale) : -96.0;
+	/* Print stats */
+	sprintf(s1, "RxAudioStats: Pk %5.1f  Avg Pwr %3.0f  Min %3.0f  Max %3.0f  dBFS  ClipCnt %u",
```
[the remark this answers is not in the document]
davidgsd: In this specific case the float format specifiers have a set width, and the clip_cnt var is an unsigned short so should never be more than a few chars wide and thus the 100 bytes size of s1 should be way more than enough.

channels/chan_simpleusb.c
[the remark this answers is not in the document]
Allan-N: I think it would be better to default the GPIO pin usage to "off", period.  We don't know what hardware implementations are out in the wild and whether this new usage will trigger any conflicts.  It's a simple enough configuration change for those who do have the enhanced hardware.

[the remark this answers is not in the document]
davidgsd: There are probably a relatively small number of users who are doing something with GPIO1, and if we wanted to be sure none of them could be inconvenienced by having the GPIO set if they are clipping their ADC, then yes the feature should default to off. In this case however I suspect that the benefit to AllStar as whole could be greater to enable the feature by default (since most users will just go with defaults and not bother to enable a feature if they have no immediate need to). So I would see it as a question of does ASL want to prioritize putting something in place that can over time help prevent clipping and thereby make a significant improvement in overall audio quality on the network, or, does ASL not want to do that in order to prevent any possible unexpected result to some very small number of users. To best make such a decision might require a bit of quantification of the benefits vs. potential inconveniences. I do believe that many 1,000's of users would greatly benefit over time from increased audio quality that results from AllStar users being more generally aware of clipping and audio levels, and I highly suspect that the number of simpleusb users who might have something unexpected happen after an upgrade (where they did not see the update notes about the checkrxaudio GPIO configuration setting) is probably relatively very small - probably < 0.1% of users. Thus from my experience I think the benefits would significantly outweigh the potential inconvenience. Also URI makers may be less likely to support a Clip LED if the feature is not enabled by default.

If other reviewers agree that the use of a GPIO should default to off I'll update the code, though I do think that this feature is a bit different than most in that it's intent is to protect the long-term audio quality of the ASL network as a whole, which is important and should be prioritized. In any other case I would not want to be the cause of a possible inconvenience to even a single user, but as it is now there many users who are inconvenienced on a daily basis when they have to ride their volume controls up and down on nets or larger QSOs due to one or two users having excessively loud/distorted. or overly quiet audio levels. There are a combination of approaches that could help address this over time, but I think that defining a Clip LED as a standard (and enabled by default) feature is a good start and a sort of acknowledgement that audio quality on ASL is a definite priority.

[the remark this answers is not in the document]
InterLinked1: Another potential suggestion... leave it disabled in the code and in configs/samples, but enable it for simpleusb in the configs/rpt version. In the code, it can stay disabled by default but the rpt sample config version could have it uncommented to enable by default.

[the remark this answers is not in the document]
Allan-N: Q? isn't the clipping info available without enabling the GPIO pin?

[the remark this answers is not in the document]
davidgsd: Re. is the clipping info available separately from the GPIO function, when I first implemented this I did not separate the stats collection and Clip LED output functionality but I then realized the stats data is very useful in the tune-menu and via AMI so I will update that now so that the statistics are always available.

And re. leaving disabled by default if the config is not present in the user's .conf file, but then enabling in the configs/rpt .conf files, if I understand that correctly it would result in it being enabled in new installs (where user will be reviewing and changing the various configs settings during node set up and thus will have a chance to review everything), but then not enabled by default for existing installs who do an ASL upgrade. That makes a lot of sense, I'll set that up now.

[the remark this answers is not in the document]
Allan-N: Re: having the statistics always available ... YES!

Re: having the default GPIO pin usage (in code) set to disabled ... YES!

Re: enabling the GPIO pin usage by default in the configuration file ... I'm personally still hesitant here.  My preference would be to have the configuration file setting to enable/use one of the GPIO pins be commented out.  Those with hardware that supports the GPIO pin would only need to remove the ";".  and those who have hardware that uses that GPIO pin for something else won't run into conflicts.  But, that's "my" preference.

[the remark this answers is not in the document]
InterLinked1: > And re. leaving disabled by default if the config is not present in the user's .conf file, but then enabling in the configs/rpt .conf files, if I understand that correctly it would result in it being enabled in new installs (where user will be reviewing and changing the various configs settings during node set up and thus will have a chance to review everything), but then not enabled by default for existing installs who do an ASL upgrade. That makes a lot of sense, I'll set that up now.

Maybe I didn't state that super clearly... what I meant is that the configuration files in this repo go into configs/samples de facto, but there are also slightly different versions in configs/rpt "curated" for ASL.

The configs/samples files are not supposed to be ASL-catered at all, and hence I think the feedback on not trying to push users in a particular direction "for the sake of the network" would apply here. Same for the code, which isn't inherently ASL-specific either.

The configs in the rpt directory are intended for people who want an "ASL-recommended set of configs", and so I think it's fine to be a bit more hand holding there and set the default to 1 in that simpleusb.conf if you like, giving us the best of both worlds.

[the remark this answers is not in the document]
davidgsd: I updated the code to not enable GPIO output by default in the .c files, so that after an ASL update GPIO writes will not result unless a user manually updates their .conf file to enable the clipledgpio parameter. For new installs the parameter is still set to 1 in simpleusb.conf. To clarify, on a new node when ASL3 is installed for the first time and it creates a new default .conf file, is that copied from the repository configs/rpt/simpleusb.conf?

channels/chan_usbradio.c
```diff
@@ -549,6 +554,16 @@ static int hidhdwconfig(struct chan_usbradio_pvt *o)
 		o->hid_gpio_loc = 1;		/* For ALL GPIO */
 		o->valid_gpios = 1;			/* for GPIO 1 */
 	}
+	/* validate clipledgpio setting (Clip LED GPIO#) */
+	if (o->clipledgpio) {
+		if (o->clipledgpio >= GPIO_PINCOUNT || !(o->valid_gpios & (1 << (o->clipledgpio - 1)))) {
+			ast_log(LOG_ERROR, "Channel %s: clipledgpio = GPIO%d not supported\n", o->name, o->clipledgpio);
+			o->clipledgpio = 0;
+		}
+		else {
```
InterLinked1, line 563: } else {

configs/samples/simpleusb.conf.sample
```diff
@@ -90,6 +90,10 @@
 
 ;duplex3 = 0                ; duplex 3 gain setting (0 to disable)
 
+;clipledgpio = 1            ; Enable ADC Clip Detect feature to use a GPIO output (0 to disable).
```
InterLinked1, line 93: I know it's commented out by default but maybe for consistency both configs/samples files should have it set to 0 here? Just to avoid confusion

res/res_usbradio.c
```diff
@@ -732,6 +732,95 @@ struct timeval ast_radio_tvnow(void)
...
+			dpk, tpwr, dmin, dmax, clipcnt);
+	if (fd) {
+		ast_cli(fd, "%s\n", s1);
+	}
+	else {
```
InterLinked1, line 819: } else {

InterLinked1 (changes requested):

channels/chan_simpleusb.c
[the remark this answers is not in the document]
InterLinked1: Might be a good question for @jxmx and the package maintainers... if just running the rpt_install.sh script, no, it uses the files in configs/samples only. But I think the ASL3 wrappers likely use the rpt files instead.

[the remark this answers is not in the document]
Allan-N: > I think the ASL3 wrappers likely use the rpt files instead.

The default ASL3 configuration files come from .../app_rpt/configs/rpt/

channels/chan_usbradio.c
davidgsd: just updated this.

res/res_usbradio.c
davidgsd: just updated this.

configs/samples/simpleusb.conf.sample
davidgsd: Per our earlier comments I updated it to disabled by default in the .c files, and it is consistent between the configs/rpt/ and .sample files. I think it makes sense to leave it disabled by default on existing nodes ie. during updates, but that for new node installs the default (for at least the simpleusb driver) should be enabled, because during a new install is when users will be reviewing the config settings and selecting the values that make sense for them (and setting up any GPIO events). If the setting were not enabled in a new install though, I think there is a tendency that with the people who are less proficient in setting audio levels are probably also more likely to be less proficient with other config settings, and thus never make use of a feature that could alert them to audio level issues.

InterLinked1: > Per our earlier comments I updated it to disabled by default in the .c files, and it is consistent between the configs/rpt/ and .sample files. I think it makes sense to leave it disabled by default on existing nodes ie. during updates, but that for new node installs the default (for at least the simpleusb driver) should be enabled, because during a new install is when users will be reviewing the config settings and selecting the values that make sense for them (and setting up any GPIO events). If the setting were not enabled in a new install though, I think there is a tendency that with the people who are less proficient in setting audio levels are probably also more likely to be less proficient with other config settings, and thus never make use of a feature that could alert them to audio level issues.

I agree with your logic, but that's what the rpt files are for. The samples files aren't supposed to be judgmental like that, they are really just documentation, hence why the difference is a bit confusing from that perspective.

davidgsd: OK I think I understand what you're saying about the .sample files being more general to Asterisk and just documenting what configs are there rather than recommending specific values (correct me if I'm wrong). So I just updated that in configs/samples/simpleusb.conf.sample to 0, which now matches usbradio.conf.sample.

res/res_usbradio.c
InterLinked1, line 819: I'm still seeing this on 2 separate lines?
(should be `} else {`

InterLinked1 (changes requested):

res/res_usbradio.c
davidgsd: OK just fixed that last dangling else. (Nice to see this style preferred in C now as I remember when this more PHPish bracket style was frowned upon)

InterLinked1: > OK just fixed that last dangling else. (Nice to see this style preferred in C now as I remember when this more PHPish bracket style was frowned upon)

Interesting... that's K&R syntax, which is mostly what the Asterisk coding style is (with a few deviations).

InterLinked1 (approved):

res/res_usbradio.c
davidgsd: Must have been just the San Diego satellite communications companies I worked for in the 90s'-00's (before I moved into PHP). I just looked in my K&R book and they do indeed not put brackets on their own lines 👍

channels/chan_simpleusb.c
```diff
@@ -2090,6 +2104,13 @@ static struct ast_frame *simpleusb_read(struct ast_channel *c)
 				if (f1->datalen - src >= l) {	
 					/* enough to fill a frame */
 					memcpy(o->simpleusb_write_buf + o->simpleusb_write_dst, (char *) f1->data.ptr + src, l);
+
+					/* TBR - below appears to be an attempt to match levels to the original CM108
+					 * IC which has been out of production for over 10 years. Scaling audio to 
+					 * 109.375% will result in clipping! Any adjustments for CM1xxx gain differences
+					 * should be made in the mixer settings, not in the audio stream itself.
+					 */
+#if 0
```
Allan-N, line 2113: I very much appreciate the thoughts/comments but I question commenting out this block of code.  Won't this impact everyone with a non-C108_PRODUCT_ID device requiring them to make adjustments to their [mixer] settings?  If this is true, we will likely end up with lots of nodes with low audio.  If we do make this change now then we will need a very **BIG** release note.

channels/chan_usbradio.c
```diff
@@ -2031,6 +2045,14 @@ static struct ast_frame *usbradio_read(struct ast_channel *c)
...
 		ast_log(LOG_WARNING, "USB read channel [%s] was not stuck.\n", o->name);
 	}
+
+	/* TBR - below appears to be an attempt to match levels to the original CM108
+	 * IC which has been out of production for over 10 years. Scaling all rx audio to 
+	 * 80% would result in a 20% loss in dynamic range, added quantization noise, and 
+	 * outgoing IAX audio reduced by 2dB. Any adjustments for CM1xxx IC gain differences
+	 * should be made in the mixer settings, not in the audio stream itself.
+	 */
+#if 0
```
Allan-N, line 2055: Please see my comments in chan_simpleusb.c re: concerns about this #if ... #endif block.

channels/chan_usbradio.c
```diff
@@ -2096,6 +2135,12 @@ static struct ast_frame *usbradio_read(struct ast_channel *c)
 	}
 #endif
 
+	/* TBR - below appears to be an attempt to match levels to the original CM108
+	 * IC which has been out of production for over 10 years. Scaling audio to 110%
+	 * will result in clipping! Any adjustments for CM1xxx IC gain differences
+	 * should be made in the mixer settings, not in the audio stream itself.
+	 */
+#if 0
```
Allan-N, line 2143: Please see my comments in chan_simpleusb.c re: concerns about this #if ... #endif block.

Allan-N (changes requested):

channels/chan_simpleusb.c
KB4MDD: This is going to break most asl3 nodes and my nodes.  This might be appropriate on new installs we some sort of legacy option.

I would have to revisit all my sites after this update.

I think this should be in another pull request.  It is going to hold uo these changes.

Allan-N: Maybe we need an 'audio_scaling' scaling variable that enables/disables this behavior.  We set the default value to be TRUE/yes (to protect existing configurations that don't have the variable in their .conf files).  Then, we add 'audio_scaling = no' to the template "simpleusb.conf" file that will be used for "new" installations.

davidgsd: These sections that I if 0'd out are a major, major issue and bug, as they **will introduce distortion** in Tx audio on all ASL nodes that use any URI made in the past 10 years. I cannot stress enough how serious this is.

If incoming IAX audio happens to be less than 90%, which is usually the case for **most** users, you don't get distortion because these blocks of code are scaling by ~110%, which is only a  0.83dB level difference, but any IAX audio between 0dBFS and -0.83dBFS **will be digitally clipped and distorted** ie. sine waves squared off at the top and bottom (which has a very negative impact on audio quality and is very noticeable) on any and all nodes using ASL that have any URI which is not using the original 10+ year-old CM108 IC.

Once this digital clipping is introduced, it doesn't matter what any of your mixer settings are -- the Tx audio **will be** distorted.  Let me know if you have any questions on this or if anything I have explained above is not clear.

Fixing this now will greatly improve audio quality while bringing Tx audio levels to exactly where they should be, which is difference of -0.83dB. Differences of less than 1dB are usually not even perceptible thus fixing this will **not** cause "low" audio.

I will open an bug report on this now.

davidgsd: Bug report opened: https://github.com/AllStarLink/app_rpt/issues/399

channels/chan_usbradio.c
davidgsd: Bug report opened: https://github.com/AllStarLink/app_rpt/issues/399

channels/chan_usbradio.c
davidgsd: Bug report opened: https://github.com/AllStarLink/app_rpt/issues/399

channels/chan_simpleusb.c
davidgsd: Also see this post for additional notes: https://community.allstarlink.org/t/asl-usb-audio-notes/21832/

davidgsd: > This is going to break most asl3 nodes and my nodes. This might be appropriate on new installs we some sort of legacy option.
> I would have to revisit all my sites after this update.

Hi Danny, just wanted to clarify, once this issue is fixed there will be essentially zero perceptible difference in tx audio levels. The channel drivers are now multiplying all outgoing audio by 1.1, which results in only a 0.83dB difference in audio levels - and which also results in 0.83 dB of clipping and distortion added to any incoming IAX audio signals that have peaks above -0.83dBFS.  Thus fixing this will result in less than 1dB of difference in audio levels, which with most repeaters having 50-60dB of dynamic range, is essentially completely unnoticeable. And your repeaters would then all have almost 1 dB **less** distortion, which **will** make a noticeable improvement difference in audio quality. 

Thus to be as clear, fixing this will absolutely not break anything, and you would not have to revisit any of your sites. Please see the bug report I opened for more discussion and notes on this.

The only thing that is broken is the current implementation, which is creating distortion, as well as added quantization noise. These hacks are absolutely, unquestionably inappropriate, benefit no one, and should have never been in ASL. Fortunately they are extremely easy to remove, and will result only in an extremely slight reduction in Tx audio volume, while significantly improving audio quality for all ASL users.

davidgsd: PR has been updated to re-enable the audio stream scaling hacks in chan_simplusb.c and chan_usbradio.c. Moved the call to check_rx_audio() to above where the 0.8x rx audio scaling is done in chan_usbradio.c.

include/asterisk/res_usbradio.h
```diff
@@ -433,3 +445,67 @@ void ast_radio_time(time_t *second);
...
+ * detected.
+ *
+ * Example output message:
+ *   RxAudioStats: Pk -2.1  Avg Pwr -32  Min -60  Max -12  dBFS  ClipCnt 0
+ *
+ * Results are scaled to double precision 0.0-1.0 and converted to log (dB)
+ * ie. 10*log10(scaledVal) for power levels.
+ *
+ * \author  		NR9V
+ * \param fd		File descriptor to print to, or if 0 print using ast_verbose()
```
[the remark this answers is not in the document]
davidgsd: Writing to STDIN would seem to be a very non-standard use case that I'd think would never happen in app_rpt at least with this function. And looks like asterisk sets stdin and out to /dev/null/:
# pidof asterisk
787
# ls -l /proc/787/fd
total 0
lr-x------ 1 asterisk asterisk 64 Sep 15 12:11 0 -> /dev/null
l-wx------ 1 asterisk asterisk 64 Sep 15 12:11 1 -> /dev/null
Thus I think the current behavior of the function to call ast_verbose() if fd=0 is probably the best thing to do.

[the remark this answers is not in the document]
Allan-N: I'm not objecting to using the fd==0 test.  All I was suggesting is that you have a function where you are passed a file descriptor and "0" is a valid file descriptor.  The function should not make any assumptions about whether FD 0 is STDIN or was setup for a real file.  On the other hand, you will never have an FD==-1.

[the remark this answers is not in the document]
davidgsd: I do see the value of keeping things generic regardless of the actual usage of a function in context, so have updated that to allow 0, and print using ast_verbose only if fd < 0.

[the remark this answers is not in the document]
InterLinked1: > Thus I think the current behavior of the function to call ast_verbose() if fd=0 is probably the best thing to do.

I looked at the code and it seems you already made a change here but the comments don't line up with that, so just wanted to add that I agree with Allan, never use 0 as an indicator for lack of file descriptor, -1 is the standard value for "no FD", including in Asterisk.

> And looks like asterisk sets stdin and out to /dev/null:

I suspect that's because you're running a daemonized Asterisk, which is more typical, but STDOUT should be present when Asterisk is run in the foreground.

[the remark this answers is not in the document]
davidgsd: I did update the function header comments to in the previous commit:
` * \param fd		File descriptor to print to, or if -1 print using ast_verbose()`

But as the function was in fact using ast_verbose for any fd < 0, I just updated that now and pushed to have that line read:
 * \param fd		File descriptor to print to, or if < 0 print using ast_verbose()

Allan-N (approved):
Looks good!

KB4MDD (approved):
This looks good.

Allan-N:
The fix for this issue has been incorporated into ASL3-Asterisk 20.9.3+asl3-3.0.5-1

PR #406, opened 2024-09-17
Fix LastKeyedTime/LastTxKeyedTime info in RPT events

Allan-N:
The rpt_manager_trigger() function uses ctime(3) to add formatted date/time strings for LastKeyedTime and LastTxKeyedTime.  Unfortunately, the returned strings include trailing '\n' characters that add empty lines to the output. This presents an issue as an empty line typically indicates the end of the AMI output.

This change removes the extra (unexpected) '\n' characters.

### DIFF
--- a/apps/app_rpt/rpt_manager.c
+++ b/apps/app_rpt/rpt_manager.c
@@ -18,8 +18,7 @@
 
 extern struct rpt rpt_vars[MAXRPTS];
 
-static char *
-ctime_no_newline(const time_t *clock)
+static char* ctime_no_newline(const time_t *clock)
 {
 	static char buf[32];
 	const char *cp;
--- a/apps/app_rpt/rpt_manager.c
+++ b/apps/app_rpt/rpt_manager.c
@@ -18,6 +18,22 @@
 
 extern struct rpt rpt_vars[MAXRPTS];
 
+static char *
+ctime_no_newline(const time_t *clock)
+{
+	static char buf[32];
+	const char *cp;
+	size_t len;
+
+	cp = ctime_r(clock, buf);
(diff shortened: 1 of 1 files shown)


KB4MDD (approved):
This looks good.

apps/app_rpt/rpt_manager.c
```diff
@@ -18,6 +18,22 @@
 
 extern struct rpt rpt_vars[MAXRPTS];
 
+static char *
+ctime_no_newline(const time_t *clock)
```
InterLinked1, line 22: static char* should be on the same line as the function name, with no space between char and *

InterLinked1 (changes requested):

apps/app_rpt/rpt_manager.c
Allan-N: Fixed

apps/app_rpt/rpt_manager.c
```diff
@@ -18,8 +18,7 @@
 
 extern struct rpt rpt_vars[MAXRPTS];
 
-static char *
-ctime_no_newline(const time_t *clock)
+static char* ctime_no_newline(const time_t *clock)
```
InterLinked1, line 21: Actually misspoke here, sorry... what I meant is the * should be on the function name, not the char. Sorry for the mixup!

InterLinked1 (changes requested):

apps/app_rpt/rpt_manager.c
Allan-N: Everyone has their coding style :-)

InterLinked1 (approved):

PR #409, opened 2024-09-27
Tune menus should allow specifying "node" to update

Allan-N:
Adds a new "-n node" argument to the `simpleusb-tune-menu` and `radio-tune-menu` commands that allow one to [pre-]select the active device being displayed and updated.


KB4MDD (approved):
This looks good.

InterLinked1 (approved):

PR #410, opened 2024-09-27
Tune menus will not always save settings

Allan-N:
When the tune menus request that the current configuration settings be saved there is an assumption that each variable will be represented in the .conf file.  There is no reason that any variables missing from the .conf files cannot be added.

### DIFF
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -3198,6 +3198,36 @@ static void _menu_txb(int fd, struct chan_simpleusb_pvt *o, const char *str)
 }
 
 /*!
+ * \brief Update the tune settings to the configuration file.
+ * \param filename	The configuration file being updated (e.g. "simpleusb.conf").
+ * \param category	The category being updated (e.g. "12345").
+ * \param variable	The variable being updated (e.g. "rxboost").
+ * \param value		The value being updated (e.g. "yes").
+ */
+static int tune_variable_update(const char *filename, struct ast_category *category,
+								const char *variable, const char *value)
+{
+	int	res;
+
+	res = ast_variable_update(category, variable, value, NULL, 0);
+	if (res != 0) {
+		struct ast_variable *var;
+
+		/* if we could not find/update the variable, create new */
+		var = ast_variable_new(variable, value, filename);
+		if (var != NULL) {
+			/* and append */
+			ast_variable_append(category, var);
+			return 0;
+		}
+
+		return -1;
+	}
+
+	return 0;
+}
+
+/*!
  * \brief Write tune settings to the configuration file. If the device EEPROM is enabled, the settings are  saved to EEPROM.
  * \param o Channel private.
  */
@@ -3216,25 +3246,25 @@ static void tune_write(struct chan_simpleusb_pvt *o)
 	}
 
 #define CONFIG_UPDATE_STR(field) \
-	if (ast_variable_update(category, #field, o->field, NULL, 0)) { \
+	if (tune_variable_update(CONFIG, category, #field, o->field)) { \
 		ast_log(LOG_WARNING, "Failed to update %s\n", #field); \
 	}
 
 #define CONFIG_UPDATE_INT(field) { \
 	char _buf[15]; \
 	snprintf(_buf, sizeof(_buf), "%d", o->field); \
-	if (ast_variable_update(category, #field, _buf, NULL, 0)) { \
+	if (tune_variable_update(CONFIG, category, #field, _buf)) { \
 		ast_log(LOG_WARNING, "Failed to update %s\n", #field); \
 	} \
 }
 
 #define CONFIG_UPDATE_BOOL(field) \
-	if (ast_variable_update(category, #field, o->field ? "yes" : "no", NULL, 0)) { \
+	if (tune_variable_update(CONFIG, category, #field, o->field ? "yes" : "no")) { \
 		ast_log(LOG_WARNING, "Failed to update %s\n", #field); \
 	}
 	
 #define CONFIG_UPDATE_SIGNAL(key, field) \
-	if (ast_variable_update(category, #key, signal_type[o->field], NULL, 0)) { \
+	if (tune_variable_update(CONFIG, category, #key, signal_type[o->field])) { \
 		ast_log(LOG_WARNING, "Failed to update %s\n", #field); \
 	}
 
@@ -3258,7 +3288,7 @@ static void tune_write(struct chan_simpleusb_pvt *o)
 		CONFIG_UPDATE_INT(rxondelay);
 		CONFIG_UPDATE_INT(txoffdelay);
 		if (ast_config_text_file_save2(CONFIG, cfg, "chan_simpleusb", 0)) {
-			ast_log(LOG_WARNING, "Failed to save config\n");
+			ast_log(LOG_WARNING, "Failed to save config %s.\n", CONFIG);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


channels/chan_simpleusb.c
```diff
@@ -3197,6 +3197,36 @@ static void _menu_txb(int fd, struct chan_simpleusb_pvt *o, const char *str)
...
 
+/*!
+ * \brief Update the tune settings to the configuration file.
+ * \param filename	The configuration file being updated (e.g. "simpleusb.conf").
+ * \param category	The category being updated (e.g. "12345").
+ * \param variable	The variable being updated (e.g. "rxboost").
+ * \param value		The value being updated (e.g. "yes").
+ */
+static int tune_variable_update(const char *filename, struct ast_category *category,
+								const char *variable, const char *value)
```
InterLinked1, line 3208: Seems like a lot of indentation here?

channels/chan_simpleusb.c
```diff
@@ -3197,6 +3197,36 @@ static void _menu_txb(int fd, struct chan_simpleusb_pvt *o, const char *str)
 	return;
 }
 
+/*!
+ * \brief Update the tune settings to the configuration file.
+ * \param filename	The configuration file being updated (e.g. "simpleusb.conf").
+ * \param category	The category being updated (e.g. "12345").
+ * \param variable	The variable being updated (e.g. "rxboost").
+ * \param value		The value being updated (e.g. "yes").
+ */
```
InterLinked1, line 3206: retval is not documented

channels/chan_simpleusb.c
```diff
@@ -3197,6 +3197,36 @@ static void _menu_txb(int fd, struct chan_simpleusb_pvt *o, const char *str)
...
+static int tune_variable_update(const char *filename, struct ast_category *category,
+								const char *variable, const char *value)
+{
+	int	res;
+
+	res = ast_variable_update(category, variable, value, NULL, 0);
```
InterLinked1, line 3212: Asterisk coding style prefers this kind of control flow:

```
if (failed) {
  return -1;
}
// continue if success
```

There shouldn't need to be more than one level of blocks with this approach, as opposed to up to two now (basically, invert your conditions)

channels/chan_usbradio.c
```diff
@@ -4388,6 +4388,36 @@ static void tune_rxctcss(int fd, struct chan_usbradio_pvt *o, int intflag)
...
+ * \brief Update the tune settings to the configuration file.
+ * \param filename	The configuration file being updated (e.g. "usbradio.conf").
+ * \param category	The category being updated (e.g. "12345").
+ * \param variable	The variable being updated (e.g. "rxboost").
+ * \param value		The value being updated (e.g. "yes").
+ */
+static int tune_variable_update(const char *filename, struct ast_category *category,
+								const char *variable, const char *value)
+{
+	int	res;
```
InterLinked1, line 4401: should just be one space between int and res

channels/chan_usbradio.c
```diff
@@ -4388,6 +4388,36 @@ static void tune_rxctcss(int fd, struct chan_usbradio_pvt *o, int intflag)
...
 
+/*!
+ * \brief Update the tune settings to the configuration file.
+ * \param filename	The configuration file being updated (e.g. "usbradio.conf").
+ * \param category	The category being updated (e.g. "12345").
+ * \param variable	The variable being updated (e.g. "rxboost").
+ * \param value		The value being updated (e.g. "yes").
+ */
+static int tune_variable_update(const char *filename, struct ast_category *category,
+								const char *variable, const char *value)
```
InterLinked1, line 4399: same thing here with indentation

channels/chan_usbradio.c
```diff
@@ -4388,6 +4388,36 @@ static void tune_rxctcss(int fd, struct chan_usbradio_pvt *o, int intflag)
...
+{
+	int	res;
+
+	res = ast_variable_update(category, variable, value, NULL, 0);
+	if (res != 0) {
```
InterLinked1, line 4404: same re: conditions

channels/chan_usbradio.c
```diff
@@ -4463,7 +4493,7 @@ static void tune_write(struct chan_usbradio_pvt *o)
 		CONFIG_UPDATE_SIGNAL(txmixa, txmixa, mixer_type);
 		CONFIG_UPDATE_SIGNAL(txmixb, txmixb, mixer_type);
 		if (ast_config_text_file_save2(CONFIG, cfg, "chan_usbradio", 0)) {
-			ast_log(LOG_WARNING, "Failed to save config\n");
+			ast_log(LOG_WARNING, "Failed to save config %s.\n", CONFIG);
```
InterLinked1, line 4496: Personally, I don't think periods are needed at the end of log messages, it doesn't help readability any, though I don't know if that's a hard guideline, so this is more of a suggestion if you agree or you can ignore if not

InterLinked1 (changes requested):

channels/chan_simpleusb.c
Allan-N: This file appears to be using 4 character tabstops and the indentation is simply to have the args align.

channels/chan_usbradio.c
Allan-N: That was one <tab>.  Changed to a <space>

channels/chan_simpleusb.c
Allan-N: Reworked

channels/chan_usbradio.c
Allan-N: I think this is OK.

channels/chan_usbradio.c
Allan-N: Reworked

channels/chan_usbradio.c
Allan-N: Poof!

channels/chan_simpleusb.c
Allan-N: Added.

channels/chan_simpleusb.c
```diff
@@ -3197,6 +3197,37 @@ static void _menu_txb(int fd, struct chan_simpleusb_pvt *o, const char *str)
...
+		return 0;
+	}
+
+		/* if we could not find/update the variable, create new */
+		var = ast_variable_new(variable, value, filename);
+	if (var == NULL) {
+		return -1;
+	}
+
+			/* and append */
```
InterLinked1, line 3226: Indentation is off in this function now (at least looking at it on GitHub)

channels/chan_usbradio.c
```diff
@@ -4388,6 +4388,37 @@ static void tune_rxctcss(int fd, struct chan_usbradio_pvt *o, int intflag)
...
+		return 0;
+	}
+
+		/* if we could not find/update the variable, create new */
+		var = ast_variable_new(variable, value, filename);
+	if (var == NULL) {
+		return -1;
+	}
+
+			/* and append */
```
InterLinked1, line 4417: Same thing with indentation - just need to remove some tabs

InterLinked1 (changes requested):

channels/chan_simpleusb.c
Allan-N: What you are seeing is the difference between the 4-character tabstops used in the file and the default/typical 8-character tabstops.  GitHub doesn't know what tabstop settings were/are used in the file.

channels/chan_usbradio.c
Allan-N: tabstops=4 vs. tabstops=8

InterLinked1: A tab is a tab, the alignment should be correct no matter what the tabstop is.
I took another look and I think there is an extra tab on some lines when there shouldn't be. e.g. 3220, 3229, etc.

Allan-N:
My bad.  The app I used to copy changes from a working directory to the one I was using for the PR was compressing whitespace.  That kept from seeing changes that also needed to be committed.

channels/chan_usbradio.c
Allan-N: Fixed.

InterLinked1 (approved):

KB4MDD (approved):
This looks good.

Issue #399, opened 2024-09-14
Simpleusb and Usbradio drivers are introducing digital clipping into all Tx audio


davidgsd:
While implementing an Rx Audio Statistics feature I stumbled across a major issue in the channel drivers. See comments in https://github.com/AllStarLink/app_rpt/pull/396. The sections of code that I if 0'd out and added comments to will introduce distortion in Tx audio on all ASL nodes that use any URI made in the past 10 years. 

If incoming IAX peak audio levels happen to be less than -0.83dbFS, which is usually the case with most users, distortion will not occur because these blocks of code are scaling by "only" ~110%, which is only a 0.83dB level difference, but any IAX audio between -0.83dBFS and 0dBFS **will be** digitally clipped and distorted ie. sine waves squared off at the top and bottom (which has a negative impact on audio quality and is definitely noticeable), on any and all nodes that have any URI that is not using the original 10+ year-old CM108 IC.

Once this digital clipping is introduced, it doesn't matter what any of the mixer settings are -- Tx audio **will** be distorted. Having been a EE and audio engineer for 30 years I am amazed that these hacks ever made it into the ASL code 10 years ago and that it went unnoticed until now. Let me know if anyone has any questions. I can understand that for anyone without an audio engineering background it might not be clear why these sections of code are an issue so I'd be happy to chat or to share results of various tests, eg. sending a 0dBFS sine wave test tone over IAX and then recording from a URI DAC output into a wave editor.

Fixing this now will greatly improve audio quality while bringing Tx audio levels to exactly where they should be (which is a difference of -0.83dB). Differences of less than 1dB are usually not even perceptible thus fixing this will not cause "low" audio or any issue for existing nodes, but will noticeably improve audio quality on any node that has a significant number of users or where incoming IAX audio levels do sometimes exceed -0.83dBFS.

davidgsd:
Also see this post for additional notes: https://community.allstarlink.org/t/asl-usb-audio-notes/21832/

davidgsd:
Below are the code excerpts in question:

chan_simpleusb.c:
```
					/* TBR - below appears to be an attempt to match levels to the original CM108
					 * IC which has been out of production for over 10 years. Scaling audio to 
					 * 109.375% will result in clipping! Any adjustments for CM1xxx gain differences
					 * should be made in the mixer settings, not in the audio stream itself.
					 */
					/* Adjust the audio level for CM119 A/B devices */
[...]

PR #411, opened 2024-09-27
SimpleUSB: show RX level

Allan-N:
The `simpleusb-tune-menu` shows the current TX levels but only reports the current RX level when adjustments are being made.  The information is available and should be reported.

### DIFF
--- a/utils/simpleusb-tune-menu.c
+++ b/utils/simpleusb-tune-menu.c
@@ -747,6 +747,7 @@ static int astgetresp(char *cmd)
 	int rxboost = 0, preemphasis = 0, deemphasis = 0;
 	int plfilter = 0, pttmode = 0, carrierfrom = 0, ctcssfrom = 0;
 	int rxondelay = 0, txoffdelay = 0;
+	int rxmixerset = 0, micplaymax = 0, spkrmax = 0, micmax = 0;
 	int result;
 	char str[256];
 
@@ -755,14 +756,15 @@ static int astgetresp(char *cmd)
 	for (;;) {
 
 		/* get device parameters from Asterisk */
-		if (astgetline(COMMAND_PREFIX "tune menu-support 0", str, sizeof(str) - 1)) {
-			printf("The setup information for chan_simpleusb could not be retrieved!\n\n");
+		if (astgetline(COMMAND_PREFIX "tune menu-support 0+4", str, sizeof(str) - 1)) {
+			printf("The chan_simpleusb setup information could not be retrieved!\n\n");
 			printf("Verify that Asterisk is running and chan_simpleusb is loaded.\n\n");
 			exit(255);
 		}
-		if (sscanf(str, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d", 
+		if (sscanf(str, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d", 
 			&txmixaset, &txmixbset, &echomode, &rxboost, &preemphasis, &deemphasis, 
-			&plfilter, &pttmode, &carrierfrom, &ctcssfrom, &rxondelay, &txoffdelay) != 12) {
+			&plfilter, &pttmode, &carrierfrom, &ctcssfrom, &rxondelay, &txoffdelay,
+			&rxmixerset, &micplaymax, &spkrmax, &micmax) != 16) {
 			fprintf(stderr, "Error parsing device parameters: %s\n", str);
 			exit(255);
 		}
@@ -774,7 +776,7 @@ static int astgetresp(char *cmd)
 		}
 
 		printf("1) Select active USB device\n");
-		printf("2) Set Rx Voice Level (using display)\n");
+		printf("2) Set Rx Voice Level using display (currently '%d')\n", rxmixerset);
 		if (keying) {
 			printf("3) Set Transmit A Level (currently '%d') and send test tone\n", txmixaset);
 		} else {
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -3326,10 +3326,22 @@ static void tune_menusupport(int fd, struct chan_simpleusb_pvt *o, const char *c
 	option_verbose = 0;
 	switch (cmd[0]) {
 	case '0':					/* return audio processing configuration */
-		ast_cli(fd, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d\n", 
-			o->txmixaset, o->txmixbset, o->echomode, o->rxboost, o->preemphasis, 
-			o->deemphasis, o->plfilter, o->invertptt, o->rxcdtype, o->rxsdtype, 
-			o->rxondelay, o->txoffdelay);
+		/* note: to maintain backward compatibility for those expecting a specific # of
+		   values to be returned (and in a specific order).  So, we only add to the end
+		   of the returned list.  Also, once an update has been released we can't change
+		   the format/content of any previously returned string */
+		if (!strcmp(cmd, "0+4")) {
+			ast_cli(fd, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d\n", 
+				o->txmixaset, o->txmixbset, o->echomode, o->rxboost, o->preemphasis, 
+				o->deemphasis, o->plfilter, o->invertptt, o->rxcdtype, o->rxsdtype, 
+				o->rxondelay, o->txoffdelay, o->rxmixerset,
+				o->micplaymax, o->spkrmax, o->micmax);
+		} else {
+			ast_cli(fd, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d\n", 
+				o->txmixaset, o->txmixbset, o->echomode, o->rxboost, o->preemphasis, 
+				o->deemphasis, o->plfilter, o->invertptt, o->rxcdtype, o->rxsdtype, 
+				o->rxondelay, o->txoffdelay);
+		}
 		break;
 	case '1':					/* return usb device name list */
 		for (x = 0, oy = simpleusb_default.next; oy && oy->name; oy = oy->next, x++) {
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -3696,9 +3696,14 @@ static void _menu_print(int fd, struct chan_usbradio_pvt *o)
 		ast_cli(fd, "off.\n");
 	}
 
+	if (o->rxdemod == RX_AUDIO_FLAT) {
+		ast_cli(fd, "Rx Level currently set to %d\n", (int) ((o->rxvoiceadj * 200.0) + .5));
+	} else {
+		ast_cli(fd, "Rx Level currently set to %d\n", o->rxmixerset);
+	}
+	ast_cli(fd, "Rx Squelch currently set to %d\n", o->rxsquelchadj);
 	ast_cli(fd, "Tx Voice Level currently set to %d\n", o->txmixaset);
 	ast_cli(fd, "Tx Tone Level currently set to %d\n", o->txctcssadj);
-	ast_cli(fd, "Rx Squelch currently set to %d\n", o->rxsquelchadj);
 	return;
 }
 
@@ -3963,10 +3968,24 @@ static void tune_menusupport(int fd, struct chan_usbradio_pvt *o, const char *cm
 		}
 	switch (cmd[0]) {
 	case '0':					/* return audio processing configuration */
-		ast_cli(fd, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d\n", 
-			flatrx, txhasctcss, o->echomode, o->rxboost, o->txboost,
-			o->rxcdtype, o->rxsdtype, o->rxondelay, o->txoffdelay,
-			o->txprelim, o->txlimonly, o->rxdemod, o->txmixa, o->txmixb);
+		/* note: to maintain backward compatibility for those expecting a specific # of
+		   values to be returned (and in a specific order).  So, we only add to the end
+		   of the returned list.  Also, once an update has been released we can't change
+		   the format/content of any previously returned string */
+		if (!strcmp(cmd, "0+9")) {
+			ast_cli(fd, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%f,%d,%d,%d,%d,%d,%d,%d\n", 
+				flatrx, txhasctcss, o->echomode, o->rxboost, o->txboost,
+				o->rxcdtype, o->rxsdtype, o->rxondelay, o->txoffdelay,
+				o->txprelim, o->txlimonly, o->rxdemod, o->txmixa, o->txmixb,
+				o->rxmixerset, o->rxvoiceadj, o->rxsquelchadj, o->txmixaset,
+				o->txmixbset, o->txctcssadj, o->micplaymax, o->spkrmax,
+				o->micmax);
+		} else {
+			ast_cli(fd, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d\n", 
+				flatrx, txhasctcss, o->echomode, o->rxboost, o->txboost,
+				o->rxcdtype, o->rxsdtype, o->rxondelay, o->txoffdelay,
+				o->txprelim, o->txlimonly, o->rxdemod, o->txmixa, o->txmixb);
+		}
 		break;
 	case '1':					/* return usb device name list */
 		for (x = 0, oy = usbradio_default.next; oy && oy->name; oy = oy->next, x++) {
--- a/utils/radio-tune-menu.c
+++ b/utils/radio-tune-menu.c
@@ -78,6 +78,7 @@ static const char * const sd_signal_type[] = {"no", "usb", "usbinvert", "dsp", "
 static const char * const demodulation_type[] = {"no", "speaker", "flat"};
 
 /*! \brief mixer type */
+enum { TX_OUT_OFF, TX_OUT_VOICE, TX_OUT_LSD, TX_OUT_COMPOSITE, TX_OUT_AUX };
 static const char * const mixer_type[] = {"no", "voice", "tone", "composite", "auxvoice"};
 
 /*! \brief command prefix for Asterisk - simpleusb channel driver access */
@@ -975,6 +976,10 @@ static void options_menu(void)
 	int rxboost = 0, txboost = 0, carrierfrom = 0, ctcssfrom = 0;
 	int rxondelay = 0, txoffdelay = 0, txprelim = 0, txlimonly = 0;
 	int rxdemod = 0, txmixa = 0, txmixb = 0;
+	int rxmixerset = 0, rxsquelchadj = 0;
+	float rxvoiceadj = 0;
+	int txmixaset = 0, txmixbset = 0, txctcssadj = 0;
+	int micplaymax = 0, spkrmax = 0, micmax = 0;
 	char str[256];
 	int result;
 
@@ -983,15 +988,17 @@ static void options_menu(void)
 	for (;;) {
 
 		/* get device parameters from Asterisk */
-		if (astgetline(COMMAND_PREFIX "tune menu-support 0", str, sizeof(str) - 1)) {
+		if (astgetline(COMMAND_PREFIX "tune menu-support 0+9", str, sizeof(str) - 1)) {
 			printf("The setup information for chan_usbradio could not be retrieved!\n\n");
 			printf("Verify that Asterisk is running and chan_usbradio is loaded.\n\n");
 			exit(255);
 		}
-		if (sscanf(str, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d", 
+		if (sscanf(str, "%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%d,%f,%d,%d,%d,%d,%d,%d,%d", 
 			&flatrx, &txhasctcss, &echomode, &rxboost, &txboost,
 			&carrierfrom, &ctcssfrom, &rxondelay, &txoffdelay,
-			&txprelim, & txlimonly, &rxdemod, &txmixa, &txmixb) != 14) {
+			&txprelim, &txlimonly, &rxdemod, &txmixa, &txmixb,
+			&rxmixerset, &rxvoiceadj, &rxsquelchadj, &txmixaset,
+			&txmixbset, &txctcssadj, &micplaymax, &spkrmax, &micmax) != 23) {
 			fprintf(stderr, "Error parsing device parameters: %s\n", str);
 			exit(255);
 		}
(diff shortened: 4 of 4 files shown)


Allan-N:
Note: minimal changes were needed for SimpleUSB as only one variable was involved.  I think I can do the same for USBRadio but multiple settings (and calculations) are in play.

Please advise if you think we parity between the two is essential.

InterLinked1 (approved):

InterLinked1:
> Note: minimal changes were needed for SimpleUSB as only one variable was involved. I think I can do the same for USBRadio but multiple settings (and calculations) are in play.
> 
> Please advise if you think we parity between the two is essential.

Not knowing the context, I think parity is a good thing in general, but I'll defer to @KB4MDD

utils/simpleusb-tune-menu.c
```diff
@@ -774,7 +785,7 @@ static int astgetresp(char *cmd)
 		}
 
 		printf("1) Select active USB device\n");
-		printf("2) Set Rx Voice Level (using display)\n");
+		printf("2) Set Rx Voice Level w/display (currently '%d')\n", rxmixerset);
```
KB4MDD, line 788: Allen how do you feel about replacing 'w/display' with 'with display'?

KB4MDD:
There are a number of adjustments in chan_usbradio.  If you want to make the set Rx Voice Level the same as chan_simpleusb, that would be good.  That would have those options the same and we could evaluate the need to add the current setting to the other commands.  The P option can be used to see the current settings of the other items.

utils/simpleusb-tune-menu.c
Allan-N: Done.

Allan-N:
I have updated the code so that both SimpleUSB and USBRadio will now display the current settings.

InterLinked1 (approved):

davidgsd:
I had noticed this inconsistency before, nice to see this now being added. Question though, is the rxvoiceadj parameter actually used in any actual audio processing?  If so, where? I know the parameter exists and can be read/written to the conf file, but after what seemed a thorough search of the various C files it appears the variable is _not actually used_ in any calculations. If that is correct, then it would not be very useful to show that cfg instead of rxmixerset in the rxdemod == RX_AUDIO_FLAT case. (From what I can tell, all use of the rxvoiceadj parameter in USBRadio is dead code that has no actual effect on any audio functionality and should be removed, and in SimpleUSB it has no significant effect other than causing clipping and should be totally removed from there also.)

Allan-N:
>Question though, is the rxvoiceadj parameter actually used in any audio processing? If so, where?

I see many references to o->rxvoiceadj in channels/chan_usbradio.c

davidgsd:
> I see many references to o->rxvoiceadj in channels/chan_usbradio.c

Hi Allen, as I mentioned above, "I know the parameter exists and can be read/written to the conf file, but after what seemed a thorough search of the various C files it appears the variable is _not used_ in any calculations".

My question is not if there are references to rxvoiceadj in the code. It is whether rxvoiceadj is actually used for anything that will have **any effect whatsoever on audio levels or related USBRadio functionality**.  

For example, if I add a parameter named xyz_adj to the config files, allow it to be read and written from the .conf files, and displayed and set in a utility menu, but that configuration parameter is _not actually used in any way during processing of any audio_, then what is the purpose of that parameter? Well of course it would have no purpose or reason to exist in the code if it doesn't actually have some real effect on some audio processing function or logic.

From my review of the code, rxvoiceadj is not used in any way that would affect any audio or other processing done in USBRadio.  If you disagree, could you please point out where this variable is used in such a way?  Again, note that reading the parameter to/from a conf file, or EEPROM, or displaying it or editing it in a menu, does not mean that it actually affects actual channel driver audio/DSP processing functionality.

Presumably rxvoiceadj was indeed used in some actual audio processing functionality at some point in the distant past, but it appears that whatever that functionality was, it has since been removed, but the rest of the code relating to the parameter was not.  Thus it appears the parameter now does nothing but take up space.  Therefore, rather than continue to copy and paste this parameter into other code, I would suggest we take a closer look at what it is actually doing, if anything. I'm pretty sure you will find that it is indeed not used for any real purpose, ie. it is a 'vestigial' parameter and any references to it in new or old code that imply it affects "Rx Voice Level" are incorrect.

Maybe I'm wrong and my greps of the C files somehow missed it being assigned to some other variable or function that actually does use it in a way that affects Rx voice level, so in that case I apologize for asking you to show me where in the code this variable actually does something like that. However I'd bet $20 that this parameter does not in fact in any way affect rx voice levels, or anything else.

I happened to have been looking at this parameter as part of my investigations into bug report https://github.com/AllStarLink/app_rpt/issues/399, and noted there the following a couple days ago: "USBRadio does not appear to use the rxvoiceadj parameter to do any audio scaling. It references the parameter in a number of places, reading it and saving it to/from the conf files and optionally a CM1xx attached EEPROM, and printing the value in different places, but from what I can tell the value is never used in any kind of calculation or processing of audio. xpmr.c has a copy of rxvoiceadj (pChan->prxVoiceAdjust) but this also is never actually used. Thus it seems pretty clear to me that all code in chan_usbradio.c and xmpr.c referencing these variables is dead code, that just makes it harder for everyone to understand what the various channel driver settings and variables actually do, and that really should be deleted."

Allan-N:
>Hi Allen, as I mentioned above, "I know the parameter exists and can be read/written to the conf file, but after what seemed a thorough search of the various C files it appears the variable is not used in any calculations".
>
>My question is not if there are references to rxvoiceadj in the code. It is whether rxvoiceadj is actually used for anything that will have any effect whatsoever on audio levels or related USBRadio functionality.

Yes, looking more carefully and I agree.  It looks as if the variable (and related usage) is only read or written but never actually used.  Same for channels/xpmr/xpmr.c and channels/chan_voter.c.  But, it does look like the variable is used in chan_simpleusb.c.

But, having made the observation here, I don't think this is the place to drive its removal.  Let's file another issue :-)

Allan-N:
Filed https://github.com/AllStarLink/app_rpt/issues/417

Allan-N:
@davidgsd

You had added some comments to https://github.com/AllStarLink/app_rpt/issues/399 that have me thinking. Specifically, that the CM108's variants allow setting the dBV values but not with the granularity of 0-999 levels that we have been presenting in the UI.  To me, that means if I had been using a value of 500 but found the level to be just a bit low then I might raise it to 525 ... not knowing why raising the value by 10 can't be detected but raising the value by 15 could be heard/measured.

Q? do you know if there's a place in the SimpleUSB / USBRadio channel drivers that knows what divisor is being used?  I'm thinking about how we might display both the 0-999 value, the dBV value, and the interval between each step.

With my latest, work-in-progress, changes I have the simpleusb-tune-menu showing :
```
  2) Set Rx Voice Level using display (currently '500')
```
Would this be better ?
```
  2) Set Rx Voice Level using display (currently '500', -12.0 dBV)
```
But, how would one convey that you need to increment/decrement the level by 26.32 before it will make a difference?  Would this be too confusing?
```
  2) Set Rx Voice Level using display (currently '500', -12.0 dBV, 486...512)
```
and, yes, we would be computing the low...high range for the current bucket

Thoughts?

davidgsd:
Hi Allan, I think there are 2 steps that would be great to do, the first is to show the actual mixer settings read and written to the CM1xx, and then a 2nd more involved step would then be to fully characterize the A/AH vs. B CM1xx ICs and then also show what those mixer settings correspond to in actual voltages. The dBV values I mentioned before are what should be the case according to their datasheets (see https://allscan.info/images/CM1xx-specs-2000x2752.png), however there are some discrepancies in some parts of the datasheets as well as in actual observed behavior. I may have some time this week to  run a script that loops through all the gain settings on both a CM108 AH and B while I apply test signals of known amplitudes to the Micin line, and confirm what signal levels actually result into ASL.

To show the actual CM1xx mixer settings (eg. set in chan_simpleusb.c :: mixer_write()) I added the following to susb_tune() after the first call to mixer_write(o);
```
			/*	mixer_write() writes the following to the CM1xx:
					o->micplaymax, o->spkrmax, o->micmax
					mic_setting =  o->rxmixerset * o->micmax / 1000;
				Then sets rxvoiceadj as follows:
					// get interval step size
					f = 1000.0 / (float) o->micmax;
					o->rxvoiceadj = 1.0 + (modff(((float) o->rxmixerset) / f, &f1) * .187962);
			*/
			ast_cli(fd, "micplaymax %d, spkrmax %d, micmax %d, micsetting %d stepsize %0.3f rxvoiceadj %0.06f\n",
				o->micplaymax, o->spkrmax, o->micmax, (o->rxmixerset * o->micmax / 1000), 
				(1000.0 / (float) o->micmax), o->rxvoiceadj);
```

With that in place when you then run the susbtune utility and change the rxmixerset value you'll see exactly what the setting gets translated to into the CM1xx.  I'll do a bit more testing today on this and expand on this comment in a few hours but for now this should maybe help show the parameters of interest.

davidgsd:
Example output of above log message on CM108B:
```
3040-asl3*CLI> susb tune rx 500
Changed setting on RX Channel to 500
micplaymax 31, spkrmax 37, micmax 35, micsetting 17 stepsize 28.571
3040-asl3*CLI> susb tune rx 600
Changed setting on RX Channel to 600
micplaymax 31, spkrmax 37, micmax 35, micsetting 21 stepsize 28.571
```

davidgsd:
I finished characterizing the CM108B. I calibrated a signal generator such that 0dB = the IC's max rated input which is 2.88Vpp. Then took the following measurements:
Vin=2.88Vpp = 2.04Vrms (max input spec) = 6.2dBVrms
`rxmixset step siggendB	Max		Avg
0			0	0	    -10.8  	-14
30			1	0		-9.8	-13
60			2	0		-8.8	-12
...
180			6	0		-4.8	-8
...
300			10	0		-0.8	-4
330			11	0		 0.0	-3
...
380			13	-3	    -0.4	-3
...
470			16	-6		-0.4	-3
500			17  -10 	-3.4	-6
600			21	-12		-1.3	-4
700			24	-20		-6.3	-9
800			28	-30		-12.1	-15  (65.87mV)
900			31	-30		-9.1	-12
999			34	-30		-6.1	-9`

Conclusions:
CM108B has 35 gain steps
At lowest gain (step 0), 0dB sig in results in measured signal of -10.8 dBFS => -10.8 dB attenuation
At highest gain (step 34), -30dB sig in results in measured signal of -6.1 dBFS => 23.9 dB gain
=> Gain referenced to stated max input (6.2dBVrms) is -11 to 24 dB = 35dB, => 1dB / step
=> Gain referenced to 0dBV is -11 to 24 dB 
=> Supported input range for 0dBFS is -18 to ~6*
	*At low gain settings, input clipping occurs prior to the ADC.
	Therefore, it appears that gain settings of below 300 have no practical use as distortion could be occurring before the signal gets to the ADC.

Will repeat this on the CM108AH tomorrow.  BTW example of commands used:
`3040-asl3*CLI> susb tune rx 500
Changed setting on RX Channel to 500
micplaymax 31, spkrmax 37, micmax 35, micsetting 17 stepsize 28.571
3040-asl3*CLI> susb tune menu-support Y
RxAudioStats: Pk  -3.4  Avg Pwr  -6  Min  -6  Max  -6  dBFS  ClipCnt 0
3040-asl3*CLI> susb tune rx 600
Changed setting on RX Channel to 600
micplaymax 31, spkrmax 37, micmax 35, micsetting 21 stepsize 28.571`
The rx audio stats function comes in very handy for this kind of testing. It's also nice for showing distortion of a sine wave test signal, which should have exactly a 3dB difference in peak - Avg pwr. Once that starts going down you know there is distortion occurring somewhere.

davidgsd:
It seems that the reason C-Media went from 16 1.5dB ADC gain steps in A/AH to 36 1dB steps in the B's was to make it work more like a volume control rather than simply a gain trim for boosting low level signals. (That also makes it more consistent with the DAC mixer range and steps.) For apps such as ASL which do not have a 'volume' knob on the ADC side, the extra ~12dB of attenuation available in the B's does not translate to usable attenuation, unless someone was really clipping the CM1xx mixer input. That input does seem to have some compression going on, with rxmixerset=0 I had to put 11.5Vrms into it (16.25Vpk / 21.2dBV), to get a max. reading of -7.2dBFS which is quite a lot of voltage to put into something that runs on a 5V supply. Each 1dB increase in rxmixerset then only yielded a ~0.6dB increase in the measured input peak value. Even with 16Vpk going in I had to set rxmixerset to 300 before I got a peak reading of 0dBFS. So there are some interesting nonlinearities happening in the lower gain range. There could actually be some sort of useful compression/limiting characteristic there. I may try adding a more precise distortion measurement output for use with a sine wave input which should further clarify if there may be a valid use for the CM1xxB lower gain settings.

davidgsd:
Looks like rxmixerset of 325 is the lowest setting that can be used where the full ADC dynamic range is available and no distortion occurs in the CM1xxB mixer. In the following test output I added a Crest Factor stat to the RxAudioStats output. That should be 3dB, and anytime goes below ~2.8dB but ClipCnt is 0 that indicates clipping/distortion in the CM1xxB mixer rather than in the ADC. I'll take a look at the IAX waveforms tomorrow also.

```
3040-asl3*CLI> susb tune rx 300
Changed setting on RX Channel to 300
micplaymax 31, spkrmax 37, micmax 35, micsetting 10 stepsize 28.571
3040-asl3*CLI> susb tune menu-support y
RxAudioStats: Pk  -2.0  Avg Pwr  -5.1  Min  -5  Max  -5  dBFS  CF  3.07 dB  ClipCnt 0
RxAudioStats: Pk  -1.9  Avg Pwr  -5.0  Min  -5  Max  -5  dBFS  CF  3.16 dB  ClipCnt 0
RxAudioStats: Pk  -1.3  Avg Pwr  -4.4  Min  -5  Max  -4  dBFS  CF  3.12 dB  ClipCnt 0
RxAudioStats: Pk  -1.1  Avg Pwr  -4.2  Min  -4  Max  -4  dBFS  CF  3.10 dB  ClipCnt 0
RxAudioStats: Pk  -0.8  Avg Pwr  -3.6  Min  -4  Max  -4  dBFS  CF  2.83 dB  ClipCnt 0
RxAudioStats: Pk  -0.8  Avg Pwr  -3.5  Min  -4  Max  -3  dBFS  CF  2.72 dB  ClipCnt 0
RxAudioStats: Pk  -0.8  Avg Pwr  -3.3  Min  -3  Max  -3  dBFS  CF  2.52 dB  ClipCnt 0
RxAudioStats: Pk  -0.7  Avg Pwr  -3.0  Min  -3  Max  -3  dBFS  CF  2.36 dB  ClipCnt 0
RxAudioStats: Pk  -0.5  Avg Pwr  -2.9  Min  -3  Max  -3  dBFS  CF  2.44 dB  ClipCnt 0
RxAudioStats: Pk  -0.5  Avg Pwr  -2.5  Min  -3  Max  -2  dBFS  CF  2.06 dB  ClipCnt 0
RxAudioStats: Pk  -0.4  Avg Pwr  -2.4  Min  -3  Max  -2  dBFS  CF  1.97 dB  ClipCnt 0
RxAudioStats: Pk  -0.4  Avg Pwr  -2.2  Min  -2  Max  -2  dBFS  CF  1.85 dB  ClipCnt 0
RxAudioStats: Pk  -0.3  Avg Pwr  -2.1  Min  -2  Max  -2  dBFS  CF  1.82 dB  ClipCnt 0
RxAudioStats: Pk  -0.3  Avg Pwr  -1.9  Min  -2  Max  -2  dBFS  CF  1.61 dB  ClipCnt 0
RxAudioStats: Pk  -0.3  Avg Pwr  -1.7  Min  -2  Max  -2  dBFS  CF  1.47 dB  ClipCnt 0
RxAudioStats: Pk  -0.1  Avg Pwr  -1.5  Min  -2  Max  -1  dBFS  CF  1.39 dB  ClipCnt 0
RxAudioStats: Pk  -0.1  Avg Pwr  -1.3  Min  -1  Max  -1  dBFS  CF  1.19 dB  ClipCnt 0
3040-asl3*CLI>
3040-asl3*CLI> susb tune rx 350
Changed setting on RX Channel to 350
micplaymax 31, spkrmax 37, micmax 35, micsetting 12 stepsize 28.571
3040-asl3*CLI> susb tune menu-support y
RxAudioStats: Pk  -2.2  Avg Pwr  -5.2  Min  -5  Max  -5  dBFS  CF  3.07 dB  ClipCnt 0
RxAudioStats: Pk  -0.0  Avg Pwr  -3.7  Min  -5  Max  -3  dBFS  CF  3.67 dB  ClipCnt 0
RxAudioStats: Pk   0.0  Avg Pwr  -2.5  Min  -3  Max  -2  dBFS  CF  2.48 dB  ClipCnt 263
RxAudioStats: Pk   0.0  Avg Pwr  -2.5  Min  -3  Max  -2  dBFS  CF  2.52 dB  ClipCnt 198
RxAudioStats: Pk  -0.0  Avg Pwr  -2.8  Min  -3  Max  -3  dBFS  CF  2.82 dB  ClipCnt 0
RxAudioStats: Pk  -0.0  Avg Pwr  -2.8  Min  -3  Max  -3  dBFS  CF  2.83 dB  ClipCnt 0
RxAudioStats: Pk  -0.0  Avg Pwr  -2.8  Min  -3  Max  -3  dBFS  CF  2.83 dB  ClipCnt 0
RxAudioStats: Pk  -0.0  Avg Pwr  -2.8  Min  -3  Max  -3  dBFS  CF  2.83 dB  ClipCnt 0
RxAudioStats: Pk  -0.0  Avg Pwr  -3.0  Min  -3  Max  -3  dBFS  CF  3.02 dB  ClipCnt 0
RxAudioStats: Pk  -0.0  Avg Pwr  -3.1  Min  -3  Max  -3  dBFS  CF  3.05 dB  ClipCnt 0
RxAudioStats: Pk  -0.0  Avg Pwr  -3.3  Min  -3  Max  -3  dBFS  CF  3.26 dB  ClipCnt 0
RxAudioStats: Pk  -0.2  Avg Pwr  -3.4  Min  -4  Max  -3  dBFS  CF  3.20 dB  ClipCnt 0
RxAudioStats: Pk  -0.6  Avg Pwr  -3.7  Min  -4  Max  -4  dBFS  CF  3.07 dB  ClipCnt 0
RxAudioStats: Pk  -0.6  Avg Pwr  -3.7  Min  -4  Max  -4  dBFS  CF  3.07 dB  ClipCnt 0
RxAudioStats: Pk  -0.6  Avg Pwr  -3.7  Min  -4  Max  -4  dBFS  CF  3.07 dB  ClipCnt 0
RxAudioStats: Pk  -0.6  Avg Pwr  -3.7  Min  -4  Max  -4  dBFS  CF  3.07 dB  ClipCnt 0
RxAudioStats: Pk  -0.6  Avg Pwr  -3.7  Min  -4  Max  -4  dBFS  CF  3.07 dB  ClipCnt 0
RxAudioStats: Pk  -0.6  Avg Pwr  -3.7  Min  -4  Max  -4  dBFS  CF  3.07 dB  ClipCnt 0
RxAudioStats: Pk  -0.6  Avg Pwr  -3.7  Min  -4  Max  -4  dBFS  CF  3.07 dB  ClipCnt 0
3040-asl3*CLI>
[2024-10-07 01:44:09.668] NOTICE[15882]: dnsmgr.c:225 dnsmgr_refresh: dnssrv: host 'register.allstarlink.org' changed from 34.105.111.212:443 to 162.248.92.131:443
3040-asl3*CLI> susb tune rx 325
Changed setting on RX Channel to 325
micplaymax 31, spkrmax 37, micmax 35, micsetting 11 stepsize 28.571
3040-asl3*CLI> susb tune menu-support y
RxAudioStats: Pk  -1.6  Avg Pwr  -4.7  Min  -5  Max  -5  dBFS  CF  3.07 dB  ClipCnt 0
RxAudioStats: Pk  -1.6  Avg Pwr  -4.7  Min  -5  Max  -5  dBFS  CF  3.07 dB  ClipCnt 0
RxAudioStats: Pk  -0.7  Avg Pwr  -4.6  Min  -5  Max  -4  dBFS  CF  3.85 dB  ClipCnt 0
RxAudioStats: Pk  -0.1  Avg Pwr  -3.1  Min  -3  Max  -3  dBFS  CF  2.98 dB  ClipCnt 0
RxAudioStats: Pk  -0.1  Avg Pwr  -3.1  Min  -3  Max  -3  dBFS  CF  2.94 dB  ClipCnt 0
RxAudioStats: Pk  -0.1  Avg Pwr  -3.1  Min  -3  Max  -3  dBFS  CF  2.94 dB  ClipCnt 0
RxAudioStats: Pk  -0.1  Avg Pwr  -3.1  Min  -3  Max  -3  dBFS  CF  2.94 dB  ClipCnt 0
RxAudioStats: Pk  -0.1  Avg Pwr  -3.1  Min  -3  Max  -3  dBFS  CF  2.94 dB  ClipCnt 0
RxAudioStats: Pk  -0.1  Avg Pwr  -3.1  Min  -3  Max  -3  dBFS  CF  2.94 dB  ClipCnt 0
RxAudioStats: Pk  -0.1  Avg Pwr  -3.1  Min  -3  Max  -3  dBFS  CF  2.94 dB  ClipCnt 0
RxAudioStats: Pk  -0.1  Avg Pwr  -3.1  Min  -3  Max  -3  dBFS  CF  2.94 dB  ClipCnt 0


```

Allan-N:
With the information you provided (and the suggested extra logging/reporting to the channel driver) it looks like it will be easy to know how the 0-999 levels are divided up.  That means I can easily report :
```
  2) Set Rx Voice Level using display (currently '500', 8/16, 486...512)
```
What are your thoughts on "8/16" ?  on "486...512"?  Does either/both help?  or do they make the level reporting too confusing?

Reporting dBV may be out-of-reach, at least for this change ... but might be good to include in some documentation.

davidgsd:
Hi Allan, I just finished characterizing the CM108AH ADC. Results:
Vin=2.88Vpp = 2.04Vrms (max input spec) = 6.2dBVrms = "0dB" setting on signal generator (1KHz sine wave input)
```
rxmixset step siggendB	Max	Avg
0	0	-6	-0.8  	-3.8
70	1	-7	-0.2	-3.2
250	4	-12	-0.5	-3.6
500	8	-18	-0.3	-3.4
750	12	-24	-0.1	-3.3
999	15	-30	-1.2	-4.8
```
Conclusions:
CM108AH has 15 rxmixer gain steps
At lowest gain (step 0), -6dB sig in results in measured signal of -0.8 dBFS => 5.2 dB gain
At highest gain (step 15), -30dB sig in results in measured signal of -1.2 dBFS => 28.8 dB gain
=> Gain referenced to stated max input (6.2dBVrms) is 5 to 29 dB = 24dB range => 1.5dB / step
=> Supported input range for 0dBFS is -23 to 1 dBVrms

Example commands used during tests:
```
3040-asl3*CLI> susb tune rx 500
Changed setting on RX Channel to 500
micplaymax 127, spkrmax 151, micmax 16, micsetting 8 stepsize 62.500
3040-asl3*CLI> susb tune menu-support Y
RxAudioStats: Pk  -0.3  Avg Pwr  -3.4  Min  -3  Max  -3  dBFS  CF  3.13 dB  ClipCnt 0
3040-asl3*CLI> susb tune rx 750
Changed setting on RX Channel to 750
micplaymax 127, spkrmax 151, micmax 16, micsetting 12 stepsize 62.500
3040-asl3*CLI> susb tune menu-support Y
RxAudioStats: Pk  -0.1  Avg Pwr  -3.3  Min  -3  Max  -3  dBFS  CF  3.19 dB  ClipCnt 0
```

Summarizing the key points of all above testing, 
* The CM108AH has an input range of -23 to 1 dBVrms, 
corresponding to rxmixerset values of 999 to 0,
corresponding to micsetting values of 15 to 0 (1.5dB steps)
* The CM108B has a usable input range (supporting full ADC dynamic range) of -18 to 6 dBVrms, 
corresponding to rxmixerset values of 999 to 325,
corresponding to micsetting values of 34 to 11 (1dB steps)
* According to the datasheets, ADC input and rx mixer specs are identical for CM108AH and CM119A, and identical for CM108B and CM119B.

I also plan to characterize the DAC side of the 3 main variants (CM108 A/AH, CM119A) as there is significant variation there in stated output range, and supposedly the same number of gain steps per the datasheets (0 - -45 dB), yet from above debug prints we see that AH has micplaymax 127 & spkrmax 151 vs. micplaymax 31 & spkrmax 37. The A/AH datasheets say nothing about having 151 txmix steps so I'll see if I can solve that mystery tomorrow.

Options for what could be shown to users:
* Existing 0-999 numbers. Of course these should continue to be shown as they allow simple enough iterative adjustments.
* Actual CM1xx IC setting steps.  This would be good to show somewhere as it shows exactly what the hardware is configured for. However it's not a number that is otherwise useful or intuitive so it probably doesn't deserve too much attention.
* dB or dBV values.  This is much more useful as it tells you a real number and makes subsequent adjustments much faster because dB are standard calibrated units. eg. if you see from the level meter or rx stats that your current level is -12 dB and you want to be at -3 dB, you just add 9 to the rxmix setting and you're done.  No more making random guesses about what 0-999 number might actually correspond to 9dB louder.

Note that dBV units have all the advantages of dB but in addition are absolute rather than relative units, that correspond to actual voltages that can be easily measured by anyone with a DMM and can easily be cross-referenced to various radio datasheets or other audio equipment datasheets. Thus anyone designing a repeater system or other node can easily confirm what their gains should be based on standard specifications that most manufacturers provide and have that as a starting point when configuring node and other equipment settings.

URIs do not provide these specs and I seem to be the first person to actually have measured what gain ranges are fully and properly supported. The CM1xx datasheets are entirely unclear about this and have some contradictions. I will plan to add this info to an "Advanced" section in the ASL manual so that this data is readily available in a nice table format. With that in place various people with audio, commercial radio, or systems engineering backgrounds will immediately be able to see how the various pieces connect together and what the gain structures are. URIs do not generally alter Rx levels other than some provide a trim pot, so this info should then clarify things for many interfaces.

With the above in mind, dBV is definitely the way to go. It's the standard in pro audio equipment and there is nothing complicated about it. Most people won't know what exact voltage a certain dBV value corresponds to, but they wouldn't know that for an arbitrary 0-999 number either, and it's very easy to look up a dBV value and see what it means, and dB units in general are much clearer. i.e. most hams are used to dB units such as with S-meters which have S 0-9 (6dB per step), then S9+10dB, S9+20dB, etc.  And a good number of people have general familiarity with decibels and know for example that a -3dB difference makes something sound half as loud.  dBV are also very simple to show, i.e. in your example,
`2) Set Rx Voice Level using display (currently '500' (-11dBV))
`
The added dBV value is very short and simple, and though it would probably be ignored by most people, that ~20% of hams with electronics, engineering, or audio backgrounds will be quite impressed and will find that extra bit of information very useful.  It is however a significant task to store the conversions between 0-999, CM1xx steps, and the actual gain ranges for all IC variants, so I wouldn't suggest doing this now. Probably in the next week I can add a page with all this data to the ASL Manual, and at that point the ASL code itself could just refer people to that link rather than try to show all that info in the menus and support display and input of the rx and tx mix values in multiple units, which could make the menus a bit messy/confusing. Would probably take some experimentation to make it work in a simple and elegant way that best suited both "amateurs" and audio/engineering pros. There could for example be a single menu setting that would enable dBV units, or any other desired units (%, 0-999, V, mV, dBu) and then the other various menus would follow that. That would minimize extra clutter. Anyhow just some ideas off the top of my head...

KB4MDD (approved):
This looks good.

PR #413, opened 2024-10-02
chan_usrp:  Add TX/RX debug messages

KB4MDD:
This updates chan_usrp to have debug messages for TX and RX on and off. The messages closely match those in chan_simpleusb and chan_usbradio. This indication is needed to help users troubleshoot problems when using this driver.

I also added a queue overflow warning message should the internal queue overflow.  This should never happen, but if it does, it needs to be reported.

### DIFF
--- a/channels/chan_usrp.c
+++ b/channels/chan_usrp.c
@@ -371,9 +371,11 @@ static int usrp_indicate(struct ast_channel *ast, int cond, const void *data, si
 	switch (cond) {
 	case AST_CONTROL_RADIO_KEY:
 		p->txkey = 1;
+		ast_debug(1, "Channel %s: ACRK TX ON.\n", ast_channel_name(ast));
 		break;
 	case AST_CONTROL_RADIO_UNKEY:
 		p->txkey = 0;
+		ast_debug(1, "Channel %s: ACRUK TX OFF.\n", ast_channel_name(ast));
 		break;
 	case AST_CONTROL_HANGUP:
 		return -1;
@@ -593,6 +595,7 @@ static int usrp_xwrite(struct ast_channel *ast, struct ast_frame *frame)
 				remque((struct qelem *) qp);
 				ast_free(qp);
 			}
+			ast_log(LOG_WARNING, "Channel %s: Receive queue exceeds threshold\n", ast_channel_name(ast));
 			if (pvt->rxkey) {
 				pvt->rxkey = 1;
 			}
@@ -610,6 +613,7 @@ static int usrp_xwrite(struct ast_channel *ast, struct ast_frame *frame)
 				fr.delivery.tv_sec = 0;
 				fr.delivery.tv_usec = 0;
 				ast_queue_frame(ast, &fr);
+				ast_debug(1, "Channel %s: RX ON\n", ast_channel_name(ast));
 			}
 			pvt->rxkey = MAX_RXKEY_TIME;
 			qp = pvt->rxq.qe_forw;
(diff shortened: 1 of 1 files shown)


channels/chan_usrp.c
```diff
@@ -593,6 +595,7 @@ static int usrp_xwrite(struct ast_channel *ast, struct ast_frame *frame)
 				remque((struct qelem *) qp);
 				ast_free(qp);
 			}
+			ast_log(LOG_WARNING, "Channel %s: Receive queue exceeds threshold\n", ast_channel_name(ast));
```
InterLinked1, line 598: Could we log the threshold value itself too here?
I know it's supposed to be static, but just in case somebody changes it

channels/chan_usrp.c
```diff
@@ -662,6 +666,7 @@ static int usrp_xwrite(struct ast_channel *ast, struct ast_frame *frame)
 		fr.delivery.tv_sec = 0;
 		fr.delivery.tv_usec = 0;
 		ast_queue_frame(ast, &fr);
+		ast_debug(1, "Channel %s: RX OFF\n", ast_channel_name(ast));
```
InterLinked1, line 669: Won't this debug message get triggered for every single frame processed?

channels/chan_usrp.c
KB4MDD: I added the threshold to the message.

channels/chan_usrp.c
KB4MDD: This only occurs when the rxkey variable counts down to 1.  The code that follows will decrement it to zero.  Previous code sets it to a number above 1 so that this action only occurs one time.

InterLinked1 (approved):

channels/chan_usrp.c
```diff
@@ -371,9 +371,11 @@ static int usrp_indicate(struct ast_channel *ast, int cond, const void *data, si
 	switch (cond) {
 	case AST_CONTROL_RADIO_KEY:
 		p->txkey = 1;
+		ast_debug(1, "Channel %s: ACRK TX ON.\n", ast_channel_name(ast));
```
Allan-N, line 374: What is ACRK ?

channels/chan_usrp.c
```diff
@@ -371,9 +371,11 @@ static int usrp_indicate(struct ast_channel *ast, int cond, const void *data, si
...
 		break;
 	case AST_CONTROL_RADIO_UNKEY:
 		p->txkey = 0;
+		ast_debug(1, "Channel %s: ACRUK TX OFF.\n", ast_channel_name(ast));
```
Allan-N, line 378: What is ARCUK ?

channels/chan_usrp.c
```diff
@@ -593,6 +595,7 @@ static int usrp_xwrite(struct ast_channel *ast, struct ast_frame *frame)
 				remque((struct qelem *) qp);
 				ast_free(qp);
 			}
+			ast_log(LOG_WARNING, "Channel %s: Receive queue exceeds the threshold of %lu\n", ast_channel_name(ast), USRP_VOICE_FRAME_SIZE);
```
Allan-N, line 598: Does it help knowing anything about how much we are exceeding the threshold ?

channels/chan_usrp.c
KB4MDD: The only part of that I understand is the K.

It ment something to someone else.

channels/chan_usrp.c
KB4MDD: Not really, this should never happen.  Sending UDP packets does not block.

If this occurs something bad has happened.

channels/chan_usrp.c
Allan-N: My first thought when I noticed these two messages was that one indicated keying (K, TX ON) and the other un-keying (UK, TX OFF).  Looking more carefully at the other drivers and the messages here are not unique to usrp.

channels/chan_usrp.c
Allan-N: ACRK == Keying, ARCUK == UnKeying ???

Allan-N (approved):

PR #423, opened 2024-11-10
Fix crash if "rpt.conf" cannot be loaded

Allan-N:
Note: the crash that prompted this fix was a typo in the category declaration for a node (e.g. "[1234]\(node-NOTmain)").

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -5511,7 +5511,10 @@ static int load_config(int reload)
 
 	cfg = ast_config_load("rpt.conf", config_flags);
 	if (!cfg) {
-		ast_log(LOG_NOTICE, "Unable to open radio repeater configuration rpt.conf.  Radio Repeater disabled.\n");
+		ast_log(LOG_ERROR, "Unable to open radio repeater configuration rpt.conf.  Radio Repeater disabled.\n");
+		return -1;
+	} else if (cfg == CONFIG_STATUS_FILEINVALID) {
+		ast_log(LOG_ERROR, "Errors detected in the radio repeater configuration rpt.conf.  Radio Repeater disabled.\n");
 		return -1;
 	}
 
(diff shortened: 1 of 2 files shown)


apps/app_rpt.c
```diff
@@ -5514,6 +5514,10 @@ static int load_config(int reload)
 		ast_log(LOG_NOTICE, "Unable to open radio repeater configuration rpt.conf.  Radio Repeater disabled.\n");
 		return -1;
 	}
+	if (cfg == CONFIG_STATUS_FILEINVALID) {
```
InterLinked1, line 5517: Not that it changes the code at all, but stylistically, in Asterisk modules this would be an else if chained off the previous if.

apps/app_rpt.c
```diff
@@ -5514,6 +5514,10 @@ static int load_config(int reload)
...
+		ast_log(LOG_NOTICE, "Errors detected in the radio repeater configuration rpt.conf.  Radio Repeater disabled.\n");
```
InterLinked1, line 5518: This should be a warning or even an error, not a notice

apps/app_rpt.c
```diff
@@ -6232,6 +6236,9 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		if (b1)
 			ast_shrink_phone_number(b1);
 		cfg = ast_config_load("rpt.conf", config_flags);
+		if (cfg == CONFIG_STATUS_FILEINVALID) {
```
InterLinked1, line 6239: Shouldn't there also be a warning here? If this happens, the config was valid at some point but is now invalid, which seems pretty bad

apps/app_rpt/rpt_config.c
```diff
@@ -733,6 +733,11 @@ void load_rpt_vars(int n, int init)
 		ast_log(LOG_NOTICE, "Unable to open radio repeater configuration rpt.conf.  Radio Repeater disabled.\n");
 		pthread_exit(NULL);
 	}
+	if (cfg == CONFIG_STATUS_FILEINVALID) {
```
InterLinked1, line 736: else if / warning or error, not notice

InterLinked1 (changes requested):

apps/app_rpt.c
Allan-N: Here, I'm not concerned.  No specific warning was being logged if the file was removed so I took the same approach if the configuration was modified and is no longer valid.  But, regardless, you will get the "Cannot find specified system node %s\n" log message issued at the end of the function.

InterLinked1 (approved):

KB4MDD (approved):
Looks good.

PR #432, opened 2024-12-07
Add support for a info message on the echolink connect page

firealarmss:
Adds a feature to display a set message (from the config file) on the echolink app when connected.


channels/chan_echolink.c
```diff
@@ -20,6 +20,9 @@
  * --------
  * 02/5/23 - Danny Lloyd, KB4MDD <kb4mdd@arrl.net>
  * corrected memory leaks
+ * --------
+ * 12/7/24 - Caleb, K4PHP <caleb.k4php@gmail.com>
```
InterLinked1, line 24: Generally speaking these types of comments are heavily discouraged nowadays. Authorship is shown in the commit history

channels/chan_echolink.c
```diff
@@ -1820,6 +1825,23 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
 			ast_log(LOG_WARNING, "Exceeded buffer size");
 			return;
 		}
+
+		if (instp->mymessage[0] != '\0') {
```
InterLinked1, line 1829: Add a comment in English explaining what you are trying to do here (convert all the "\\n" to "\n")
Also, I'm curious why you choose to modify the original buffer here, it seems like the memmove's will therefore only happen the first time? In that case why not choose to do that when the buffer is initialized to begin with? That might be a little bit more logical. Alternately, if you wanted to leave the original buffer intact, you could move the snprintf calls into the loop.

channels/chan_echolink.c
firealarmss: I followed your header format, fair assumption IMO; resolved

channels/chan_echolink.c
```diff
@@ -1820,6 +1822,25 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
 			ast_log(LOG_WARNING, "Exceeded buffer size");
 			return;
 		}
+
+		i += j;
+
+		if (instp->mymessage[0] != '\0') {
+			char *p = instp->mymessage;
+			while ((p = strstr(p, "\\n")) != NULL) {
+				*p = '\n';  // Replace the literal \n with a newline character
```
Allan-N, line 1831: The [Asterisk coding guidelines ](https://docs.asterisk.org/Development/Policies-and-Procedures/Coding-Guidelines/) include :
```
* Don't use C++ type (//) comments.
```
and we're trying to follow those conventions.

Allan-N (changes requested):

channels/chan_echolink.c
firealarmss: This should be resolved.

channels/chan_echolink.c
```diff
@@ -1820,6 +1822,25 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
 			ast_log(LOG_WARNING, "Exceeded buffer size");
 			return;
 		}
+
+		i += j;
+
+		if (instp->mymessage[0] != '\0') {
+			char *p = instp->mymessage;
```
KB4MDD, line 1829: I recommend moving your \n replacement to the store_config routine.  This would be around line 3847 where you store the message from the config file.  The replacement only needs to happen one time.  (The existing code will only do that replacement once, but we will use a few cycles testing the message every time we send the info message.)

Allan-N: @firealarmss What do you think about the requested change?  If OK, can you update your changes?

Murf: Would be very nice thanks :)

channels/chan_echolink.c
```diff
@@ -3818,6 +3839,13 @@ static int store_config(struct ast_config *cfg, char *ctg)
 		ast_copy_string(instp->myname, val, sizeof(instp->myname));
 	}
 
+	val = ast_variable_retrieve(cfg, ctg, "message");
+	if (!val) {
```
InterLinked1, line 3843: These 5 lines could be simplified to just:

`ast_copy_string(instp->mymessage, S_OR(val, ""), sizeof(instp->mymessage));`

configs/rpt/echolink.conf
```diff
@@ -21,6 +21,9 @@ height = 0							; 0=10 1=20 2=40 3=80 4=160 5=320 6=640 7=1280 8=2560 9=5120 (A
 gain = 0							; Gain in db (0-9)
 dir = 0								; 0=omni 1=45deg 2=90deg 3=135deg 4=180deg 5=225deg 6=270deg 7=315deg 8=360deg (Direction)
 
+; Message to show on the echolink connect page. Use \n for new lines.
```
InterLinked1, line 24: Now that I think about this, I think it's more common to allow the same config option to be specified multiple times, and have future usage append rather than replace (I believe I've seen this in a few other places). Not saying it *has* to be done that way, but it's a bit more conventional than having the user put \n in the config, and have it all on one line, especially if it's long.

Murf: Can this be added now anyway please? It can be fine tuned later.

Allan-N: >I think it's more common to allow the same config option to be specified multiple times, and have future usage append rather than replace (I believe I've seen this in a few other places).

Really?  I have not seen that usage in any of the .conf files I have encountered.  Also, how would this work with AMI's UpdateConfig?

InterLinked1: I can't find a good example now but there are some options like this. For example, in chan_dahdi, additional cadences are added by specifying `cadence=XXX` multiple times.
This was more of a musing but to clarify, I'm fine with it not being implemented this way, it was just a passing thought. The other comments are more important.

Allan-N (approved):
I'm OK taking the current changes AS-IS and take care of the added changes/enhancements in a separate PR.

channels/chan_echolink.c
mkmer: Should we make this change to all similar functions above and below this, or make this one just a bit different?

mkmer:
I was able to apply the suggested changes.

channels/chan_echolink.c
```diff
@@ -1822,25 +1822,6 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
...
-
-		if (instp->mymessage[0] != '\0') {
-			char *p = instp->mymessage;
-			while ((p = strstr(p, "\\n")) != NULL) {
-				*p = '\n'; /* Replace the literal \n with a newline character */
-				memmove(p + 1, p + 2, strlen(p + 2) + 1); /* Shift the string to remove the \n */
-				p++; 
-			}
-
-			j = snprintf(pkt + i, sizeof(pkt) - i, "%s\n\n", instp->mymessage);
```
Allan-N: Doesn't this code need to stay?

Allan-N (changes requested):

channels/chan_echolink.c
mkmer: Probably - Got a bit delete happy :)

mkmer: I believe this is back as it should be.

channels/chan_echolink.c
```diff
@@ -1820,7 +1822,20 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
 			ast_log(LOG_WARNING, "Exceeded buffer size");
 			return;
 		}
-		
+
+		i += j;
```
Allan-N, line 1826: What do you think about moving this `i += j;` line to inside the `if (instp->mymessage[0] != '\0') {}` below and removing the `i += j;` at the bottom of the "if"?  That way we only adjust "i" when needed.

mkmer: How about keeping track of `i` and removing the strlen() calls just below that?
Let me know what you think of the update.

channels/chan_echolink.c
```diff
@@ -1821,16 +1823,39 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
...
 			cp = ast_strdup((*(struct el_node **) nodep)->pvt->linkstr);
 			if (cp) {
-				i = strlen(pkt);
-				strncat(pkt + i, cp, sizeof(pkt) - i);
+				j = snprintf(pkt + i, sizeof(pkt) - i, "%s", cp);
 				ast_free(cp);
+
+				if (j >= sizeof(pkt) - i) {
+					ast_log(LOG_WARNING, "Exceeded buffer size");
+					return;
```
mkmer, line 1854: Wondering if `return` is the right thing to do when the buffer is exceeded.  Should we just keep going and send a truncated message + log a warning?

mkmer: Maybe use a `goto` to log once, then send what we have

KB4MDD: I think starting at line 1828, the following returns should send the text that was built.

The return does not hurt the operation, we just stop sending status updates to the echolink clients.

Allan-N: An alternative, maybe update the "Exceeded buffer size" message(s) to provide a bit more context (e.g. "Buffer size exceeded: \"message\" too long", \"linked node list\" too long, etc)

channels/chan_echolink.c
```diff
@@ -1813,24 +1815,47 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
...
 		if (j >= sizeof(pkt) - i) {
 			ast_log(LOG_WARNING, "Exceeded buffer size");
-			return;
+			goto send_message;
 		}
 		
+		i += j;
+		
+		if (instp->mymessage[0] != '\0') {
+			j = snprintf(pkt + i, sizeof(pkt) - i, "%s\n\n", instp->mymessage);
```
Allan-N, line 1829: I like the i += snprintf() coding you used below.  You should do the same here (and elsewhere)

channels/chan_echolink.c
```diff
@@ -1813,24 +1815,47 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
...
+			}
+			
+			i += j;
+		}
+
 		if ((*(struct el_node **) nodep)->pvt && (*(struct el_node **) nodep)->pvt->linkstr) {
-			i = strlen(pkt);
-			strncat(pkt + i, "Systems Linked:\r", sizeof(pkt) - i);
+			i += snprintf(pkt + i, sizeof(pkt) - i, "Systems Linked:\r");
+
```
Allan-N, line 1841: I'd drop this blank line

channels/chan_echolink.c
```diff
@@ -1813,24 +1815,47 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
 		i = snprintf(pkt, sizeof(pkt), "oNDATA\rWelcome to Allstar Node %s\r", instp->astnode);
```
Allan-N, line 1815: I like the i += snprintf() coding you used below. You should do the same here (and elsewhere)

channels/chan_echolink.c
```diff
@@ -1813,24 +1815,47 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
...
+
+			if (j >= sizeof(pkt) - i) {
+				ast_log(LOG_WARNING, "Exceeded buffer size");
+				goto send_message;
+			}
+			
+			i += j;
+		}
+
 		if ((*(struct el_node **) nodep)->pvt && (*(struct el_node **) nodep)->pvt->linkstr) {
```
Allan-N, line 1839:
```
		cp = (*(struct el_node **) nodep)->pvt ? (*(struct el_node **) nodep)->pvt->linkstr : NULL;
 		if (cp) {
```
and I don't see any need for the strdup()

Allan-N (changes requested):

channels/chan_echolink.c
```diff
@@ -218,6 +218,7 @@ do not use 127.0.0.1
 
 #define	DELIMCHR ','
 #define	QUOTECHR 34
+#define EL_INIT_BUFFER = sizeof("oNDATA\rWelcome to Allstar Node \rEcholink Node %s\rNumber \r \rSystems Linked:\r) + 300 /* initial string buffer size for ast_str buffer */
```
Allan-N: Change the `%s` to `123456`

channels/chan_echolink.c
```diff
@@ -1799,42 +1802,39 @@ static void lookup_node_callsign(const void *nodep, const VISIT which, void *clo
...
+
+		ast_str_set(pkt, 0, "oNDATA\rWelcome to Allstar Node %s\r", instp->astnode);
+		ast_str_append(pkt, 0, "Echolink Node %s\rNumber %u\r \r", instp->mycall, instp->mynode);
+
+		if (instp->mymessage[0] != '\0') {
+			ast_str_append(pkt, 0, "%s\n\n", instp->mymessage);
 		}
-		
+
 		if ((*(struct el_node **) nodep)->pvt && (*(struct el_node **) nodep)->pvt->linkstr) {
```
Allan-N, line 1825: Put the `cp = ...` assignment here, then check for `if (cp) {`, and then you don't need the assignment/check below

channels/chan_echolink.c
```diff
@@ -1799,42 +1802,39 @@ static void lookup_node_callsign(const void *nodep, const VISIT which, void *clo
...
+				ast_str_append(pkt, 0, "%s", cp);
+				}
 		}
-		sendto(instp->audio_sock, pkt, strlen(pkt), 0, (struct sockaddr *) &sin, sizeof(sin));
+
+		sendto(instp->audio_sock, ast_str_buffer(pkt), ast_str_len(pkt), 0, (struct sockaddr *) &sin, sizeof(sin));
 		
 		instp->tx_ctrl_packets++;
 		(*(struct el_node **) nodep)->tx_ctrl_packets++;
+		ast_free(pkt);
```
Allan-N, line 1837: This line could go right after the sendto()

channels/chan_echolink.c
```diff
@@ -1801,60 +1802,40 @@ static void lookup_node_callsign(const void *nodep, const VISIT which, void *clo
...
-		}
-		offset += snprintf(pkt + offset, sizeof(pkt) - offset, "Echolink Node %s\rNumber %u\r \r", instp->mycall, instp->mynode);
-		if (offset >= sizeof(pkt)) {
-			ast_log(LOG_WARNING, "Exceeded buffer size");
-			goto send_message;
-		}
 
-		if (instp->mymessage[0] != '\0') {
-			offset += snprintf(pkt + offset, sizeof(pkt) - offset, "%s\n\n", instp->mymessage);
+		ast_str_set(pkt, 0, "oNDATA\rWelcome to Allstar Node %s\r", instp->astnode);
```
InterLinked1, line 1818: Should it really say "Allstar node"... Echolink isn't specific to ASL, right?
Somebody could use this outside of ASL and not want that message there?

channels/chan_echolink.c
```diff
@@ -1801,60 +1802,40 @@ static void lookup_node_callsign(const void *nodep, const VISIT which, void *clo
...
-				goto send_message;
-			}
-
-			cp = (*(struct el_node **) nodep)->pvt ? (*(struct el_node **) nodep)->pvt->linkstr : NULL;
-			if (cp) {
-				offset += snprintf(pkt + offset, sizeof(pkt) - offset, "%s", cp);
-
-				if (offset >= sizeof(pkt)) {
-					ast_log(LOG_WARNING, "Exceeded buffer size");
+			if (cp = (*(struct el_node **) nodep)->pvt ? (*(struct el_node **) nodep)->pvt->linkstr : NULL){
```
InterLinked1, line 1828: space before {
Also, could you do the cast outside of the conditional? Maybe at the top of the whole function, especially given I see many, many casts all right here for the same thing. This is also very hard to read.

channels/chan_echolink.c
```diff
@@ -1799,42 +1802,39 @@ static void lookup_node_callsign(const void *nodep, const VISIT which, void *clo
...
-			if (cp) {
-				i = strlen(pkt);
-				strncat(pkt + i, cp, sizeof(pkt) - i);
-				ast_free(cp);
-			}
+			ast_str_append(pkt, 0, "Systems Linked:\r");
+
+			if (cp = (*(struct el_node **) nodep)->pvt ? (*(struct el_node **) nodep)->pvt->linkstr : NULL){
+				ast_str_append(pkt, 0, "%s", cp);
+				}
```
InterLinked1, line 1830: line indented too far?

channels/chan_echolink.c
```diff
@@ -217,6 +218,7 @@ do not use 127.0.0.1
 
 #define	DELIMCHR ','
 #define	QUOTECHR 34
+#define EL_INIT_BUFFER = sizeof("oNDATA\rWelcome to Allstar Node \rEcholink Node %s\rNumber \r \rSystems Linked:\r") + 300 /* initial string buffer size for ast_str buffer */
```
InterLinked1, line 221: ast_str will dynamically expand... we don't need to be super precise about how big it is, and this duplicates the phrasing from send_info. I think much of the tree will just start with 256 or something like that

channels/chan_echolink.c
```diff
@@ -1799,42 +1802,39 @@ static void lookup_node_callsign(const void *nodep, const VISIT which, void *clo
 static void send_info(const void *nodep, const VISIT which, const int depth)
 {
 	struct sockaddr_in sin;
-	char pkt[5120], *cp;
+	struct ast_str *pkt;
+	char *cp;
 	struct el_instance *instp = (*(struct el_node **) nodep)->instp;
-	int i, j;
+
+	pkt = ast_str_create(EL_INIT_BUFFER);
```
InterLinked1, line 1809: This can fail

channels/chan_echolink.c
```diff
@@ -1799,42 +1802,39 @@ static void lookup_node_callsign(const void *nodep, const VISIT which, void *clo
...
 
 	if ((which == leaf) || (which == postorder)) {
```
InterLinked1, line 1812: blank line

channels/chan_echolink.c
mkmer: Do we want it to "make sense" why we started with this "size" or just magic number it?

channels/chan_echolink.c
mkmer: It is showing the Allstar node (%s) that your connected to.... I don't think it anything else makes sense here...

KB4MDD: I would prefer to leave Allstar, but if it is removed, we will need that configurable.  

Also, I think it should be spelled "AllStar".

channels/chan_echolink.c
KB4MDD: You have to pass an initial length.  Why not pick a reasonable value.

mkmer:
I was searching this code base an i'll be... they used ast_str...
https://github.com/AllStarLink/app_rpt/blob/da38bf2e4a88fbe3454bf89b62a93885e8de086e/channels/chan_echolink.c#L2350
and
https://github.com/AllStarLink/app_rpt/blob/da38bf2e4a88fbe3454bf89b62a93885e8de086e/channels/chan_echolink.c#L2473

Thing is:
-  not checking for failure and should (right?),
-  not freeing the pointer on exit in either case?

If you agree this is "wrong", I'll create a PR to fix it.

InterLinked1:
> I was searching this code base an i'll be... they used ast_str...
> 
> https://github.com/AllStarLink/app_rpt/blob/da38bf2e4a88fbe3454bf89b62a93885e8de086e/channels/chan_echolink.c#L2350
> 
> 
> and
> https://github.com/AllStarLink/app_rpt/blob/da38bf2e4a88fbe3454bf89b62a93885e8de086e/channels/chan_echolink.c#L2473
> 
> Thing is:
> 
> * not checking for failure and should (right?),
> * not freeing the pointer on exit in either case?
> 
> If you agree this is "wrong", I'll create a PR to fix it.

alloca allocates on the stack, not from the heap, so no need to free. It can be slightly more dangerous so usually you want to be more careful with the size. alloca should always succeed unless it crashes the program, you can't really handle it.

channels/chan_echolink.c
```diff
@@ -899,7 +902,7 @@ static int rtcp_make_sdes(unsigned char *pkt, int pkt_len, const char *call, con
 	ap += l;
 
 	if (astnode) {
-		snprintf(line, EL_CALL_SIZE + EL_NAME_SIZE, "Allstar %s", astnode);
+		snprintf(line, EL_CALL_SIZE + EL_NAME_SIZE, "AllStar %s", astnode);
```
Allan-N, line 905: Maybe capture the length returned by snprintf() and use that instead of having to call strlen() ?

channels/chan_echolink.c
```diff
@@ -1601,14 +1604,14 @@ static int el_text(struct ast_channel *ast, const char *text)
 			j = 0;
 			k = 0;
 			for (x = 0; x < i; x++) {
-				/* Process allstar node numbers - skip over those that begin with '3' which are echolink */
+				/* Process AllStar node numbers - skip over those that begin with '3' which are echolink */
 				if ((*(strs[x] + 1) != '3')) {
 					if (strlen(pkt + k) >= 32) {
 						k = strlen(pkt);
```
Allan-N, line 1610: strncat() expects you to pass the size of the buffer.  In this case, these two lines should be :
```
strncat(pkt, "\r ", pkt_len);
```
with no need for "k"

channels/chan_echolink.c
```diff
@@ -1601,14 +1604,14 @@ static int el_text(struct ast_channel *ast, const char *text)
...
 						strncat(pkt, "\r    ", pkt_len - k);
 					}
 					if (!j++) {
-						strncat(pkt, "Allstar:", pkt_len - strlen(pkt));
+						strncat(pkt, "AllStar:", pkt_len - strlen(pkt));
```
Allan-N, line 1614: Again, the length of the string buffer is `pkt_len`, not `pkt_len - strlen()`

channels/chan_echolink.c
Allan-N: Alternatively, you could use `strncat(pkt + k, ..., pkt_len - k)`

channels/chan_echolink.c
```diff
@@ -1799,40 +1802,34 @@ static void lookup_node_callsign(const void *nodep, const VISIT which, void *clo
...
-			i = strlen(pkt);
-			strncat(pkt + i, "Systems Linked:\r", sizeof(pkt) - i);
-			cp = ast_strdup((*(struct el_node **) nodep)->pvt->linkstr);
-			if (cp) {
-				i = strlen(pkt);
-				strncat(pkt + i, cp, sizeof(pkt) - i);
-				ast_free(cp);
-			}
+
+		if (cp = cp = (*(struct el_node **) nodep)->pvt ? (*(struct el_node **) nodep)->pvt->linkstr : NULL) {
```
Allan-N: cp = cp = ?

Allan-N (changes requested):

Allan-N (approved):

mkmer (approved):

mkmer (approved):

KB4MDD (approved):

Allan-N (approved):

mkmer (approved):

channels/chan_echolink.c
Allan-N: I also agree that the message should say "AllStar node ######".  Remember, we taking connections from the EchoLink network and connecting them to the AllStar network.  The message is to inform those on the EchoLink network that they have been connected to the AllStar network.

Also, the [AllStar] node number being reported is associated with the `astnode` variable from the `echolink.conf` file.

InterLinked1: I think making it configurable is always good, but if that's out of scope of this PR, that can be done separately. I think that text was there before as it is anyways.

I say that because (correct me if I'm wrong), somebody could be using this module in a separate network (or not in a network even?) that isn't Allstar, connected to Echolink. It's not like `app_rpt` is only for AllStarLink users. Yes, I realize, in practice probably 99.99% of the user base might be, but there may be people experimenting on the edge for whom this might not apply.

Allan-N: Sure, we can make the string configurable.  But, let's leave that, and any other changes associated with EchoLink <--> EchoLink connections, as an issue for "later".

InterLinked1 (approved):

PR #433, opened 2024-12-12
[ASL2 PATCH] Update morse code character timing

Allan-N:
An ASL2 change (from https://github.com/AllStarLink/ASL-Asterisk/pull/19) was not fully merged into the ASL3 source base.  That PR included an correction to the morse code character timing, specifically the inter character delay.  This change updates the timing to match what's commonly understood (the space between two letters is equal to three dots).

### DIFF
--- a/apps/app_rpt/rpt_channel.c
+++ b/apps/app_rpt/rpt_channel.c
@@ -341,9 +341,9 @@ int send_morse(struct ast_channel *chan, char *string, int speed, int freq, int 
 
 	/* Establish timing releationships */
 
-	dashtime = 3 * dottime;
+	dashtime = dottime * 3;
 	intralettertime = dottime;
-	interlettertime = dottime * 4;
+	interlettertime = dottime * 3;
 	interwordtime = dottime * 7;
 
 	for (; (*string) && (!res); string++) {


encbar5 (changes requested):
@Allan-N A clarification - from looking at the code change, this is not trying to adjust the relative length of the "Dah" but rather the space between two letters, here called "interlettertime".

I disagree with this code change for the following reasons:

The [international recommendation for morse code](https://www.itu.int/rec/R-REC-M.1677-1-200910-I/)  lists the following spacings
- a dash is equal to three dots
- the space between the signals forming the same letter is equal to one dot
- the space between two letters is equal to three dots
- the space between two words is equal to seven dots

The variable `interlettertime` seems like it is representing the spacing between letters, but it actually is not. 

If you look at line 395, the actual spacing inserted between letters is `interlettertime - intralettertime`. Before the changes `interlettertime` is `4` and `intralettertime` is `1`, so the difference is `3` "Dits" which is correct.

I believe the reason it is being done this way is that the logic is always inserting the `intralettertime` after every letter, no matter if it is at the end of the word or not. See line 388. 

Therefore this change would introduce a spacing between words of two dots instead of three.

This change should either be reverted, or if anything the variable `interlettertime` could be renamed to indicate its actual usage.

Allan-N:
@encbar5 Thank you for your comments AND you are correct. This PR will be closed as not-to-be-fixed.

Allan-N:
Not sure where I was off when I looked at the changes earlier today.  This change is NOT needed.

Allan-N:
Hmmm... The idea of changing the interlettertime to be "3" kinda makes sense.  And, with that we only need to change the one place it's used to not subtract intralettertime.  But, the end result would be no change to the generated morse code audio so we might as well leave the code alone.

Allan-N:
Nope.  I think the change *IS* correct.  Why?  After each "Dit" or "Dah" we add the `intralettertime`.  Between each character we also add the `interlettertime - intralettertime` time.  With the pre-patch values that would be the equivalent of 1 + 4 - 1 = "4" dot times.  We want "3".  If we change `interlettertime` to 3 then we would have the expected 1 + 3 - 1 = "3" dot times.

encbar5:
@Allan-N Ah yes, sorry, you have the correct logic.

encbar5 (approved):

JimZAH:
Just a comment, thanks for adding this patch back in. I was going to raise a PR but haven't had the time (due to a house move).

KB4MDD (approved):

PR #444, opened 2025-01-08
Address channel crashing on rapid connect/disconnect

mkmer:
Addressing issue #422 
1 - remove channel locks around ```ast_sendtext()```.  This made triggering the crash much more difficult (aka things are executing faster) but did not eliminate it.
2 - remove "extra" send_newkey() -> it's extra, and everything appears to work after removal.
3 - test channel state still up before sending text messages.  I can no longer crash asterisk after this addition.


mkmer:
This may or may not be the "right way" to fix the crash.  In my test case it makes it impossible to crash.

mkmer:
Forget it... still possible to crash (just harder)

PR #445, opened 2025-01-08
Address channel crashing on rapid connect/disconnect 

mkmer:
Addressing https://github.com/AllStarLink/app_rpt/issues/422
There appears to be a race condition between  ```process_link_channels``` https://github.com/AllStarLink/app_rpt/blob/5101b1bb1338144ecd2ea00f7d4c7ffd18104a25/apps/app_rpt.c#L4495-L4497 detecting and closing a channel while  ```rpt_exec``` is still mid setup on the connection. https://github.com/AllStarLink/app_rpt/blob/5101b1bb1338144ecd2ea00f7d4c7ffd18104a25/apps/app_rpt.c#L6154 is still "starting up" the channel.

Simply allowing ```process_link_channels``` to deal with sending the new key will solve this issue, but may introduce other issues that I currently do not observe.

Alternately, we could make sure the channel is not null in the ```send_newkey()``` function, but I'm not sure locking the channel prevents Asterisk from cleaning in up (it just gives us a warning and cleans up anyway from what I can tell).

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -6751,11 +6751,6 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 			donodelog_fmt(myrpt,"LINK%s,%s", l->phonemode ? "(P)" : "", l->name);
 		}
 		doconpgm(myrpt, l->name);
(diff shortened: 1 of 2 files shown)


mkmer:
Well.. I guess this doesn't send audio - just like the original bug :(

PR #446, opened 2025-01-08
Test ast channel state and address crash

mkmer:
Issue: https://github.com/AllStarLink/app_rpt/issues/422
Final attempt - this appears to cover the race and prevent crashes all while not breaking the crazy hard to understand NEWKEY logic

If anything, I hope it helps with alternate solutions.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -6144,7 +6144,7 @@ static void answer_newkey_helper(struct rpt *myrpt, struct ast_channel *chan, in
 	rpt_mutex_lock(&myrpt->blocklock);
 	if (ast_channel_state(chan) != AST_STATE_UP) {
 		ast_answer(chan);
-		if (!phone_mode) {
+		if (!phone_mode && (ast_channel_state(chan) == AST_STATE_UP)) {
 			send_newkey(chan);
 		}
 	}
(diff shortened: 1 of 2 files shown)


InterLinked1:
I don't think this is quite right. For one, what's the basis of requiring the channel have to be answered in order to do the logic? It may prevent the crashes observed but it seems likely to also prevent it from running when it should.

(If you're checking if the state is up to see if it exists, that isn't the way to do that, that just tells you if answer supervision has already been returned to the caller)

mkmer:
I was thinking if no one is there, why try to send text messages... it is actually answered just before we try to send the text messages, but if the other end immediately hangs up AND the channel logic happens to acknowledge the hang up before this logic finished (probably a race to locking/unlocking) we end up trying to send to a disposed of channel.
Somehow we need to deal with the race when connections are flaky.

mkmer:
I feel like the answer may be to kill the thread for:
 https://github.com/AllStarLink/app_rpt/blob/5101b1bb1338144ecd2ea00f7d4c7ffd18104a25/apps/app_rpt.c#L6154 
when we get here:
https://github.com/AllStarLink/app_rpt/blob/5101b1bb1338144ecd2ea00f7d4c7ffd18104a25/apps/app_rpt.c#L4495

If it's not finished when we hang up, why keep running?
I'm not sure how to do that...

mkmer:
I have found a "better" way - will submit a different PR for discussion.

PR #449, opened 2025-01-10
app_rpt: Exit rpt_exec if call can not be answered

mkmer:
app_rpt: Exit rpt_exec if call can not be answered
Clean up link status as we could not start the call:  Prevents periodic_process_link() from forcing "new key from 2 to 0"
Fixes: #422

I found a few answer examples in the Asterisk base code - it looks like we need to be sure to succeed at answering the phone to be sure the channel code is not cleaning up.
Also noticed that we set up the channel before we actually answer the call, so if it fails, reset the link registers ```l->```.

I can no longer force a crash with the rapid connect/disconnect.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -6737,7 +6737,17 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		gettimeofday(&myrpt->lastlinktime, NULL);
 		rpt_mutex_lock(&myrpt->blocklock);
 		if (ast_channel_state(chan) != AST_STATE_UP) {
-			ast_answer(chan);
+			if (ast_answer(chan)){
+				ast_debug(1, "Stopped PBX on %s, state: %i :Could not answer\n", ast_channel_name(chan),ast_channel_state(chan));
+				l->connected = 0;
+				l->thisconnected = 0;
+				l->newkey = 0;
+				l->newkeytimer = 0;
+				rpt_mutex_unlock(&myrpt->blocklock);
+				rpt_mutex_unlock(&myrpt->lock);
+				ast_channel_pbx_set(chan, NULL);
+				pthread_exit(NULL); 
+
 			if (l->name[0] > '9') {
 				if (ast_safe_sleep(chan, 500) == -1) {
 					return -1;


mkmer:
While this PR does fix the problem I've been seeing, it may make for a more robust solution to answer the call BEFORE we create the link https://github.com/AllStarLink/app_rpt/blob/5101b1bb1338144ecd2ea00f7d4c7ffd18104a25/apps/app_rpt.c#L6666-L6671

See the comment indicates we've established a call, when in reality it's only ringing at this point and no one has picked it up yet.
So we create all of the link data as if it were picked up, but the other end can still hang up before we answer.

mkmer:
This may be easier to communicate the alternate method:
https://github.com/AllStarLink/app_rpt/compare/master...mkmer:app_rpt:Alternate-issue-442-fix

While it's bypasses a lot more code - It seems like it's the more correct way to handle setting up phone calls?

InterLinked1:
> I found a few answer examples in the Asterisk base code - it looks like we need to be sure to succeed at answering the phone to be sure the channel code is not cleaning up.

Where are you seeing this? There is absolutely *no* requirement that channels be answered in Asterisk.

If that is the basis of this logic, I think it is merely coincidence / addressing a symptom that this has resolved the issue in this case.

mkmer:
Yah - that's the fun part :( 
I tried doing it with a macro on my personal node but the macro doesn't seem to go "fast" enough.
I am hitting connect and disconnect on allscan page as quickly as possible - sometimes the client doesn't even recognize the connect event.  
It usually takes 4 or 5 times minimum before I get a crash.  First you should see a few couldn't send NEWKEY errors, and 2 to 0 warnings.

mkmer:
Here are the links for what I was "cribbing" from:
Here are a few I checked out...

https://github.com/asterisk/asterisk/blob/48df8b63759995e2cddc2331ec344b79/apps/app_privacy.c#L108-L110

https://github.com/asterisk/asterisk/blob/4ef68fbce8df8b63759995e2cddc2331ec344b79/apps/app_read.c#L248-L253

https://github.com/asterisk/asterisk/blob/4ef68fbce8df8b63759995e2cddc2331ec344b79/apps/app_talkdetect.c#L125-L131

I really don't have a very good understanding of what is "right".  Just trying a few things that look like they might be right.

mkmer:
OK, Maybe what I have in the alternate helps solve 2 issues: 
1) it doesn't initialize the link data if we can't answer the call (fixing to 2 to 0 black magic)
2) It sets the channel reference count before we start using trying to use it, then references it when done fiddling with it.

I think 2 is what you were referring to in previous conversations - I'll be happy to test whatever you come up with :)

Maybe we only need 2 to fix the crash.

mkmer:
Sharing one more log related to this (using the alternate code).  It shows the channel is in state 4 (ringing), we answer it and the answer function returns non 0 (I know this because it executed the new disconnect code "could not answer"), but the channel moves to 6(AST_STATE_UP).
```
[2025-01-10 17:39:04.607] DEBUG[1758988][C-00000059] app_rpt.c: Ansering the call
[2025-01-10 17:39:04.607] DEBUG[1758988][C-00000059] chan_iax2.c: Answering IAX2 call
[2025-01-10 17:39:04.960] DEBUG[1750536][C-00000059] chan_iax2.c: Immediately destroying 1727, having received hangup
[2025-01-10 17:39:04.960] DEBUG[1758988][C-00000059] app_rpt.c: Stopped PBX on IAX2/149.154.11.243:24241-1727, state: 6 :Could not answer
[2025-01-10 17:39:04.960] DEBUG[1758988][C-00000059] pbx.c: Spawn extension (radio-secure,287893,1) exited non-zero on 'IAX2/149.154.11.243:24241-1727'
[2025-01-10 17:39:04.960] DEBUG[1758988][C-00000059] channel.c: Soft-Hanging (0x10) up channel 'IAX2/149.154.11.243:24241-1727'
[2025-01-10 17:39:04.960] DEBUG[1758988][C-00000059] channel.c: Channel 0x7f6d8800dba0 'IAX2/149.154.11.243:24241-1727' hanging up.  Refs: 2
[2025-01-10 17:39:04.960] DEBUG[1758988][C-00000059] chan_iax2.c: We're hanging up IAX2/149.154.11.243:24241-1727 now...
[2025-01-10 17:39:04.960] DEBUG[1758988][C-00000059] chan_iax2.c: Really destroying IAX2/149.154.11.243:24241-1727 now...
[2025-01-10 17:39:04.960] DEBUG[1758988][C-00000059] chan_iax2.c: schedule decrement of callno used for 149.154.11.243 in 60 seconds
[2025-01-10 17:39:04.960] DEBUG[1758988][C-00000059] channel.c: Channel 0x7f6d8800dba0 'IAX2/149.154.11.243:24241-1727' destroying

```
My point here is that the answer function knows something about the channel as it is return non 0. 
It's a lot to dig through.  Hopefully the debug helps you visualize the logic flow.

mkmer:
This will not be merged - more of a way to communicate my "ideas" - mostly bad ones ;)

mkmer:
@Allan-N 
> @mkmer Can I ask how you are exercising the rapid connects/disconnects? Are you using a single server? multiple servers? have you automated this with code/scripts/...?

I finally have a AMI script that can force the issue.  Looks like this:
Just 0.06 seconds between connect and disconnect messages will get every other connection.

```
        action = Action('Command\nCommand: rpt cmd 29972 ilink 13 287890',)
        print(f"Sending Action connect ")
        voter = client.send_action(action)  # get the AMI voter details
        response = voter.response
        print(f"Repsonse {response}")
        time.sleep(0.06)
        action = Action('Command\nCommand: rpt cmd 29972 ilink 11 287890',)
        print(f"Sending Action disconnect ")
        voter = client.send_action(action)  # get the AMI voter details
        response = voter.response
        print(f"Repsonse {response}")


        time.sleep(1)
```

This is using the Python AMI library, but you could get there using the Allmon backend stuff too.

PR #451, opened 2025-01-11
app_rpt: Don't hard hangup channels currently executing PBX.

InterLinked1:
Resolves: #422


mkmer:
Running it - I get these warning:
```
[2025-01-11 07:10:55.853] WARNING[1800895]: app_rpt.c:4149 safe_hangup: Channel IAX2/149.154.11.243:24439-11626 still has a PBX, requesting hangup for it
[2025-01-11 07:10:56.415] WARNING[1800877]: app_rpt.c:5770 rpt_master: rpt_thread restarted on node 287893
[2025-01-11 07:11:24.458] WARNING[1801040][C-0000002d]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text IAX2/149.154.11.243:24439-11615 on !NEWKEY1!
[2025-01-11 07:11:24.459] WARNING[1801040][C-0000002d]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text IAX2/149.154.11.243:24439-11615 on !NEWKEY1!
[2025-01-11 07:11:26.200] WARNING[1801037]: app_rpt.c:3102 periodic_process_links: 0x7fefc00202a0 newkeytimer expired on connected node, setting newkey from 2 to 0.
[2025-01-11 07:11:31.581] WARNING[1801037]: app_rpt.c:4149 safe_hangup: Channel IAX2/149.154.11.243:24439-8232 still has a PBX, requesting hangup for it
[2025-01-11 07:11:32.418] WARNING[1800877]: app_rpt.c:5770 rpt_master: rpt_thread restarted on node 287893
[2025-01-11 07:11:46.419] WARNING[1800877]: app_rpt.c:5770 rpt_master: rpt_thread restarted on node 287893
[2025-01-11 07:12:16.422] WARNING[1800877]: app_rpt.c:5770 rpt_master: rpt_thread restarted on node 287893
```
First part is close to the scenario - newkey fails are a sign your "close".
PROBLEM? -> Now for every disconnect I receive the restart warning.

I'm fairly sure this one hit the spot (see safe hangup is in between "send text" messages.  

```
[2025-01-11 07:18:28.399] WARNING[1802771][C-00000023]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text !NEWKEY1! on IAX2/149.154.11.243:24139-7620
[2025-01-11 07:18:28.401] WARNING[1802771][C-00000023]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text !NEWKEY1! on IAX2/149.154.11.243:24139-7620
[2025-01-11 07:18:30.192] WARNING[1802733]: app_rpt.c:3102 periodic_process_links: 0x7f81b00026c0 newkeytimer expired on connected node, setting newkey from 2 to 0.
[2025-01-11 07:18:45.693] WARNING[1802788][C-00000024]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text !NEWKEY1! on IAX2/149.154.11.243:24139-2382
[2025-01-11 07:18:45.693] WARNING[1802733]: app_rpt.c:4149 safe_hangup: Channel IAX2/149.154.11.243:24139-2382 still has a PBX, requesting hangup for it
[2025-01-11 07:18:45.694] WARNING[1802788][C-00000024]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text !NEWKEY1! on IAX2/149.154.11.243:24139-2382
```

I am not able to crash so far.

apps/app_rpt.c
```diff
@@ -3226,7 +3226,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 		if ((!l->chan) && (!l->retrytimer) && l->outbound && (l->retries++ < l->max_retries) && (l->hasconnected)) {
 			if (l->chan)
 				ast_hangup(l->chan);
-			l->chan = 0;
+			l->chan = NULL;
```
mkmer, line 3229: I was going to ask about this - there are mixed instances where `l->chan = 0` and others where `l->chan = NULL`.  Maybe we should make them consistent.  0 != NULL , but I believe if tests give the same result.

apps/app_rpt.c
```diff
@@ -4140,46 +4140,73 @@ static inline void free_frame(struct ast_frame **f)
...
-			if (l->chan) {
-				ast_hangup(l->chan);
-			}
-			l->chan = NULL;
 			rpt_mutex_lock(&myrpt->lock);
+			hangup_link_chan(l);
+			rpt_mutex_unlock(&myrpt->lock);
 			return;
 		}
 		if (l->outbound && (l->retries++ < l->max_retries) && (l->hasconnected)) {
```
mkmer, line 4197: I had a comment on this line in one of the threads -> looks a bit fishy that `l->hasconnected` must be true/1 to pass the if, then we set it to 1 again?  While it doesn't hurt anything, it makes one question if one of them is not "right" (maybe the intended test was `l->connected`)
I don't think its directly an issue at this point.

apps/app_rpt.c
```diff
@@ -4205,10 +4233,12 @@ static void remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 	free_frame(&l->lastf1);
 	free_frame(&l->lastf2);
 	/* hang-up on call to device */
-	ast_hangup(l->chan);
+	hangup_link_chan(l);
```
mkmer, line 4236: Do we need `rpt_mutex_lock` / unload around this one?  Looks like all other `hangup_link_chan` have it.

mkmer:
I think the restart source is here:
https://github.com/AllStarLink/app_rpt/blob/68cf227a5f68fc4cfa56d16aae39fe8fb0bddd03/apps/app_rpt.c#L4957

I'm not sure why it's different now, the "sunny day" path is still and `ast_hangup` like it was before...

InterLinked1:
> I am not able to crash so far.

Yup, that's really the only goal of this PR. The newkeytimer thing may be a separate thing we want to resolve, but the goal here is to ensure it does not cause a crash, and to ensure no invalid memory accesses are made, like was likely happening before.

> I'm not sure why it's different now, the "sunny day" path is still an ast_hangup like it was before...

I think those are for the node channels, the channel executing `Rpt()` isn't technically a node channel, it's just one of the "links" that's ephemerally connected to it. That is the channel that was getting pulled out unexpectedly here.

InterLinked1:
> PROBLEM? -> Now for every disconnect I receive the restart warning and all links are disconnected (not just the guy hanging up)

This is expected, I think, the code that was triggering the crash was part of code to terminate the node, so all links would get disconnected.

The entire node didn't restart previously only because Asterisk itself crashed before that could happen :)

Obviously, we will want to prevent the restarts / newkeytimer issue, but I think that would be a followup PR to this one.

apps/app_rpt.c
InterLinked1: Yup, good catch. I actually put it around the other channel by mistake, I'll switch it around.

apps/app_rpt.c
InterLinked1: I agree, that doesn't mean it was write, Jim was human, he could have made a typo :)

I think we should definitely note this to look at for the follow-up PR, I'll add a comment as well on this line to come back to it.

Allan-N:
>> PROBLEM? -> Now for every disconnect I receive the restart warning and all links are disconnected (not just the guy hanging up)

>Obviously, we will want to prevent the restarts / newkeytimer issue, but I think that would be a followup PR to this one.

One node disconnecting should not result in a crash ... but it should also not result in others being disconnected.

InterLinked1:
> > > PROBLEM? -> Now for every disconnect I receive the restart warning and all links are disconnected (not just the guy hanging up)
> 
> > Obviously, we will want to prevent the restarts / newkeytimer issue, but I think that would be a followup PR to this one.
> 
> One node disconnecting should not result in a crash ... but it should also not result in others being disconnected.

It's a bit odd, because the logic seems to be intentionally written that way. If one of the channels is hung up, then abort the node (I'm looking at `process_link_channels`, and by extension its use in `rpt`). Are you saying this isn't what's happened in the past? Or perhaps it never happened because the newkeytimer issue never happened, but the same thing would've happened if it had.

So then what it is the appropriate thing to do, continue on without those channels? If the above theory is true, it might make sense to leave the existing logic alone and fix the newkeytimer issue directly so it doesn't trigger this path to begin with.

apps/app_rpt.c
```diff
@@ -4305,14 +4337,16 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
...
 		rpt_mutex_lock(&myrpt->lock);
 		if (who == l->chan) {	/* if it was a read from rx */
 			struct ast_frame *f;
 			rpt_mutex_unlock(&myrpt->lock);
 			f = ast_read(l->chan);
 			if (!f) {
+				ast_debug(3, "Failed to read frame on %s, must've hung up\n", ast_channel_name(l->chan));
 				remote_hangup_helper(myrpt, l);
-				break;
+				return -1;
```
mkmer, line 4349: I think this is the guilty code causing the repeater to restart.  We were NOT setting "toexit" here, so I think it probably should be `return 0;` or `break`.

apps/app_rpt.c
```diff
@@ -4494,22 +4525,20 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 				}
 				if (f->subclass.integer == AST_CONTROL_HANGUP) {
 					ast_frfree(f);
+					ast_debug(3, "Received hangup frame on %s\n", ast_channel_name(l->chan));
 					remote_hangup_helper(myrpt, l);
-					break;
+					return -1;
```
mkmer, line 4530: I think this too should be `return 0;` or it will trigger repeater restart

mkmer: I have removed both of these and NOW the repeater thread stays alive when a node disconnects.
I think we only want `return -1` when we were originally setting `toexit=1`...

apps/app_rpt.c
InterLinked1: I removed the toexit variable in favor of returning a value directly. nonzero is the equivalent of toexit.

mkmer:
From what I can tell, `ast_softhangup` is the key to the race -> when `process_link_channels` interrupts `rpt_exec` routine it's known and uses `ast_softhangup`.  If it's not executing the thread we just `ast_hangup` like before. 
Viola - crash gone :)

apps/app_rpt.c
InterLinked1: As with the other one, this maps to the previous behavior, which also would have triggered repeater restart as toexit would have been set true.

apps/app_rpt.c
mkmer: Right, but there was no `toexit` in original code
4 total instances: 2 instances have toexit =1, the other 2 instances did not (from what I can tell)

InterLinked1:
> From what I can tell, `ast_softhangup` is the key to the race -> when `process_link_channels` interrupts `rpt_exec` routine it's known and uses `ast_softhangup`. If it's not executing the thread we just `ast_hangup` like before. Viola - crash gone :)

Yeah, basically. I updated the commit with the full commit description describing the thought process there.

mkmer:
I have my hub running this code with the two `return -1;` changed to `return 0;`.  
As written ANY node disconnecting restarts the repeater task and dumps all other connections (not good) and we can't use it like that.
I'll keep it running until we something new to test that doesn't dump everyone on a disconnect.

apps/app_rpt.c
InterLinked1: Oh, I see what you mean there. There were some breaks without that, I thought they had set it but had not. I updated the PR so they return 0 now. Does that resolve the exiting issue for you as well now?

mkmer: Yes - the PR now matches what I have running.  I'll follow up if we have any issues.

apps/app_rpt.c
```diff
@@ -4140,46 +4140,73 @@ static inline void free_frame(struct ast_frame **f)
...
 }
 
+/*! \brief Safely hang up any channel, even if a PBX could be running on it */
+static inline void safe_hangup(struct ast_channel *chan)
+{
+	/* myrpt is locked here, so we can trust this will be an atomic operation,
+	 * since we also lock before setting the pbx to NULL */
+	if (ast_channel_pbx(chan)) {
+		ast_log(LOG_WARNING, "Channel %s still has a PBX, requesting hangup for it\n", ast_channel_name(chan));
+		ast_softhangup(chan, 0);
```
mkmer, line 4150: Should we have a hangup code here?  0 is not enumerated in `ast_softhangup`. (not sure it matters but magic numbers are usually frown upon).
https://github.com/asterisk/asterisk/blob/4ef68fbce8df8b63759995e2cddc2331ec344b79/include/asterisk/channel.h#L1136-L1182

InterLinked1: Good catch, I confused it with another API where 0 is the "default", I've updated that.

apps/app_rpt.c
```diff
@@ -4140,46 +4140,73 @@ static inline void free_frame(struct ast_frame **f)
...
 }
 
+/*! \brief Safely hang up any channel, even if a PBX could be running on it */
+static inline void safe_hangup(struct ast_channel *chan)
+{
+	/* myrpt is locked here, so we can trust this will be an atomic operation,
+	 * since we also lock before setting the pbx to NULL */
+	if (ast_channel_pbx(chan)) {
+		ast_log(LOG_WARNING, "Channel %s still has a PBX, requesting hangup for it\n", ast_channel_name(chan));
+		ast_softhangup(chan, AST_SOFTHANGUP_EXPLICIT);
```
mkmer, line 4150: So maybe another dumb question here... Could we just softhangup all the time? Is there a need to do the "hard" hangup? Certainly would simplify this logic

InterLinked1: No, soft hangups are intended for when you don't own the channel, hard for when you do

mkmer: Actually maybe we should do this:
https://github.com/AllStarLink/app_rpt/blob/262c22398aee54328a31342c5f7c508650b3587b/apps/app_rpt.c#L4202-L4205
Here:
https://github.com/AllStarLink/app_rpt/blob/262c22398aee54328a31342c5f7c508650b3587b/apps/app_rpt.c#L4181-L4187
after hangup....

InterLinked1: I think your reasoning makes sense, but I agree, it should be a separate PR.
After this merges, please go ahead and put it up (we may also want to see if it prevents the newkey messages in the first place, which would be great).

Allan-N:
@mkmer Are you good with these changes?

mkmer:
I'm 100% good - I have had it running on 2 nodes since PR was created and have not seen a crash.
I'm planning on testing out the proposed "fixes" once this is merged and submitting an additional PR.

mkmer (approved):

Allan-N:
> I'm 100% good

Thanks.  I'm going to wrap up my review this morning and if all looks good we'll get this merged.

apps/app_rpt.c
```diff
@@ -6737,10 +6762,17 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
...
+		 * start using it, and potentially set l->chan to NULL.
+		 * Therefore, we have to be careful when using it
+		 * (lock and check for NULL). */
+
 		if (ast_channel_state(chan) != AST_STATE_UP) {
 			ast_answer(chan);
 			if (l->name[0] > '9') {
 				if (ast_safe_sleep(chan, 500) == -1) {
+					ast_debug(3, "Channel %s hung up\n", ast_channel_name(chan));
 					return -1;
```
Allan-N, line 6776: Are we return'ing here with both ->lock and ->blocklock held?

Allan-N (changes requested):

apps/app_rpt.c
mkmer: That looks bad :) 
I bet it's a tough test to trigger (failing `ast_safe_sleep` on channel). 
I believe this is a test for a webreceiver - sleep a bit then carry on.

mkmer: BTW: I think there are more of these (probably fix them in a new PR):
Killing the thread with a lock (3 locaitons):
https://github.com/AllStarLink/app_rpt/blob/262c22398aee54328a31342c5f7c508650b3587b/apps/app_rpt.c#L6940-L6962

Allan-N: Yes, over the past few days I've been noticing a bunch of code that would benefit from some attention.  But, as you noted, they may be best addressed in new PR's.

apps/app_rpt.c
```diff
@@ -6761,12 +6796,45 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
...
+			rpt_mutex_unlock(&myrpt->lock);
+			if (!strcasecmp(ast_channel_tech(chan)->type, "echolink") || !strcasecmp(ast_channel_tech(chan)->type, "tlb") || (l->name[0] > '9')) {
+				rpt_telemetry(myrpt, CONNECTED, l);
+			}
+		} else {
+			rpt_mutex_unlock(&myrpt->lock);
+		}
+
+		/* In theory, both these conditions should be true if one is, since the node thread will queue a soft hangup on this channel and then set l->chan to NULL */
+		if (!l->chan || ast_check_hangup(chan)) {
```
mkmer, line 6822: I'm thinking we need to set `newkeytimer = 0;` here since we are no longer looking for a newkey message.

apps/app_rpt.c
```diff
@@ -4433,25 +4467,22 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 				if (l->lastf2)
 					memset(l->lastf2->data.ptr, 0, l->lastf2->datalen);
 				l->dtmfed = 1;
-			}
-			if (f->frametype == AST_FRAME_TEXT) {
+			} else if (f->frametype == AST_FRAME_TEXT) {
```
mkmer, line 4470: Should this be a case switch rather than a chain of `else if ()`???

apps/app_rpt.c
mkmer: Let's deal with this later if needed.

apps/app_rpt.c
```diff
@@ -6789,12 +6822,45 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
...
+		 * it is possible that this node's thread (started at rpt() - the C function, not the dialplan one),
+		 * has already requested this channel be hungup.
+		 * We take care to not use ast_hangup if we detect the channel has a PBX,
+		 * so this channel won't have the rug pulled out from underneath it.
+		 * However, l->chan could still be set to NULL. Thus, if it's NULL here,
+		 * that means we're going to get hung up and should just abort, no point in doing telemetry.
+		 * For that reason, even though l->chan is chan if l->chan isn't NULL,
+		 * we always use chan here, rather than l->chan, since l->chan could be NULL.
+		 * This way, even if l->chan isn't NULL when we check but becomes NULL later, we're still safe. */
+		rpt_mutex_lock(&myrpt->lock);
```
mkmer, line 6837: I believe this locking logic is causing this crash:
```
[2025-01-19 19:06:19.881] WARNING[396385][C-00000024]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text !NEWKEY1! on IAX2/149.154.11.243:24135-12943
[2025-01-19 19:06:19.882] WARNING[396385][C-00000024]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text !NEWKEY1! on IAX2/149.154.11.243:24135-12943
  == Spawn extension (radio-secure, 287893, 1) exited non-zero on 'IAX2/149.154.11.243:24135-12943'
[2025-01-19 19:06:19.882] WARNING[396385][C-00000024]: channel.c:2630 ast_hangup: Hard hangup called by thread LWP 396385 on IAX2/149.154.11.243:24135-12943, while blocked by thread LWP 396291 in procedure ast_waitfor_nandfds!  Expect a failure
    -- Hungup 'IAX2/149.154.11.243:24135-12943'
```

Somehow we are hard hanging up in the midst of the lock?!?

Issue #338, opened 2024-07-02
With outstreamcmd= set, shutdown/restart of Asterisk is not reliable


jxmx:
With the "broadcastify" setup as documented with `outstreamcmd`, shutting down or restarting asterisk with `systemctl` is not reliable. It's either very slow or fails. Asterisk/app_rpt should be killing off any spawned processes at shutdown.

jxmx:
The solution here is probably not to change app_rpt, but to make a monolithic item for app_rpt to kill on shutdown.

InterLinked1:
> but to make a monolithic item for app_rpt to kill on shutdown.

What do you really mean by this, exactly?

> The solution here is probably not to change app_rpt, 

All Asterisk modules are expected to clean up anything they started. If `app_rpt` is not fully doing so when requested to unload, then it is a bug in my opinion.

jxmx:
Could be. I haven't really spent time debugging yet. What APPEARS to be happening is that Asterisk/app_rpt is killing off the parent bash/sh process but then lame is re-parenting it's PID in a way that systemd is detecting as being part of the Asterisk group of PIDs but not to Asterisk/app_rpt. I was just thinking that a quick-and-easy fix might be to make sure that the Broadcastify process is a single-shot application that talks over a socket or something to the item that actually doing the lame/ezstream process.

pcpackrat:
I noticed that when using outstreamcmd the CPU pegs at 100% when astres.sh is issued.  I stopped using it for now.  

Also the huge delay is quite normal for broadcastify style streaming.  Even when using a local streaming server.

PR #443, opened 2025-01-04
Properly kill stuff started with startoutstream()

jxmx:
NOTE: I'm just staging this for testing/review -- this is NOT ready for merge

This is to address #338

PR #452, opened 2025-01-12
app_rpt: Tighten logic for stopping outstreamcmd.

InterLinked1:
* Add warnings if outstreamcmd fails to start.
* Fix handling if fork() fails.
* Use ast_safe_fork() instead of fork().
* Close all file descriptors in child before exec.
* Refactor kill(outstreamcmd) into helper function.
* If outstreamcmd fails to execute, don't continously try to restart it in a loop. Instead, log an error and give up.

Resolves: #338

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1063,37 +1063,46 @@ static void startoutstream(struct rpt *myrpt)
 	}
 	n = finddelim(str, strs, 100);
 	if (n < 1) {
+		ast_log(LOG_ERROR, "Could not parse string '%s'\n", myrpt->p.outstreamcmd);
+		ast_free(str);
 		return;
-	}
-	if (myrpt->outstreampipe[1] != -1) {
+	} else if (myrpt->outstreampipe[1] != -1) {
+		ast_log(LOG_ERROR, "Outstream pipe already exists? (fd %d)\n", myrpt->outstreampipe[1]);
+		ast_free(str);
 		close(myrpt->outstreampipe[1]);
 		myrpt->outstreampipe[1] = -1;
 		myrpt->outstreamlasterror = 0;
 	}
+
 	if (pipe(myrpt->outstreampipe) == -1) {
 		ast_log(LOG_ERROR, "pipe() failed: %s\n", strerror(errno));
 		ast_free(str);
 		return;
-	}
-	if (fcntl(myrpt->outstreampipe[1], F_SETFL, O_NONBLOCK) == -1) {
+	} else if (fcntl(myrpt->outstreampipe[1], F_SETFL, O_NONBLOCK) == -1) {
 		ast_log(LOG_ERROR, "Cannot set pipe to NONBLOCK: %s", strerror(errno));
 		ast_free(str);
 		return;
 	}
-	if (!(myrpt->outstreampid = fork())) {
+	myrpt->outstreampid = ast_safe_fork(0);
+	if (myrpt->outstreampid == -1) {
+		ast_log(LOG_ERROR, "fork() failed: %s\n", strerror(errno));
+		ast_free(str);
+		close(myrpt->outstreampipe[1]);
+		myrpt->outstreampipe[1] = -1;
+		return;
+	} else if (!myrpt->outstreampid) {
 		close(myrpt->outstreampipe[1]);
 		if (dup2(myrpt->outstreampipe[0], fileno(stdin)) == -1) {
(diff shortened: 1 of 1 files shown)


jxmx:
As discussed in #443 this seems to work fine to address #338 in my testing.

apps/app_rpt.c
```diff
@@ -1063,49 +1063,49 @@ static void startoutstream(struct rpt *myrpt)
...
-	}
-	if (myrpt->outstreampipe[1] != -1) {
+	} else if (myrpt->outstreampipe[1] != -1) {
+		ast_log(LOG_ERROR, "Outstream pipe already exists? (fd %d)\n", myrpt->outstreampipe[1]);
 		close(myrpt->outstreampipe[1]);
 		myrpt->outstreampipe[1] = -1;
 		myrpt->outstreamlasterror = 0;
-	}
-	if (pipe(myrpt->outstreampipe) == -1) {
+	} else if (pipe(myrpt->outstreampipe) == -1) {
```
Allan-N, line 1073: Changing this line from an "if" to and "else if" will result in a change in behavior.  Before, if an outstreampipe existed then it would be closed ... followed by creating a new pipe.  Now, you fork a new process using the same (old) pipe.  Did we not want to start fresh?

Allan-N (changes requested):

apps/app_rpt.c
InterLinked1: Good catch, I meant it only for the block after that

Allan-N (approved):

apps/app_rpt.c
```diff
@@ -1063,49 +1063,51 @@ static void startoutstream(struct rpt *myrpt)
...
-	if (fcntl(myrpt->outstreampipe[1], F_SETFL, O_NONBLOCK) == -1) {
+	} else if (fcntl(myrpt->outstreampipe[1], F_SETFL, O_NONBLOCK) == -1) {
 		ast_log(LOG_ERROR, "Cannot set pipe to NONBLOCK: %s", strerror(errno));
 		ast_free(str);
 		return;
 	}
-	if (!(myrpt->outstreampid = fork())) {
+	myrpt->outstreampid = ast_safe_fork(0);
+	if (myrpt->outstreampid == -1) {
+		ast_log(LOG_ERROR, "fork() failed: %s\n", strerror(errno));
```
Allan-N, line 1086: Don't we need an `ast_free(str)` here?

apps/app_rpt.c
```diff
@@ -1063,49 +1063,51 @@ static void startoutstream(struct rpt *myrpt)
 	}
 	n = finddelim(str, strs, 100);
 	if (n < 1) {
+		ast_log(LOG_ERROR, "Could not parse string '%s'\n", myrpt->p.outstreamcmd);
```
Allan-N, line 1066: Don't we need an ast_free(str) here?

apps/app_rpt.c
```diff
@@ -3515,6 +3517,18 @@ static inline void process_command(struct rpt *myrpt)
...
 
+static inline void stop_outstream(struct rpt *myrpt)
+{
+	if (myrpt->outstreampid > 0) {
+		int res = kill(myrpt->outstreampid, SIGTERM);
+		if (res) {
+			ast_log(LOG_ERROR, "Cannot kill outstream process %d for node %s: %s\n", (int) myrpt->outstreampid, myrpt->name, strerror(errno));
+		} else {
+			ast_debug(3, "Sent SIGTERM to process %d\n", (int) myrpt->outstreampid);
+		}
```
Allan-N, line 3528: Should we also add `myrpt->outstreampid = 0;` here to indicate that the outstream process is no longer running?

InterLinked1: I thought about that, but this is already done at at line 5770.
I don't really like the idea of checking if the process is still running at every loop iteration, but since it's already be done and setting it there, wasn't sure if that needed to be duplicated. Open to thoughts on that though, fine with it being done either way.

Allan-N: The code around line 5770 is checking to see if an outstream process was started (we have a pid) and that the process is still running (the kill checks for that).  Both of those make checks make some sense.

But, regardless of what we do with outstreampid, it looks like we're going to get "a" debug message.  Do we want to see :

- No outstreampid exists yet
- PID xxx not currently running

I guess that since we just killed the processs we had running the 2nd message would make more sense.

Allan-N (approved):

PR #453, opened 2025-01-13
Clarify newkey logic with enum

mkmer:
After some struggles understanding newkey logic, I had made these changes for my own sanity.

- add enum for newkey values
- use "good" name for last_keytimer
- Fix comments for 2 to 0 forcing.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -7302,7 +7302,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		}
 		keyed = myrx;
 		update_timer(&myrpt->rxlingertimer, elap, 0);
-		if ((myrpt->newkey == RADIO_KEY_NOT_ALLOWED) && keyed && (!myrpt->rxlingertimer)) {
+		if ((myrpt->rpt_newkey == RADIO_KEY_NOT_ALLOWED) && keyed && (!myrpt->rxlingertimer)) {
 			myrpt->rerxtimer = 0;
 			keyed = 0;
 		}
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3119,6 +3119,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 		update_timer(&l->rxlingertimer, elap, 0);
 
 		/* Update the timer, checking if it expired just now. */
+		newkeytimer_last = l->newkeytimer;
 		update_timer(&l->newkeytimer, elap, 0);
 
 		/* Some reverse-engineering comments here from NA debugging issue #46 (inbound calls being keyed when they shouldn't be)
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -5910,7 +5910,7 @@ static inline int exec_chan_read(struct rpt *myrpt, struct ast_channel *chan, ch
 	if (f->frametype == AST_FRAME_VOICE) {
 		struct ast_frame *f1;
 		int ismuted;
-		if (myrpt->rpt_newkey == RADIO_KEY_NOT_ALLOWED){
+		if (myrpt->rpt_newkey == RADIO_KEY_NOT_ALLOWED) {
 			myrpt->rxlingertimer = ((myrpt->iaxkey) ? RX_LINGER_TIME_IAXKEY : RX_LINGER_TIME);
 			if (!*keyed) {
 				*keyed = 1;
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3119,7 +3119,6 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 		update_timer(&l->rxlingertimer, elap, 0);
 
 		/* Update the timer, checking if it expired just now. */
-		x = l->newkeytimer;
 		update_timer(&l->newkeytimer, elap, 0);
 
 		/* Some reverse-engineering comments here from NA debugging issue #46 (inbound calls being keyed when they shouldn't be)
@@ -7303,7 +7302,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		}
 		keyed = myrx;
 		update_timer(&myrpt->rxlingertimer, elap, 0);
-		if ((myrpt->newkey == 2) && keyed && (!myrpt->rxlingertimer)) {
+		if ((myrpt->newkey == RADIO_KEY_NOT_ALLOWED) && keyed && (!myrpt->rxlingertimer)) {
 			myrpt->rerxtimer = 0;
 			keyed = 0;
 		}
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1730,16 +1730,16 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		return;
 	}
 	if (!strcmp(str, NEWKEYSTR)) {
-		if ((!mylink->newkey) || mylink->newkeytimer) {
+		if ((!mylink->link_newkey) || mylink->newkeytimer) {
 			mylink->newkeytimer = 0;
-			mylink->newkey = 1;
-			send_old_newkey(mylink->chan);
+			mylink->link_newkey = RADIO_KEY_ALLOWED_REDUNDANT;
+			send_newkey_redundant(mylink->chan);
 		}
 		return;
 	}
 	if (!strcmp(str, NEWKEY1STR)) {
 		mylink->newkeytimer = 0;
-		mylink->newkey = 2;
+		mylink->link_newkey = RADIO_KEY_NOT_ALLOWED;
 		return;
 	}
 	if (!strncmp(str, IAXKEYSTR, strlen(IAXKEYSTR))) {
@@ -2202,14 +2202,14 @@ static int handle_remote_data(struct rpt *myrpt, const char *str)
 	if (!strcmp(str, DISCSTR))
 		return 0;
 	if (!strcmp(str, NEWKEYSTR)) {
-		if (!myrpt->newkey) {
-			send_old_newkey(myrpt->rxchannel);
-			myrpt->newkey = 1;
+		if (!myrpt->rpt_newkey) {
+			send_newkey_redundant(myrpt->rxchannel);
+			myrpt->rpt_newkey = RADIO_KEY_ALLOWED_REDUNDANT;
 		}
 		return 0;
 	}
 	if (!strcmp(str, NEWKEY1STR)) {
-		myrpt->newkey = 2;
+		myrpt->rpt_newkey = RADIO_KEY_NOT_ALLOWED;
 		return 0;
 	}
 	if (!strncmp(str, IAXKEYSTR, strlen(IAXKEYSTR))) {
@@ -2333,7 +2333,7 @@ static int attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	l->connecttime = 0;
 	l->thisconnected = 0;
 	l->iaxkey = 0;
-	l->newkey = 0;
+	l->link_newkey = RADIO_KEY_ALLOWED;
 
 	cap = ast_format_cap_alloc(AST_FORMAT_CAP_FLAG_DEFAULT);
 	if (!cap) {
@@ -2350,7 +2350,7 @@ static int attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	l->lastrealrx = 0;
 	l->rxlingertimer = ((l->iaxkey) ? RX_LINGER_TIME_IAXKEY : RX_LINGER_TIME);
 	l->newkeytimer = NEWKEYTIME;
-	l->newkey = 2;
+	l->link_newkey = RADIO_KEY_NOT_ALLOWED;
 	while ((f1 = AST_LIST_REMOVE_HEAD(&l->textq, frame_list)))
 		ast_frfree(f1);
 	if (l->chan) {
@@ -2939,6 +2939,7 @@ static inline void dump_rpt(struct rpt *myrpt, const int lasttx, const int laste
 	ast_debug(2, "myrpt->sleepreq = %d\n", (int) myrpt->sleepreq);
 	ast_debug(2, "myrpt->p.parrotmode = %d\n", (int) myrpt->p.parrotmode);
 	ast_debug(2, "myrpt->parrotonce = %d\n", (int) myrpt->parrotonce);
+	ast_debug(2, "myrpt->rpt_newkey =%d\n", myrpt->rpt_newkey);
 
 	zl = myrpt->links.next;
 	while (zl != &myrpt->links) {
@@ -2954,7 +2955,7 @@ static inline void dump_rpt(struct rpt *myrpt, const int lasttx, const int laste
 		ast_debug(2, "        link->retrytimer %d\n", zl->retrytimer);
 		ast_debug(2, "        link->retries = %d\n", zl->retries);
 		ast_debug(2, "        link->reconnects = %d\n", zl->reconnects);
-		ast_debug(2, "        link->newkey = %d\n", zl->newkey);
+		ast_debug(2, "        link->link_newkey = %d\n", zl->link_newkey);
 		zl = zl->next;
 	}
 
@@ -3105,7 +3106,7 @@ static inline void update_timer(int *timer_ptr, int elap, int end_val)
 static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 {
 	struct ast_frame *f;
-	int x;
+	int newkeytimer_last;
 	struct rpt_link *l = myrpt->links.next;
 	while (l != &myrpt->links) {
 		int myrx, mymaxct;
@@ -3122,10 +3123,11 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 		update_timer(&l->newkeytimer, elap, 0);
 
 		/* Some reverse-engineering comments here from NA debugging issue #46 (inbound calls being keyed when they shouldn't be)
-		 * This if statement executes if the timer just expired.
-		 * This does NOT include cases like in handle_link_data where we set newkeytimer = 0 explicitly + set newkey to 1 or 2 (because then x == 0 here)
+		 * This if statement executes if the newkeytimer just expired.
+		 * This does NOT include cases like in handle_link_data where we set newkeytimer = 0 explicitly + set newkey 
+		 * to RADIO_KEY_ALLOWED_REDUNDANT or RADIO_KEY_NOT_ALLOWED (because then newkeytimer_last == 0 here)
 		 */
-		if (x > 0 && !l->newkeytimer) {
+		if (newkeytimer_last > 0 && !l->newkeytimer) { /* Translation: We were timing, and it just expired */
 			/* Issue #46 background:
 			 *
 			 * There is a kind of "handshake" that happens when setting up the IAX2 trunk between two nodes,
@@ -3150,7 +3152,9 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			 * Note that the above depiction separates the TX and RX, but there are only 3 text frames involved.
 			 * In issue #46, 3 text frames are sent, but only 2 are really "received".
 			 * And it so happens that the text frame that B doesn't get from A is exactly the text frame
-			 * that is responsible for setting newkeytimer=0 and newkey=2, i.e. if this doesn't happen, then we'll hit the WARNING case in the below if statement.
+			 * that is responsible for setting newkeytimer=0 and newkey=RADIO_KEY_NOT_ALLOWED, i.e. if this doesn't happen, 
+			 * then we'll hit the WARNING case in the below if statement.  Because this code sets l->link_newkey to RADIO_KEY_ALLOWED,
+			 * There is an unintended radio keyup.
 			 *
 			 * Note that all of these comments are from spending hours debugging this issue and reverse-engineering, but at this point I'm pretty confident
 			 * about these parts of the code, even though I'm not Jim Dixon and he didn't comment any of this code originally.
@@ -3163,26 +3167,30 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			if (l->thisconnected) {
 				/* We're connected, but haven't received a NEWKEY1STR text frame yet...
 				 * The newkeytimer expired on a connected (~answered?) node, i.e. handle_link_data hasn't yet gotten called
-				 * to set newkeytimer = 0 and newkey to non-zero, i.e. we haven't received a text frame with NEWKEY1STR over the IAX2 channel yet.
+				 * to set newkeytimer = 0 and newkey to RADIO_KEY_NOT_ALLOWED, i.e. we haven't received a text frame with NEWKEY1STR over the IAX2 channel yet.
 				 */
-				if (l->newkey == 2) {
+				if (l->link_newkey == RADIO_KEY_NOT_ALLOWED) {
 					/* This can ripple to have consequences down the line, namely we might start writing voice frames
 					 * across the IAX2 link because of this, basically causing us to be transmitting (keyed).
 					 * If this happens, this indicates a problem upstream, and we should emit a warning here
 					 * since undesired behavior will likely ensue.
+					 * We probably SHOULD just hangup the line right here if we are connected and did not receive the ~!NEWKEY1!
+					 * message as something is wrong with the text messaging part of the connection.
 					 */
-					ast_log(LOG_WARNING, "%p newkeytimer expired on connected node, setting newkey from 2 to 0.\n", l);
-					l->newkey = 0;
+					ast_log(LOG_WARNING, "%p newkeytimer expired on connected node, setting newkey from RADIO_KEY_NOT_ALLOWED to RADIO_KEY_ALLOWED.\n", l);
+					l->link_newkey = RADIO_KEY_ALLOWED;
 				}
 			} else {
-				/* If not connected yet (maybe a slow link connection?), wait another NEWKEYTIME ms */
+				/* If not connected yet (maybe a slow link connection?), wait another NEWKEYTIME ms (forever! - probably should limit
+				 * the number of retries here) 
+				 */
 				l->newkeytimer = NEWKEYTIME;
 			}
 		}
 		if ((l->linkmode >1) && (l->linkmode < 0x7ffffffe)) {
 			update_timer(&l->linkmode, elap, 1);
 		}
-		if ((l->newkey == 2) && l->lastrealrx && (!l->rxlingertimer)) {
+		if ((l->link_newkey == RADIO_KEY_NOT_ALLOWED) && l->lastrealrx && (!l->rxlingertimer)) {
 			rxunkey_helper(myrpt, l);
 		}
 
@@ -3233,7 +3241,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 				ast_debug(7, "@@@@ node %s sent node string %s to node %s\n", myrpt->name, lstr, l->name);
 			}
 		}
-		if (l->newkey == 1) {
+		if (l->link_newkey == RADIO_KEY_ALLOWED_REDUNDANT) {
 			if ((l->retxtimer += elap) >= REDUNDANT_TX_TIME) {
 				l->retxtimer = 0;
 				if (l->chan && l->phonemode == 0) {
@@ -4324,7 +4332,7 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 				*totx = 0;
 			if (l->phonemode == 0 && l->chan && (l->lasttx != *totx)) {
 				if (*totx && !l->voterlink) {
-					if (l->newkey < 2)
+					if (l->link_newkey != RADIO_KEY_NOT_ALLOWED)
 						ast_indicate(l->chan, AST_CONTROL_RADIO_KEY);
 				} else {
 					ast_indicate(l->chan, AST_CONTROL_RADIO_UNKEY);
@@ -4365,7 +4373,7 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 
 				l->rxlingertimer = ((l->iaxkey) ? RX_LINGER_TIME_IAXKEY : RX_LINGER_TIME);
 
-				if ((l->newkey == 2) && (!l->lastrealrx)) {
+				if ((l->link_newkey == RADIO_KEY_NOT_ALLOWED) && (!l->lastrealrx)) {
 					rxkey_helper(myrpt, l);
 				}
 				if (((l->phonemode) && (l->phonevox)) || (!strcasecmp(ast_channel_tech(l->chan)->type, "echolink"))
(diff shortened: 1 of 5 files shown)


Allan-N (commented):
While I support adding an `enum` I think we can do better with the enum value naming.  What do the KEYUP_ALLOWED, NOKEYUP_ALLOWED, and OLD_KEYUP_ALLOWED names mean?  And, at the very least, it would be good to add some comments in `app_rpt.h` about each value and when it should be used.

mkmer:
I don't disagree. These were names that made sense to me with limited understanding as I was trying to understand the logic.
I will be happy to change to whatever the team would like.

apps/app_rpt/app_rpt.h
```diff
@@ -342,6 +342,12 @@ struct rpt_chan_stat {
 #define NEWKEY1STR "!NEWKEY1!"
 #define IAXKEYSTR "!IAXKEY!"
 
+enum newkey {
```
InterLinked1, line 345: Maybe name it `rpt_newkey` so it has a common prefix?

apps/app_rpt/app_rpt.h
```diff
@@ -421,7 +427,7 @@ struct rpt_link {
 	char newkey;
 	char iaxkey;
 	int linkmode;
-	int newkeytimer;
+	newkey newkeytimer;
```
InterLinked1, line 430: Should be prefixed with "enum", it's not a typedef

Allan-N (commented):
I just noticed your enum value comments :-)

... and now that I know more about the variable/values "I" might suggest a few renames.

- KEYUP_ALLOWED : OK
- OLD_KEYUP_ALLOWED --> KEYUP_ALLOWED_OLD
- NOKEYUP_ALLOWED --> KEYUP_NOT_ALLOWED

and, I will add that I'm not wild about the "_OLD"

Also, the app_rpt.c handling of `newkey == 1` is odd ... and because the handling differs from `newkey == 2` that makes me question _OLD even more.

But, before making changes I think we should hear what @InterLinked1 and @KB4MDD think about the naming?

apps/app_rpt/app_rpt.h
```diff
@@ -342,6 +342,12 @@ struct rpt_chan_stat {
 #define NEWKEY1STR "!NEWKEY1!"
 #define IAXKEYSTR "!IAXKEY!"
 
+enum newkey {
+	KEYUP_ALLOWED, /*Radio keyup is allowed from link */
+	OLD_KEYUP_ALLOWED, /*Old "NEWKEY!" - radio keyups allowed from link */
+	NOKEYUP_ALLOWED /*On initial connection before reciept of "NEWKEY1!" message - no radio keyups are allowed */
+	}
```
InterLinked1, line 349: This line shouldn't be indented

apps/app_rpt/app_rpt.h
```diff
@@ -342,6 +342,12 @@ struct rpt_chan_stat {
 #define NEWKEY1STR "!NEWKEY1!"
 #define IAXKEYSTR "!IAXKEY!"
 
+enum newkey {
+	KEYUP_ALLOWED, /*Radio keyup is allowed from link */
```
InterLinked1, line 346: There should be a space after the beginning of the comment.

For each of these, use `/*!< my comment */` to make it a valid doxygen comment, and please also add a doxygen comment for the enum as a whole

I would also highly recommend each enum value has a common prefix to make it easy to identify they all belong to the same enum. The prefix could be `NEWKEY` or something like that which makes sense

apps/app_rpt.c
```diff
@@ -7239,7 +7239,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 
 		if (rem_rx && !myrpt->remoterx) {
 			myrpt->remoterx = 1;
-			if (myrpt->newkey < 2)
+			if (myrpt->newkey < NOKEYUP_ALLOWED)
```
InterLinked1, line 7242: Does it really make sense to say "less than an enum value"? Semantically, it's not clear what that means.

There is meaning to the values so it makes sense as numbers obviously - I would probably handle this case (`myrpt->newkey < NOKEYUP_ALLOWED`) by defining a macro next to the enum, named based on what this condition is trying to capture. Somebody else might able to recommend something more concrete better than I.

InterLinked1 (changes requested):
I very much support this change, I was going to do it myself but not having a radio background I wasn't super confident about what the values actually meant, so thanks for doing this.

Mostly stylistic recommendations here.

InterLinked1 (changes requested):
Also, please squash all your commits together into a single commit, and please rename the commit to be prefixed with the module name, e.g. `app_rpt: Use enum for newkey variable.`

KB4MDD:
I agree with @Allan-N, I think the enums should setup with KEYUP.  I don't know the purpose of _OLD value - it would be nice to give it a good name if we understand how it works.

mkmer:
> I agree with @Allan-N, I think the enums should setup with KEYUP. I don't know the purpose of _OLD value - it would be nice to give it a good name if we understand how it works.

I called it old key because of this code:
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt/rpt_channel.c#L463-L483

I suppose we should rename this function too?

apps/app_rpt.c
mkmer: Seems like `!=` will solve the same and make more sense for the enum.

mkmer:
> Also, the app_rpt.c handling of `newkey == 1` is odd ... and because the handling differs from `newkey == 2` that makes me question _OLD even more.

Best I can tell, there was an "old" mechanism for !NEWKEY!, I don't see anywhere in the current code base where it is sent other than a response to someone sending it, so this is probably hanging around for comparability which may no longer be needed (who really knows if there are super old versions out there for sure...).  I could be missing it.

To me, it looks like !NEWKEY1! mechanism keeps a remote from triggering TX until the channel receives the message, then we set `l->newkey` to 0 audio begins to flow.
First we set up the channel and flag `l->newkey = KEYUP_NOT_ALLOWED` (2)
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L6686-L6692

Then we send a couple of !NEWKEY1! messages:
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L6740-L6763

The other side of the link then responds with !NEWKEY1! for answering the call:
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L4454-L4468

The key to control happening is here:
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L4488-L4499

If it is  (>=2) we ignore `AST_CONTROL_RADIO_KEY` - hence I called it `KEYUP_NOT_ALLOWED`

There are a few things in here that should also get #DEFINES to clear things up:
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L6759
`l->name[0] <= '9'` -> That's the first character of the node id (right?) What type are we really looking for?

Same thing here:
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L6742
`l->name > '9'`

I think that means it's not numeric - maybe a web receiver?  

Then there is this `!= 0` 
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L6691C47-L6691C57

What nodes have 0 for the first character?

This is a LOT!  I feel like putting it in words may help me verify my thoughts and allow others to confirm I'm not out in the weeds.

Allan-N:
Looking over app_rpt.c and all of the place where we check `newkey == 1` you will comparisons to `REDUNDANT_TX_TIME`.  All references to REDUNDANT_TX_TIME relate to `newkey == 1`.  Given this, I would like to suggest :
- KEYUP_ALLOWED_OLD --> KEYUP_ALLOWED_REDUNDANT

apps/app_rpt.c
```diff
@@ -1677,14 +1677,14 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
...
-			send_old_newkey(mylink->chan);
+			mylink->newkey = KEYUP_ALLOWED_REDUNDANT;
+			send_newkey_redundant(mylink->chan);
 		}
 		return;
 	}
 	if (!strcmp(tmp, NEWKEY1STR)) {
 		mylink->newkeytimer = 0;
-		mylink->newkey = 2;
+		mylink->newkey = KEYUP_NOT_ALLOWED;
```
Allan-N, line 1687: OK.  I have to ask the question here.  Do we have KEYUP_NOT_ALLOWED and KEYUP_ALLOWED flipped?  or am I just not reading the code correctly?

I would have thought that one would start off with the `newkey` state being "not" allowed until we have completed our negotiations.  We start off with `newkey == 0` (KEYUP_NOT ALLOWED).  Then, when we receive NEWKEY1STR we move to the `newkey == 2` (KEYUP_ALLOWED) state.  And, if we get the "old" NEWKEYSTR we use the in-between `newkey == 1` (KEYUP_ALLOWED_REDUNDANT).

I'm thinking that we should have :
```
enum rpt_newkey { /*!< Repeter connection newkey handshake */
 	KEYUP_NOT_ALLOWED,
 	KEYUP_ALLOWED_REDUNDANT,
 	KEYUP_ALLOWED,
 }
```
This way, any non-zero value would be "allowed".

Most certainly need to review how `newkey` is used ... and the comments (many that still refer to 0, 1, 2) before making any changes.

Allan-N: Or, do we start off with the status allowing keyup's (KEYUP_ALLOWED) and give the connection a couple of seconds to negotiate.  Then, if the negotiation time (newkeytimer?) elapses and we haven't heard from the remote side then we block keyups (KEYUP_NOT_ALLOWED)?

Just reading the code, trying to understand how this works, trying to understand the minimal (and a bit confusing) comments.

mkmer: This is why I'm sharing :)  It is VERY counter intuitive to me as well.

At this point I don't think we are changing operation (not intentionally anyway).  Goal being to document function, then maybe we can change it later?  It's amazing how words (enum) change how one looks at a test case or value assignments.

I believe this is "key" to what it is doing;

https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L4295-L4300

Essentially if its 2, `AST_CONTROL_RADIO_KEY` is not allowed - although `AST_CONTROL_RADIO_UNKEY` is.
I think this has to do with it being an IAX full duplex link channel (via some convoluted tests) - this is configured by both ends of an IAX call sending `!NEWKEY1!`.  

TBH, I'm still a bit lost on everything it is doing.  I'm planing on adding some debug to try and follow what is happening.

Allan-N: That block of code helps to explain how the `newkey` value is used.  The next question to answer is what state we start off in (and why)? and how/why we transition between the states?  Adding some comments to explain this, after we've figured it out, will certainly help.

apps/app_rpt.c
```diff
@@ -4294,7 +4294,7 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 				*totx = 0;
 			if (l->phonemode == 0 && l->chan && (l->lasttx != *totx)) {
 				if (*totx && !l->voterlink) {
-					if (l->newkey < 2)
+					if (l->newkey != KEYUP_NOT_ALLOWED
```
Allan-N, line 4297: Looks like you dropped the close paren ... and to conform to the current coding conventions it would be good to add { } braces.

apps/app_rpt.c
```diff
@@ -4522,10 +4522,10 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 					fac_frame(f, fac);
 				}
 				/* foop */
-				if (l->chan && (l->lastrx || (!altlink(myrpt, l))) && ((l->newkey < 2) || l->lasttx || strcasecmp(ast_channel_tech(l->chan)->type, "IAX2"))) {
+				if (l->chan && (l->lastrx || (!altlink(myrpt, l))) && ((l->newkey != KEYUP_NOT_ALLOWED) || l->lasttx || strcasecmp(ast_channel_tech(l->chan)->type, "IAX2"))) {
 					/* Reverse-engineering comments from NA debugging issue #46:
-					 * We may be receiving frames from channel drivers but we discard them and don't pass them on if newkey hasn't been set to 2 yet.
-					 * Of course if handle_link_data is never called to set newkey to 2 and stop newkeytimer, then at some point, we'll
+					 * We may be receiving frames from channel drivers but we discard them and don't pass them on if newkey hasn't been set to <2 yet.
```
Allan-N, line 4527: Not wild about having the <2 here.  Maybe "... if newkey has not yet been negotiated` ?

apps/app_rpt.c
```diff
@@ -4522,10 +4522,10 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
...
+					 * Of course if handle_link_data is never called to set newkey to <2 and stop newkeytimer, then at some point, we'll
```
Allan-N, line 4528: Another <2 :-)  Here, I don't know what to say.  In an earlier comment I questioned/suggested that the code may start off blocking keyup but give up and allow keying if some period of time has elapsed and the negotiation doesn't happen.

Allan-N (changes requested):

apps/app_rpt.c
mkmer: I went with `!= KEYUP_NOT_ALLOWED` which in this case means it's 0 (KEYUP_ALLOWED) or 1 (KEYUP_ALLOWED_REDUNDANT) is True.  

Probably should update the comment....

apps/app_rpt.c
mkmer: > Or, do we start off with the status allowing keyup's (KEYUP_ALLOWED) and give the connection a couple of seconds to negotiate. Then, if the negotiation time (newkeytimer?) elapses and we haven't heard from the remote side then we block keyups (KEYUP_NOT_ALLOWED)?
> 
> Just reading the code, trying to understand how this works, trying to understand the minimal (and a bit confusing) comments.

From what I can tell, we startup with KEYUP_ALLOWED on the caller side.  On the receiver side we poke KEYUP_NOT_ALLOWED when we get the channel link is started up:
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L6691-L6693

The receiver side then sends  !NEWKEY1! and caller would set `->newkey' = 2 (KEYUP_NOT_ALLOWED)

mkmer: I have pushed a correction to the comments using the values (and ENUM) for now.
The original comment https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L3093-L3094
is incorrectly describing what actually happens when `handle_link_data` receives !NEWKEY1!.

apps/app_rpt.c
```diff
@@ -3043,17 +3043,18 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
...
 			l->newkeytimer = 0;
 
 		/* Some reverse-engineering comments here from NA debugging issue #46 (inbound calls being keyed when they shouldn't be)
 		 * This if statement executes if the timer just expired.
-		 * This does NOT include cases like in handle_link_data where we set newkeytimer = 0 explicitly + set newkey to 1 or 2 (because then x == 0 here)
+		 * This does NOT include cases like in handle_link_data where we set newkeytimer = 0 explicitly + set newkey to 1 or 2 
+		 * (because then newkeytimer_last == 0 here)
 		 */
-		if (x > 0 && !l->newkeytimer) {
+		if (newkeytimer_last > 0 && !l->newkeytimer) {
```
InterLinked1, line 3057: Use the enum value here (not > 0)

apps/app_rpt.c
mkmer: The receiving node initializes here:
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L6686-L6692

The Calling node initializes here when it's created:
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt/rpt_link.c#L670-L683

and here when it's answered

https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L4454-L4464

apps/app_rpt.c
```diff
@@ -3078,7 +3079,8 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			 * Note that the above depiction separates the TX and RX, but there are only 3 text frames involved.
 			 * In issue #46, 3 text frames are sent, but only 2 are really "received".
 			 * And it so happens that the text frame that B doesn't get from A is exactly the text frame
-			 * that is responsible for setting newkeytimer=0 and newkey=2, i.e. if this doesn't happen, then we'll hit the WARNING case in the below if statement.
+			 * that is responsible for setting newkeytimer=0 and newkey=2 (KEYUP_NOT_ALLOWED), i.e. if this doesn't happen, then we'll hit the WARNING case
```
InterLinked1, line 3082: You can replace values in the comments with the enum values as well

apps/app_rpt.c
```diff
@@ -3091,16 +3093,16 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			if (l->thisconnected) {
 				/* We're connected, but haven't received a NEWKEY1STR text frame yet...
 				 * The newkeytimer expired on a connected (~answered?) node, i.e. handle_link_data hasn't yet gotten called
-				 * to set newkeytimer = 0 and newkey to non-zero, i.e. we haven't received a text frame with NEWKEY1STR over the IAX2 channel yet.
+				 * to set newkeytimer = 0 and newkey to 2 KEYUP_NOT_ALLOWED, i.e. we haven't received a text frame with NEWKEY1STR over the IAX2 channel yet.
```
InterLinked1, line 3096: Again, no more integers here

apps/app_rpt.c
```diff
@@ -4522,11 +4524,11 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 					fac_frame(f, fac);
 				}
 				/* foop */
-				if (l->chan && (l->lastrx || (!altlink(myrpt, l))) && ((l->newkey < 2) || l->lasttx || strcasecmp(ast_channel_tech(l->chan)->type, "IAX2"))) {
+				if (l->chan && (l->lastrx || (!altlink(myrpt, l))) && ((l->newkey != KEYUP_NOT_ALLOWED) || l->lasttx || strcasecmp(ast_channel_tech(l->chan)->type, "IAX2"))) {
 					/* Reverse-engineering comments from NA debugging issue #46:
-					 * We may be receiving frames from channel drivers but we discard them and don't pass them on if newkey hasn't been set to 2 yet.
-					 * Of course if handle_link_data is never called to set newkey to 2 and stop newkeytimer, then at some point, we'll
-					 * set newkey = 0 forcibly (see comments in that part of the code for more info), which will cause us to start passing on the voice frames here.
+					 * We may be receiving frames from channel drivers but we discard them and don't pass them on if newkey hasn't been set to <2 (!= KEYUP_NOT_ALLOWED) yet.
```
InterLinked1, line 4529: ... and here

apps/app_rpt/app_rpt.h
```diff
@@ -342,6 +342,12 @@ struct rpt_chan_stat {
 #define NEWKEY1STR "!NEWKEY1!"
 #define IAXKEYSTR "!IAXKEY!"
 
+enum rpt_newkey { /*!< Repeter connection newkey handshake */
```
InterLinked1, line 345: Move the first comment above the enum, and no ! needed.
Should be:

`/*! \brief Repeater connection newkey handshake state */`

apps/app_rpt/app_rpt.h
```diff
@@ -342,6 +342,12 @@ struct rpt_chan_stat {
...
+	KEYUP_ALLOWED, /*!< Radio keyup is allowed from link */
```
InterLinked1, line 346: More comments describing when a repeater is in each state might be nice :)

InterLinked1 (changes requested):
In the commit message, enum should be all lowercase. `ENUM` in Asterisk refers to something completely different :)

apps/app_rpt.c
mkmer: This is the `newkeytimer` not `newkey` here.  You're just seeing if the timer expired on this call and "one shotting" the logic (right?)

InterLinked1: Ah, okay, my bad - all good then here!

apps/app_rpt.c
```diff
@@ -3091,16 +3093,16 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			if (l->thisconnected) {
 				/* We're connected, but haven't received a NEWKEY1STR text frame yet...
 				 * The newkeytimer expired on a connected (~answered?) node, i.e. handle_link_data hasn't yet gotten called
-				 * to set newkeytimer = 0 and newkey to non-zero, i.e. we haven't received a text frame with NEWKEY1STR over the IAX2 channel yet.
+				 * to set newkeytimer = 0 and newkey to KEYUP_NOT_ALLOWED, i.e. we haven't received a text frame with NEWKEY1STR over the IAX2 channel yet.
 				 */
-				if (l->newkey == 2) {
+				if (l->newkey == KEYUP_NOT_ALLOWED) {
```
Allan-N, line 3098: If you read these comments, look at what we're checking, and the new state we're setting one might think that we have the wrong labels for 0 and 2?  To have the code match the comments I would have expected :
```
if (l->newkey == KEYUP_ALLOWED) {
    // we've haven't received our newkey ack, stop transmitting
    l->newkey = KEYUP_NOT_ALLOWED;
}
```

Should "KEYUP_ALLOWED = 0" and "KEYUP_NOT_ALLOWED = 2" ?

mkmer: OK - I also find it confusing as heck... (And I may be way out in the weeds)  I certainly chased my tail around and around.

2 = `KEYUP_NOT_ALLOWED`.  
`l->newkey == 2` inhibits sending AST_CONTROL_RADIO_KEY (see notes / code example above). 

What was/is happening in the crash "test case" -> we start a connection firing off `rpt_exec` logic.  In there it sets  up the l->xxx stuff and sets `l->newkey = 2;` and `l->newkeytimer !=0` before the connection is even answered (although answering doesn't matter here). 

IF the caller hangs up before  `rpt_exec` is finished, `l->newkey = 2`  is stuck that way and `l->connected` is true.  When the `l->newkeytimer` finishes its causing the above logic to trigger.  I think this will self heal when the channel finally closes and the `*l` is removed from the list and freed - aka: this code not necessary.  

I believe this logic is no longer doing (or never did) anything for us, the jitter buffer fixed root problem.  

The point of this PR is to change numbers to enum and NOT change the logic - so if it was a 2, it's now KEYUP_NOT_ALLOWED, etc.  

Maybe the next PR reset `l->connected` on soft hangups and make sure this code is not triggering (ever) - then remove it in a 3rd PR.  Not sure.

Allan-N: > The point of this PR is to change numbers to enum and NOT change the logic - so if it was a 2, it's now KEYUP_NOT_ALLOWED, etc.

I'm not suggesting that we change the logic.  What I'm questioning is (my?) choice of labels for the 0, 1, 2 values.  The change that was made was to have 0 == KEYUP_ALLOWED and 2 == KEYUP_NOT_ALLOWED.  I'm wondering if we got this backwards and if "0" should have been labeled KEYUP_NOT_ALLOWED (and "1" should have been labeled KEYUP_ALLOWED).  There are at least a few places where flipping the label names (not the effective values) would have the code match what one would expect to be happening (like this case and in handle_link_data/handle_remote_data)

mkmer: This is the commit where comments were added: 
[00101d3add13edc55b6a688439d51606d61dcd53](https://github.com/AllStarLink/app_rpt/commit/00101d3add13edc55b6a688439d51606d61dcd53)

mkmer: Gotcha - I though you were looking to change the code.
I think it's right because of this:
https://github.com/AllStarLink/app_rpt/blob/604328d3c689a9ac28df22a76d359f8c8ac8eaff/apps/app_rpt.c#L4295-L4300

if `l->newkey ==2` we don't send `ast_indicate(l->chan, AST_CONTROL_RADIO_KEY);` 
To me that means "KEY NOT ALLOWED" 
Should we call it something else?

Allan-N: In this case, 2==keyup_not_allowed makes sense.  I shall continue to ponder the logic and naming.

mkmer: OK - I just noticed how I mixed up `->newkey`!
There are `l->newkey` and `myrpt->newkey` vars.  I have them mushed together.
The LINK sets up `=2` when the call comes, but the repeater has to hear the message to set it's var to 2.

I'm working on fixing this up, and maybe revising my thoughts.

mkmer:
I think I've untangled this a bit.  I went with 2 enums one for the repeater structure and one for the link structure.  Hopefully that will help be a bit more clear which one we are using.  If I've done it right, I think it will explain it self :)

apps/app_rpt/app_rpt.h
```diff
@@ -342,6 +342,20 @@ struct rpt_chan_stat {
 #define NEWKEY1STR "!NEWKEY1!"
 #define IAXKEYSTR "!IAXKEY!"
 
+/*! \brief Repeater link connection newkey handshake state */
```
InterLinked1, line 345: I'm not really sure I like the idea of duplicating the same enum for two different uses, but I'm curious to hear what others think about this.

An alternate approach might be use the same enum, but name the variable in the struct differently e.g. myrpt->rpt_newkey and l->link_newkey or something like that. Not saying that's better per se, just another approach I wanted to throw out there.

mkmer: I really got a bit carried away with trying to keep the two vars separate.  I don't think it makes sense either now that I'm looking at it.   I suppose I others like it, I can put it back easy enough.

Allan-N: Wait, there are two "different" newkey variables?  That most certainly adds to the confusion.

I agree that duplicating the same enum doesn't make any sense ... as long as we have the same 3 values.  But, if one flavor uses 2 values and the other 3 then, yes, split 'em up.

I also agree that it would help to clear things up if we used different variable names.  The suggested `rpt_newkey` and `link_newkey` names work for me.

mkmer: That makes sense - I have updated the structures to `rpt_link -> link_newkey` and `rpt->rpt_newkey`.
The both appear to use the same 0,1,2 values, so one enum should be "right" in this case.

mkmer:
As far as I can tell all this newkey business is doing is deciding if we send `ast_indicate(l->chan, AST_CONTROL_RADIO_KEY);` down the channel.  0,1=YES 2=NO.  When `newkey==2`, there is a different path to actually keying up the repeater.  I changed the enum names to `RADIO_KEY_xxx` because it doesn't stop keyup of the repeater and I was feeling like that KEYUP was confusing too.

As you dig around, let's see if it helps...

Allan-N:
Since we (currently) only have one enum, maybe change `enum rpt_newkey` to `enum newkey`

apps/app_rpt/app_rpt.h
mkmer: I've added some comments to these - I think it's more clear.

Allan-N (approved):

apps/app_rpt.c
```diff
@@ -3090,7 +3091,7 @@ static inline void update_timer(int *timer_ptr, int elap, int end_val)
 static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 {
 	struct ast_frame *f;
-	int x;
+	int newkeytimer_last;
```
InterLinked1, line 3094: Shouldn't this be the enum?

apps/app_rpt/rpt_channel.h
```diff
@@ -42,4 +42,4 @@ int send_link_pl(struct rpt *myrpt, char *txt);
 /*! \brief send newkey request */
 void send_newkey(struct ast_channel *chan);
 
-void send_old_newkey(struct ast_channel *chan);
+void send_newkey_redundant(struct ast_channel *chan);
```
InterLinked1, line 45: Since you have a good understanding of what these mean, could we maybe add a brief description for this function, and update the other one?

InterLinked1 (changes requested):

apps/app_rpt.c
mkmer: I don't think so.  This is a timer value that was in a "dumb" variable name `x`.  Maybe I've missed something.

apps/app_rpt/rpt_channel.h
mkmer: I have an understanding... not sure it's good :)
Hopefully the new comments help clarify.

apps/app_rpt.c
```diff
@@ -5899,7 +5910,7 @@ static inline int exec_chan_read(struct rpt *myrpt, struct ast_channel *chan, ch
 	if (f->frametype == AST_FRAME_VOICE) {
 		struct ast_frame *f1;
 		int ismuted;
-		if (myrpt->newkey == 2) {
+		if (myrpt->rpt_newkey == RADIO_KEY_NOT_ALLOWED){
```
InterLinked1, line 5913: space before `{`

apps/app_rpt/rpt_channel.h
```diff
@@ -39,7 +39,16 @@ int send_usb_txt(struct rpt *myrpt, char *txt);
 /*! \brief send asterisk frame text message on the current tx channel */
 int send_link_pl(struct rpt *myrpt, char *txt);
 
-/*! \brief send newkey request */
+/*! \brief send newkey request NEWKEY1STR to caller.  When a call is initiated
+ * l->link_newkey is set to RADIO_KEY_NOT_ALLOWED, and l->newkeytimer is activate.
+ * If the timer expires before receiving NEWKEY1STR, l->link_newkey is set to RADIO_KEY_ALLOWED.
+ * NEWKEY1STR causes the reppeater to keyup based on AST_FRAME_VOICE detected on IAX2 links.
```
InterLinked1, line 45: repeater

apps/app_rpt/rpt_channel.h
```diff
@@ -39,7 +39,16 @@ int send_usb_txt(struct rpt *myrpt, char *txt);
...
+*/
 void send_newkey(struct ast_channel *chan);
 
-void send_old_newkey(struct ast_channel *chan);
+/*! \brief send newkey request NEWKEYSTR to caller.  This appears to be a legacy message.
+ * If NEWKEYSTR is received on the link, l->link_newkey is set RADIO_KEY_ALLOWED_REDUNDANT,
+ * the l->newkeytimer is disabled, and NEWKEYSTR is echoed to the caller.
+*/
+
```
InterLinked1, line 53: Can remove blank line here

InterLinked1 (changes requested):

InterLinked1 (approved):

PR #457, opened 2025-01-17
app_rpt: remove duplicate timer code adding inline update_timer

mkmer:
There appears to be many duplicate instances of the timer update logic.  This may be a cleaner less repetitive way?

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3023,6 +3023,21 @@ static inline void rxunkey_helper(struct rpt *myrpt, struct rpt_link *l)
 	}
 }
 
+/*! \brief Do timer value update, limit to end_val */
+static inline void update_timer(int *timer_ptr, int elap, int end_val)
+{
+	if (!timer_ptr){
+		return;
+	}
+	if (*timer_ptr > end_val) {
+		*timer_ptr -= elap;
+	}
+	if (*timer_ptr < end_val) {
+		*timer_ptr = end_val;
+	}
+}
+
+
 static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 {
 	struct ast_frame *f;
@@ -3036,18 +3051,11 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			ast_write(l->chan, f);
 			ast_frfree(f);
 		}
-
-		if (l->rxlingertimer)
-			l->rxlingertimer -= elap;
-		if (l->rxlingertimer < 0)
-			l->rxlingertimer = 0;
+		update_timer(&l->rxlingertimer, elap, 0);
 
 		/* Update the timer, checking if it expired just now. */
 		x = l->newkeytimer;
-		if (l->newkeytimer)
-			l->newkeytimer -= elap;
-		if (l->newkeytimer < 0)
-			l->newkeytimer = 0;
+		update_timer(&l->newkeytimer, elap, 0);
 
 		/* Some reverse-engineering comments here from NA debugging issue #46 (inbound calls being keyed when they shouldn't be)
 		 * This if statement executes if the timer just expired.
@@ -3107,19 +3115,14 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 				l->newkeytimer = NEWKEYTIME;
 			}
 		}
-		if ((l->linkmode > 1) && (l->linkmode < 0x7ffffffe)) {
-			l->linkmode -= elap;
-			if (l->linkmode < 1)
-				l->linkmode = 1;
+		if ((l->linkmode >1) && (l->linkmode < 0x7ffffffe)) {
+			update_timer(&l->linkmode, elap, 1);
 		}
 		if ((l->newkey == 2) && l->lastrealrx && (!l->rxlingertimer)) {
 			rxunkey_helper(myrpt, l);
 		}
 
-		if (l->voxtotimer)
-			l->voxtotimer -= elap;
-		if (l->voxtotimer < 0)
-			l->voxtotimer = 0;
+		update_timer(&l->voxtotimer, elap, 0);
 
 		if (l->lasttx != l->lasttx1) {
 			if ((!l->phonemode) || (!l->phonevox))
@@ -3142,11 +3145,9 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 				myrx = myrx || l->wasvox;
 		}
 		l->lastrx = myrx;
-		if (l->linklisttimer) {
-			l->linklisttimer -= elap;
-			if (l->linklisttimer < 0)
-				l->linklisttimer = 0;
-		}
+
+		update_timer(&l->linklisttimer,elap, 0);
+
 		if ((!l->linklisttimer) && (l->name[0] != '0') && (!l->isremote)) {
 			struct ast_frame lf;
 			char lstr[MAXLINKLIST];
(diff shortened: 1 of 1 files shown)


apps/app_rpt.c
```diff
@@ -3023,6 +3023,18 @@ static inline void rxunkey_helper(struct rpt *myrpt, struct rpt_link *l)
 	}
 }
 
+/*! \brief Do timer value update, limit to end_val */
+static inline void update_timer(int* timer, int elap, int end_val)
+{
+	if (*timer > end_val) {
```
encbar5, line 3029: If this function is trying to replace the repetitive code, then I think this would need to be a null check instead of a comparison.
`if (timer)`

mkmer: There are two different "timers" some stop at 1, some stop at 0.

`if (*timer > 0){}` solves the same as `if (*timer){}` in the case were we only subtract if its `>0`.  If it's `<0` we don't subtract in either case (right?)
  
The timers that stop at 1 always used `if (*timer > 1){}`

We could use 1 for all, or make 2 different version -> one for `==0` end and one for `==1` end
I was thinking one function with variable end points made more sense and kept redundant code down.

mkmer: BWT: The example your provided is missing that is the test is against the VALUE of the pointer not the pointer itself?  Your example would test the pointer rather than the value of the pointer.

apps/app_rpt.c
```diff
@@ -3107,19 +3112,14 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 				l->newkeytimer = NEWKEYTIME;
 			}
 		}
-		if ((l->linkmode > 1) && (l->linkmode < 0x7ffffffe)) {
-			l->linkmode -= elap;
-			if (l->linkmode < 1)
-				l->linkmode = 1;
+		if (l->linkmode < 0x7ffffffe) {
```
Allan-N, line 3115: This one's complicated.  If l->linkmode < 1 (..., -1, 0) then you don't want to adjust the value.

apps/app_rpt.c
Allan-N: As written, the function assumes that "timer" (maybe should be named "timer_ptr") will NOT be NULL.  For safety, you could use :
```
    if (timer && (*timer > end_val)) {
```

Allan-N (changes requested):

apps/app_rpt.c
mkmer: While it would be hard to call with a NULL pointer, it's probably worth the CPU to test for it :)

apps/app_rpt.c
mkmer: Well, it was `> 1` and `< "a really  big positive number"` because 0x7F(blah) is sign bit 31 = 0.
Since `>1` is in the function, it's a bit redundant to do it outside too.  
That said, I think it does make sense for "clarity" to test >1 outside the function so it's easier to see what we are really testing before processing the `-=elap`.  
(maybe?)

apps/app_rpt.c
```diff
@@ -3023,6 +3023,21 @@ static inline void rxunkey_helper(struct rpt *myrpt, struct rpt_link *l)
 	}
 }
 
+/*! \brief Do timer value update, limit to end_val */
+static inline void update_timer(int* timer_ptr, int elap, int end_val)
```
Allan-N, line 3027: The whitespace in this function could be cleaned up a bit.

apps/app_rpt.c
```diff
@@ -3427,49 +3370,33 @@ static inline int update_timers(struct rpt *myrpt, const int elap, const int tot
...
 	} else {
-		if (myrpt->parrottimer)
-			myrpt->parrottimer -= elap;
-		if (myrpt->parrottimer < 0)
-			myrpt->parrottimer = 0;
+		update_timer(&myrpt->parrottimer, elap, 0);
 	}
-	update_macrotimer(myrpt, elap);
-	update_dtmf_local_timer(myrpt, elap);
+	update_macrotimer(myrpt, elap, 0);
```
Allan-N, line 3398: Looks like you missed this one.

Allan-N (changes requested):

apps/app_rpt.c
Allan-N: This one is a bit interesting.  Have a look at how l->linkmode is set in rpt_config.c.  There's a couple of values that appear to be special (0, 1, 0x7ffffffe, 0x7fffffff).

apps/app_rpt.c
Allan-N: Ideally, we'd have the compiler flag any usage where a valid pointer is not passed.  Personally, I'm OK leaving out null pointer check but having the safeguard in doesn't hurt.

apps/app_rpt.c
InterLinked1, line 3027: The * should go with timer_ptr, not int, per Asterisk coding conventions.
Also, the commit message could just be `app_rpt: Refactor timer update code` since this is a clear-cut refactor

InterLinked1 (changes requested):

apps/app_rpt.c
mkmer: Looks more like I mangled than missed it :(

apps/app_rpt.c
mkmer: Sure would be nice to have a linting tool to check this stuff for us (especially me - sloppy haven't synthesized c code in 10 years type of person)

apps/app_rpt.c
mkmer: Yes - this is more or less bad practice to make double use of a var like this....  It's almost like we only have 64 bits to play with :)
https://github.com/AllStarLink/app_rpt/blob/87dc9741ad2740d36f30dd4046635c239feab9c7/apps/app_rpt/rpt_link.c#L34-L45

Looks like we use it to set a link mode, then use it as a hang_timer if we're in "demand mode" `(==1)` AND linkmodedynamic.
https://github.com/AllStarLink/app_rpt/blob/87dc9741ad2740d36f30dd4046635c239feab9c7/apps/app_rpt/rpt_telemetry.c#L51-L52

I'd say these should be broken up to be more explicit, maybe add a `->link_hangtimer`? Not the easiest magic to decode for sure.

That said, I don't think this PR breaks how it works today.

Allan-N: >Yes - this is more or less bad practice to make double use of a var like this...

I agree

>That said, I don't think this PR breaks how it works today.

I agree

Allan-N (approved):

apps/app_rpt.c
mkmer: VSCode has a C format too (right click -> choose format) that could be run to force formatting (at least).  I gave it a try here on app_rpt.c :
https://github.com/AllStarLink/app_rpt/compare/master...mkmer:app_rpt:lint?expand=1

This is just the default VSCODE format - VSCode could be configured in the repo to run it before a local check in helping keep the format in an asterisk format or whatever.  Same could happen with a other rules recommended in the Asterisk guide.  Maybe the Asterisk guys already have these tools setup?  I don't see them right off the bat. 

As you guys get more "helpers" on the project it could help a bunch and save your precious time in reviews.

I've been VERY impressed with the HomeAssistant project and how they are wrangling 2000+ developers into following the "rules" for the project using these tools.

KB4MDD: This info here is helpful.
https://docs.asterisk.org/Historical-Documentation/Development/Historical-Policies-and-Procedures/Code-Review/Coding-Guidelines/

apps/app_rpt.c
```diff
@@ -3023,6 +3023,21 @@ static inline void rxunkey_helper(struct rpt *myrpt, struct rpt_link *l)
...
 }
 
+/*! \brief Do timer value update, limit to end_val */
+static inline void update_timer(int *timer_ptr, int elap, int end_val)
+{
+	if (!timer_ptr){
+		return;
+	}
+	if (*timer_ptr > end_val) {
+			*timer_ptr -= elap;
```
InterLinked1, line 3033: There is an extra indent on this line

InterLinked1 (changes requested):

InterLinked1 (approved):

Issue #459, opened 2025-01-20
app_rpt: "fast hangup" crashing with latest dev.


mkmer:
There is quite a bit of additional code beyond what I had tested earlier in PR https://github.com/AllStarLink/app_rpt/pull/451

Now when executing the "fast disconnect" test I receive this crash:
 ```
[2025-01-19 19:24:50.540] WARNING[420715][C-00000024]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text !NEWKEY1! on IAX2/149.154.11.243:24357-1639
[2025-01-19 19:24:50.542] WARNING[420715][C-00000024]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text !NEWKEY1! on IAX2/149.154.11.243:24357-1639
  == Spawn extension (radio-secure, 287893, 1) exited non-zero on 'IAX2/149.154.11.243:24357-1639'
[2025-01-19 19:24:50.542] WARNING[420715][C-00000024]: channel.c:2630 ast_hangup: Hard hangup called by thread LWP 420715 on IAX2/149.154.11.243:24357-1639, while blocked by thread LWP 420617 in procedure ast_waitfor_nandfds!  Expect a failure
```

[core-asterisk-2025-01-20T00-24-50Z-full.txt](https://github.com/user-attachments/files/18471067/core-asterisk-2025-01-20T00-24-50Z-full.txt)
[core-asterisk-2025-01-20T00-24-50Z-locks.txt](https://github.com/user-attachments/files/18471068/core-asterisk-2025-01-20T00-24-50Z-locks.txt)
[core-asterisk-2025-01-20T00-24-50Z-brief.txt](https://github.com/user-attachments/files/18471069/core-asterisk-2025-01-20T00-24-50Z-brief.txt)
[core-asterisk-2025-01-20T00-24-50Z-info.txt](https://github.com/user-attachments/files/18471072/core-asterisk-2025-01-20T00-24-50Z-info.txt)

**Comment by mkmer** [comment]
This may be a result of me setting up my dev environment wrong too.

**Comment by mkmer** [comment]
This is the offending code:
https://github.com/AllStarLink/app_rpt/blob/e4dee9018642037100f981eea2ad96efddd4b877/apps/app_rpt.c#L6848-L6852

Commenting it out eliminates the crash but we end up with a Max retries error indefinitely:
```
[...]

PR #460, opened 2025-01-20
app_rpt: Acquire blocklock mutex when hard hanging up link channel.

InterLinked1:
Resolves: #459

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4154,7 +4154,7 @@ static inline void free_frame(struct ast_frame **f)
 }
 
 /*! \brief Safely hang up any channel, even if a PBX could be running on it */
-static inline void safe_hangup(struct ast_channel *chan)
+static inline void safe_hangup(struct rpt *myrpt, struct ast_channel *chan)
 {
 	/* myrpt is locked here, so we can trust this will be an atomic operation,
 	 * since we also lock before setting the pbx to NULL */
@@ -4163,15 +4163,19 @@ static inline void safe_hangup(struct ast_channel *chan)
 		ast_softhangup(chan, AST_SOFTHANGUP_EXPLICIT);
 	} else {
 		ast_debug(3, "Hard hanging up channel %s\n", ast_channel_name(chan));
+		/* Another thread could be servicing the channel right now...
+		 * so acquire the block lock to ensure we have exclusive control first. */
+		rpt_mutex_lock(&myrpt->blocklock);
 		ast_hangup(chan);
+		rpt_mutex_unlock(&myrpt->blocklock);
 	}
 }
 
 /*! \note myrpt->lock must be held when calling */
-static inline void hangup_link_chan(struct rpt_link *l)
+static inline void hangup_link_chan(struct rpt *myrpt, struct rpt_link *l)
 {
 	if (l->chan) {
-		safe_hangup(l->chan);
+		safe_hangup(myrpt, l->chan);
 		l->chan = NULL;
 	}
 }
@@ -4196,20 +4200,20 @@ static void remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 			else
 				l->disctime = DISC_TIME;
 			rpt_mutex_lock(&myrpt->lock);
-			hangup_link_chan(l);
+			hangup_link_chan(myrpt, l);
 			rpt_mutex_unlock(&myrpt->lock);
 			return;
 		}
 
 		if (l->retrytimer) {
 			rpt_mutex_lock(&myrpt->lock);
-			hangup_link_chan(l);
+			hangup_link_chan(myrpt, l);
 			rpt_mutex_unlock(&myrpt->lock);
 			return;
 		}
 		if (l->outbound && (l->retries++ < l->max_retries) && (l->hasconnected)) {
 			rpt_mutex_lock(&myrpt->lock);
-			hangup_link_chan(l);
+			hangup_link_chan(myrpt, l);
 			l->hasconnected = 1; /*! \todo BUGBUG XXX l->hasconnected has to be true to get here, why set it again? Is this a typo? */
 			l->retrytimer = RETRY_TIMER_MS;
 			l->elaptime = 0;
@@ -4248,7 +4252,7 @@ static void remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 
 	rpt_mutex_lock(&myrpt->lock);
 	/* hang-up on call to device */
-	hangup_link_chan(l);
+	hangup_link_chan(myrpt, l);
 	rpt_mutex_unlock(&myrpt->lock);
 
 	ast_hangup(l->pchan);


mkmer:
I needed to add `myrpt` to all of the `hangup_link_chan()` calls.  After that it still crashes with a ton of these (see end of message):
This appears to be directly a result of calling `ast_check_hangup()` here:
https://github.com/AllStarLink/app_rpt/blob/34572442a503b539a92fa2619cc1e75847c044d8/apps/app_rpt.c#L6852

If I remove it, this crash goes away and we get a different error/crash (same as case 2 in the issue comments):
```
[2025-01-20 12:45:04.793] WARNING[756092][C-00000023]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text !NEWKEY1! on IAX2/149.154.11.243:24141-4553
[2025-01-20 12:45:04.795] WARNING[756092][C-00000023]: app_rpt/rpt_channel.c:470 send_newkey: Failed to send text !NEWKEY1! on IAX2/149.154.11.243:24141-4553
  == Spawn extension (radio-secure, 287893, 1) exited non-zero on 'IAX2/149.154.11.243:24141-4553'
    -- Hungup 'IAX2/149.154.11.243:24141-4553'
```


Ton of errors to follow:
```
[2025-01-20 12:34:13.402] ERROR[740647]: channel.c:3091 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.404] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x32b) [0x55d8b0f5b4db]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.404] ERROR[740647]: channel.c:3118 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.405] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x47c) [0x55d8b0f5b62c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.405] ERROR[740647]: channel.c:3120 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.407] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x4ad) [0x55d8b0f5b65d]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.407] ERROR[740647]: channel.c:3039 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.408] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0xbc) [0x55d8b0f5b26c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.408] ERROR[740647]: channel.c:3055 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.410] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x167) [0x55d8b0f5b317]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.410] ERROR[740647]: channel.c:3084 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.411] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x248) [0x55d8b0f5b3f8]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.411] ERROR[740647]: channel.c:3091 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.413] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x32b) [0x55d8b0f5b4db]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.413] ERROR[740647]: channel.c:3118 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.414] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x47c) [0x55d8b0f5b62c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.414] ERROR[740647]: channel.c:3120 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.416] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x4ad) [0x55d8b0f5b65d]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.416] ERROR[740647]: channel.c:3039 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.417] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0xbc) [0x55d8b0f5b26c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.417] ERROR[740647]: channel.c:3055 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.419] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x167) [0x55d8b0f5b317]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.419] ERROR[740647]: channel.c:3084 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.420] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x248) [0x55d8b0f5b3f8]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.420] ERROR[740647]: channel.c:3091 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.422] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x32b) [0x55d8b0f5b4db]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.422] ERROR[740647]: channel.c:3118 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.423] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x47c) [0x55d8b0f5b62c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.423] ERROR[740647]: channel.c:3120 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.424] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x4ad) [0x55d8b0f5b65d]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.425] ERROR[740647]: channel.c:3039 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.426] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0xbc) [0x55d8b0f5b26c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.426] ERROR[740647]: channel.c:3055 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.427] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x167) [0x55d8b0f5b317]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.427] ERROR[740647]: channel.c:3084 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.428] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x248) [0x55d8b0f5b3f8]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.428] ERROR[740647]: channel.c:3091 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.429] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x32b) [0x55d8b0f5b4db]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.429] ERROR[740647]: channel.c:3118 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.430] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x47c) [0x55d8b0f5b62c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.430] ERROR[740647]: channel.c:3120 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.431] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x4ad) [0x55d8b0f5b65d]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.431] ERROR[740647]: channel.c:3039 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.432] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0xbc) [0x55d8b0f5b26c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.432] ERROR[740647]: channel.c:3055 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.432] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x167) [0x55d8b0f5b317]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.432] ERROR[740647]: channel.c:3084 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.433] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x248) [0x55d8b0f5b3f8]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.433] ERROR[740647]: channel.c:3091 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.434] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x32b) [0x55d8b0f5b4db]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.434] ERROR[740647]: channel.c:3118 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.435] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x47c) [0x55d8b0f5b62c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.435] ERROR[740647]: channel.c:3120 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.435] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x4ad) [0x55d8b0f5b65d]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.435] ERROR[740647]: channel.c:3039 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.436] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0xbc) [0x55d8b0f5b26c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.436] ERROR[740647]: channel.c:3055 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.437] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x167) [0x55d8b0f5b317]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.437] ERROR[740647]: channel.c:3084 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.437] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x248) [0x55d8b0f5b3f8]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.437] ERROR[740647]: channel.c:3091 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.438] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x32b) [0x55d8b0f5b4db]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.438] ERROR[740647]: channel.c:3118 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.438] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x47c) [0x55d8b0f5b62c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.438] ERROR[740647]: channel.c:3120 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.439] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x4ad) [0x55d8b0f5b65d]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.439] ERROR[740647]: channel.c:3039 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.440] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0xbc) [0x55d8b0f5b26c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.440] ERROR[740647]: channel.c:3055 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.440] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x167) [0x55d8b0f5b317]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.440] ERROR[740647]: channel.c:3084 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.441] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x248) [0x55d8b0f5b3f8]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.441] ERROR[740647]: channel.c:3091 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.441] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x32b) [0x55d8b0f5b4db]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.441] ERROR[740647]: channel.c:3118 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.442] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x47c) [0x55d8b0f5b62c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.442] ERROR[740647]: channel.c:3120 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.442] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x4ad) [0x55d8b0f5b65d]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.442] ERROR[740647]: channel.c:3039 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.443] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1a3) [0x55d8b0f1a823]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0xbc) [0x55d8b0f5b26c]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]

[2025-01-20 12:34:13.443] ERROR[740647]: channel.c:3055 ast_waitfor_nandfds: FRACK!, Failed assertion bad magic number 0x0 for object 0x7ff7a80014f0 (0)
[2025-01-20 12:34:13.443] ERROR[740647]:   Got 7 backtrace records
# 0: /usr/sbin/asterisk(__ao2_unlock+0x13a) [0x55d8b0f1a96a]
# 1: /usr/sbin/asterisk(ast_waitfor_nandfds+0x167) [0x55d8b0f5b317]
# 2: /usr/sbin/asterisk(ast_waitfor_n+0x14) [0x55d8b0f5bb74]
# 3: /usr/lib/asterisk/modules/app_rpt.so(+0x1fdbe) [0x7ff7d4a1fdbe]
# 4: /usr/sbin/asterisk(+0x1bf26c) [0x55d8b106d26c]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x891c4) [0x7ff7d86bf1c4]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x10985c) [0x7ff7d873f85c]
```

InterLinked1:
What's line 3091 in your channel.c?

mkmer:
3091 looks like this:
```
        for (x = 0; x < n; x++) {
                ast_channel_lock(c[x]);
                for (y = 0; y < ast_channel_fd_count(c[x]); y++) {
                        fdmap[max].fdno = y;  /* fd y is linked to this pfds */
                        fdmap[max].chan = x;  /* channel x is linked to this pfds */
                        max += ast_add_fd(&pfds[max], ast_channel_fd(c[x], y));
                }
                CHECK_BLOCKING(c[x]);
                ast_channel_unlock(c[x]);
        }
```
`ast_channel_unlock()` is at 3091

mkmer:
OK - I think I've compiled with debug and better back trace flags -> lets see if this is any more helpful:
[core-asterisk-2025-01-20T21-05-59Z-full.txt](https://github.com/user-attachments/files/18482489/core-asterisk-2025-01-20T21-05-59Z-full.txt)
[core-asterisk-2025-01-20T21-05-59Z-info.txt](https://github.com/user-attachments/files/18482491/core-asterisk-2025-01-20T21-05-59Z-info.txt)
[core-asterisk-2025-01-20T21-05-59Z-locks.txt](https://github.com/user-attachments/files/18482492/core-asterisk-2025-01-20T21-05-59Z-locks.txt)
[core-asterisk-2025-01-20T21-05-59Z-thread1.txt](https://github.com/user-attachments/files/18482493/core-asterisk-2025-01-20T21-05-59Z-thread1.txt)
[core-asterisk-2025-01-20T21-05-59Z-brief.txt](https://github.com/user-attachments/files/18482494/core-asterisk-2025-01-20T21-05-59Z-brief.txt)

mkmer:
I had a theory that we are trying to hang up the channel twice -> once due to `return -1;` and once due to the `softhangup()` code.

Commenting out `softhangup` stops the crash when we execute  `ast_check_hangup(chan)`.  but we still crash with the `return -1;`

Wish I understood what the return codes are used for - did a bit of digging and can't find any docs.

mkmer:
SO, changing this:
https://github.com/AllStarLink/app_rpt/blob/e4dee9018642037100f981eea2ad96efddd4b877/apps/app_rpt.c#L4163

to `ast_softhangup_no_lock()` AND 

https://github.com/AllStarLink/app_rpt/blob/e4dee9018642037100f981eea2ad96efddd4b877/apps/app_rpt.c#L6848

removing `ast_check_hangup(chan)` but still returning -1 also prevents a crash.

Again, I'm not sure if it's the "right way", but it doesn't crash :)

mkmer:
This is the set of changes (probably clearer this way):
https://github.com/mkmer/app_rpt/tree/hangup_crash_fix

InterLinked1:
> SO, changing this:
> 
> https://github.com/AllStarLink/app_rpt/blob/e4dee9018642037100f981eea2ad96efddd4b877/apps/app_rpt.c#L4163
> 
> to `ast_softhangup_no_lock()` AND
> 
> https://github.com/AllStarLink/app_rpt/blob/e4dee9018642037100f981eea2ad96efddd4b877/apps/app_rpt.c#L6848
> 
> removing `ast_check_hangup(chan)` but still returning -1 also prevents a crash.
> 
> Again, I'm not sure if it's the "right way", but it doesn't crash :)

I don't think `ast_softhangup_no_lock` is correct. It's intended for usage where the channel is already locked, which isn't the case here.

To be clear, the channel *may* be locked by another thread, which seems probable if it's being serviced by another thread too. But no_lock is if the current thread has already locked the channel, which it doesn't appear we have.

`ast_check_hangup` doesn't acquire the channel lock, so using `ast_check_hangup_locked` may be the more appropriate thing to do in that case. Could you scrap both of our changes from HEAD thus far and try changing your `ast_check_hangup` to `ast_check_hangup_locked` to see if that helps?

mkmer:
I think I found at least the description:
 * \param execute a function callback to execute the application. It should return
 *                non-zero if the channel needs to be hung up.
non-zero is calling a hangup.  
Maybe we shouldn't be hanging up at the `softhangup` location if the PBX has it, then `return -1;` will do it's job?

mkmer:
This is updated to what may be more appropriate (still has `ast_check_hangup()` removed :
https://github.com/mkmer/app_rpt/tree/hangup_crash_fix

Essentially remove the `softhangup`, if the rpt_exec thread still has control on a hangup message, don't hangup as it will be managed by the pbx with `return -1;`

Still not really sure why `ast_check_hangup()` is causing issues.
This is a crash report for the above code with the `||ast_check_hangup()` included. 
[core-asterisk-2025-01-21T13-51-44Z-info.txt](https://github.com/user-attachments/files/18491850/core-asterisk-2025-01-21T13-51-44Z-info.txt)
[core-asterisk-2025-01-21T13-51-44Z-locks.txt](https://github.com/user-attachments/files/18491851/core-asterisk-2025-01-21T13-51-44Z-locks.txt)
[core-asterisk-2025-01-21T13-51-44Z-thread1.txt](https://github.com/user-attachments/files/18491852/core-asterisk-2025-01-21T13-51-44Z-thread1.txt)
[core-asterisk-2025-01-21T13-51-44Z-brief.txt](https://github.com/user-attachments/files/18491853/core-asterisk-2025-01-21T13-51-44Z-brief.txt)
[core-asterisk-2025-01-21T13-51-44Z-full.txt](https://github.com/user-attachments/files/18491854/core-asterisk-2025-01-21T13-51-44Z-full.txt)

InterLinked1:
> That didn't work either. I think there are two crash sources (this is my theory anyway - having a difficult time following the asterisk side)
> 
> #1: We have a channel hangup message, which caused `softhangup` to execute while `rpt_exec` is still working. When we `return -1;` in rpt_exec, the asterisk side of things tries to hangup the channel in a deadlock due to softhangup? (or it force hangs up stuff). This is why `return 0;` or `softhangup_nolock()` make the crash go away (maybe?).

Returning 0 from a dialplan application (`Rpt` in this case) indicates dialplan execution can continue.
-1 indicates that dialplan application should stop, and the channel will then typically be hung up. The KEEPALIVE thing is a separate thing mostly unique to `Rpt` where the channel would be kept alive after the dialplan application had returned.

A softhangup can be called from any thread to queue a hangup on a channel (typically owned by another thread). It will typically cause internal applications like `ast_read` to return -1, NULL, a NULL frame, etc. - think of it like closing a file descriptor that another thread is using, or signaling it, then causing it to return and then exit.

The PBX thread normally owns a channel, in which case returning -1 would be sufficient to hang it up. But due to the whole keepalive thing where a separate thread is then responsible for the channel, that changes thngs there.

> #2 Is simply testing the channel with `ast_check_hangup` is crashing - this one really has me lost. Unless `chan` has already become NULL (I don't know). Looking at the asterisk function, it makes no sense.

chan wouldn't be NULL (or it would be visible in the backtrace). But if chan itself is invalid for some reason, then that could indeed be an issue.

> Is there a description of the expected return codes from rpt_exec (or any xxx_exec) somewhere? I've hunted a bit and don't see one.

As I said above, 0 means continue and -1 means (failure). Failure can be handled by the TryExec application, but otherwise dialplan will terminate and the channel is hungup.

There used to be more return values like AST_PBX_KEEPALIVE, but those have mostly been removed.

mkmer:
Is this the code that calls `rpt_exec()` and deals with the return code?
https://github.com/asterisk/asterisk/blob/50a25ac8474d7900ba59a68ed4fd942074082435/apps/app_dial.c#L2082-L2092

mkmer:
I'm not sure what the next step is on this one.  While there are a couple of ways to "not crash", it sounds like they are not the right way to deal with the deadlock on the channel.

InterLinked1:
> Is this the code that calls rpt_exec() and deals with the return code?
https://github.com/asterisk/asterisk/blob/50a25ac8474d7900ba59a68ed4fd942074082435/apps/app_dial.c#L2082-L2092

Nope, Dial is another application, just like Rpt. The code would be part of pbx.c

> I'm not sure what the next step is on this one. While there are a couple of ways to "not crash", it sounds like they are not the right way to deal with the deadlock on the channel.

Yeah, I think what would probably be most helpful, at least to me, is if there is a way I can reliably reproduce this myself. That was instrumental in getting the other change resolved as we could quickly test and iterate. If there's any way to do that here, that would be great.

Unfortunately, I have to deal with moving some equipment, and I'm going to be out of town next week, so I might be a bit slow in being able to look into this.

mkmer:
You need the server node and a client node:
This python script (I hacked it together - so never mind comments they are probably wrong.)
[fastconnect.zip](https://github.com/user-attachments/files/18512459/fastconnect.zip)
uses pyami_asterisk library: https://pypi.org/project/asterisk-ami/

If you have another way to issue AMI commands, just use the commands from the script.
`action_string = f"Command\nCommand: rpt cmd {MY_NODE} ilink 13 {TARGET_NODE}"`
`action_string = f"Command\nCommand: rpt cmd {MY_NODE} ilink 11 {TARGET_NODE}"`

Anything under .10 seconds typically will trigger the failed to send !NEWKEY1! error then soon after a 2 to 0 message. (and crash anywhere in between)

I found you can't issue them through the command line and you can't create a macro with both connect/disconnect in it - neither are fast enough to get it to fail.

Thinking about it a bit more -> maybe putting a sleep in the right spot in the `rpt_exec` function could make it stay there long enough to trigger the condition more easily.  

I'll be here when you have something :)

mkmer:
Oh, one more thing: my last iteration (remove `softhangup` from `safe_hangup`, and `return -1;` has some problem where my remote node becomes disconnected even with Permanent  checked - no crash on either end.  So whatever is going on, `return -1;` by itself may be doing something odd and keeping my remote node from connecting long enough it times out on retries.

mkmer:
> > Is this the code that calls rpt_exec() and deals with the return code?
> > https://github.com/asterisk/asterisk/blob/50a25ac8474d7900ba59a68ed4fd942074082435/apps/app_dial.c#L2082-L2092
> 
> Nope, Dial is another application, just like Rpt. The code would be part of pbx.c
> 
> > I'm not sure what the next step is on this one. While there are a couple of ways to "not crash", it sounds like they are not the right way to deal with the deadlock on the channel.
> 
> Yeah, I think what would probably be most helpful, at least to me, is if there is a way I can reliably reproduce this myself. That was instrumental in getting the other change resolved as we could quickly test and iterate. If there's any way to do that here, that would be great.
> 
> Unfortunately, I have to deal with moving some equipment, and I'm going to be out of town next week, so I might be a bit slow in being able to look into this.

I don't know how I got that link... this looks more like the spot:
https://github.com/asterisk/asterisk/blob/50a25ac8474d7900ba59a68ed4fd942074082435/main/pbx.c#L4353-L4375

If we exit non 0, It looks like we fall down to the `ast_hangup()` while waiting on a `soft_hangup` in the other thread.
Look forward to what you figure out.

mkmer:
```
[2025-01-20 11:24:06.322] WARNING[692197][C-00000024]: channel.c:2630 ast_hangup: Hard hangup called by thread LWP 692197 on IAX2/149.154.11.243:24501-5692, while blocked by thread LWP 692009 in procedure ast_waitfor_nandfds!  Expect a failure
    -- Hungup 'IAX2/149.154.11.243:24501-5692'
```
I really believe this is pointing to us using `ast_hangup` because of the `return -1;` while the soft hangup is doing it's thing.
I don't think `rpt_exec` has the channel locked which allows `ast_softhangup` to get a lock.  Then before it can finish, the result of `return -1;` is executing a hard hangup code (quoted above).

mkmer:
@InterLinked1 Any new thoughts on how we can fix this?

InterLinked1:
> ```
> [2025-01-20 11:24:06.322] WARNING[692197][C-00000024]: channel.c:2630 ast_hangup: Hard hangup called by thread LWP 692197 on IAX2/149.154.11.243:24501-5692, while blocked by thread LWP 692009 in procedure ast_waitfor_nandfds!  Expect a failure
>     -- Hungup 'IAX2/149.154.11.243:24501-5692'
> ```
> 
> I really believe this is pointing to us using `ast_hangup` because of the `return -1;` while the soft hangup is doing it's thing. I don't think `rpt_exec` has the channel locked which allows `ast_softhangup` to get a lock. Then before it can finish, the result of `return -1;` is executing a hard hangup code (quoted above).

Sorry, it's been a bit since I've looked at this. I think `while blocked by thread LWP 692009 in procedure ast_waitfor_nandfds` is key here, what is that thread doing here exactly? It's indeed true that this probably shouldn't be happening at that point in time.

The best way would be if would could get a backtrace from that scenario, is that something you can reproduce easily?

We can force it to crash at that line if you add `ast_assert(0)` at channel.c, line 2630, whenever that condition is triggered. That should ensure a core is dumped and allow you to get a backtrace from that instant.

InterLinked1:
> [core-asterisk-2025-02-20T00-53-49Z-full.txt](https://github.com/user-attachments/files/18877800/core-asterisk-2025-02-20T00-53-49Z-full.txt) [core-asterisk-2025-02-20T00-53-49Z-info.txt](https://github.com/user-attachments/files/18877801/core-asterisk-2025-02-20T00-53-49Z-info.txt) [core-asterisk-2025-02-20T00-53-49Z-locks.txt](https://github.com/user-attachments/files/18877802/core-asterisk-2025-02-20T00-53-49Z-locks.txt) [core-asterisk-2025-02-20T00-53-49Z-thread1.txt](https://github.com/user-attachments/files/18877803/core-asterisk-2025-02-20T00-53-49Z-thread1.txt) [core-asterisk-2025-02-20T00-53-49Z-brief.txt](https://github.com/user-attachments/files/18877804/core-asterisk-2025-02-20T00-53-49Z-brief.txt)
> 
> This is the core dump from adding `assert(0);` to the hangup routine. Nothing interesting appeared in the debug log file - just asterisk restart message.
> 
> I placed the assert(0) here: https://github.com/asterisk/asterisk/blob/d63c3e80fcfc3fc73220482d1794222a9d9b8029/main/channel.c#L2628-L2632 Except I'm on 20.x so it's not exactly at 2632. I hope it's the right spot...

This looks suspicious to me, if it's crashing in timing.c. But apparently handle is NULL there and it shouldn't be.

Just to confirm, do you have a timing module loaded? Weird crashes can occur in Asterisk if you don't have one loaded, and there isn't anything that enforces you have one loaded.

mkmer:
I'm fairly sure it's loaded.  I will recompile everything just to make sure I didn't cross pollinate one of the files or something odd like that.
I'm still a bit suspect on where you wanted me to put the `assert(0)` - the line numbers didn't match up quite right.  There is an `assert(0)` a few lines before related to the "expect problems" error.

InterLinked1:
> [2025-01-20 11:24:06.322] WARNING[692197][C-00000024]: channel.c:2630 ast_hangup: Hard hangup called by thread LWP 692197 on IAX2/149.154.11.243:24501-5692, while blocked by thread LWP 692009 in procedure ast_waitfor_nandfds!  Expect a failure

Yeah, that should do it, interesting. Just wanted to make sure that wasn't it!

Also, the line is the one containing the error above (Hard hangup)

mkmer:
Sorry - want to be clear so my tests are helpful:
The error message is here:
https://github.com/asterisk/asterisk/blob/2d7948fa483c8703cf9948811fd67147f4d560d4/main/channel.c#L2622-L2626

And I added and `assert(0)` here (before the `->hangup()`)

https://github.com/asterisk/asterisk/blob/2d7948fa483c8703cf9948811fd67147f4d560d4/main/channel.c#L2629-L2631

So - L2626 should have given us a core dump right? (We got the error message).  Does it keep going after the `assert(0)` and over write the core when we actually crash? (I'm not familiar with how this all works).

InterLinked1:
> Sorry - want to be clear so my tests are helpful: The error message is here: https://github.com/asterisk/asterisk/blob/2d7948fa483c8703cf9948811fd67147f4d560d4/main/channel.c#L2622-L2626
> 
> And I added and `assert(0)` here (before the `->hangup()`)
> 
> https://github.com/asterisk/asterisk/blob/2d7948fa483c8703cf9948811fd67147f4d560d4/main/channel.c#L2629-L2631
> 
> So - L2626 should have given us a core dump right? (We got the error message). Does it keep going after the `assert(0)` and over write the core when we actually crash? (I'm not familiar with how this all works).

Ah, I didn't realize there was already an assert (my bad, should've looked).

It should crash assuming you compiled with `DO_CRASH`. You can run menuselect and go to Developer Compile Flags and make sure that is selected. That will force Asterisk to dump core when an assertion fails.

mkmer:
So that's the problem - no `DO_CRASH` so it keeps going and over writes the core when it actually crashes.  

I don't see a "DO CRASH" flag in menu select under compiler flags.

InterLinked1:
> So that's the problem - no `DO_CRASH` so it keeps going and over writes the core when it actually crashes.
> 
> I don't see a "DO CRASH" flag in menu select under compiler flags.

That probably means you didn't run `configure` with developer mode enabled, so those options aren't available.

I should update the README probably but you can use `--devmode` with `phreaknet install` if you are running that way. Otherwise, rerun configure like so: https://github.com/InterLinked1/phreakscript/blob/master/phreaknet.sh#L1380

Allan-N:
FYI : making sure that `build-asl3` also has a devmode is on my TODO list (we all gotta be on the same page)

mkmer:
This set is with `DO_CRASH` and developer build + better backtrace enabled.  

[core-asterisk-2025-02-21T18-55-59Z-full.txt](https://github.com/user-attachments/files/18914236/core-asterisk-2025-02-21T18-55-59Z-full.txt)
[core-asterisk-2025-02-21T18-55-59Z-info.txt](https://github.com/user-attachments/files/18914237/core-asterisk-2025-02-21T18-55-59Z-info.txt)
[core-asterisk-2025-02-21T18-55-59Z-locks.txt](https://github.com/user-attachments/files/18914238/core-asterisk-2025-02-21T18-55-59Z-locks.txt)
[core-asterisk-2025-02-21T18-55-59Z-thread1.txt](https://github.com/user-attachments/files/18914239/core-asterisk-2025-02-21T18-55-59Z-thread1.txt)
[core-asterisk-2025-02-21T18-55-59Z-brief.txt](https://github.com/user-attachments/files/18914240/core-asterisk-2025-02-21T18-55-59Z-brief.txt)

InterLinked1:
> This set is with `DO_CRASH` and developer build + better backtrace enabled.
> 
> [core-asterisk-2025-02-21T18-55-59Z-full.txt](https://github.com/user-attachments/files/18914236/core-asterisk-2025-02-21T18-55-59Z-full.txt) [core-asterisk-2025-02-21T18-55-59Z-info.txt](https://github.com/user-attachments/files/18914237/core-asterisk-2025-02-21T18-55-59Z-info.txt) [core-asterisk-2025-02-21T18-55-59Z-locks.txt](https://github.com/user-attachments/files/18914238/core-asterisk-2025-02-21T18-55-59Z-locks.txt) [core-asterisk-2025-02-21T18-55-59Z-thread1.txt](https://github.com/user-attachments/files/18914239/core-asterisk-2025-02-21T18-55-59Z-thread1.txt) [core-asterisk-2025-02-21T18-55-59Z-brief.txt](https://github.com/user-attachments/files/18914240/core-asterisk-2025-02-21T18-55-59Z-brief.txt)

Thanks, this provides a lot of clarity and confirms what I was thinking.

The crash is occuring because when `ast_hangup` is called, nobody else should be using the channel. However, `app_rpt` is still using it, was we inferred already. See thread 59 in the backtrace. Specifically at line 5357:

```
myrpt->scram++;
rpt_mutex_lock(&myrpt->blocklock);
who = ast_waitfor_n(cs1, n, &ms);
rpt_mutex_unlock(&myrpt->blocklock);
if (who == NULL) {
	ms = 0;
}
```

It's stuck in `ast_waitfor_n` at that point, and we need to make sure that in this case, we synchronously kick it out of there and wait for the `rpt` thread to exit, before returning to the PBX.

I haven't given any thought to the best way to do that yet (need to get back to a meeting now), but that's what's going on here.

mkmer:
Well... 
I think it's ok to NOT to `return -1` from the thread, as the thread with the `softhangup` will hang it up, no need to tell upstream there was a hangup...  I don't think we need to deal with hangup in 2 places... But as usual, I'm still learning the ins and outs.

InterLinked1:
> Well... I think it's ok to NOT to `return -1` from the thread, as the thread with the `softhangup` will hang it up, no need to tell upstream there was a hangup... I don't think we need to deal with hangup in 2 places... But as usual, I'm still learning the ins and outs.

That might work in this particular scenario, but that's not really the "correct" solution. It will be a bit more work to adjust things but it's the right thing to do here.

mkmer:
Adding a `sleep(2);` or larger as needed on Line 6817 helps make triggering the crash easier.  
https://github.com/AllStarLink/app_rpt/blob/7b926425281203552b48840f9c5092af669263e9/apps/app_rpt.c#L6815-L6818

Since this is a race condition, it may be harder to trigger on faster hardware.

InterLinked1:
> ```
> [2025-01-20 11:24:06.322] WARNING[692197][C-00000024]: channel.c:2630 ast_hangup: Hard hangup called by thread LWP 692197 on IAX2/149.154.11.243:24501-5692, while blocked by thread LWP 692009 in procedure ast_waitfor_nandfds!  Expect a failure
>     -- Hungup 'IAX2/149.154.11.243:24501-5692'
> ```
> 
> I really believe this is pointing to us using `ast_hangup` because of the `return -1;` while the soft hangup is doing it's thing. I don't think `rpt_exec` has the channel locked which allows `ast_softhangup` to get a lock. Then before it can finish, the result of `return -1;` is executing a hard hangup code (quoted above).

So far, I'm not able to reproduce this.
Could you provide a minimally reproducible config and the full logs (not backtrace) from a run of this?
Also, in your logger.conf, please set `dateformat=%F %T.%3q`

mkmer:
Check out https://github.com/AllStarLink/app_rpt/pull/525 -> 100% solves the race.

InterLinked1:
> Check out #525 -> 100% solves the race.

Agreed, that seems like a good solution since it can't return -1 with it in the list. I wasn't sure if moving it later would have an adverse impact on the telemetry or something related but this is great if it works!

I'd still like to try capturing this in a test though so in the future we'll be able to detect regressions, etc. In a new Asterisk system with two nodes, how do you get into this state? I was thinking that you establish an incoming call to the "server" node, then use the two ilink commands, but that doesn't seem to do it, so I think I'm missing something...

mkmer:
The process is to "call" and "hangup" quickly with they ilink command - nothing established to start with.  Ideally, the hangup command arrives just after the link list is updated, and before the `return -1` logic is evaluated (AND waitfor_n has had a chance to updates it's list to include the new channel).

I believe it happens MUCH faster/easier on a node with some audio traffic and at list one other link causing the `waitfor_n` to make a new list much faster (now that I "understand" what the root cause was).

InterLinked1:
> The process is to "call" and "hangup" quickly with the ilink command - nothing established to start with. Ideally, the hangup command arrives just after the link list is updated, and before the `return -1` logic is evaluated (AND waitfor_n has had a chance to updates it's list to include the new channel).
> 
> I believe it happens MUCH faster/easier on a node with some audio traffic and at list one other link causing the `waitfor_n` to make a new list much faster (now that I "understand" what the root cause was).

I have the test executing the commands now but there is too much time passing between the 13 and 11 I think. The machine is from 2006, so it's possible it's just so slow with everything going on. I'm going to keep tweaking to see if I can bring them closer.

From the beginning, I've noticed these messages constantly:
`app_rpt/rpt_channel.c: time out dtmf_local_timer=1`

Do you see this too with debug set to 10? Wondering if I need an option to disable this or if that's just normal.

InterLinked1:
> The process is to "call" and "hangup" quickly with the ilink command - nothing established to start with. Ideally, the hangup command arrives just after the link list is updated, and before the `return -1` logic is evaluated (AND waitfor_n has had a chance to updates it's list to include the new channel).
> 

Okay, I actually do have the right timing, but nothing is crashing for me.

I'm curious, what happens if you manually run this test on your node?

extensions.conf:
```
[wait]
exten => s,1,Answer(,i)
	same => n,Wait(999)
	same => n,Hangup()

; Manually launch test using 'channel originate Local/s@wait extension ilink@timer'
[timer]
exten => ilink,1,Answer(,i)
	same => n,Wait(3) ; Rpt() currently rejects calls until 2 seconds after module load
	same => n,System(asterisk -rx "rpt cmd 456 ilink 13 123")
	same => n,Wait(0.01)
	same => n,System(asterisk -rx "rpt cmd 456 ilink 11 123")
	same => n,Hangup()

[repeaters]
exten => 123,1,Set(CALLERID(num)=456)
	same => n,Rpt(${EXTEN})
```

rpt.conf:
```
[general]
node_lookup_method = file

[nodes]
123 = radio@127.0.0.1/123,NONE ; server
456 = radio@127.0.0.1/456,NONE ; client

[123]
rxchannel = DAHDI/pseudo

[456]
rxchannel = DAHDI/pseudo
```

iax.conf:
```
[general]
bandwidth=high
debug=yes
authdebug=yes

[radio]
type=user
username=radio
context=repeaters
```

modules.conf:
```
[modules]
autoload=no

require => app_dial
require => app_rpt
require => app_userevent
require => chan_dahdi
require => chan_iax2
require => codec_ulaw
require => func_callerid
require => pbx_config
require => res_curl
load => res_crypto
```

logger.conf:
```
[general]
dateformat=%F %T.%3q
```

mkmer:
I think this is finally making a bit of sense now (after looking at the code for "other stuff":
https://github.com/AllStarLink/app_rpt/blob/f99565842cc60dc71736c127621c64423d88b6bb/apps/app_rpt.c#L7269
https://github.com/AllStarLink/app_rpt/blob/f99565842cc60dc71736c127621c64423d88b6bb/apps/app_rpt.c#L5373
We are registering a channel with `ast_waitfor_n()` before `rpt_exec()` is finished.  Then the `return -1` is pulling a hard hangup and clearing the channel pointers which is "not ok".
Maybe we can delay adding to the link list until the very end of `rpt_exec()`?

> > The process is to "call" and "hangup" quickly with the ilink command - nothing established to start with. Ideally, the hangup command arrives just after the link list is updated, and before the `return -1` logic is evaluated (AND waitfor_n has had a chance to updates it's list to include the new channel).
> > I believe it happens MUCH faster/easier on a node with some audio traffic and at list one other link causing the `waitfor_n` to make a new list much faster (now that I "understand" what the root cause was).
> 
> I have the test executing the commands now but there is too much time passing between the 13 and 11 I think. The machine is from 2006, so it's possible it's just so slow with everything going on. I'm going to keep tweaking to see if I can bring them closer.
> 
> From the beginning, I've noticed these messages constantly: `app_rpt/rpt_channel.c: time out dtmf_local_timer=1`
> 
> Do you see this too with debug set to 10? Wondering if I need an option to disable this or if that's just normal.

YES - I took that debug line out.  I don't think it makes sense as it's "normal".

mkmer:
It usually takes more than 1 try to get it to crash (which makes more sense to me now).  Usually the 3rd or 4th time is the charm.  Maybe just add 4 loops to the dialplan.

InterLinked1:
> It usually takes more than 1 try to get it to crash (which makes more sense to me now). Usually the 3rd or 4th time is the charm. Maybe just add 4 loops to the dialplan.

Is that still true with the test I provided? I believe it happens on your system, but that's different from working with other configurations.
I also added the `sleep(2)` as you suggested but that didn't seem to help.

I am getting a bunch of these, so that's one issue uncovered so far, but I don't think this is related:
`[Mar 02 12:15:54] WARNING[432247]: asterisk.asterisk:249 errReceived: Asterisk 127.0.0.1 received error: WARNING: High fence violation of 0x7ff61c029dc0 allocated at app_rpt/rpt_link.c rpt_update_links() line 582`

InterLinked1:
Update: I was able to get it to fail once just now, but it didn't the time after.
I'll keep playing around with trying to make it reproduce more consistently, but at least I can see I think I'm on the right track here.

Allan-N:
Fixed with https://github.com/AllStarLink/app_rpt/pull/525

PR #461, opened 2025-01-21
app_rpt: Remove unreachable code

mkmer:
l->chan can't be true if it was false.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3229,9 +3229,6 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			break;
 		}
 		if ((!l->chan) && (!l->retrytimer) && l->outbound && (l->retries++ < l->max_retries) && (l->hasconnected)) {
-			if (l->chan)
-				ast_hangup(l->chan);
-			l->chan = NULL;
 			rpt_mutex_unlock(&myrpt->lock);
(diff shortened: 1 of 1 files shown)


InterLinked1 (approved):

Allan-N (approved):

apps/app_rpt.c
```diff
@@ -3229,9 +3229,6 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			break;
 		}
 		if ((!l->chan) && (!l->retrytimer) && l->outbound && (l->retries++ < l->max_retries) && (l->hasconnected)) {
-			if (l->chan)
-				ast_hangup(l->chan);
-			l->chan = NULL;
```
encbar5, line 3234: Why was line 3234 removed? This does not look like dead code.

encbar5 (changes requested):

apps/app_rpt.c
mkmer: I believe this is what the code is doing here:
If `((!l->chan)` aka `l->chan = NULL`, 
let's execute `if (l->chan)` which will be always false, 
then lets set `l->chan = NULL;` which it already was (right?)

encbar5: Fair point.

encbar5 (approved):
Are we intending on running a static analyzer on all the code?

PR #463, opened 2025-01-21
Multiple files: correct spelling

mkmer:
Typos everywhere!
Correct spelling in multiple files.
Activate codespell on the repository
A user can run `codespell -w --skip="./.git" --ignore-words=./dev/.codespellignore <file>` check/fix a file locally before committing.


README.md
```diff
@@ -62,7 +62,7 @@ If you want to manually install app_rpt et al., here is how:
 
 ### Compiling
 
-First, detection of the ALSA library needs to be readded to the build system, by applying the following patch: https://github.com/InterLinked1/phreakscript/blob/master/patches/alsa.diff
+First, detection of the ALSA library needs to be read to the build system, by applying the following patch: https://github.com/InterLinked1/phreakscript/blob/master/patches/alsa.diff
```
InterLinked1, line 65: This is not a typo, maybe should be re-added if readded is not correct?

apps/app_rpt/rpt_bridging.c
```diff
@@ -246,7 +246,7 @@ int __rpt_request(void *data, struct ast_format_cap *cap, enum rpt_chan_type cha
 	 * rpt_mutex_lock(&myrpt->lock);
 	 * afterwards,
 	 * if flags & RPT_LINK_CHAN.
-	 * This might not be necessary, but if it is, this should be readded. */
+	 * This might not be necessary, but if it is, this should be read. */
```
InterLinked1, line 249: not a typo

.clang-format
```diff
@@ -0,0 +1,191 @@
+---
```
InterLinked1, line 1: Could .clang-format and .codespellignore be moved into a subdirectory? Would be nice to not clutter up the root dir with various config files.

InterLinked1 (changes requested):

README.md
mkmer: Oops... I read that twice - read-ded  is all I could see :)

.clang-format
mkmer: I have always seen them in the root in other repos.  I'll see if we can move them.

Allan-N: Could we move the directory to `.github/support` (I don't know there are any rules about what can/cannot live there).  Another possibility would be changing `support` to `.support`.

apps/app_gps.c
```diff
@@ -54,7 +54,7 @@ a 'portable tent'.
 - - QTH
 . - X
 / - Dot
-0-9 Numerial Boxes
+0-9 Numeral Boxes
```
Allan-N, line 57: I wondered if the original spelling was intended to be "numerical" (vs. "numeral").  Looking at [APRS Display Symbols](http://aprs.net/vm/DOS/SYMBOLS.HTM) the symbol could be either "Numeral circle" or "NUMBERED CIRCLE".

KB4MDD: Changed to numbered.

encbar5:
1. If this pull request is focused on spelling, then the commit should probably not have the .clang-format file in it
2. Why would we only check spelling on master? Why would we not check it on development branches?

mkmer:
> 1. If this pull request is focused on spelling, then the commit should probably not have the .clang-format file in it

Fair point - I was working on both and this file "slipped' in.  It really shouldn't be in this PR.
> 
>     2. Why would we only check spelling on master? Why would we not check it on development branches?

Master branch is all we have, maybe we should have a dev branch too?  
Releases are tagged in the Master, so we can always get back to specific released code.  
Also, The Asterisk guys appear to run this way (with a tagged master rather than a dev branch).

It is a good question, I'll review the `on push` and `on pull request` directives.  
PR looks right because `master` is all we PR typically.

encbar5 (approved):

mkmer (commented):
Looks like we're good to go with this.
There were a few spelling errors that remained in `app_gps.c` - go figure :)

Allan-N (approved):

InterLinked1 (approved):

PR #464, opened 2025-01-22
app_gps: Code updates and documentation

KB4MDD:
This updates app_gps with the following changes:

Update the code to use the asterisk coding guidelines. Added documentation to all functions.
Added documentation to gps.conf.
Added functions: GPS_READ and APRS_SENDTT.
Added the cli command gps status.
Added the ability to select the primary or alternate symbol table. Replaced depreciated ast_gethostbyname.
Addressed issues with module unload not completing timely.

GPS_READ() function replaces the need to create the gps.dat file stored in the /tmp directory.  The application previously wrote to the disk once per second, which included a system call to rename a temporary file.

app_rpt and chan_echolink have been updated to use GPS_READ to retrieve the gps data, instead of reading a data file.

APRS_SENDTT(stanza, overlay) = callsign, function replaces the pipe that was used by app_rpt to send APRStt requests to app_gps.

APSTAR is now used as the tocall for both APRS an APRStt posts.  This is being registered as the tocall for AllStar Link.

If elevation is available, it is included in the APRS post.

The cli command 'gps status' will allow the user to check the status of the attached GPS device.  It tells the user if the GPS device is connected, if it is locked to the satellites, and the current and default position.

The unlocked warning message is now only printed when it first occurs. The message is turned back on when we receive a lock on the satellite.

The gps.conf file now has an entry for icontable.  The default is '\' primary or '/' for the alternate symbol table.

For APRStt to work, the aprstt entry in app_rpt needs to be enabled.  That option has been added to rtp.conf and commented out.

There were a number of refinements to the code, along with fixes for memory leaks.

### DIFF
--- a/apps/app_gps.c
+++ b/apps/app_gps.c
@@ -513,16 +513,16 @@ static void *aprs_connection_thread(void *data)
 		if (sockfd < 0) {
 			ast_log(LOG_ERROR, "Error opening socket: %s\n", strerror(errno));
 			ast_mutex_unlock(&aprs_socket_lock);
-			/* Wait 500ms before trying again. */
-			usleep(500000);
+			/* Wait 1 second before trying again. */
+			sleep(1);
 			continue;
 		}
 		
 		if (ast_sockaddr_resolve_first_af(&addr, server, PARSE_PORT_IGNORE, AST_AF_INET)) {
 			ast_log(LOG_WARNING, "Server %s cannot be found!\n", server);
 			ast_mutex_unlock(&aprs_socket_lock);
-			/* Wait 500ms before trying again. */
-			usleep(500000);
+			/* Wait 1 second before trying again. */
+			sleep(1);
 			continue;
 		}
 		ast_sockaddr_set_port(&addr, atoi(port));
@@ -530,8 +530,8 @@ static void *aprs_connection_thread(void *data)
 		if (ast_connect(sockfd, &addr) < 0) {
 			ast_log(LOG_WARNING, "Cannot connect to server %s. Error: %s\n", server, strerror(errno));
 			ast_mutex_unlock(&aprs_socket_lock);
-			/* Wait 500ms before trying again. */
-			usleep(500000);
+			/* Wait 1 second before trying again. */
+			sleep(1);
 			continue;
 		}
 
@@ -1154,11 +1154,9 @@ static void *aprs_sender_thread(void *data)
 		/* See if we need to send live GPS or the default */
 		if (current_gps_position.is_valid) {
 			selected_position = current_gps_position;
-		} else {
-			if (this_def_position.is_valid && !ehlert) {
+		} else if (this_def_position.is_valid && !ehlert) {
 				selected_position = this_def_position;
 				selected_position.last_updated = time(NULL);
-			}
 		}
 		ast_mutex_unlock(&position_update_lock);
 		/* 
@@ -1166,12 +1164,11 @@ static void *aprs_sender_thread(void *data)
 		 * The last_updated time must be current so that
 		 * we know we are getting good GPS information.
 		 */
-		if (selected_position.is_valid && (selected_position.last_updated + GPS_VALID_SECS) >= time(NULL)) {
-			if (now >= (lastupdate + my_update_secs)) {
-				report_aprs(ctg, selected_position.latitude, selected_position.longitude, selected_position.elevation);
-				lastupdate = now;
-				my_update_secs = interval;
-			}
+		if (selected_position.is_valid && (selected_position.last_updated + GPS_VALID_SECS) >= time(NULL) &&
+			now >= (lastupdate + my_update_secs)) {
+			report_aprs(ctg, selected_position.latitude, selected_position.longitude, selected_position.elevation);
+			lastupdate = now;
+			my_update_secs = interval;
 		}
 		/* wait 1 second */
 		sleep(1);
@@ -1499,11 +1496,9 @@ static int gps_read_helper(struct ast_channel *chan, const char *cmd, char *data
 	/* See if we need to send live GPS or the default */
 	if (current_gps_position.is_valid) {
 		selected_position = current_gps_position;
-	} else {
-		if (general_def_position.is_valid) {
+	} else if (general_def_position.is_valid) {
 			selected_position = general_def_position;
 			selected_position.last_updated = time(NULL);
-		}
 	}
 	ast_mutex_unlock(&position_update_lock);
 	/* 
--- a/apps/app_gps.c
+++ b/apps/app_gps.c
@@ -238,8 +238,8 @@
 /*! Defines */
 #define	APRS_DEFAULT_SERVER "rotate.aprs.net"
 #define	APRS_DEFAULT_PORT "14580"
-#define	APRS_DEFAULT_COMMENT "AllStar Link Node"
-#define	APRSTT_DEFAULT_COMMENT "AllStar Link TT Report"
+#define	APRS_DEFAULT_COMMENT "Asterisk/app_rpt Node"
+#define	APRSTT_DEFAULT_COMMENT "Asterisk/app_rpt TT Report"
 #define	APRSTT_DEFAULT_OVERLAY '0'
 #define APRS_DEFAULT_ICON_TABLE '/'		/* primary table */
 #define	APRS_DEFAULT_ICON '-'			/* house */
@@ -283,11 +283,11 @@ static int gps_unlock_shown = 0;
  * \brief Position information structure
  */
 struct position_info {
-	int is_valid:1;			/* contains valid values indicator */
-	char latitude[25];		/* latitude format DDMM.SSS */
-	char longitude[25];		/* longitude format DDDMM.SSS */
-	char elevation[25];		/* elevation format VVVV.V */
-	time_t last_updated;	/* the time these values were last updated */
+	unsigned int is_valid:1;	/* contains valid values indicator */
+	char latitude[25];			/* latitude format DDMM.SSS */
+	char longitude[25];			/* longitude format DDDMM.SSS */
+	char elevation[25];			/* elevation format VVVV.V */
+	time_t last_updated;		/* the time these values were last updated */
 };
 
 static struct position_info current_gps_position;
@@ -296,7 +296,7 @@ static struct position_info general_def_position;
 /*!
  *\brief Enum for aprs sender info type.
  */
-enum Sender_Type {
+enum APRS_Sender_Type {
 	APRS,
 	APRSTT
 };
@@ -306,7 +306,7 @@ enum Sender_Type {
  */
 struct aprs_sender_info {
 	AST_LIST_ENTRY(aprs_sender_info) list;
-	enum Sender_Type type;					/* sender type enum */
+	enum APRS_Sender_Type type;					/* sender type enum */
 	char stanza[50]; 						/* stanza associated with this aprs thread */
 	pthread_t thread_id;					/* thread id for this sender */
 	ast_cond_t condition;					/* condition indicator for this sender */
@@ -496,7 +496,7 @@ static void *aprs_connection_thread(void *data)
 		}
 		sockfd = socket(AF_INET, SOCK_STREAM, 0);
 		if (sockfd < 0) {
-			ast_log(LOG_ERROR, "Error opening socket. Error: %s\n", strerror(errno));
+			ast_log(LOG_ERROR, "Error opening socket: %s\n", strerror(errno));
 			ast_mutex_unlock(&aprs_socket_lock);
 			/* Wait 500ms before trying again. */
 			usleep(500000);
@@ -524,9 +524,7 @@ static void *aprs_connection_thread(void *data)
 		sprintf(buf, "user %s pass %s vers Asterisk app_gps_V3\n", call, password);
 		
 		if (send(sockfd, buf, strlen(buf), 0) < 0) {
-			ast_log(LOG_WARNING, "Can not send sign on to server. Error: %s\n", strerror(errno));
-			close(sockfd);
-			sockfd = -1;
+			ast_log(LOG_WARNING, "Can not send sign on to server: %s\n", strerror(errno));
 			ast_mutex_unlock(&aprs_socket_lock);
 			continue;
 		}
@@ -553,7 +551,6 @@ static void *aprs_connection_thread(void *data)
 				break;
 			}
 			if (fds[0].revents & POLLIN) {
-				memset(buf, 0, sizeof(buf));
 				if (recv(sockfd, buf, sizeof(buf) - 1, 0) > 0) {
 					ast_debug(4, "APRS-IS: %s\n", buf);
 				}
@@ -588,7 +585,7 @@ static int report_aprs(char *ctg, char *lat, char *lon, char *elev)
 {
 	struct ast_config *cfg = NULL;
 	char *call, *comment, icon, icon_table;
-	char power, height, gain, dir, *val, servercall[200], buf[350], *cp;
+	char power, height, gain, dir, *val, servercall[20], buf[350], *cp;
 	struct ast_flags zeroflag = { 0 };
 	char elev_str[25];
 	float elev_f;
@@ -680,8 +677,8 @@ static int report_aprs(char *ctg, char *lat, char *lon, char *elev)
 	/* Reduce the precision of latitude and longitude
 	 *
 	 * Latitude is expressed as a fixed 8-character field, in degrees and decimal
-     * minutes (to two decimal places), followed by the letter N for north or 
-     * S for south.
+	 * minutes (to two decimal places), followed by the letter N for north or 
+	 * S for south.
 	 * Longitude is expressed as a fixed 9-character field, in degrees and decimal
 	 * minutes (to two decimal places), followed by the letter E for east or 
 	 * W for west.
@@ -698,11 +695,12 @@ static int report_aprs(char *ctg, char *lat, char *lon, char *elev)
 	}
 	
 	/* Setup optional elevation */
-	memset(elev_str, 0, sizeof(elev_str));
 	elev_f = 0;
 	sscanf(elev, "%f", &elev_f);
 	if (elev_f > 0) {
 		snprintf(elev_str, sizeof(elev_str), "/A=%06.0f", elev_f * 3.28);
+	} else {
+		elev_str[0] = '\0';
 	}
 
 	snprintf(buf, sizeof(buf), "%s>APSTAR,TCPIP*,qAC,%s:!%s%c%s%cPHG%d%d%d%d%s%s\r\n",
@@ -716,7 +714,7 @@ static int report_aprs(char *ctg, char *lat, char *lon, char *elev)
 		return -1;
 	}
 	if (send(sockfd, buf, strlen(buf), 0) < 0) {
-		ast_log(LOG_WARNING, "Can not send APRS (GPS) data. Error: %s\n", strerror(errno));
+		ast_log(LOG_WARNING, "Can not send APRS (GPS) data: %s\n", strerror(errno));
 		ast_mutex_unlock(&aprs_socket_lock);
 		return -1;
 	}
@@ -749,7 +747,7 @@ static int report_aprstt(char *ctg, char *lat, char *lon, char *theircall, char 
 {
 	struct ast_config *cfg = NULL;
 	char *call, *comment;
-	char *val, basecall[300], buf[300], buf1[100], *cp;
+	char *val, basecall[20], buf[300], buf1[100], *cp;
 	time_t t;
 	struct tm *tm;
 	struct ast_flags zeroflag = { 0 };
@@ -793,8 +791,8 @@ static int report_aprstt(char *ctg, char *lat, char *lon, char *theircall, char 
 	/* Reduce the precision of latitude and longitude
 	 *
 	 * Latitude is expressed as a fixed 8-character field, in degrees and decimal
-     * minutes (to two decimal places), followed by the letter N for north or 
-     * S for south.
+	 * minutes (to two decimal places), followed by the letter N for north or 
+	 * S for south.
 	 * Longitude is expressed as a fixed 9-character field, in degrees and decimal
 	 * minutes (to two decimal places), followed by the letter E for east or 
 	 * W for west.
@@ -826,7 +824,7 @@ static int report_aprstt(char *ctg, char *lat, char *lon, char *theircall, char 
 	}
 	
 	if (send(sockfd, buf, strlen(buf), 0) < 0) {
-		ast_log(LOG_WARNING, "Can not send APRS (APSTAR) data. Error: %s\n", strerror(errno));
+		ast_log(LOG_WARNING, "Can not send APRS (APSTAR) data: %s\n", strerror(errno));
 		ast_mutex_unlock(&aprs_socket_lock);
 		return -1;
 	}
@@ -899,14 +897,14 @@ static void *gps_reader(void *data)
 	/* Open the serial port configured for the GPS device */
 	fd = open(comport, O_RDWR);
 	if (fd == -1) {
-		ast_log(LOG_WARNING, "Cannot open serial port %s: Error %s\n", comport, strerror(errno));
+		ast_log(LOG_WARNING, "Cannot open serial port %s: %s\n", comport, strerror(errno));
 		goto err;
 	}
 
 	if (has_comport) {
 		memset(&mode, 0, sizeof(mode));
 		if (tcgetattr(fd, &mode)) {
-			ast_log(LOG_WARNING, "Unable to get serial parameters on %s: Error %s\n", comport, strerror(errno));
+			ast_log(LOG_WARNING, "Unable to get serial parameters on %s: %s\n", comport, strerror(errno));
 			close(fd);
 			goto err;
 		}
@@ -924,7 +922,7 @@ static void *gps_reader(void *data)
 		cfsetispeed(&mode, baudrate);
 		cfsetospeed(&mode, baudrate);
 		if (tcsetattr(fd, TCSANOW, &mode)) {
-			ast_log(LOG_WARNING, "Unable to set serial parameters on %s: Error %s\n", comport, strerror(errno));
+			ast_log(LOG_WARNING, "Unable to set serial parameters on %s: %s\n", comport, strerror(errno));
 			close(fd);
 			goto err;
 		}
@@ -933,7 +931,7 @@ static void *gps_reader(void *data)
 	/* Give the device a few milliseconds to come on-line */
 	usleep(100000);
 
-	/*! \todo we need to detail with someone unplugging the device */
+	/*! \todo we need to deal with someone unplugging the device */
 	
 	while (run_forever) {
 		
@@ -1063,7 +1061,7 @@ static void *aprs_sender_thread(void *data)
 	time_t now, lastupdate;
 	struct ast_flags zeroflag = { 0 };
 	struct position_info this_def_position, selected_position;
-	struct aprs_sender_info *sender_entry = (struct aprs_sender_info *) data;
+	struct aprs_sender_info *sender_entry = data;
 	
 	ctg = ast_strdupa(sender_entry->stanza);
 	
@@ -1185,7 +1183,7 @@ static void *aprstt_sender_thread(void *data)
 	int i, j, ttlist, ttoffset, ttslot, myoffset;
 	char *ctg, c;
 	char *val, *deflat, *deflon, *defelev, ttsplit, *ttlat, *ttlon;
-	char fname[200], lat[100], theircall[100], overlay;
+	char fname[200], lat[25], theircall[20], overlay;
 	FILE *mfp;
 	struct stat mystat;
 	time_t now, lastupdate;
@@ -1193,7 +1191,7 @@ static void *aprstt_sender_thread(void *data)
 	struct position_info this_def_position, selected_position;
 	struct timeval tv;
 	struct timespec ts = {0};
-	struct aprs_sender_info *sender_entry = (struct aprs_sender_info *) data;
+	struct aprs_sender_info *sender_entry = data;
 	
 	ctg = ast_strdupa(sender_entry->stanza);
 	
@@ -1292,55 +1290,55 @@ static void *aprstt_sender_thread(void *data)
 	if (stat(fname, &mystat) == -1) {
 		mfp = fopen(fname, "w");
 		if (!mfp) {
-			ast_log(LOG_ERROR, "Can not create aprstt common block file %s. Error: %s\n", fname, strerror(errno));
+			ast_log(LOG_ERROR, "Can not create aprstt common block file %s: %s\n", fname, strerror(errno));
 			pthread_exit(NULL);
 		}
 		memset(&ttempty, 0, sizeof(ttempty));
 		for (i = 0; i < ttlist; i++) {
 			if (fwrite(&ttempty, 1, sizeof(ttempty), mfp) != sizeof(ttempty)) {
-				ast_log(LOG_ERROR, "Error initializing aprtss common block file %s. Error: %s\n", fname, strerror(errno));
+				ast_log(LOG_ERROR, "Error initializing aprtss common block file %s: %s\n", fname, strerror(errno));
 				fclose(mfp);
 				pthread_exit(NULL);
 			}
 		}
 		fclose(mfp);
 		if (stat(fname, &mystat) == -1) {
-			ast_log(LOG_ERROR, "Unable to stat new aprstt common block file %s: Error: %s\n", fname, strerror(errno));
+			ast_log(LOG_ERROR, "Unable to stat new aprstt common block file %s: %s\n", fname, strerror(errno));
 			pthread_exit(NULL);
 		}
 	}
 	if (mystat.st_size < (sizeof(struct ttentry) * ttlist)) {
 		mfp = fopen(fname, "r+");
 		if (!mfp) {
-			ast_log(LOG_ERROR, "Can not open aprstt common block file %s. Error: %s\n", fname, strerror(errno));
+			ast_log(LOG_ERROR, "Can not open aprstt common block file %s: %s\n", fname, strerror(errno));
 			pthread_exit(NULL);
 		}
 		memset(&ttempty, 0, sizeof(ttempty));
 		if (fseek(mfp, 0, SEEK_END)) {
-			ast_log(LOG_ERROR, "Can not seek aprstt common block file %s. Error: %s\n", fname, strerror(errno));
+			ast_log(LOG_ERROR, "Can not seek aprstt common block file %s: %s\n", fname, strerror(errno));
 			pthread_exit(NULL);
 		}
 		for (i = mystat.st_size; i < (sizeof(struct ttentry) * ttlist); i += sizeof(struct ttentry)) {
 			if (fwrite(&ttempty, 1, sizeof(ttempty), mfp) != sizeof(ttempty)) {
-				ast_log(LOG_ERROR, "Error growing aprtss common block file %s. Error: %s\n", fname, strerror(errno));
+				ast_log(LOG_ERROR, "Error growing aprtss common block file %s: %s\n", fname, strerror(errno));
 				fclose(mfp);
 				pthread_exit(NULL);
 			}
 		}
 		fclose(mfp);
 		if (stat(fname, &mystat) == -1) {
-			ast_log(LOG_ERROR, "Unable to stat updated aprstt common block file %s. Error: %s\n", fname, strerror(errno));
+			ast_log(LOG_ERROR, "Unable to stat updated aprstt common block file %s: %s\n", fname, strerror(errno));
 			pthread_exit(NULL);
 		}
 	}
 	mfp = fopen(fname, "r+");
 	if (!mfp) {
-		ast_log(LOG_ERROR, "Can not open aprstt common block file %s. Error: %s\n", fname, strerror(errno));
+		ast_log(LOG_ERROR, "Can not open aprstt common block file %s: %s\n", fname, strerror(errno));
 		pthread_exit(NULL);
 	}
 	ttentries = mmap(NULL, mystat.st_size, PROT_READ | PROT_WRITE, MAP_SHARED, fileno(mfp), 0);
 	if (ttentries == NULL) {
-		ast_log(LOG_ERROR, "Cannot map aprtss common file %s. Error: %s\n", fname, strerror(errno));
+		ast_log(LOG_ERROR, "Cannot map aprtss common file %s: %s\n", fname, strerror(errno));
 		pthread_exit(NULL);
 	}
 	
@@ -1483,8 +1481,6 @@ static int gps_read_helper(struct ast_channel *chan, const char *cmd, char *data
 {
 	struct position_info selected_position;
 	
-	memset(buf, 0, len);
-	
 	ast_mutex_lock(&position_update_lock);
 	selected_position.is_valid = 0;
 	/* See if we need to send live GPS or the default */
@@ -1506,6 +1502,7 @@ static int gps_read_helper(struct ast_channel *chan, const char *cmd, char *data
 		return 0;
 	}
 
+	*buf = '\0';
 	return -1;
 }
 /*!
--- a/apps/app_gps.c
+++ b/apps/app_gps.c
@@ -1821,10 +1821,7 @@ static int load_module(void)
 		ast_log(LOG_ERROR, "Cannot create APRStt sender thread %s", sender_entry->section);
 		return -1;
 	}
-	/* 	if (!sender_entry) {
-		return -1;
-	}
-
+	/* 
 	 * See if we have sections other than general. 
 	 * If present, create aprs processing threads for those sections 
 	 */
--- a/apps/app_gps.c
+++ b/apps/app_gps.c
@@ -348,7 +348,7 @@ static time_t time_monotonic(void)
  *
  * \retval 			Returns number of substrings found.
  */
-static int explode_string(char *str, char *strp[], int limit, char delim, char quote)
+static int explode_string(char *str, char *strp[], int limit, const char delim, const char quote)
 {
 	int i, l, inquo;
 
@@ -598,7 +598,7 @@ static void *aprs_connection_thread(void *data)
  * \retval 0		Success
  * \retval -1		Failure
  */
-static int report_aprs(char *ctg, char *lat, char *lon, char *elev)
+static int report_aprs(const char *ctg, const char *lat, const char *lon, const char *elev)
 {
 	struct ast_config *cfg = NULL;
 	char *call, *comment, icon, icon_table;
@@ -760,7 +760,7 @@ static int report_aprs(char *ctg, char *lat, char *lon, char *elev)
  * \retval -1		Failure
  */
 
-static int report_aprstt(char *ctg, char *lat, char *lon, char *theircall, char overlay)
+static int report_aprstt(const char *ctg, const char *lat, const char *lon, const char *theircall, const char overlay)
 {
 	struct ast_config *cfg = NULL;
 	char *call, *comment;
--- a/apps/app_gps.c
+++ b/apps/app_gps.c
@@ -210,7 +210,7 @@
 		</synopsis>
 		<description>
 			<para>Read current or default GPS position.
-			Returns a string in the format epoch, space, latitude, spaces,
+			Returns a string in the format monotonic time, space, epoch time, space, latitude, spaces,
 			longitude, spaces, and elevation.</para>
 		</description>
 	</function>
@@ -238,6 +238,7 @@
 /*! Defines */
 #define	APRS_DEFAULT_SERVER "rotate.aprs.net"
 #define	APRS_DEFAULT_PORT "14580"
+#define APRS_IS_VERSION "Asterisk app_gps_V3"
 #define	APRS_DEFAULT_COMMENT "Asterisk/app_rpt Node"
 #define	APRSTT_DEFAULT_COMMENT "Asterisk/app_rpt TT Report"
 #define	APRSTT_DEFAULT_OVERLAY '0'
@@ -287,7 +288,8 @@ struct position_info {
 	char latitude[25];			/* latitude format DDMM.SSS */
 	char longitude[25];			/* longitude format DDDMM.SSS */
 	char elevation[25];			/* elevation format VVVV.V */
-	time_t last_updated;		/* the time these values were last updated */
+	time_t last_updated;		/* the ephoh time these values were last updated */
+	time_t last_updated_mono;	/* the monotonic time these values were last updated */
 };
 
 static struct position_info current_gps_position;
@@ -536,7 +538,7 @@ static void *aprs_connection_thread(void *data)
 		}
 
 		/* Log into the APRS-IS server */
-		sprintf(buf, "user %s pass %s vers Asterisk app_gps_V3\n", call, password);
+		snprintf(buf, sizeof(buf), "user %s pass %s vers " APRS_IS_VERSION "\n", call, password);
 		
 		if (send(sockfd, buf, strlen(buf), 0) < 0) {
 			ast_log(LOG_WARNING, "Can not send sign on to server: %s\n", strerror(errno));
@@ -762,7 +764,7 @@ static int report_aprstt(char *ctg, char *lat, char *lon, char *theircall, char 
 {
 	struct ast_config *cfg = NULL;
 	char *call, *comment;
-	char *val, basecall[20], buf[300], buf1[100], *cp;
+	char *val, basecall[20], buf[300], buf1[20], *cp;
 	time_t t;
 	struct tm *tm;
 	struct ast_flags zeroflag = { 0 };
@@ -826,8 +828,8 @@ static int report_aprstt(char *ctg, char *lat, char *lon, char *theircall, char 
 	t = time(NULL);
 	tm = gmtime(&t);
 	
-	sprintf(buf1, "%s-12", theircall);
-	sprintf(buf, "%s>APSTAR:;%-9s*%02d%02d%02dz%s%c%sA%s\r\n",
+	snprintf(buf1, sizeof(buf1), "%s-12", theircall);
+	snprintf(buf, sizeof(buf), "%s>APSTAR:;%-9s*%02d%02d%02dz%s%c%sA%s\r\n",
 			call, buf1, tm->tm_hour, tm->tm_min, tm->tm_sec, lat, overlay, lon, comment);
 	
 	ast_mutex_lock(&aprs_socket_lock);
@@ -901,7 +903,7 @@ static void *gps_reader(void *data)
 	int res, i, n, fd, has_comport = 0;
 	struct termios mode;
 	struct position_info *selected_info;
-	time_t now;
+	time_t now_mono;
 
 	if (comport) {
 		has_comport = 1;
@@ -974,9 +976,9 @@ static void *gps_reader(void *data)
 			selected_info = &general_def_position;
 			
 		} else {
-			now = time_monotonic();
+			now_mono = time_monotonic();
 			/* Check for no data receiption */
-			if (current_gps_position.last_updated + GPS_VALID_SECS < now) {
+			if (current_gps_position.last_updated_mono + GPS_VALID_SECS < now_mono) {
 				ast_mutex_lock(&position_update_lock);
 				current_gps_position.is_valid = 0;
 				ast_mutex_unlock(&position_update_lock);
@@ -1036,6 +1038,7 @@ static void *gps_reader(void *data)
 			snprintf(current_gps_position.longitude, sizeof(current_gps_position.longitude) - 1, "%s%s", strs[4], strs[5]);
 			snprintf(current_gps_position.elevation, sizeof(current_gps_position.elevation) - 1, "%s%s", strs[9], strs[10]);
 			current_gps_position.last_updated = time(NULL);
+			current_gps_position.last_updated_mono = now_mono;
 			ast_mutex_unlock(&position_update_lock);
 			
 			selected_info = & current_gps_position;
@@ -1073,7 +1076,7 @@ static void *aprs_sender_thread(void *data)
 	char *ctg;
 	char *val, *deflat, *deflon, *defelev;
 	int interval, my_update_secs, ehlert;
-	time_t now, lastupdate;
+	time_t now_mono, lastupdate_mono;
 	struct ast_flags zeroflag = { 0 };
 	struct position_info this_def_position, selected_position;
 	struct aprs_sender_info *sender_entry = data;
@@ -1143,11 +1146,11 @@ static void *aprs_sender_thread(void *data)
 	 }
 	
 	memset(&selected_position, 0, sizeof(selected_position));
-	lastupdate = time_monotonic();
+	lastupdate_mono = time_monotonic();
 	my_update_secs = GPS_UPDATE_SECS;
 	
 	while (run_forever) {
-		now = time_monotonic();
+		now_mono = time_monotonic();
 		
 		ast_mutex_lock(&position_update_lock);
 		selected_position.is_valid = 0;
@@ -1157,6 +1160,7 @@ static void *aprs_sender_thread(void *data)
 		} else if (this_def_position.is_valid && !ehlert) {
 				selected_position = this_def_position;
 				selected_position.last_updated = time(NULL);
+				selected_position.last_updated_mono = now_mono;
 		}
 		ast_mutex_unlock(&position_update_lock);
 		/* 
@@ -1164,10 +1168,10 @@ static void *aprs_sender_thread(void *data)
 		 * The last_updated time must be current so that
 		 * we know we are getting good GPS information.
 		 */
-		if (selected_position.is_valid && (selected_position.last_updated + GPS_VALID_SECS) >= time(NULL) &&
-			now >= (lastupdate + my_update_secs)) {
+		if (selected_position.is_valid && (selected_position.last_updated_mono + GPS_VALID_SECS) >= now_mono &&
+			now_mono >= (lastupdate_mono + my_update_secs)) {
 			report_aprs(ctg, selected_position.latitude, selected_position.longitude, selected_position.elevation);
-			lastupdate = now;
+			lastupdate_mono = now_mono;
 			my_update_secs = interval;
 		}
 		/* wait 1 second */
@@ -1431,6 +1435,7 @@ static void *aprstt_sender_thread(void *data)
 			if (this_def_position.is_valid) {
 				selected_position = this_def_position;
 				selected_position.last_updated = now;
+				selected_position.last_updated_mono = time_monotonic();
 			}
 		}
 		ast_mutex_unlock(&position_update_lock);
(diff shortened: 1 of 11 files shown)


apps/app_rpt/rpt_functions.c
```diff
@@ -1714,28 +1714,25 @@ int function_cop(struct rpt *myrpt, char *param, char *digitbuf, int command_sou
...
-		else
-			fprintf(fp, "%s\n", argv[1]);
-		fclose(fp);
-		if (myatoi(argv[0]) == 63)
-			rpt_telemetry(myrpt, ARB_ALPHA, (void *) argv[1]);
+		
+		if (!ast_custom_function_find("APRS_SENDTT")) {
+			ast_log(LOG_WARNING, "app_gps is not loaded.  APRSTT failed\n");
+		} else {
+			memset(func, 0, sizeof(func));
```
InterLinked1, line 1726: No memset needed, similar comments as above

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2337,21 +2337,19 @@ void *rpt_tele_thread(void *this)
...
 	case STATS_GPS_LEGACY:
-		fp = fopen(GPSFILE, "r");
-		if (!fp) {
+		
+		/* If the app_gps custom function GPS_READ exists, read the GPS position */
+		if (!ast_custom_function_find("GPS_READ")) {
 			break;
-		}
-		if (fstat(fileno(fp), &mystat) == -1) {
+		}
```
InterLinked1, line 2344: trailing whitespace

apps/app_gps.c
```diff
@@ -30,212 +30,307 @@
...
 
-#define	GPS_DEFAULT_SERVER "second.aprs.net"
-#define	GPS_DEFAULT_PORT "10151"
-#define	GPS_DEFAULT_COMMENT "Asterisk app_rpt server"
-#define	TT_DEFAULT_COMMENT "Asterisk app_rpt user"
-#define	TT_DEFAULT_OVERLAY '0'
+/*! Defines */
+#define	APRS_DEFAULT_SERVER "rotate.aprs.net"
+#define	APRS_DEFAULT_PORT "14580"
+#define	APRS_DEFAULT_COMMENT "AllStar Link Node"
```
InterLinked1, line 241: These modules are not supposed to be specific to AllStar, so I'm not sure I like that it's hardcoded. Maybe something more generic, like "Link Node"?
Can these be configurable through the config file?

apps/app_gps.c
```diff
@@ -30,212 +30,307 @@
...
-* Returns number of substrings found.
-*/
+/* Message flags */
+static int gps_unlock_shown = 0;
+
+/*!
+ * \brief Position information structure
+ */
+struct position_info {
+	int is_valid:1;			/* contains valid values indicator */
```
InterLinked1, line 286: Typically we use `unsigned int:1` for bitfield, I'm not really sure what difference it makes though

apps/app_gps.c
```diff
@@ -30,212 +30,307 @@
...
+	time_t last_updated;	/* the time these values were last updated */
+};
+
+static struct position_info current_gps_position;
+static struct position_info general_def_position;
 
+/*!
+ *\brief Enum for aprs sender info type.
+ */
+enum Sender_Type {
```
InterLinked1, line 299: `enum aprs_sender_type` maybe?

apps/app_gps.c
```diff
@@ -293,194 +387,305 @@ static int getserialchar(int fd)
...
+			 */
+			res = ast_poll(fds, 1, 500);
+			if (res == 0) {
+				continue;
+			}
+			if (res < 0 || fds[0].revents & POLLHUP) {
+				break;
+			}
+			if (fds[0].revents & POLLIN) {
+				memset(buf, 0, sizeof(buf));
```
InterLinked1, line 556: Not necessary to memset prior to recv

apps/app_gps.c
```diff
@@ -293,194 +387,305 @@ static int getserialchar(int fd)
...
 		return -1;
 	}
-	ast_config_destroy(cfg);
-	cfg = NULL;
 
-	ast_copy_string(basecall, call, sizeof(basecall));
-	cp = strchr(basecall, '-');
-	if (cp)
-		*cp = 0;
+    /* Setup the server call sign
```
InterLinked1, line 666: Misaligned first line

apps/app_gps.c
```diff
@@ -293,194 +387,305 @@ static int getserialchar(int fd)
...
+			strcat(servercall, "S");
+		}
+	} else {
+		strcat(servercall, "-VS");
+	}
+	
+	/* Reduce the precision of latitude and longitude
+	 *
+	 * Latitude is expressed as a fixed 8-character field, in degrees and decimal
+     * minutes (to two decimal places), followed by the letter N for north or
```
InterLinked1, line 683: These 2 lines seem to be missing a tab indent

apps/app_gps.c
```diff
@@ -491,34 +696,57 @@ static int report_aprs(char *ctg, char *lat, char *lon)
 		*(cp + 3) = lon[strlen(lon) - 1];
 		*(cp + 4) = 0;
 	}
+	
+	/* Setup optional elevation */
+	memset(elev_str, 0, sizeof(elev_str));
```
InterLinked1, line 701: If elev_f > 0, the memset isn't necessary.
I would just do `elev_str[0] = '\0'` in the false branch to cover the other case

apps/app_gps.c
```diff
@@ -528,36 +756,49 @@ static int report_aprstt(char *ctg, char *lat, char *lon, char *theircall, char
...
 	cp = strchr(basecall, '-');
-	if (cp)
+	if (cp) {
 		*cp = 0;
+	}
+	
+	/* Reduce the precision of latitude and longitude
+	 *
+	 * Latitude is expressed as a fixed 8-character field, in degrees and decimal
+     * minutes (to two decimal places), followed by the letter N for north or
```
InterLinked1, line 796: Alignment

apps/app_gps.c
```diff
@@ -568,56 +809,104 @@ static int report_aprstt(char *ctg, char *lat, char *lon, char *theircall, char
...
-	else
+	} else {
 		comport = "/dev/null";
+	}
 
+	/* Open the serial port configured for the GPS device */
 	fd = open(comport, O_RDWR);
 	if (fd == -1) {
-		ast_log(LOG_WARNING, "Cannot open serial port %s\n", comport);
+		ast_log(LOG_WARNING, "Cannot open serial port %s: Error %s\n", comport, strerror(errno));
```
InterLinked1, line 902: I like that we print out the actual error, but no need to explicitly prepend "Error" to the error

apps/app_gps.c
```diff
@@ -635,51 +924,57 @@ static void *gpsthread(void *data)
...
+			ast_log(LOG_WARNING, "Unable to set serial parameters on %s: Error %s\n", comport, strerror(errno));
 			close(fd);
 			goto err;
 		}
 	}
+	
+	/* Give the device a few milliseconds to come on-line */
 	usleep(100000);
 
+	/*! \todo we need to detail with someone unplugging the device */
```
InterLinked1, line 936: deal instead of detail?

apps/app_gps.c
```diff
@@ -693,316 +988,448 @@ static void *gpsthread(void *data)
...
 	struct stat mystat;
-	time_t now, was, lastupdate;
+	time_t now, lastupdate;
 	struct ttentry *ttentries, ttempty;
-	struct ast_flags zeroflag = { 0 };
-
+	struct position_info this_def_position, selected_position;
+	struct timeval tv;
+	struct timespec ts = {0};
+	struct aprs_sender_info *sender_entry = (struct aprs_sender_info *) data;
```
InterLinked1, line 1196: The cast isn't necessary

apps/app_gps.c
```diff
@@ -1012,85 +1439,265 @@ static void *gps_tt_thread(void *data)
...
+ * \param	len		Length of the buffer.
+ *
+ * \retval	0		Success
+ * \retval	-1		Failure
+ */
+static int gps_read_helper(struct ast_channel *chan, const char *cmd, char *data, char *buf, size_t len)
+{
+	struct position_info selected_position;
+	
+	memset(buf, 0, len);
```
InterLinked1, line 1486: Not necessary. If we return -1, just do `*buf = '\0'` first, if we really want to be nice

apps/app_gps.c
```diff
@@ -1012,85 +1439,265 @@ static void *gps_tt_thread(void *data)
...
+ * \param	data	Pointer to the data passed to the function.
+ * \param	value	Pointer to value that represents the callsign.
+ *
+ * \retval	0		Success
+ * \retval	-1		Failure
+ */
+static int aprs_sendtt_helper(struct ast_channel *chan, const char *function, char *data, const char *value)
 {
-	/*! \todo ??? Why does this even exist? */
+		char *parse;
```
InterLinked1, line 1528: Looks like these first few lines are indented extra

apps/app_gps.c
```diff
@@ -1012,85 +1439,265 @@ static void *gps_tt_thread(void *data)
...
+
 	return 0;
 }
 
+/*!
+ * \brief Asterisk function setup struct for gps_read_helper
+ */
+static struct ast_custom_function gps_read_function = {
+	.name = "GPS_READ",
+	.synopsis = "Read GPS position",
```
InterLinked1, line 1582: I think the synopsis and desc fields are deprecated and shouldn't be set... the xml documentation replaced these

apps/app_gps.c
```diff
@@ -1012,85 +1439,265 @@ static void *gps_tt_thread(void *data)
...
+ * \param e				Asterisk cli entry.
+ * \param cmd			Cli command type.
+ * \param a				Pointer to asterisk cli arguments.
+ * \return	Cli success or failure.
+ */
+static char *handle_cli_status(struct ast_cli_entry *e, int cmd, struct ast_cli_args *a)
+{
+	switch (cmd) {
+	case CLI_INIT:
+		e->command = "gps status";
```
InterLinked1, line 1608: How about `gps show status`, to be more in line with the CLI naming conventions?

apps/app_gps.c
```diff
@@ -1012,85 +1439,265 @@ static void *gps_tt_thread(void *data)
...
-	pthread_join(gps_ttthread, NULL);
-	ast_debug(1, "General threads have exited\n");
-	res = ast_unregister_application(app);
+		ast_debug(2, "Waiting for gps_reader_thread to exit\n");
+		pthread_join(gps_reader_thread_id, NULL);
+		ast_free(comport);
+	}
+	
+	/* Shutdown and clean up sender threads */
+	AST_LIST_TRAVERSE_SAFE_BEGIN(&aprs_sender_list, sender_entry, list) {
```
InterLinked1, line 1666: Rather than a safe traversal, you can use REMOVE_HEAD - example: https://github.com/asterisk/asterisk/blob/316693ded9e5b02423301646db9b2e5bb93b5692/main/chanvars.c#L117

apps/app_gps.c
```diff
@@ -1151,27 +1767,90 @@ static int load_module(void)
...
-		debug = ast_true(val);
-
-	ast_pthread_create(&aprs_thread, NULL, aprsthread, NULL);
-	if (comport)
-		ast_pthread_create(&gps_thread, NULL, gpsthread, NULL);
-	ast_pthread_create(&gps_subthread, NULL, gps_sub_thread, ast_strdup("general"));
-	ast_pthread_create(&gps_ttthread, NULL, gps_tt_thread, ast_strdup("general"));
+	}
+	
+	memset(&current_gps_position, 0, sizeof(current_gps_position));
```
InterLinked1, line 1774: Static variables are already 0-initialized, so this  technically isn't necessary

InterLinked1 (changes requested):

apps/app_rpt.c
```diff
@@ -1545,28 +1545,38 @@ static inline void collect_function_digits_post(struct rpt *myrpt, int res, cons
...
+			ast_log(LOG_WARNING, "app_gps is not loaded.  APRSTT failed\n");
 		} else {
-			fprintf(fp, "%s %c\n", aprscall, overlay);
-			fclose(fp);
-			rpt_telemetry(myrpt, ARB_ALPHA, (void *) aprscall);
-		}
+			memset(func, 0, sizeof(func));
+			snprintf(func, sizeof(func) -1, "APRS_SENDTT(%s,%c)", !myrpt->p.aprstt[0] ? "general" : myrpt->p.aprstt, overlay);
+			/* execute the APRS_SENDTT function in app_gps*/
+			if (!ast_func_write(NULL, func, aprscall)) {
```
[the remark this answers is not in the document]
KB4MDD: I will guess that a small number of people use app_gps.  It is not loaded by default in our sample configs.  I hope that this work, along with documentation in the manual will increase the usage.

mkmer (commented):
Nice work!  This should really help save the memory cards on PI's.
Just a few comments around format stuff.

apps/app_rpt/rpt_functions.c
```diff
@@ -1714,28 +1714,25 @@ int function_cop(struct rpt *myrpt, char *param, char *digitbuf, int command_sou
...
 			break;
 		}
-		if (argc > 2)
-			fprintf(fp, "%s %c\n", argv[1], *argv[2]);
-		else
-			fprintf(fp, "%s\n", argv[1]);
-		fclose(fp);
-		if (myatoi(argv[0]) == 63)
-			rpt_telemetry(myrpt, ARB_ALPHA, (void *) argv[1]);
+
```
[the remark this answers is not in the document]
KB4MDD: Good find.  I have this corrected.

apps/app_rpt/rpt_functions.c
```diff
@@ -1714,28 +1714,25 @@ int function_cop(struct rpt *myrpt, char *param, char *digitbuf, int command_sou
...
 		}
-		if (argc > 2)
-			fprintf(fp, "%s %c\n", argv[1], *argv[2]);
-		else
-			fprintf(fp, "%s\n", argv[1]);
-		fclose(fp);
-		if (myatoi(argv[0]) == 63)
-			rpt_telemetry(myrpt, ARB_ALPHA, (void *) argv[1]);
+		
+		if (!ast_custom_function_find("APRS_SENDTT")) {
```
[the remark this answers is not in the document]
KB4MDD: corrected.

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2767,25 +2764,20 @@ void rpt_telemetry(struct rpt *myrpt, int mode, void *data)
 			send_tele_link(myrpt, mystr);
 			return;
 		case STATS_GPS:
-			fp = fopen(GPSFILE, "r");
-			if (!fp) {
+			
+			/* If the app_gps custom function GPS_READ exists, read the GPS position */
+			if (!ast_custom_function_find("GPS_READ")) {
```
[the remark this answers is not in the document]
KB4MDD: It would be normal for the GPS_READ function to not be found in the event that the user is not using a gps or aprs (ie apps_gps is not loaded.) I suspect very few people have it loaded.

KB4MDD:
@InterLinked1 @mkmer 

Regarding adding warning messages if the new functions cannot be found.  I will be glad to add that.   I will note that if I don't check for the function and app_gps is not loaded, the core function handler will send a warning message that the function does not exist.  (I ran into this and decided to add the check.)

I will note that this is really only an issue if the enables a dtmf command that uses one of the functions.  If they are doing that, I would like to think they would have enabled app_gps.

What is the preference - leave as is, add a warning message that the function is not there (preserving my if), or removing my test and allowing the core function handler to report the problem?

apps/app_rpt/rpt_functions.c
```diff
@@ -1714,28 +1714,25 @@ int function_cop(struct rpt *myrpt, char *param, char *digitbuf, int command_sou
...
+			ast_log(LOG_WARNING, "app_gps is not loaded.  APRSTT failed\n");
+			return DC_COMPLETE;
+		}
+		memset(func, 0, sizeof(func));
+		snprintf(func, sizeof(func) -1, "APRS_SENDTT(%s,%c)", !myrpt->p.aprstt ? "general" : myrpt->p.aprstt, 
+			argc > 2 ? argv[3][0] : ' ');
```
mkmer, line 1729: Looking at this again, Is something wrong with this test?  argv[3] is argument 4 right?  should it be `> 3?`

apps/app_rpt.c
```diff
@@ -4938,28 +4945,36 @@ static void *rpt(void *this)
...
+			unsigned long long u;
+			char gps_data[100];
+			char lat[100], lon[100], elev[100];
+
 			myrpt->lastgpstime = t;
-			fp = fopen(GPSFILE, "r");
-			if (!fp)
+			
+			/* If the app_gps custom function GPS_READ exists, read the GPS position */
+			if (!ast_custom_function_find("GPS_READ")) {
```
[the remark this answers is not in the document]
KB4MDD: I see.  I will let you and @InterLinked1 make a recommendation.  We have it in that one place - but not the others.

apps/app_rpt.c
```diff
@@ -1545,28 +1545,37 @@ static inline void collect_function_digits_post(struct rpt *myrpt, int res, cons
...
-			ast_copy_string(fname, APRSTT_PIPE, sizeof(fname) - 1);
-		else
-			snprintf(fname, sizeof(fname) - 1, APRSTT_SUB_PIPE, myrpt->p.aprstt);
-		fp = fopen(fname, "w");
-		if (!fp) {
-			ast_log(LOG_WARNING, "Can not open APRSTT pipe %s: %s\n", fname, strerror(errno));
+		ast_debug(1, "APRStt got string %s callsign %s overlay %c\n", cmd, aprscall, overlay);
+		
+		if (!ast_custom_function_find("APRS_SENDTT")) {
+			ast_log(LOG_WARNING, "app_gps is not loaded.  APRSTT failed\n");
```
Allan-N, line 1571: Is it APRSTT or APRStt ?

channels/chan_echolink.c
```diff
@@ -3366,9 +3363,8 @@ static void *el_reader(void *data)
 				}
 				mylat = instp->lat;
 				mylon = instp->lon;
-				fp = fopen(GPSFILE, "r");
-				if (fp && (fstat(fileno(fp), &mystat) != -1) && (mystat.st_size < 100)) {
-					if (fscanf(fp, "%u %f%c %f%c", &u, &lat, &latc, &lon, &lonc) == 5) {
+				if (ast_custom_function_find("GPS_READ") && !ast_func_read(NULL, "GPS_READ()", gps_data, sizeof(gps_data))) {
+					if (sscanf(gps_data, "%llu %f%c %f%c", &u, &lat, &latc, &lon, &lonc) == 5) {
```
Allan-N, line 3367: Here, you appear to be looking for latc and lonc ... but other places where you scan the GPS_READ data you don't.  Why?

KB4MDD: This is how echolink worked originally.  They expect the lat and lon information in a different format than is used elsewhere.  We pull off the direction indicator (N, S, E, W) for use later in the routine.

apps/app_rpt/rpt_functions.c
```diff
@@ -1714,28 +1714,25 @@ int function_cop(struct rpt *myrpt, char *param, char *digitbuf, int command_sou
...
-		if (argc > 2)
-			fprintf(fp, "%s %c\n", argv[1], *argv[2]);
-		else
-			fprintf(fp, "%s\n", argv[1]);
-		fclose(fp);
-		if (myatoi(argv[0]) == 63)
-			rpt_telemetry(myrpt, ARB_ALPHA, (void *) argv[1]);
+		
+		if (!ast_custom_function_find("APRS_SENDTT")) {
+			ast_log(LOG_WARNING, "app_gps is not loaded.  APRSTT failed\n");
```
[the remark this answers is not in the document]
KB4MDD: Corrected to APRStt

configs/rpt/gps.conf
```diff
@@ -1,35 +1,90 @@
...
+;
+; ttcomment =                 ; Text to be displayed associated with this TT station
+; ttlat =                     ; Fixed latitude in decimal degrees - if not specified 'lat' is used for the default
+; ttlon =                     ; Fixed longitude in decimal degrees - if not specified 'lon' is used for the default
+; ttelev =                    ; Elevation of Antenna in Meters (*NOT* HAAT) - if not specified 'elev' is used for the default
+; ttlist =                    ; Number of touch tone entries to maintain.  Default is 10.
+; ttoffset =                  ; TT Offset from main lat/lon value. Default is 10.
+; ttsplit =                   ; yes = center reports around lat/lon origin, no = (default) only use ttoffset
+
+;
```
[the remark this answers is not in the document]
KB4MDD: I will defer to your judgement on using templates.  For 99% of users, they will only have one stanza [general].  The application does support multiple stanzas for cases where the user has a node at a different location.  

Templates would work, however, all of the values would be different except the APRIS-IS server settings for the second or third node.

For the sample [1999] node, I did copy all of the settings.  Maybe I need to drop the ones that would not be used on the second node.  Let me know your thoughts and whether to rework for templates.

configs/rpt/gps.conf
```diff
@@ -1,35 +1,90 @@
...
 ; Configuration for app_gps
 ;
 [general]
-; See https://wiki.allstarlink.org/wiki/Gps.conf
-call = W1AW-1               ; callsign (including SSID) for APRS purposes
+call = W1AW-1               ; callsign (including SSID) for APRS
```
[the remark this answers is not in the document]
KB4MDD: I set the call to INVALID-1.  Added a comment to change this.

comment was moved down so that call, password, server and port are the first 4 entries.

apps/app_rpt.c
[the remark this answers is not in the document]
InterLinked1: It really would depend on whether the configuration is valid or not. `app_gps` is optional and `app_rpt` can run without it (no hard dependency). Is there a valid/correct configuration of `app_rpt` that would trigger the warning? If so, then maybe not. But if this means the user has configured something that won't work, I think a warning is appropriate.

InterLinked1:
> @InterLinked1 @mkmer
> 
> Regarding adding warning messages if the new functions cannot be found. I will be glad to add that. I will note that if I don't check for the function and app_gps is not loaded, the core function handler will send a warning message that the function does not exist. (I ran into this and decided to add the check.)
> 
> I will note that this is really only an issue if the enables a dtmf command that uses one of the functions. If they are doing that, I would like to think they would have enabled app_gps.
> 
> What is the preference - leave as is, add a warning message that the function is not there (preserving my if), or removing my test and allowing the core function handler to report the problem?

In this case, keep as is (checking before call). We don't want to emit spurious warnings either.

configs/rpt/gps.conf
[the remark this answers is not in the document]
Allan-N: Looking at your current `rpt.conf` changes I see that you have `;aprstt = general` in the `[node-main]` template.  If I remove the ";" from this line then it will be part of the stanza for every node that uses the template.  Is it OK to have multiple nodes reporting the same [general] APRStt info?

Templates work really well if you have some information that needs to be shared across stanzas with each stanza being able to have it's own unique settings.  I am also loathe to go back down the path of folks having to copy/paste and entire block of text only to change one line.

My thinking is that if we want to support multiple nodes reporting different information then we should really be looking at templates ... and do it now before we've deployed one solution and need to switch to another.

Happy to discuss this more.  Also happy to help.

apps/app_rpt.c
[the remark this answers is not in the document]
KB4MDD: I have made adjustments to the warning messages, along with adding a response over the air should the user enable a command to read the location and app_gps is not loaded.

apps/app_gps.c
```diff
@@ -30,212 +30,307 @@
...
-*/
+/* Message flags */
+static int gps_unlock_shown = 0;
+
+/*!
+ * \brief Position information structure
+ */
+struct position_info {
+	int is_valid:1;			/* contains valid values indicator */
+	char latitude[25];		/* latitude format DDMM.SSS */
```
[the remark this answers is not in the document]
KB4MDD: I am fixing these.  Some of those were from the original code.

configs/rpt/gps.conf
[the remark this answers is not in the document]
KB4MDD: It is ok for every node to use aprstt = general.  In most cases that would be the correct solution.  It would only need to be different if someone has nodes in different geographic areas.

apps/app_gps.c
KB4MDD: I changed the defaults to Asterisk/app_rpt.  They however are customizable.  You would not want to use those defaults.

apps/app_rpt.c
```diff
@@ -1545,28 +1545,37 @@ static inline void collect_function_digits_post(struct rpt *myrpt, int res, cons
...
 }
 
+/*!
+ * \brief Send APRStt (Touchtone) to app_gps for processing.
+ * This routine takes the received APRStt touchtone digits
+ * and translates them to a callsign.  The results are
+ * sent to app_gps using the APRS_SENDTT function for 
+ * processing and posting to the APRS-IS server.
+ *
+ * \param rpt		pointer to repeater struct.
```
InterLinked1, line 1555: It's actually named myrpt

apps/app_gps.c
```diff
@@ -30,212 +30,307 @@
...
+			longitude, spaces, and elevation.</para>
+		</description>
+	</function>
+	<function name="APRS_SENDTT" language="en_US">
+		<synopsis>
+			Report a callsign to APRS.
+		</synopsis>
+		<syntax>
+			<parameter name="stanza" required="true">
+				<para>Stanza to use.</para>
```
InterLinked1, line 223: Stanza of what? This should be a bit more specific. Also, Asterisk configuration typically refers to them as "sections" (and as "contexts" for the dialplan specifically)

apps/app_gps.c
```diff
@@ -30,212 +30,307 @@
...
+			</parameter>
+			<parameter name="overlay" required="true">
+				<para>Overlay to report. Number 1 to 9.</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>Nothing is returned.</para>
+		    <example title="Sending a report.">
```
InterLinked1, line 231: Indent the 3 lines for the example

apps/app_gps.c
```diff
@@ -30,212 +30,307 @@
...
+				exten => s,1,APRS_SENDTT(general,1)=WB6NIL
```
InterLinked1, line 232: The function needs to be written by the `Set` or `MSet` application to do what I think you're trying to there:

You could do:
`exten => s,1,Set(APRS_SENDTT(general,1)=WB6NIL)`

apps/app_gps.c
```diff
@@ -30,212 +30,307 @@
...
+	time_t last_updated;		/* the time these values were last updated */
+};
+
+static struct position_info current_gps_position;
+static struct position_info general_def_position;
 
+/*!
+ *\brief Enum for aprs sender info type.
+ */
+enum APRS_Sender_Type {
```
InterLinked1, line 299: all lowercase, per Asterisk convention would be nice. Is there a reason to use this case?

apps/app_gps.c
```diff
@@ -30,212 +30,307 @@
...
+	APRSTT
+};
+
+/*!		
+ * \brief Structure to track APRS and APRStt sender threads.
+ */
+struct aprs_sender_info {
+	AST_LIST_ENTRY(aprs_sender_info) list;
+	enum APRS_Sender_Type type;					/* sender type enum */
+	char stanza[50]; 						/* stanza associated with this aprs thread */
```
InterLinked1, line 310: Is this referring to a config section in rpt.conf?

apps/app_gps.c
```diff
@@ -640,46 +927,52 @@ static void *gpsthread(void *data)
...
-				strs[9] = astr;
-			} else
-				strs[9] = "000.0";
-			strs[10] = "M";
+			
+			selected_info = &general_def_position;
+			
 		} else {
-			c = 0;
+			time(&now);
```
InterLinked1, line 962: This usage is deprecated, in new code `now = time(NULL)` is preferred

apps/app_gps.c
```diff
@@ -693,316 +986,448 @@ static void *gpsthread(void *data)
...
+				ast_log(LOG_NOTICE, "GPS locked\n");
+				gps_unlock_shown = 0;
+			}
+			
+			ast_mutex_lock(&position_update_lock);
+			current_gps_position.is_valid = 1;
+			snprintf(current_gps_position.latitude, sizeof(current_gps_position.latitude) - 1, "%s%s", strs[2], strs[3]);
+			snprintf(current_gps_position.longitude, sizeof(current_gps_position.longitude) - 1, "%s%s", strs[4], strs[5]);
+			snprintf(current_gps_position.elevation, sizeof(current_gps_position.elevation) - 1, "%s%s", strs[9], strs[10]);
+			time(&current_gps_position.last_updated);
```
InterLinked1, line 1023: Similar here with time

apps/app_gps.c
```diff
@@ -693,316 +986,448 @@ static void *gpsthread(void *data)
...
-		sprintf(buf, "/bin/mv %s %s > /dev/null 2>&1", GPS_WORK_FILE, GPS_DATA_FILE);
-		ast_safe_system(buf);
+		
+		ast_debug(5, "Got latitude: %s, longitude: %s, elevation: %s from: %s\n", 
+			selected_info->latitude, selected_info->longitude, selected_info->elevation,
+			(selected_info == &current_gps_position) ? "GPS" : "Default");
 	}
-	close(fd);
+	
+	if (fd) {
```
InterLinked1, line 1034: `if (fd != -1)`

And do we then need to set to -1 after? (I haven't checked, maybe not)

apps/app_gps.c
```diff
@@ -1151,27 +1759,88 @@ static int load_module(void)
...
+	sender_entry = ast_calloc(1, sizeof(*sender_entry));
+	sender_entry->type = APRS;
+	strcpy(sender_entry->stanza, "general");
+	ast_mutex_init(&sender_entry->lock);
+	ast_cond_init(&sender_entry->condition, 0);
+	AST_LIST_INSERT_TAIL(&aprs_sender_list, sender_entry, list);
+	ast_pthread_create(&sender_entry->thread_id, NULL, aprs_sender_thread, sender_entry);
+	
+	/* Create the aprs tt sender thread for 'general' */
+	sender_entry = ast_calloc(1, sizeof(*sender_entry));
```
InterLinked1, line 1801: Check for allocation failure

apps/app_gps.c
```diff
@@ -1151,27 +1759,88 @@ static int load_module(void)
...
+	}
+
+	/* Create the aprs connection thread */
+	ast_pthread_create(&aprs_connection_thread_id, NULL, aprs_connection_thread, NULL);
+	/* If we have a comport specified, start the GPS processing thread */
+	if (comport) {
+		ast_pthread_create(&gps_reader_thread_id, NULL, gps_reader, NULL);
+	}
+	/* Create the aprs sender thread for 'general' */
+	sender_entry = ast_calloc(1, sizeof(*sender_entry));
```
InterLinked1, line 1792: Check for allocation failure

apps/app_gps.c
```diff
@@ -1151,27 +1759,88 @@ static int load_module(void)
...
+			strcpy(general_def_position.elevation, "000.0M");
+		}
+	}
+
+	/* Create the aprs connection thread */
+	ast_pthread_create(&aprs_connection_thread_id, NULL, aprs_connection_thread, NULL);
+	/* If we have a comport specified, start the GPS processing thread */
+	if (comport) {
+		ast_pthread_create(&gps_reader_thread_id, NULL, gps_reader, NULL);
+	}
```
InterLinked1, line 1790: Need to handle pthread_create failure

apps/app_gps.c
```diff
@@ -1151,27 +1759,88 @@ static int load_module(void)
...
 	while ((ctg = ast_category_browse(cfg, ctg)) != NULL) {
-		if (ctg == NULL)
+		if (ctg == NULL) {
 			continue;
-		/* Don't clobber gps_thread */
-		ast_pthread_create_detached(&gps_section_thread, NULL, gps_sub_thread, ast_strdup(ctg));
-		ast_pthread_create_detached(&gps_section_thread, NULL, gps_tt_thread, ast_strdup(ctg));
+		}
+		/* Create the aprs sender thread for this category */
+		sender_entry = ast_calloc(1, sizeof(*sender_entry));
```
InterLinked1, line 1817: alloc can fail

apps/app_gps.c
```diff
@@ -1151,27 +1759,88 @@ static int load_module(void)
...
+		ast_pthread_create(&sender_entry->thread_id, NULL, aprs_sender_thread, sender_entry);
+		
+		/* Create the aprs tt sender thread for this category */
+		sender_entry = ast_calloc(1, sizeof(*sender_entry));
+		sender_entry->type = APRSTT;
+		ast_copy_string(sender_entry->stanza, ctg, sizeof(sender_entry));
+		ast_mutex_init(&sender_entry->lock);
+		ast_cond_init(&sender_entry->condition, 0);
+		AST_LIST_INSERT_TAIL(&aprs_sender_list, sender_entry, list);
+		ast_pthread_create_detached(&sender_entry->thread_id, NULL, aprstt_sender_thread, sender_entry->stanza);
```
InterLinked1, line 1832: can fail

apps/app_gps.c
```diff
@@ -1151,27 +1759,88 @@ static int load_module(void)
...
+		sender_entry = ast_calloc(1, sizeof(*sender_entry));
+		sender_entry->type = APRS;
+		ast_copy_string(sender_entry->stanza, ctg, sizeof(sender_entry->stanza));
+		ast_mutex_init(&sender_entry->lock);
+		ast_cond_init(&sender_entry->condition, 0);
+		AST_LIST_INSERT_TAIL(&aprs_sender_list, sender_entry, list);
+		ast_pthread_create(&sender_entry->thread_id, NULL, aprs_sender_thread, sender_entry);
+		
+		/* Create the aprs tt sender thread for this category */
+		sender_entry = ast_calloc(1, sizeof(*sender_entry));
```
InterLinked1, line 1826: can fail

InterLinked1 (changes requested):

apps/app_gps.c
KB4MDD: I have replaced stanza with 'section'.

apps/app_gps.c
KB4MDD: All instances have been changed.

apps/app_rpt.c
```diff
@@ -4976,28 +4982,36 @@ static void *rpt(void *this)
 
 		time(&t);
```
Allan-N, line 4983: Haven't we been down this path before?  Since we are using this time for a comparison then shouldn't we been be using rpt_time() ?

apps/app_gps.c
```diff
@@ -1783,53 +1781,86 @@ static int load_module(void)
...
 	AST_LIST_INSERT_TAIL(&aprs_sender_list, sender_entry, list);
-	ast_pthread_create(&sender_entry->thread_id, NULL, aprstt_sender_thread, sender_entry);
-	/* 
-	 * See if we have stanzas other than general. 
-	 * If present, create aprs processing threads for those stanzas 
+	if (ast_pthread_create(&sender_entry->thread_id, NULL, aprstt_sender_thread, sender_entry)) {
+		ast_log(LOG_ERROR, "Cannot create APRStt sender thread %s", sender_entry->section);
+		return -1;
+	}
+	/* 	if (!sender_entry) {
```
InterLinked1, line 1824: is this supposed to be here?

KB4MDD: Sorry - a copy and paste error.  Corrected.

configs/rpt/gps.conf
```diff
@@ -1,35 +1,92 @@
...
+                            ;  South America: soam.aprs2.net
+                            ;  Europe/Africa: euro.aprs2.net
+                            ;  Asia: asia.aprs2.net
+                            ;  Oceania: aunz.aprs2.net
+							;  Global Hamnet: aprs.hc.r1.ampr.org
+port = 14580                ; port on APRS-IS server to send data
 interval = 600              ; Beacon interval in seconds
-icon = r                    ; A CAR (default) Icon to be displayed for station on APRS display
+                            ;  Mobile 60 to 120, infrastructure 1200, fixed base 1800
+comment = AllStar Node 1999 ; Text to be displayed for this station
```
Allan-N, line 21: This "feels" like a per-node configuration (not really "general" if we have more than one node)

configs/rpt/gps.conf
```diff
@@ -1,35 +1,92 @@
...
 comport = /dev/ttyS0        ; Serial port for GPS receiver (specify this only if using GPS receiver)
-baudrate = 4800             ; Baud rate for GPS receiver (specify this only if using GPS receiver)
-debug = n                   ; set this for debug output
+baudrate = 4800             ; Baud rate for GPS receiver (specify this only if using GPS receiver), default 4800
 freq = 145.000              ; Display Frequency of station
-tone = 100.0                ; CTCSS tone of station (0.0 for none)
-lat = 12.3456               ; Fixed (default) latitude in decimal degrees
-lon = -123.4567             ; Fixed (default) longitude in decimal degrees
+tone = 100.0                ; CTCSS tone of station (0.0 for none) 
+lat = 12.3456               ; Fixed latitude in decimal degrees (positive for North, negative for South)
```
Allan-N, line 31: How does this work?  When is this lat/lon/elev used when you are using reading the info from a GPS?  Is the idea that one would setup the config either comport+baudrate OR lat/lon/elev?

apps/app_rpt.c
KB4MDD: Was that ast_radio_time?

I don't think any of the timing in this application is critical.  Being seconds off are not a problem with APRS reporting.
 
If you think we should move to `clock_gettime(CLOCK_MONOTONIC, &ts)`, I will make those changes.

configs/rpt/gps.conf
```diff
@@ -1,35 +1,92 @@
...
+;
+; These settings are for APRStt (touch tone) mode.  APRStt must be enabled in rpt.conf
+;
+; ttcomment =                 ; Text to be displayed associated with this TT station
+; ttlat =                     ; Fixed latitude in decimal degrees - if not specified 'lat' is used for the default
+; ttlon =                     ; Fixed longitude in decimal degrees - if not specified 'lon' is used for the default
+; ttelev =                    ; Elevation of Antenna in Meters (*NOT* HAAT) - if not specified 'elev' is used for the default
+; ttlist =                    ; Number of touch tone entries to maintain.  Default is 10.
+; ttoffset =                  ; TT Offset from main lat/lon value. Default is 10.
+; ttsplit =                   ; yes = center reports around lat/lon origin, no = (default) only use ttoffset
```
Allan-N, line 90: Here's a general comment about the gps.conf file.  It looks to me like we have a bunch of settings that can are grouped together.  We have the call, password, server, port, and interval that all relate to reporting the information.  We have comport, baudrate for the connected GPS receiver.  We have freq, tone, power, height, gain, and dir that are about the station.  Maybe put the related settings together with a comment block?

apps/app_rpt.c
Allan-N: Definitely, we should be using a CLOCK_MONOTONIC time anywhere we are checking elapsed time.

and, yes, it very well could be ast_radio_time.  I picked up rpt_time when looking at some of the PRs from a while back.

KB4MDD: I could do this for few places, but the program is using epoch for a reason.  Some data is stored that needs epoch.  The CLOCK_MONOTONIC is based off of boot time and would not work where we need to report epoch.

Would you like me to change the places where CLOCK_MONOTONIC would work?

configs/rpt/gps.conf
KB4MDD: [general] was used by the original code.  I left it as such so that existing users would not be affected.  

The ability to use multiple sections was never documented.  There was one reference by Jim on Community about that feature.  

If you have two nodes at one site and want to have a different 'comment', you would need to sections.

In all reality, the APRS-IS settings should be in a [settings] section.  You could then configure one section per node.  This however would break existing users.

I am open to discussion.

apps/app_rpt.c
Allan-N: I think any place where we are looking for the elapsed time needs to use CLOCK_MONOTONIC.  If there are places where we want a "date/time" stamp then, yes, using a time based on the epoch makes sense.  And, if we have a single time that we're trying to use for both purposes then maybe they need to be split.

configs/rpt/gps.conf
KB4MDD: I have both setup in my install.  It pulls from the GPS as primary and falls back to the defaults lat/lon/elev when the GPS is not available.  This replicates the original intent by Jim (not saying that this is correct, but it does work.)

configs/rpt/gps.conf
KB4MDD: The call is used for both aprs-is login and reporting.

I will adjust the other items.

configs/rpt/gps.conf
Allan-N: So, the lat/lon/elev are used when we can't acquire GPS data.  Adding a comment above these 3 settings would clear this up.

configs/rpt/gps.conf
Allan-N: I think it's fine if an old config will mostly work.  But, we can also have our configuration help new folks start off on a clearer path.  I also think it's important for us to provide a good example for both the simple and more complex cases.

KB4MDD: Let me know what changes you have in mind for the conf file since the last update.

configs/rpt/gps.conf
KB4MDD: Changes made for review.

apps/app_rpt.c
KB4MDD: Changes made in the appropriate places.

apps/app_gps.c
```diff
@@ -693,316 +1001,446 @@ static void *gpsthread(void *data)
...
-					}
-				}
+		now = time_monotonic();
+		
+		ast_mutex_lock(&position_update_lock);
+		selected_position.is_valid = 0;
+		/* See if we need to send live GPS or the default */
+		if (current_gps_position.is_valid) {
+			selected_position = current_gps_position;
+		} else {
```
mkmer, line 1157: `else if {` ?

apps/app_gps.c
```diff
@@ -693,316 +1001,446 @@ static void *gpsthread(void *data)
...
 			}
 		}
-		if ((!gotfiledata) && (!ehlert) && deflat && deflon) {
+		ast_mutex_unlock(&position_update_lock);
+		/* 
+		 * See if it is time to send the position report
+		 * The last_updated time must be current so that
+		 * we know we are getting good GPS information.
+		 */
+		if (selected_position.is_valid && (selected_position.last_updated + GPS_VALID_SECS) >= time(NULL)) {
```
mkmer, line 1169: I don't know - maybe all in one if statement?   I don't see a reason for the second `if`

apps/app_gps.c
```diff
@@ -293,194 +402,302 @@ static int getserialchar(int fd)
...
-			close(sockfd);
-			sockfd = -1;
-			ast_mutex_unlock(&gps_lock);
+		ast_sockaddr_set_port(&addr, atoi(port));
+		
+		if (ast_connect(sockfd, &addr) < 0) {
+			ast_log(LOG_WARNING, "Cannot connect to server %s. Error: %s\n", server, strerror(errno));
+			ast_mutex_unlock(&aprs_socket_lock);
+			/* Wait 500ms before trying again. */
 			usleep(500000);
```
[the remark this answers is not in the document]
KB4MDD: Would you prefer 1 second?

I would like for it to retry.  That is helpful for mobile users.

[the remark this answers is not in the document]
mkmer: First off - it has been this way forever, so it may not be worth messing with it.  I also may be misinterpreting what is going on here.

Usually when trying to connect to a server there is a backoff function - say 1 second for the first retry, then (timer * 2^number of retries) for the back off (exponential backoff) so 1,2,4,8,16,32...etc. up to a limit. The when you do succeed, the counter goes back to 0, and the next retry is back to 1 second.  If you wanna get real crazy, they toss in a random number multiplier too (consider a server resets and 10,000 clients try to connect at the same time - the random part will get everyone to offset a bit allowing the server to breath a bit).

configs/rpt/gps.conf
KB4MDD: Your templated version is adopted.  Thank you for the help.

apps/app_gps.c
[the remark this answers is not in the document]
mkmer: I don't think we should hold up the PR for this. 
I'll look for some examples and we can review it again then.  
There are probably other examples of this general problem (like `permanent link reconnect` retries, etc) I suppose as long as we have a maximum retry count, the servers will catch a break as clients give up. 

Sometimes questionable things (like this one) cause flaky behavior that is hard to pin point when it's used at scale.

mkmer (approved):

Allan-N (commented):
I'm making one of my last passes over these changes and, again, want to raise the issue of using a monotonic clock.  I have observed a bunch of places in the code where we are still comparing the time returned by the GPS_READ function and comparing that to what's returned by a call to time(&t).  In these instances I think we really want to get away fro time(&t).

Here's a thought.  When we are saving the current time for future calls to GPS_READ we should store TWO timers (e.g. extend the `position_info` struct to have both values).  One would be epoch based and one would be monotinic.  The calls to GPS_READ should return BOTH time values.  Then, if the code needs to report when the data was captured it will use the epoch value.  But, if the code is checking the elapsed time it would use the monotonic value.

And, as an aside, you might want to think about having a single/shared function to check for the availability of the GPS_READ function, to make the call, and to return the results (or an indication that the data is not available).

apps/app_gps.c
```diff
@@ -293,194 +402,302 @@ static int getserialchar(int fd)
...
+			ast_log(LOG_WARNING, "Cannot connect to server %s. Error: %s\n", server, strerror(errno));
+			ast_mutex_unlock(&aprs_socket_lock);
+			/* Wait 1 second before trying again. */
+			sleep(1);
 			continue;
 		}
-		sprintf(buf, "user %s pass %s vers \"Asterisk app_gps\"\n", call, password);
+
+		/* Log into the APRS-IS server */
+		sprintf(buf, "user %s pass %s vers Asterisk app_gps_V3\n", call, password);
```
Allan-N, line 539: Please use snprintf()

apps/app_gps.c
```diff
@@ -568,49 +822,97 @@ static int report_aprstt(char *ctg, char *lat, char *lon, char *theircall, char
 		*(cp + 3) = lon[strlen(lon) - 1];
 		*(cp + 4) = 0;
 	}
-	time(&t);
+	
+	t = time(NULL);
 	tm = gmtime(&t);
+	
 	sprintf(buf1, "%s-12", theircall);
```
Allan-N, line 829: Please use snprintf()

apps/app_gps.c
InterLinked1: I also wonder if stuff like "Asterisk app_gps_V3" should be a macro at the top of the module?

apps/app_gps.c
mkmer: Is this `snprintf()` one of the things we should be doing on everything we touch? 
Wondering if we should start a running list of "things to update for devs" as we work on refactors. 
So far I have 2: adding `{}` and use `snprintf()` - probably more?

Allan-N: >Is this snprintf() one of the things we should be doing on everything we touch?

IMHO, yes.

[Secure Coding Guide](https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Introduction.html)

mkmer: I don't disagree at all.  Buffer overruns is the #1 code hunt for hackers. 
I was thinking we should have a "cheat sheet" for what we want to fix if we touch something, and what a reviewer should look for if it is touched.

apps/app_gps.c
KB4MDD: Time issues have been addressed.
All sprintf's have been changed to snprintf.
APRS-IS version string is now a define.

mkmer (approved):

apps/app_rpt.c
```diff
@@ -4976,28 +4982,36 @@ static void *rpt(void *this)
...
 		while (t >= (myrpt->lastgpstime + GPS_UPDATE_SECS)) {
```
Allan-N, line 4984: Should "t" and "->lastgpstime" be switched to the monotonic time?

apps/app_rpt.c
```diff
@@ -4976,28 +4982,36 @@ static void *rpt(void *this)
...
-			if (fscanf(fp, "%u %s %s %s", &u, lat, lon, elev) < 3)
+			}
+
+			if (sscanf(gps_data, "%llu %llu %s %s %s", &u_mono, &u_epoch, lat, lon, elev) != 5) {
 				break;
-			fclose(fp);
-			was = (time_t) u;
-			if ((was + GPS_VALID_SECS) < t)
+			}
+			was = (time_t) u_epoch;
```
Allan-N, line 5002: Should "was" be using "u_mono" ?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2337,22 +2336,22 @@ void *rpt_tele_thread(void *this)
...
-		elev[0] = 0;
-		if (fscanf(fp, "%u %s %s %s", &u, lat, lon, elev) < 3) {
+
+		if (sscanf(gps_data, "%llu %llu %s %s %s", &u_mono, &u_epoch, lat, lon, elev) != 5) {
 			break;
 		}
-		fclose(fp);
-		was = (time_t) u;
+		
+		was = (time_t) u_epoch;
```
Allan-N, line 2354: Should "was" and "t" be using monotonic time?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2767,25 +2765,20 @@ void rpt_telemetry(struct rpt *myrpt, int mode, void *data)
...
-			if (fscanf(fp, "%u %s %s %s", &k, lat, lon, elev) < 3) {
-				fclose(fp);
+
+			if (sscanf(gps_data, "%llu %llu %s %s %s", &u_mono, &u_epoch, lat, lon, elev) != 5) {
 				break;
 			}
-			fclose(fp);
-			was = (time_t) k;
+
+			was = (time_t) u_epoch;
```
Allan-N, line 2781: Should "was" and "t" be using the monotonic time?

channels/chan_echolink.c
```diff
@@ -3366,10 +3363,9 @@ static void *el_reader(void *data)
...
 				mylat = instp->lat;
 				mylon = instp->lon;
-				fp = fopen(GPSFILE, "r");
-				if (fp && (fstat(fileno(fp), &mystat) != -1) && (mystat.st_size < 100)) {
-					if (fscanf(fp, "%u %f%c %f%c", &u, &lat, &latc, &lon, &lonc) == 5) {
-						was = (time_t) u;
+				if (ast_custom_function_find("GPS_READ") && !ast_func_read(NULL, "GPS_READ()", gps_data, sizeof(gps_data))) {
+					if (sscanf(gps_data, "%llu %llu %f%c %f%c", &u_mono, &u_epoch, &lat, &latc, &lon, &lonc) == 6) {
+						was = (time_t) u_epoch;
 						if ((was + GPS_VALID_SECS) >= now) {
```
Allan-N, line 3369: Should "was" and "now" be using monotonic time?

channels/chan_echolink.c
```diff
@@ -567,6 +566,20 @@ static char show_nodes_usage[] = "Usage: echolink show nodes\n";
 
 static char show_stats_usage[] = "Usage: echolink show stats\n";	
 
+/*!
```
mkmer, line 569: Why can't we use the one from rpt_utils.c?

channels/chan_echolink.c
```diff
@@ -567,6 +566,20 @@ static char show_nodes_usage[] = "Usage: echolink show nodes\n";
...
+ * \brief Get system monotonic 
+ * This returns the CLOCK_MONOTONIC time
+ * \retval		Monotonic seconds.
+ */
+static time_t time_monotonic(void)
```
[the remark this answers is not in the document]
KB4MDD: This would place a dependency on this channel driver to the application.  I however don't have an opinion.  This function first existed in res_usbradio.

apps/app_rpt.c
```diff
@@ -4996,11 +4997,12 @@ static void *rpt(void *this)
 				break;
 			}
 
-			if (sscanf(gps_data, "%llu %llu %s %s %s", &u_mono, &u_epoch, lat, lon, elev) != 5) {
+			/* gps_data format monotonic time, epoch, latitude, longitude, elevation */
+			if (sscanf(gps_data, "%llu %*u %s %s %s", &u_mono, lat, lon, elev) != 4) {
```
Allan-N, line 5001: 5 formats, 4 variables, and checking != 4 :-(

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2347,13 +2347,14 @@ void *rpt_tele_thread(void *this)
 			break;
 		}
 
-		if (sscanf(gps_data, "%llu %llu %s %s %s", &u_mono, &u_epoch, lat, lon, elev) != 5) {
+		/* gps_data format monotonic time, epoch, latitude, longitude, elevation */
+		if (sscanf(gps_data, "%llu %*u %s %s %s", &u_mono, lat, lon, elev) != 4) {
```
Allan-N, line 2351: 5 formats, 4 variables, and checking != 4 :-(

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2774,13 +2775,14 @@ void rpt_telemetry(struct rpt *myrpt, int mode, void *data)
 				break;
 			}
 
-			if (sscanf(gps_data, "%llu %llu %s %s %s", &u_mono, &u_epoch, lat, lon, elev) != 5) {
+			/* gps_data format monotonic time, epoch, latitude, longitude, elevation */
+			if (sscanf(gps_data, "%llu %*u %s %s %s", &u_mono, lat, lon, elev) != 4) {
```
Allan-N, line 2779: 5 formats, 4 variables, and checking != 4 :-(

channels/chan_echolink.c
```diff
@@ -3364,9 +3379,11 @@ static void *el_reader(void *data)
 				mylat = instp->lat;
 				mylon = instp->lon;
 				if (ast_custom_function_find("GPS_READ") && !ast_func_read(NULL, "GPS_READ()", gps_data, sizeof(gps_data))) {
-					if (sscanf(gps_data, "%llu %llu %f%c %f%c", &u_mono, &u_epoch, &lat, &latc, &lon, &lonc) == 6) {
-						was = (time_t) u_epoch;
-						if ((was + GPS_VALID_SECS) >= now) {
+					/* gps_data format monotonic time, epoch, latitude, longitude, elevation */
+					if (sscanf(gps_data, "%llu %*u %f%c %f%c", &u_mono, &lat, &latc, &lon, &lonc) == 5) {
```
Allan-N, line 3383: 6 formats, 5 variables, and checking == 5 :-(

Allan-N (changes requested):

apps/app_rpt.c
KB4MDD: The second variable is %*u.  As I understand this, this means to skip that number.  My tests seem to indicate that it returns 4 properly.  Am I missing something?

Allan-N: Ya know.  I've been coding for a bunch of years and have never seen/used the '*' assignment-suppression character used.  Now I understand and all is good.

Allan-N (approved):

channels/chan_echolink.c
Allan-N: Never mind.  %*u

apps/app_rpt/rpt_telemetry.c
Allan-N: Never mind.  %*u

apps/app_rpt/rpt_telemetry.c
Allan-N: Never mind.  %*u

mkmer (approved):

apps/app_gps.c
```diff
@@ -1012,85 +1452,259 @@ static void *gps_tt_thread(void *data)
...
+	if (ast_strlen_zero(args.overlay)) {
+		ast_log(LOG_ERROR, "APRS_SENDTT requires an overlay\n");
+		return -1;
+	}
+	if (ast_strlen_zero(value)) {
+		ast_log(LOG_ERROR, "APRS_SENDTT requires a callsign\n");
+		return -1;
+	}
+	/* Find the section */
+	AST_LIST_TRAVERSE(&aprs_sender_list, sender_entry, list) {
```
InterLinked1, line 1569: Why does this list not have a lock? It can be called from the function write callback, so without any other locking, this list could be accessed simultaneously be multiple threads.

KB4MDD: Thank you for finding this.  It has been corrected.

apps/app_gps.c
```diff
@@ -1566,11 +1567,13 @@ static int aprs_sendtt_helper(struct ast_channel *chan, const char *function, ch
 		return -1;
 	}
 	/* Find the section */
+	ast_mutex_lock(&aprs_sender_list_lock);
```
InterLinked1, line 1570: Sorry, I didn't mean just this part, I meant every instance of aprs_sender_list.
Linked lists come with their own lock, you disabled it using the NOLOCK variant, but if you remove that, then you can RDLOCK or WRLOCK the list as appropriate when using it. There's no need for a separately maintained mutex for the list, that can then be removed.

apps/app_gps.c
```diff
@@ -1151,27 +1778,118 @@ static int load_module(void)
...
+	/* Create the aprs sender thread for 'general' */
+	sender_entry = ast_calloc(1, sizeof(*sender_entry));
+	if (!sender_entry) {
+		return -1;
+	}
+	sender_entry->type = APRS;
+	strcpy(sender_entry->section, "general");
+	ast_mutex_init(&sender_entry->lock);
+	ast_cond_init(&sender_entry->condition, 0);
+	AST_LIST_INSERT_TAIL(&aprs_sender_list, sender_entry, list);
```
InterLinked1, line 1825: This needs to be surrounded by a WRLOCK

apps/app_gps.c
```diff
@@ -1151,27 +1778,118 @@ static int load_module(void)
...
+	/* Create the aprs tt sender thread for 'general' */
+	sender_entry = ast_calloc(1, sizeof(*sender_entry));
+	if (!sender_entry) {
+		return -1;
+	}
+	sender_entry->type = APRSTT;
+	strcpy(sender_entry->section, "general");
+	ast_mutex_init(&sender_entry->lock);
+	ast_cond_init(&sender_entry->condition, 0);
+	AST_LIST_INSERT_TAIL(&aprs_sender_list, sender_entry, list);
```
InterLinked1, line 1840: This needs to be surrounded by a WRLOCK

apps/app_gps.c
```diff
@@ -1151,27 +1778,118 @@ static int load_module(void)
...
+		/* Create the aprs sender thread for this category */
+		sender_entry = ast_calloc(1, sizeof(*sender_entry));
+		if (!sender_entry) {
+			return -1;
+		}
+		sender_entry->type = APRS;
+		ast_copy_string(sender_entry->section, ctg, sizeof(sender_entry->section));
+		ast_mutex_init(&sender_entry->lock);
+		ast_cond_init(&sender_entry->condition, 0);
+		AST_LIST_INSERT_TAIL(&aprs_sender_list, sender_entry, list);
```
InterLinked1, line 1862: This needs to be surrounded by a WRLOCK

apps/app_gps.c
```diff
@@ -1151,27 +1778,118 @@ static int load_module(void)
...
+		/* Create the aprs tt sender thread for this category */
+		sender_entry = ast_calloc(1, sizeof(*sender_entry));
+		if (!sender_entry) {
+			return -1;
+		}
+		sender_entry->type = APRSTT;
+		ast_copy_string(sender_entry->section, ctg, sizeof(sender_entry));
+		ast_mutex_init(&sender_entry->lock);
+		ast_cond_init(&sender_entry->condition, 0);
+		AST_LIST_INSERT_TAIL(&aprs_sender_list, sender_entry, list);
```
InterLinked1, line 1877: This needs to be surrounded by a WRLOCK

InterLinked1 (changes requested):

apps/app_gps.c
KB4MDD: Sorry - I did not think the lock was needed in module load.

InterLinked1: Strictly speaking, it isn't really, but it's probably good to be consistent in usage

InterLinked1 (approved):

PR #468, opened 2025-01-26
modules.conf: update for Asterisk 22.x.x

Allan-N:
A number of Asterisk modules were deprecated (and removed) in 21.x.x and 22.x.x.  These changes update the ASL3 provided "modules.conf" file to sync with those changes as well as some overdue cleanup.

*** NOTE ***
The file currently includes some added content and a bunch of questions. For those reviewing the changes, your comments and guidance are welcomed.

### DIFF
--- a/configs/rpt/modules.conf
+++ b/configs/rpt/modules.conf
@@ -1,132 +1,60 @@
 ;
-; Asterisk configuration file
+; Asterisk module loading configuration file
 ;
-; Module Loader configuration file
-;
-; By default DIAL does NOT load every module, only what is needed
 
+;
 ; You can enable or disable any of the asterisk modules
-; All modules are compiled and installed.
+;
+; To enable a module, use  : load = module_name.so
+; To disable a module, use : noload = module_name.so
+;
 
-; To enable a module: load => module_name.so
-; To disable a module: noload => module_name.so
-
-; You will want to enable the channel driver modules you will be using.
-; There are below in the Channel Driver section
-; The most common Channel drivers for app_rpt are:
-; chan_echolink.so   echolink channel driver
-; chan_simpleusb.so  Simple USB Radio Interface Channel Drive
-; chan_usbradio.so   USB Console Channel Driver
-; chan_usrp.so       USRP Channel Module
-; chan_voter.so      radio Voter channel driver
+;
+; For ASL (app_rpt), you need to enable the channel driver modules you will
+; be using. You will find the modules in the "Channel Driver" section, below.
+;
+; The most common ASL (app_rpt) channel drivers are:
+;
+;   chan_dahdi.so      DAHDI Telephony
+;   chan_echolink.so   Echolink Channel Driver
+;   chan_simpleusb.so  Simple USB Radio Interface Channel Driver
+;   chan_usbradio.so   USB Console Channel Driver
+;   chan_usrp.so       USRP Channel Module
+;   chan_voter.so      Voter Radio Channel Driver
+;
 
 [modules]
-autoload=no
+autoload = no
 
-;load => func_frame_trace            ; Debug
+;load = func_frame_trace                ; Debug
 
 ; Applications
-noload => app_adsiprog.so           ; Asterisk ADSI Programming Application
-noload => app_alarmreceiver.so      ; Alarm Receiver for Asterisk
-noload => app_amd.so                ; Answering Machine Detection Application
-load => app_authenticate.so         ; Authentication Application
-noload => app_cdr.so                ; Tell Asterisk to not maintain a CDR for
-noload => app_chanisavail.so        ; Check channel availability
-noload => app_channelredirect.so    ; Channel Redirect
-noload => app_chanspy.so            ; Listen to the audio of an active channel
-noload => app_controlplayback.so    ; Control Playback Application
-noload => app_dahdibarge.so         ; Barge in on channel application
-noload => app_dahdiras.so           ; DAHDI RAS Application
-noload => app_dahdiscan.so          ; Scan Zap channels application
-noload => app_db.so                 ; Database Access Functions
-load => app_dial.so                 ; Dialing Application
-noload => app_dictate.so            ; Virtual Dictation Machine
-noload => app_directed_pickup.so    ; Directed Call Pickup Application
-noload => app_directory.so          ; Extension Directory
-noload => app_disa.so               ; DISA (Direct Inward System Access) Appli
-noload => app_dumpchan.so           ; Dump Info About The Calling Channel
-noload => app_echo.so               ; Simple Echo Application
-load => app_exec.so                 ; Executes dialplan applications
-noload => app_externalivr.so        ; External IVR Interface Application
-noload => app_festival.so           ; Simple Festival Interface
-noload => app_flash.so              ; Flash channel application
-noload => app_followme.so           ; Find-Me/Follow-Me Application
-noload => app_forkcdr.so            ; Fork The CDR into 2 separate entities
-noload => app_getcpeid.so           ; Get ADSI CPE ID
-noload => app_gps.so                ; GPS interface module
-noload => app_hasnewvoicemail.so    ; Indicator for whether a voice mailbox ha
-noload => app_ices.so               ; Encode and Stream via icecast and ices
-noload => app_image.so              ; Image Transmission Application
-noload => app_lookupblacklist.so    ; Look up Caller*ID name/number from black
-noload => app_lookupcidname.so      ; Look up CallerID Name from local database
-noload => app_macro.so              ; Extension Macros
-noload => app_meetme.so             ; MeetMe conference bridge
-noload => app_milliwatt.so          ; Digital Milliwatt (mu-law) Test Applicat
-noload => app_mixmonitor.so         ; Mixed Audio Monitoring Application
-noload => app_morsecode.so          ; Morse code
-noload => app_mp3.so                ; Silly MP3 Application
-noload => app_nbscat.so             ; Silly NBS Stream Application
-noload => app_page.so               ; Page Multiple Phones
-noload => app_parkandannounce.so    ; Call Parking and Announce Application
-load => app_playback.so             ; Sound File Playback Application
-noload => app_privacy.so            ; Require phone number to be entered, if n
-noload => app_queue.so              ; True Call Queueing
-noload => app_radbridge.so          ; Radio Bridging interface module
-noload => app_random.so             ; Random goto
-noload => app_readfile.so           ; Stores output of file into a variable
-noload => app_read.so               ; Read Variable Application
-noload => app_realtime.so           ; Realtime Data Lookup/Rewrite
-noload => app_record.so             ; Trivial Record Application
-load => app_rpt.so                  ; Radio Repeater/Remote Base Application
-noload => app_sayunixtime.so        ; Say time
-noload => app_senddtmf.so           ; Send DTMF digits Application
-load => app_sendtext.so             ; Send Text Applications
-noload => app_setcallerid.so        ; Set CallerID Application
-noload => app_setcdruserfield.so    ; CDR user field apps
-noload => app_settransfercapability.so   ; Set ISDN Transfer Capability
-noload => app_sms.so                ; SMS/PSTN handler
-noload => app_softhangup.so         ; Hangs up the requested channel
-noload => app_speech_utils.so       ; Dialplan Speech Applications
-noload => app_stack.so              ; Stack Routines
-load => app_system.so               ; Generic System() application
-noload => app_talkdetect.so         ; Playback with Talk Detection
-noload => app_test.so               ; Interface Test Application
-load => app_transfer.so             ; Transfer
-noload => app_url.so                ; Send URL Applications
-noload => app_userevent.so          ; Custom User Event Application
-noload => app_verbose.so            ; Send verbose output
-noload => app_voicemail.so          ; Comedian Mail (Voicemail System)
-noload => app_waitforring.so        ; Waits until first ring after time
-noload => app_waitforsilence.so     ; Wait For Silence
-noload => app_while.so              ; While Loops and Conditional Execution
-noload => app_zapateller.so         ; Block Telemarketers with Special Informa
 
-; CDR
-noload => cdr_csv.so                ; Comma Separated Values CDR Backend
-noload => cdr_custom.so             ; Customizable Comma Separated Values CDR
-noload => cdr_manager.so            ; Asterisk Manager Interface CDR Backend
+load    = app_authenticate.so            ; Authentication Application
+load    = app_dial.so                    ; Dialing Application
+load    = app_exec.so                    ; Executes dialplan applications
+noload  = app_gps.so                     ; GPS Interface
+load    = app_playback.so                ; Sound File Playback Application
+require = app_rpt.so                     ; Radio Repeater/Remote Base Application
+load    = app_sendtext.so                ; Send and Receive Text Applications
+load    = app_system.so                  ; Generic System() application
+load    = app_transfer.so                ; Transfers a caller to another extension
 
-; Channels
-noload => chan_agent.so             ; Agent Proxy Channel
-noload => chan_alsa.so              ; ALSA Console Channel Driver
-noload => chan_beagle.so            ; Beagleboard Radio Interface Channel Driver
-load => chan_dahdi.so               ; DAHDI Telephony
-noload => chan_echolink.so          ; echolink Channel Driver
-noload => chan_features.so          ; Feature Proxy Channel
-noload => chan_gtalk.so             ; Gtalk Channel Driver
-load => chan_iax2.so                ; Inter Asterisk eXchange (Ver 2)
-noload => chan_local.so             ; Local Proxy Channel (Note: used internal
-noload => chan_oss.so               ; Channel driver for OSS sound cards
-noload => chan_phone.so             ; Generic Linux Telephony Interface driver
-noload => chan_pi.so                ; DMK Engineering "PITA" Board on Rpi2/3 Channel Driver
-load => chan_simpleusb.so           ; CM1xx USB Cards with Radio Interface Channel Driver (No DSP)
-noload => chan_sip.so               ; Session Initiation Protocol (SIP)
-noload => chan_tlb.so               ; TheLinkBox Channel Driver
-noload => chan_usbradio.so          ; CM1xx USB Cards with Radio Interface Channel Driver (DSP)
-noload => chan_usrp.so              ; GNU Radio interface USRP Channel Driver
-noload => chan_voter.so             ; Radio Voter Channel Driver
+; Channel Drivers
+
+load    = chan_dahdi.so                  ; DAHDI Telephony w/PRI & SS7 & MFC/R2
+noload  = chan_echolink.so               ; Echolink Channel Driver
+require = chan_iax2.so                   ; Inter Asterisk eXchange (Ver 2)
+noload  = chan_mobile.so                 ; Bluetooth Mobile Device Channel Driver
+noload  = chan_ooh323.so                 ; Objective Systems H323 Channel
+load    = chan_simpleusb.so              ; SimpleUSB Radio Interface Channel Driver
+noload  = chan_tlb.so                    ; TheLinkBox Channel Driver
+noload  = chan_usbradio.so               ; USB Console Channel Driver
+noload  = chan_usrp.so                   ; USRP Channel Module
+noload  = chan_voter.so                  ; Voter Radio Channel Driver
 
 ; Codecs
+
 ; CODEC          AUDIO QUALITY   BANDWIDTH (including IP and Ethernet headers)
 ; ULAW           best            87 kilobits per second (kbps)
 ; ADPCM          good            55 kbps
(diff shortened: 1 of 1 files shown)


wd5m:
I happened to notice this message in the log. Should or can res_adsi.so be changed to noload now? 

> [2025-01-26 14:47:24.720] WARNING[730005] loader.c: Module 'res_adsi' has been loaded but may be removed in a future release.

David WD5M

Allan-N:
@wd5m : you have raised a good question.

A quick scan of the Asterisk 22.1.1 sources shows that the following modules would emit that message :
- app_getcpeid.so
- app_adsiprog.so
- app_meetme.so
- res_adsi.so

jxmx (approved):

configs/rpt/modules.conf
```diff
@@ -134,80 +114,384 @@ noload => chan_voter.so             ; Radio Voter Channel Driver
...
+; Call Detail Records
+
+load   = cdr_custom.so
+
+; Channel Drivers
+
+
+; Codecs
+
+load   = codec_resample.so
```
mkmer, line 289: I don't think we need these - maybe noload just to have them "visible" if someone needs them?

configs/rpt/modules.conf
```diff
@@ -134,80 +114,384 @@ noload => chan_voter.so             ; Radio Voter Channel Driver
...
+; load directives.  Below is what was included in that file with any of the
+; above modules removed.
+;
+; Q? do we want to add any of these in the ASL3 modules.conf?
+; Q? If yes, should we add them as "load", "noload", or ";load"?
+;********************************************************************************
+
+; Applications
+
+load   = app_bridgewait.so
```
mkmer, line 270: I'm thinking noload so they are visible.  POI:we use confbridge on our hub so maybe others use it too?

configs/rpt/modules.conf
```diff
@@ -134,80 +114,384 @@ noload => chan_voter.so             ; Radio Voter Channel Driver
...
+
+; Formats
+
+load   = format_wav.so                 ; Microsoft WAV/WAV16 format (8kHz/16kHz S
+
+; Functions
+
+load   = func_callerid.so              ; Party ID related dialplan functions (Cal
+load   = func_cdr.so                   ; Call Detail Record (CDR) dialplan functi
+load   = func_strings.so
```
mkmer, line 300: We have noload func_strings.so - maybe that is wrong?

configs/rpt/modules.conf
```diff
@@ -134,80 +114,384 @@ noload => chan_voter.so             ; Radio Voter Channel Driver
...
-noload => res_speech.so             ; Generic Speech Recognition API
-load => res_timing_dahdi.so         ; DAHDI Timing Interface
-load => res_timing_timerfd.so       ; Timerfd Timing Interface is preferred for ASL3
-load => res_usbradio.so             ; ASL3 required for both simpleusb and usbradio
-
-[global]
+
+load   = res_http_websocket.so
+load   = res_musiconhold.so
+load   = res_timing_timerfd.so         ; Timerfd Timing Interface
```
mkmer, line 309: Earlier it says: Timerfd Timing Interface is preferred for ASL3. should be load.

configs/rpt/modules.conf
```diff
@@ -134,80 +114,384 @@ noload => chan_voter.so             ; Radio Voter Channel Driver
...
-load => res_smdi.so                 ; Simplified Message Desk Interface (SMDI)
-noload => res_snmp.so               ; SNMP [Sub]Agent for Asterisk
-noload => res_speech.so             ; Generic Speech Recognition API
-load => res_timing_dahdi.so         ; DAHDI Timing Interface
-load => res_timing_timerfd.so       ; Timerfd Timing Interface is preferred for ASL3
-load => res_usbradio.so             ; ASL3 required for both simpleusb and usbradio
-
-[global]
+
+load   = res_http_websocket.so
```
mkmer, line 307: I don't see the need for these. Maybe others do? websocket &

configs/rpt/modules.conf
Allan-N: This res_timing_timerfd.so line was a dup and has been removed.

configs/rpt/modules.conf
Allan-N: I had a bunch of dups here.  The only 2 remaining are codec_resample.so and codec_g722.so

configs/rpt/modules.conf
Allan-N: This func_strings.so was a dup from our previous modules.conf file.  But, the question of noload vs. load is one to ask (ASL3 "no" or  Asterisk "yes")

configs/rpt/modules.conf
Allan-N: res_musiconhold was a dup here (but it was a noload).
res_http_websocket is a Q

configs/rpt/modules.conf
KB4MDD: I think that function should be enabled.

configs/rpt/modules.conf
KB4MDD: I don't believe we are using websockets.

mkmer:
> @wd5m : you have raised a good question.
> 
> A quick scan of the Asterisk 22.1.1 sources shows that the following modules would emit that message :
> 
>     * app_getcpeid.so
> 
>     * app_adsiprog.so
> 
>     * app_meetme.so
> 
>     * res_adsi.so

I think `app_meetme` was removed already, we had to move to  `app_confbridge.so` on our hub when we moved to ASL3.

configs/rpt/modules.conf
```diff
@@ -1,213 +1,443 @@
...
+load   = codec_alaw.so                 ; A-law Coder/Decoder
+load   = codec_a_mu.so                 ; A-law and Mulaw direct Coder/Decoder
+noload = codec_dahdi.so                ; Generic DAHDI Transcoder Codec Translator
+noload = codec_g722.so                 ; ITU G.722-64kbps G722 Transcoder
+load   = codec_g726.so                 ; ITU G.726-32kbps G726 Transcoder
+load   = codec_gsm.so                  ; GSM Coder/Decoder
+noload = codec_resample.so             ; SLIN Resampling Codec
+load   = codec_ulaw.so                 ; mu-Law Coder/Decoder
 
 ; Formats
```
mkmer, line 135: We lost `noload => format_jpeg.so ; JPEG (Joint Picture Experts Group) Image` - probably ok?
We've added `load = format_mp3.so ; MP3 format [Any rate but 8000hz mono is optimal]` -> should we load this as a default? 
More thoughts here: Do we really need to load ALL of these formats - wav, sln, gsm, PCM are probably all the is "required" or maybe "recommended", the rest could be `noload` and folks could turn them on if needed? 
Probably doesn't hurt to leave it as is either.

Allan-N: >We lost noload => format_jpeg.so ; JPEG (Joint Picture Experts Group) Image - probably ok?

It was deprecated/removed.

>We've added load = format_mp3.so ; MP3 format [Any rate but 8000hz mono is optimal] -> should we load this as a default?

It was odd that we were building format_mp3.so but not including it with the other loaded formats.

>More thoughts here: Do we really need to load ALL of these formats - wav, sln, gsm, PCM are probably all the is "required" or maybe "recommended", the rest could be noload and folks could turn them on if needed?
Probably doesn't hurt to leave it as is either.

If one is concerned about our memory footprint then there are likely other modules that we don't use that would be better to "noload"

mkmer: Not sure this is the "right' time for a rehash.  My thought was more around "why load stuff we don't need for minimal operation".  Users can add as needed (like the autopatch pjsip stuff).  
I don't have a strong opinion - there are advantages/disadvantages either way :)

Allan-N: For now, this change only "loads" `format_mp3` and `func_strings`.

As for the autopatch/pjsip modules, I wish there was a way to conditionally add (load) the entire batch without requiring a big copy/paste or many edits.  But, given the choice of loading by default, the big copy/paste, or uncommenting the block I chose the latter.

mkmer (approved):

InterLinked1:
> I think `app_meetme` was removed already, we had to move to `app_confbridge.so` on our hub when we moved to ASL3.

It hasn't actually been removed yet. One of Sangoma's customers or FreePBX (forget which) raised a big stink on it, which temporarily deferred that. Haven't heard any plans of it to be removed in 23 yet, but it'll probably be removed eventually so best to avoid it.

The other 3 modules in the warning (getcpeid, adsiprog, and res_adsi) I saved from removal and and am the maintainer of now. They aren't going to go away, but Josh didn't want to change the status back from "deprecated" to "extended" which is why the warning is there. PhreakScript patches that warning to a notice since it's a bit confusing and can cause alarm for no reason.

configs/rpt/modules.conf
```diff
@@ -1,213 +1,443 @@
 ;
-; Asterisk configuration file
+; Asterisk + AllStarLink/app_rpt  module loading configuration file
```
InterLinked1, line 2: I don't agree with this line change personally. `app_rpt` is just another Asterisk module, not something "special" running along with Asterisk. AllStarLink isn't part of Asterisk either, it leverages Asterisk as part of its "packaged product", so I think this just adds confusion.

configs/rpt/modules.conf
```diff
@@ -1,213 +1,443 @@
...
+;   chan_echolink.so   Echolink Channel Driver
+;   chan_simpleusb.so  Simple USB Radio Interface Channel Driver
+;   chan_usbradio.so   USB Console Channel Driver
+;   chan_usrp.so       USRP Channel Module
+;   chan_voter.so      Voter Radio Channel Driver
+;
 
 [modules]
-autoload=no
+autoload = no
```
InterLinked1, line 27: Possible dumb question, but why all the noload's in the first place? If autoload is no, then only the modules loaded explicitly will be loaded. The noload's are all redundant and have no effect. Removing them would simplify and clean this up quite a bit, making it easier to look at and work with

For the ASL-relevant modules, makes sense to keep the noload's, so they are there for the changing, but for the most part, I think most of them could be removed

configs/rpt/modules.conf
```diff
@@ -1,213 +1,443 @@
...
+
+;********************************************************************************
+; ASL3 app_rpt configs/rpt/modules.conf (w/"deprecated" modules removed)
```
InterLinked1, line 30: configs/rpt/modules.conf shouldn't be in here IMO, that's its source file path but not something users would really know about or make sense to them

configs/rpt/modules.conf
```diff
@@ -1,213 +1,443 @@
...
+;
+; Q? should any of the commented out lines in this section be removed?
+; Q? should any of the "load"s or "noload"s in this section be changed/removed?
```
InterLinked1, line 33: A: This is highly opinionated, this template will never make all of your users happy :)
But I'll review with the lens of "the minimum a useful install would probably want"

configs/rpt/modules.conf
```diff
@@ -1,213 +1,443 @@
...
+noload = app_amd.so                    ; Answering Machine Detection Application
+load   = app_authenticate.so           ; Authentication Application
+noload = app_bridgewait.so             ; Place the channel into a holding bridge application
+noload = app_cdr.so                    ; Tell Asterisk to not maintain a CDR for the current call
+noload = app_chanisavail.so            ; Check channel availability
+noload = app_channelredirect.so        ; Redirects a given channel to a dialplan target
+noload = app_chanspy.so                ; Listen to the audio of an active channel
+noload = app_confbridge.so             ; Conference Bridge Application
+noload = app_controlplayback.so        ; Control Playback Application
+noload = app_db.so                     ; Database Access Functions
```
InterLinked1, line 51: Doesn't Tim's whitelist/blacklist dialplan code use the database? Should that be taken into account?

configs/rpt/modules.conf
```diff
@@ -1,213 +1,443 @@
...
-load => format_wav_gsm.so           ; Microsoft WAV format (Proprietary GSM)
-load => format_wav.so               ; Microsoft WAV format (8000Hz Signed Line
+
+load   = format_g723.so                ; G.723.1 Simple Timestamp File Format
+load   = format_g726.so                ; Raw G.726 (16/24/32/40kbps) data
+load   = format_g729.so                ; Raw G.729 data
+load   = format_gsm.so                 ; Raw GSM data
+load   = format_h263.so                ; Raw H.263 data
+load   = format_h264.so                ; Raw H.264 data
+load   = format_ilbc.so                ; Raw iLBC data
```
InterLinked1, line 143: Do you really need all these formats? Highly skeptical anyone is using ILBC for example, or G729, etc. And if they are, they shouldn't be

configs/rpt/modules.conf
```diff
@@ -1,213 +1,443 @@
...
+noload = res_agi.so                    ; Asterisk Gateway Interface (AGI)
+noload = res_clioriginate.so           ; Call origination and redirection from the CLI
+noload = res_config_mysql.so           ; MySQL RealTime Configuration Driver
+noload = res_convert.so                ; File format conversion CLI command
+load   = res_crypto.so                 ; Cryptographic Digital Signatures
+load   = res_curl.so                   ; cURL Resource Module (required by ASL3)
+noload = res_http_websocket.so         ; HTTP WebSocket Support
+noload = res_musiconhold.so            ; Music On Hold Resource
+load   = res_rpt_http_registrations.so ; RPT HTTP Periodic Registrations
+load   = res_smdi.so                   ; Simplified Message Desk Interface (SMDI) Resource
```
InterLinked1, line 194: I'm not objecting to it, but I'm pretty confident most ASL users aren't using SMDI

configs/rpt/modules.conf
```diff
@@ -1,213 +1,443 @@
...
+;noload = res_pjsip_t38.so                            ; PJSIP T.38 UDPTL Support
+;noload = res_pjsip_transport_websocket.so            ; PJSIP WebSocket Transport Support
+;load   = res_pjsip_xpidf_body_generator.so           ; PJSIP Extension State PIDF Provider
+;load   = res_rtp_asterisk.so                         ; Asterisk RTP Stack
+;load   = res_sorcery_astdb.so                        ; Sorcery Astdb Object Wizard
+;load   = res_sorcery_config.so                       ; Sorcery Configuration File Object Wizard
+;load   = res_sorcery_memory.so                       ; Sorcery In-Memory Object Wizard
+;load   = res_sorcery_realtime.so                     ; Sorcery Realtime Object Wizard
+
+;********************************************************************************
```
InterLinked1, line 268: As per the above, I think everything from here on after in the file could just be removed

configs/rpt/modules.conf
Allan-N: I was never planning to leave in this comment block.  It was there to prompt discussion about the changes.

configs/rpt/modules.conf
Allan-N: >Possible dumb question, but why all the noload's in the first place? If autoload is no, then only the modules loaded explicitly will be loaded. The noload's are all redundant and have no effect. Removing them would simplify and clean this up quite a bit, making it easier to look at and work with

The noload's have been included since ASL2 (and likely before).  What I don't know is the original motivation for leaving them in the file.  Thinking out loud, folks (Jim?) could have started with a configuration that enabled most of the modules and this was the path to highlighting what was disabled.  Moving forward, I don't see any harm in retaining the noload's (or changing them to ";load") as a way to show what functionality can be enabled.

configs/rpt/modules.conf
Allan-N: Tim's whitelist/blacklist uses the Asterisk (dialplan functions) database.  Would that be built-in?  part of app_db.so (not loaded)?  part of func_db.so (loaded)?  ...?

Note: I have not changed the load/noload status for any of the DB modules.

configs/rpt/modules.conf
Allan-N: The only format that I added was format_mp3.so

configs/rpt/modules.conf
InterLinked1: Since the goal is to streamline, I feel like it might overwhelm users. People that are using other functionality are not likely to use the RPT modules.conf to begin with anyways, only if they are doing ASL and nothing else. If somebody wants to see what modules there are, they can look at the latest Asterisk documentation anyways, which also has more information.

If left in, then the task is also there then to constantly keep it up to date as modules are added and removed.

The sample config does not enumerate the modules either: https://github.com/asterisk/asterisk/blob/master/configs/samples/modules.conf.sample

configs/rpt/modules.conf
InterLinked1: Yes, that is `func_db`, the module that provides those dialplan functions.

configs/rpt/modules.conf
InterLinked1: I think most of the other formats could be safely removed then. If we are already disabling most of the modules in this config, many of which are useful to many people, there is little reason in enabling obsolete codecs that nobody uses or should use

configs/rpt/modules.conf
```diff
@@ -1,213 +1,443 @@
...
+noload = app_mixmonitor.so             ; Mixed Audio Monitoring Application
+noload = app_morsecode.so              ; Morse code
+noload = app_mp3.so                    ; Silly MP3 Application
+noload = app_page.so                   ; Page Multiple Phones
+load   = app_playback.so               ; Sound File Playback Application
+noload = app_privacy.so                ; Require phone number to be entered, if no CallerID sent
+noload = app_queue.so                  ; True Call Queueing
+noload = app_read.so                   ; Read Variable Application
+noload = app_record.so                 ; Trivial Record Application
+load   = app_rpt.so                    ; Radio Repeater/Remote Base Application
```
InterLinked1, line 77: Another thought. Since this is ASL-centric, why not change `load = app_rpt` to `require = app_rpt`?

This way, Asterisk will fail to start at all if there is a critical issue with `app_rpt`.

Just a suggestion, but it's often common to do with modules that are considered "essential" to the system.

Allan-N: An interesting thought.  The gotcha is once we start down the path of identifying those modules that are "required" do we need to identify/mark others that should also be tagged?  Here, I'm leaning towards sticking with "load".

InterLinked1: It's not like they *have* to be "required", it's just an option to ensure Asterisk doesn't start up without some critical thing loaded. To me, `app_rpt` seems like the only thing that would benefit from this. It's not like `format_wav` would fail to start up for some reason. The channel drivers are all optional, etc.

`res_rpt_http_registrations` might be another one, at least for ASL users.

The benefit is it can catch issues with the system that otherwise wouldn't be noticeable until you've banged your head on the wall for an hour. I've been there a few times :)

configs/rpt/modules.conf
Allan-N: >If left in, then the task is also there then to constantly keep it up to date as modules are added and removed.

That's a fair point.  When pulling together this change I actually wrote a script to identify modules that were no longer available (either because the were removed or not compiled), modules that were compiled but not referenced, and to update the module comments.

configs/rpt/modules.conf
Allan-N: Q? when making or receiving a call do we negotiate based on the available / enabled formats?  Where do we get that list of formats?

InterLinked1: It's as defined in iax.conf + depending on which ones of those are actually loaded/available

configs/rpt/modules.conf
Allan-N: Should I also add `chan_iax2` ?

configs/rpt/modules.conf
Allan-N: In `iax.conf` I see mentions of codecs, not formats

configs/rpt/modules.conf
Allan-N: >Since the goal is to streamline, I feel like it might overwhelm users.

I don't know that the goal is to streamline.  The primary goal was to remove the modules that are gone and to add in those that new/needed.  But, while here I thought it made sense to reconcile the basic-pbx configuration.  And, lastly, many of the modules were missing or had truncated comments.

But, I do agree that there are many module references that most folks won't need.  What I don't know are the modules that the small handful of users might want because they want there configuration to support both ASL and PBX.

configs/rpt/modules.conf
InterLinked1: Yeah, if it's not there already then it would need to be loaded

configs/rpt/modules.conf
InterLinked1: > What I don't know are the modules that the small handful of users might want because they want there configuration to support both ASL and PBX.

It's going to differ for everyone, which is why I think it's a fool errand to try to anticipate that. There are hundreds of modules, thousands of possible use cases and configurations. If that's the goal, then one should just `autoload=yes` and be done. Obviously, that's not what the goal is here so it might be better to focus on that.

If they're also doing PBX stuff, then they're going to have an existing modules.conf they want to amend or they're going to customize the one provided to suit their needs.

configs/rpt/modules.conf
InterLinked1: In some cases I think they go hand in hand, but that's right, they're a bit different.
I believe codecs get used for the call, e.g. codec_ulaw, and formats are used for recording/playback, e.g. format_ulaw/format_pcm.

configs/rpt/modules.conf
Allan-N: This makes sense.  I shall prune out most of the noload's.

KB4MDD (approved):

PR #470, opened 2025-01-27
app_rpt.c: Fix truncated link list in RPT_LINK

### LINKED ISSUE #469: RPT_LINKS incomplete node list
I’m working with a network which has around 90-100 nodes connected at any one time. I want to monitor which nodes are connected and have written a script to subscribe to the `RPT_LINKS` event (Python, via the excellent pystrix package).

If I query the Allstarlink API and count the number of links connected to each of our hub nodes, I get:

```bash
$ cat nodecount.sh 
#!/usr/bin/env bash

NODES=(2167 2196 63061 601040)

t=0
for n in "${NODES[@]}"; do
    c=$(curl -s https://stats.allstarlink.org/api/stats/$n | jq '.stats .data .links | length')
    t=$(( $t + $c ))
done
echo $t
$ ./nodecount.sh 
94
```

However, the payload I receieve in the `RPT_LINKS` event has significantly fewer:

```
RPT_LINKS=69,T2167,T53396,T547462,T620131,T55512,T61706,T577817,T64077,T418191,T1999,T46348,T58753,T60148,T63466,T63783,T61578,T40064,T2196,T494130,T48907,T63429,T62169,T63863,T59629,T623770,T632850,T550113,T61739,T620133,T52375,T547487,T286982,T605872,T547482,T602201,T3889266,T555399,T3197638,T567200,T550700,T601630,T601640,T547488,T602020,T547480,T614820,T286980,T605870,T555390,T606530,T60699,T424380,T635660,T605880,T602200,T547489,T51871,T1999,T578103,T578101,T1995,T42434,T50490,T501642,T50982,T546381,T569940,T521391,T5
```

I counted them, there are indeed 69, which corresponds with `RPT_NUMLINKS`. Also node the truncated `T5` at the end. To validate that it isn’t my script, I see the same list when I run `rpt showvars NNNNNN`. The hub nodes are all connected according to our dashboards, at most 2 hops away from each other.

The topology is `2167 - 2196 - 63061`

In this case I am connected direct to 2167, but cannot see that 63061 or any of its adjacent nodes. If I connect direct to 63061 I will not see 2167 or any of its adjacent nodes. They are both connected to 2196. Both can see 2196 an its adjacent nodes in the list.

`RPT_LINKS` seems to be emitted when there is activity, the network itself is quite busy with the usual flurry of kerchunks between QSOs, I have also tried poking the hub nodes to solicit `RPT_LINKS` events between nodes, but the most I can see is around 70.

Is there any limit to the number of nodes that will be emitted with `RPT_LINKS`? Is there some configuration that would prevent an intermediate node from sending node details from one side to another (looking at 2196)?

---
### Version information

I've observed this behavior on an ASL3 node:

```
OS            : Debian GNU/Linux 12 (bookworm)
OS Kernel     : 6.1.0-28-amd64

Asterisk      : 20.10.0+asl3-3.1.0-1.deb12
ASL [app_rpt] : 3.1.0
```

<details>
<summary>Full asl-show-version 3.1.0 output</summary>

```
    $ sudo asl-show-version 
    dpkg-query: no packages found matching cockpit*
    ********** AllStarLink [ASL] Version Info **********

    OS            : Debian GNU/Linux 12 (bookworm)
    OS Kernel     : 6.1.0-28-amd64

    Asterisk      : 20.10.0+asl3-3.1.0-1.deb12
    ASL [app_rpt] : 3.1.0

    Installed ASL packages :

      Package                         Version
      ==============================  ==============================
      allmon3                         1.4.1-1.deb12
      asl3                            3.6.0-1.deb
      asl3-asterisk                   2:20.10.0+asl3-3.1.0-1.deb12
      asl3-asterisk-config            2:20.10.0+asl3-3.1.0-1.deb12
      asl3-asterisk-modules           2:20.10.0+asl3-3.1.0-1.deb12
      asl3-menu                       1.10-1.deb12
      dahdi                           1:3.1.0-2
      dahdi-dkms                      1:3.4.0-4+asl
      dahdi-linux                     1:3.4.0-4+asl
```
</details>

Also seeing it after upgrading to 3.2.0:

```
OS            : Debian GNU/Linux 12 (bookworm)
OS Kernel     : 6.1.0-28-amd64

Asterisk      : 20.11.0+asl3-3.2.0-2.deb12
ASL [app_rpt] : 3.2.0
```

<details>
<summary>Full asl-show-version 3.2.0 output</summary>

```
    $ sudo asl-show-version
    dpkg-query: no packages found matching cockpit*
    ********** AllStarLink [ASL] Version Info **********
    
    OS            : Debian GNU/Linux 12 (bookworm)
    OS Kernel     : 6.1.0-28-amd64
    
    Asterisk      : 20.11.0+asl3-3.2.0-2.deb12
    ASL [app_rpt] : 3.2.0
    
    Installed ASL packages :
    
      Package                         Version
      ==============================  ==============================
      allmon3                         1.4.2-1.deb12
      asl3                            3.6.0-1.deb
      asl3-asterisk                   2:20.11.0+asl3-3.2.0-2.deb12
      asl3-asterisk-config            2:20.11.0+asl3-3.2.0-2.deb12
      asl3-asterisk-modules           2:20.11.0+asl3-3.2.0-2.deb12
      asl3-menu                       1.11-1.deb12
      dahdi                           1:3.1.0-2
      dahdi-dkms                      1:3.4.0-4+asl
      dahdi-linux                     1:3.4.0-4+asl
```
</details>

Also seeing the same thing on an older node running `ASL Version 2.0.0-beta.6`.

mkmer:
Fixes #469 - processing received link list text message was limited to 512 characters.  Expanded to MAXLINKLIST.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1660,8 +1660,13 @@ static inline void init_text_frame(struct ast_frame *wf)
 }
 
 static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *str)
-{
-	char tmp[512], tmp1[512], cmd[300] = "", dest[300], src[30], c;
+{    
+	/* These temporary var sizes are a bit tricky:
+	 * tmp is a copy of a string that can be MAXLINKLIST (or larger)
+	 * when we recieve an 'L' it copies tmp into l->linklist which is
+	 * MAXLINKLIST long. 
+	 */
+	char tmp[MAXLINKLIST], tmp1[512], cmd[300] = "", dest[300], src[30], c;
 	int i, seq, res, ts, rest;
 	struct ast_frame wf;
 
@@ -1669,7 +1674,10 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	wf.datalen = strlen(str) + 1;
 	wf.src = "handle_link_data";
 	/* put string in our buffer */
-	ast_copy_string(tmp, str, sizeof(tmp) - 1);
+	if (strlen(str) > sizeof(tmp)) {
+		ast_log(LOG_WARNING, "Link text message data has exceeded tmp buffer size");
+	}
+	ast_copy_string(tmp, str, sizeof(tmp));
 
 	ast_debug(5, "Received text over link: '%s'\n", str);
 
@@ -1703,7 +1711,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 	if (tmp[0] == 'L') {
 		rpt_mutex_lock(&myrpt->lock);
-		strcpy(mylink->linklist, tmp + 2);
+		strcpy(mylink->linklist, tmp + 2); /* MAXLINKLIST - 2 is placed in mylink->linklist*/
 		time(&mylink->linklistreceived);
 		rpt_mutex_unlock(&myrpt->lock);
 		ast_debug(7, "@@@@ node %s recieved node list %s from node %s\n", myrpt->name, tmp, mylink->name);
@@ -1711,7 +1719,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 	if (tmp[0] == 'M') {
 		rest = 0;
-		if (sscanf(tmp, "%s %s %s %n", cmd, src, dest, &rest) < 3) {
+		if (sscanf(tmp, "%s %s %s %n", cmd, src, dest, &rest) < 3) { /*TODO: We should limit to sizeof(cmd, src, dest)*/
 			ast_log(LOG_WARNING, "Unable to parse message string %s\n", str);
 			return;
 		}
@@ -1739,7 +1747,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		return;
 	}
 	if (tmp[0] == 'T') {
-		if (sscanf(tmp, "%s %s %s", cmd, src, dest) != 3) {
+		if (sscanf(tmp, "%s %s %s", cmd, src, dest) != 3) { /*TODO: We should limit to sizeof(cmd, src, dest)*/
 			ast_log(LOG_WARNING, "Unable to parse telem string %s\n", str);
 			return;
 		}
@@ -1771,7 +1779,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 
 	if (tmp[0] == 'C') {
-		if (sscanf(tmp, "%s %s %s %s", cmd, src, tmp1, dest) != 4) {
+		if (sscanf(tmp, "%s %s %s %s", cmd, src, tmp1, dest) != 4) { /*TODO: We should limit to sizeof(cmd, src, dest, tmp1)*/
 			ast_log(LOG_WARNING, "Unable to parse ctcss string %s\n", str);
 			return;
 		}
@@ -1790,7 +1798,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 
 	if (tmp[0] == 'K') {
-		if (sscanf(tmp, "%s %s %s %d %d", cmd, dest, src, &seq, &ts) != 5) {
+		if (sscanf(tmp, "%s %s %s %d %d", cmd, dest, src, &seq, &ts) != 5) { /*TODO: We should limit to sizeof(cmd, src, dest)*/
 			ast_log(LOG_WARNING, "Unable to parse keying string %s\n", str);
 			return;
 		}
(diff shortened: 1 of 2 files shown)


Allan-N (approved):

Allan-N (approved):

apps/app_rpt.c
```diff
@@ -1660,15 +1660,23 @@ static inline void init_text_frame(struct ast_frame *wf)
 }
 
 static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *str)
-{
-	char tmp[512], tmp1[512], cmd[300] = "", dest[300], src[30], c;
+{    
+	/* These temproary var sizes are a bit tricky:
```
InterLinked1, line 1664: temporary is mispelled

apps/app_rpt.c
```diff
@@ -1660,15 +1660,23 @@ static inline void init_text_frame(struct ast_frame *wf)
...
 	struct ast_frame wf;
 
 	init_text_frame(&wf);
 	wf.datalen = strlen(str) + 1;
 	wf.src = "handle_link_data";
 	/* put string in our buffer */
+	if (strlen(str) > sizeof(tmp)){
+		ast_log(LOG_WARNING,"Link text message data has exceeded tmp buffer size");
+	}
 	ast_copy_string(tmp, str, sizeof(tmp) - 1);
```
InterLinked1, line 1680: While you're adding that check, can you remove the -1 from the following line?

apps/app_rpt/rpt_link.c
```diff
@@ -485,6 +485,13 @@ void __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, int fla
 		if (flag) {
 			snprintf(buf + spos, MAXLINKLIST - spos, "%s%c%c", l->name, mode, (l->lastrx1) ? 'K' : 'U');
 		} else {
+			
+			/* l->linklist and buff have a size of MAXLINKLIST.  We will have a problem with large networks where this exceeds MAXLINKLIST
+			 * worst case(ish) The message looks like this "<network size>, 6 digit node + T + ,"  
+			 * With MAXLINKLIST = 5120, 5120 - 4 / 8 = 638 nodes.  We can have smaller node values but generally we 
+			 * will have problems on networks > 638 nodes as is.
+			*/
```
InterLinked1, line 493: indent one more space

apps/app_rpt.c
```diff
@@ -1703,15 +1711,15 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
...
+		strcpy(mylink->linklist, tmp + 2); /* MAXLINKLIST - 2 is placed in mylink->linklist*/
 		time(&mylink->linklistreceived);
 		rpt_mutex_unlock(&myrpt->lock);
 		ast_debug(7, "@@@@ node %s recieved node list %s from node %s\n", myrpt->name, tmp, mylink->name);
 		return;
 	}
 	if (tmp[0] == 'M') {
 		rest = 0;
-		if (sscanf(tmp, "%s %s %s %n", cmd, src, dest, &rest) < 3) {
+		if (sscanf(tmp, "%s %s %s %n", cmd, src, dest, &rest) < 3) { /*TODO: We should limit to sizeof(cmd, src, dest)*/
```
InterLinked1, line 1722: It seems like tmp is mutable since we already have a copy of it (I haven't fully verified though)
Instead of the existing code, maybe we should replace with strsep for the first 3 strings and just use sscanf for the numbers? That way we can fix the issue more robustly as opposed to making a TODO?

apps/app_rpt.c
```diff
@@ -1660,15 +1660,23 @@ static inline void init_text_frame(struct ast_frame *wf)
...
+	char tmp[MAXLINKLIST], tmp1[512], cmd[300] = "", dest[300], src[30], c;
 	int i, seq, res, ts, rest;
 	struct ast_frame wf;
 
 	init_text_frame(&wf);
 	wf.datalen = strlen(str) + 1;
 	wf.src = "handle_link_data";
 	/* put string in our buffer */
+	if (strlen(str) > sizeof(tmp)){
+		ast_log(LOG_WARNING,"Link text message data has exceeded tmp buffer size");
```
InterLinked1, line 1678: space after LOG_WARNING, and before the opening brace

And then what, should we abort and not process at all or continue anyways as we are doing now?

InterLinked1 (changes requested):

apps/app_rpt.c
mkmer: Just to confirm my understanding: `ast_copy_string` copies up to size, then if the lest element is not '\0' it overwrites the last element with '\0'.

Allan-N: https://github.com/asterisk/asterisk/blob/32e6a309525cfd28133124cbe9e8c97fe9c90c52/include/asterisk/strings.h#L398-L425

mkmer:
I general, this "fixes" the issue as seen.  The questions that remain for a future PR (in my mind anyway):

- What is the max string we can receive from a text message - the initial call is a malloc, copy to pass it str in.  
- What is the max string a channel variable can accept? (Is it 5120 and that's why this value was chosen?)
- Do we need to copy it in the routine, as it's malloc, copied,  and freed around the call to either routine https://github.com/AllStarLink/app_rpt/blob/e957e50a9241e7fb358f93050b1c0d37b39b2a55/apps/app_rpt.c#L4407-L4413
- Should `l->linklist` be a fixed value or be sized to the data (this may not be nice as we would malloc/free for every message received to resize it...)
- What is the "right" size for `buf`, or should it be malloc based on the size of the links when we build it (also depending on max channel variable size)?  It could be VERY undersized if every `l->linklist` is `MAXLINKLIST` or even half. linkhttps://github.com/AllStarLink/app_rpt/blob/e957e50a9241e7fb358f93050b1c0d37b39b2a55/apps/app_rpt/rpt_link.c#L526

mkmer:
Closing this in favor of PR #474

PR #471, opened 2025-01-28
app_rpt.c: change timers of unnecessary long to int

mkmer:
Not sure how I missed the compiler warnings when I tested https://github.com/AllStarLink/app_rpt/pull/457
The changed timers are unnecessarily `long`.  We could alternately go with `long` for all timers if there is concern about expanding beyond 32767ms for any timer.  Currently all of them appear to fit in an `int`.


Allan-N:
>We could alternately go with long for all timers if there is concern about expanding beyond 32767ms for any timer.

I think your math may be off.  A [32-bit] "int" has a max value of 0x7FFFFFFF (2147483647).  If that number is in "ms" then we would have 2147483647/1000 = 2147483 seconds (24+ days).  As long as we're not computing how much time has elapsed at that scale we should be good.

Allan-N (approved):

PR #472, opened 2025-01-28
rpt_uchameleon.c: Incorrect use of sizeof()

mkmer:
This clearly should be strlen of the string.  
```
sizeof(&expect) = 8
strlen(expect) = 9
```
I suppose it "works" because the count is always 13? or the first 8 characters always match.

### DIFF
--- a/apps/app_rpt/rpt_uchameleon.c
+++ b/apps/app_rpt/rpt_uchameleon.c
@@ -51,23 +51,23 @@ int uchameleon_thread_start(struct daq_entry_tag *t)
 int uchameleon_connect(struct daq_entry_tag *t)
 {
 	int count;
-	static char *idbuf = "id\n";
-	static char *ledbuf = "led on\n";
-	static char *expect = "Chameleon";
+	static const char idbuf[] = "id\n";
+	static const char ledbuf[] = "led on\n";
+	static const char expect[] = "Chameleon";
 	char rxbuf[20];
 
 	if ((t->fd = serial_open(t->dev, B115200, 0)) == -1) {
 		ast_log(LOG_WARNING, "serial_open on %s failed!\n", t->name);
 		return -1;
 	}
-	if ((count = serial_io(t->fd, idbuf, rxbuf, strlen(idbuf), 14, DAQ_RX_TIMEOUT, 0x0a)) < 1) {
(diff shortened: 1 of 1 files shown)


apps/app_rpt/rpt_uchameleon.c
```diff
@@ -67,7 +67,7 @@ int uchameleon_connect(struct daq_entry_tag *t)
 		return -1;
 	}
 	ast_debug(3, "count = %d, rxbuf = %s\n", count, rxbuf);
-	if ((count != 13) || (strncmp(expect, rxbuf + 4, sizeof(&expect)))) {
+	if ((count != 13) || (strncmp(expect, rxbuf + 4, strlen(expect)))) {
```
Allan-N, line 70: Ick.  Why use up CPU to compute a value that's easily known by the compiler.
```
static char expect[] = "Chameleon";  // a slightly different way to represent the string

printf("sizeof(&expect)   = %ld\n", sizeof(&expect));    // returns the size of the pointer
printf("strlen(expect)    = %ld\n", strlen(expect));     // requires CPU
printf("sizeof(expect) -1 = %ld\n", sizeof(expect) - 1); // easy, peasy
```

Allan-N (changes requested):

apps/app_rpt/rpt_uchameleon.c
mkmer: Makes sense - so I suppose I'll do all of the static strings this way....

Allan-N (approved):

apps/app_rpt/rpt_uchameleon.c
```diff
@@ -51,31 +51,31 @@ int uchameleon_thread_start(struct daq_entry_tag *t)
 int uchameleon_connect(struct daq_entry_tag *t)
 {
 	int count;
-	static char *idbuf = "id\n";
-	static char *ledbuf = "led on\n";
-	static char *expect = "Chameleon";
+	static char idbuf[] = "id\n";
+	static char ledbuf[] = "led on\n";
+	static char expect[] = "Chameleon";
```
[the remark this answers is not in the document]
mkmer: I don't see why not...

KB4MDD (approved):

InterLinked1 (approved):

PR #473, opened 2025-01-28
rpt_mdc1200.c: Address issue app_rpt: Incorrect usage of sizeof() function #207

### LINKED ISSUE #207: app_rpt: Incorrect usage of sizeof() function
The following functions appear to use the sizeof() function incorrectly.

apps/app_rpt/rpt_mdc1200.c
In mdc_1200_ack_status, line 209: `memset(mdcp, 0, sizeof(&mdcp));`

This should be sizeof(mdcparams)

apps/app_rpt/rpt_uchameleon.c
In uchameleon_connect, line 73: `if ((count != 13) || (strncmp(expect, rxbuf + 4, sizeof(&expect)))) {`

expect may need to be changed to a character array or change sizeof(&expect), or the literal 9, or strlen(expect).

Note:  I am not sure why idbuf, ledbuf, and expect are declared as static.  They probably should be char arrays and const.

mkmer:
Addressing identified solution in issue fixes #207


Allan-N (approved):

KB4MDD (approved):

apps/app_rpt/rpt_mdc1200.c
```diff
@@ -206,7 +206,7 @@ void mdc1200_ack_status(struct rpt *myrpt, short UnitID)
 		ast_log(LOG_ERROR, "Cannot alloc!!\n");
 		return;
 	}
-	memset(mdcp, 0, sizeof(&mdcp));
+	memset(mdcp, 0, sizeof(*mdcp));
```
InterLinked1, line 209: This line should just be removed, we are calling calloc immediately prior to it, which zeros out memory
Additionally the error about can't alloc can be removed

InterLinked1 (changes requested):

apps/app_rpt/rpt_mdc1200.c
Allan-N: You want the error about not being able to allocate memory removed?  or just updated to reflect that it was ast_calloc() that failed vs "alloc"?

InterLinked1: Two separate things:
- memset is unnecessary. No need to fix it, the line can be removed
- The Asterisk memory manager logs errors for allocation failures. Adding our own are extraneous. That line can also be removed.

mkmer: Probably explains why no one noticed the memset not doing enough.

InterLinked1 (commented):
Looks good, could you also update your commit message formatting, e.g. to close the linked issue? Right now that's only in the PR description

Allan-N (approved):

PR #474, opened 2025-01-29
app_rpt.c: Refactor handle_link_data() and handle_remote_data()

mkmer:
This is a follow on to PR https://github.com/AllStarLink/app_rpt/pull/470 which should not be delayed.  Intent is to clean up the stuff I notice while hunting down the truncated message.

Eliminate copy of malloc `char*` to fixed length `char*`
update `sprintf()` to `snprintf()`
rpt_link.c: Calculate buffer size and malloc buffers. add function __get_buffer_size() to calculate buffer size required. add node count to __mklinklist()
add macro for OBUFSIZE and BUFSIZE calculation
eliminate `finddelim()` call as `__mklinklist()` now counts links.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1093,7 +1093,7 @@ static void startoutstream(struct rpt *myrpt)
 	if (!str) {
 		return;
 	}
-	n = finddelim(str, strs, 100);
+	n = finddelim(str, strs, ARRAY_LEN(strs));
 	if (n < 1) {
 		ast_log(LOG_ERROR, "Could not parse string '%s'\n", myrpt->p.outstreamcmd);
 		ast_free(str);
@@ -1581,7 +1581,7 @@ static inline void collect_function_digits_post(struct rpt *myrpt, int res, cons
  * \brief Send APRStt (Touchtone) to app_gps for processing.
  * This routine takes the received APRStt touchtone digits
  * and translates them to a callsign.  The results are
- * sent to app_gps using the APRS_SENDTT function for 
+ * sent to app_gps using the APRS_SENDTT function for
  * processing and posting to the APRS-IS server.
  *
  * \param myrpt		pointer to repeater struct.
@@ -1592,13 +1592,13 @@ static void do_aprstt(struct rpt *myrpt)
 	char overlay, aprscall[100], func[100];
 
 	snprintf(cmd, sizeof(cmd) - 1, "A%s", myrpt->dtmfbuf);
-	/*! \todo we need to support all 4 types of APRStt 
+	/*! \todo we need to support all 4 types of APRStt
 	 * we only support the 'A' type for call sign
 	 */
 	overlay = aprstt_xlat(cmd, aprscall);
 	if (overlay) {
 		ast_debug(1, "APRStt got string %s callsign %s overlay %c\n", cmd, aprscall, overlay);
-		
+
 		if (!ast_custom_function_find("APRS_SENDTT")) {
 			ast_log(LOG_WARNING, "app_gps is not loaded.  APRStt failed\n");
 		} else {
@@ -1607,7 +1607,7 @@ static void do_aprstt(struct rpt *myrpt)
 			if (!ast_func_write(NULL, func, aprscall)) {
 				rpt_telemetry(myrpt, ARB_ALPHA, (void *) aprscall);
 			}
-		} 
+		}
 	}
 }
 
@@ -1702,25 +1702,26 @@ static inline void init_text_frame(struct ast_frame *wf)
 
 static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *str)
 {
-	char tmp[512], tmp1[512], cmd[300] = "", dest[300], src[30], c;
+	/* XXX cmd, dst, and src should be validated. Why is remote_data src[300] in other locations?
+	* Is this a typo here?  Why would dest be any bigger than src?
+	*/
+	char tmp1[512], cmd[300] = "", dest[300], src[30], c;
 	int i, seq, res, ts, rest;
 	struct ast_frame wf;
 
 	init_text_frame(&wf);
 	wf.datalen = strlen(str) + 1;
 	wf.src = "handle_link_data";
-	/* put string in our buffer */
-	ast_copy_string(tmp, str, sizeof(tmp) - 1);
 
 	ast_debug(5, "Received text over link: '%s'\n", str);
 
-	if (!strcmp(tmp, DISCSTR)) {
+	if (!strcmp(str, DISCSTR)) {
 		mylink->disced = 1;
 		mylink->retries = mylink->max_retries + 1;
 		ast_softhangup(mylink->chan, AST_SOFTHANGUP_DEV);
 		return;
 	}
-	if (!strcmp(tmp, NEWKEYSTR)) {
+	if (!strcmp(str, NEWKEYSTR)) {
 		if ((!mylink->newkey) || mylink->newkeytimer) {
 			mylink->newkeytimer = 0;
 			mylink->newkey = 1;
@@ -1728,37 +1729,40 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		}
 		return;
 	}
-	if (!strcmp(tmp, NEWKEY1STR)) {
+	if (!strcmp(str, NEWKEY1STR)) {
 		mylink->newkeytimer = 0;
 		mylink->newkey = 2;
 		return;
 	}
-	if (!strncmp(tmp, IAXKEYSTR, strlen(IAXKEYSTR))) {
+	if (!strncmp(str, IAXKEYSTR, strlen(IAXKEYSTR))) {
 		mylink->iaxkey = 1;
 		return;
 	}
-	if (tmp[0] == 'G') {		/* got GPS data */
+	if (*str == 'G') {		/* got GPS data */
 		/* re-distribute it to attached nodes */
 		distribute_to_all_links(myrpt, mylink, src, NULL, str, &wf);
 		return;
 	}
-	if (tmp[0] == 'L') {
+	if (*str == 'L') {
 		rpt_mutex_lock(&myrpt->lock);
-		strcpy(mylink->linklist, tmp + 2);
+		if (strlen(str + 2) > sizeof(mylink->linklist) - 1) {
+			ast_log(LOG_WARNING, "Link list too long: buffer size: %ld, link size: %ld linklist: %s\n", sizeof(mylink->linklist) - 1, strlen(str + 2), str + 2);
+		}
+		ast_copy_string(mylink->linklist, str + 2, sizeof(mylink->linklist));
 		time(&mylink->linklistreceived);
 		rpt_mutex_unlock(&myrpt->lock);
-		ast_debug(7, "@@@@ node %s received node list %s from node %s\n", myrpt->name, tmp, mylink->name);
+		ast_debug(7, "@@@@ node %s received node list %s from node %s\n", myrpt->name, str, mylink->name);
 		return;
 	}
-	if (tmp[0] == 'M') {
+	if (*str == 'M') {
 		rest = 0;
-		if (sscanf(tmp, "%s %s %s %n", cmd, src, dest, &rest) < 3) {
+		if (sscanf(str, "%s %s %s %n", cmd, src, dest, &rest) < 3) { /*! \todo We should limit to sizeof(cmd, src, dest) */
 			ast_log(LOG_WARNING, "Unable to parse message string %s\n", str);
 			return;
 		}
 		if (!rest)
 			return;
-		if (strlen(tmp + rest) < 2)
+		if (strlen(str + rest) < 2)
 			return;
 		/* if is from me, ignore */
 		if (!strcmp(src, myrpt->name))
@@ -1766,21 +1770,21 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		/* if is for one of my nodes, dont do too much! */
 		for (i = 0; i < nrpts; i++) {
 			if (!strcmp(dest, rpt_vars[i].name)) {
-				ast_verb(3, "Private Text Message for %s From %s: %s\n", rpt_vars[i].name, src, tmp + rest);
-				ast_debug(1, "Node %s Got Private Text Message From Node %s: %s\n", rpt_vars[i].name, src, tmp + rest);
+				ast_verb(3, "Private Text Message for %s From %s: %s\n", rpt_vars[i].name, src, str + rest);
+				ast_debug(1, "Node %s Got Private Text Message From Node %s: %s\n", rpt_vars[i].name, src, str + rest);
 				return;
 			}
 		}
 		/* if is for everyone, at least log it */
 		if (!strcmp(dest, "0")) {
-			ast_verb(3, "Text Message From %s: %s\n", src, tmp + rest);
-			ast_debug(1, "Node %s Got Text Message From Node %s: %s\n", myrpt->name, src, tmp + rest);
+			ast_verb(3, "Text Message From %s: %s\n", src, str + rest);
+			ast_debug(1, "Node %s Got Text Message From Node %s: %s\n", myrpt->name, src, str + rest);
 		}
 		distribute_to_all_links(myrpt, mylink, src, NULL, str, &wf);
 		return;
 	}
-	if (tmp[0] == 'T') {
-		if (sscanf(tmp, "%s %s %s", cmd, src, dest) != 3) {
+	if (*str == 'T') {
+		if (sscanf(str, "%s %s %s", cmd, src, dest) != 3) { /*! \todo We should limit to sizeof(cmd, src, dest) */
 			ast_log(LOG_WARNING, "Unable to parse telem string %s\n", str);
 			return;
 		}
@@ -1811,8 +1815,8 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		return;
 	}
 
-	if (tmp[0] == 'C') {
-		if (sscanf(tmp, "%s %s %s %s", cmd, src, tmp1, dest) != 4) {
+	if (*str == 'C') {
+		if (sscanf(str, "%s %s %s %s", cmd, src, tmp1, dest) != 4) { /*! \todo We should limit to sizeof(cmd, src, tmp1, dest) */
 			ast_log(LOG_WARNING, "Unable to parse ctcss string %s\n", str);
 			return;
 		}
@@ -1830,8 +1834,8 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		return;
 	}
 
-	if (tmp[0] == 'K') {
-		if (sscanf(tmp, "%s %s %s %d %d", cmd, dest, src, &seq, &ts) != 5) {
+	if (*str == 'K') {
+		if (sscanf(str, "%s %s %s %d %d", cmd, dest, src, &seq, &ts) != 5) { /*! \todo We should limit to sizeof(cmd, src, dest) */
 			ast_log(LOG_WARNING, "Unable to parse keying string %s\n", str);
 			return;
 		}
@@ -1859,7 +1863,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 			if (myrpt->lastkeyedtime) {
 				n = (int) (now - myrpt->lastkeyedtime);
 			}
-			sprintf(tmp1, "K %s %s %d %d", src, myrpt->name, myrpt->keyed, n);
+			snprintf(tmp1, sizeof(tmp1), "K %s %s %d %d", src, myrpt->name, myrpt->keyed, n);
 			wf.data.ptr = tmp1;
 			wf.datalen = strlen(tmp1) + 1;
 			if (mylink->chan)
@@ -1887,15 +1891,15 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		rpt_mutex_unlock(&myrpt->lock);
 		return;
 	}
-	if (tmp[0] == 'I') {
-		if (sscanf(tmp, "%s %s %s", cmd, src, dest) != 3) {
+	if (*str == 'I') {
+		if (sscanf(str, "%s %s %s", cmd, src, dest) != 3) { /*! \todo We should limit to sizeof(cmd, src, dest) */
 			ast_log(LOG_WARNING, "Unable to parse ident string %s\n", str);
 			return;
 		}
 		mdc1200_notify(myrpt, src, dest);
 		strcpy(dest, "*");
 	} else {
-		if (sscanf(tmp, "%s %s %s %d %c", cmd, dest, src, &seq, &c) != 5) {
+		if (sscanf(str, "%s %s %s %d %c", cmd, dest, src, &seq, &c) != 5) {/*! \todo We should limit to sizeof(cmd, src, dest) */
 			ast_log(LOG_WARNING, "Unable to parse link string %s\n", str);
 			return;
 		}
@@ -2178,37 +2182,37 @@ static int handle_remote_dtmf_digit(struct rpt *myrpt, char c, char *keyed, int 
 	return res;
 }
 
-static int handle_remote_data(struct rpt *myrpt, char *str)
+static int handle_remote_data(struct rpt *myrpt, const char *str)
 {
-	char tmp[300], cmd[300], dest[300], src[300], c;
+	/* Should src[300] be src[30] as in handle_link_data?*/
+	char cmd[300], dest[300], src[300], c;
 	int seq, res;
 
 	/* put string in our buffer */
-	ast_copy_string(tmp, str, sizeof(tmp));
-	if (!strcmp(tmp, DISCSTR))
+	if (!strcmp(str, DISCSTR))
 		return 0;
-	if (!strcmp(tmp, NEWKEYSTR)) {
+	if (!strcmp(str, NEWKEYSTR)) {
 		if (!myrpt->newkey) {
 			send_old_newkey(myrpt->rxchannel);
 			myrpt->newkey = 1;
 		}
 		return 0;
 	}
-	if (!strcmp(tmp, NEWKEY1STR)) {
+	if (!strcmp(str, NEWKEY1STR)) {
 		myrpt->newkey = 2;
 		return 0;
 	}
-	if (!strncmp(tmp, IAXKEYSTR, strlen(IAXKEYSTR))) {
+	if (!strncmp(str, IAXKEYSTR, strlen(IAXKEYSTR))) {
 		myrpt->iaxkey = 1;
 		return 0;
 	}
 
-	if (tmp[0] == 'T')
+	if (*str == 'T')
 		return 0;
 
 #ifndef	DO_NOT_NOTIFY_MDC1200_ON_REMOTE_BASES
-	if (tmp[0] == 'I') {
-		if (sscanf(tmp, "%s %s %s", cmd, src, dest) != 3) {
+	if (*str == 'I') {
+		if (sscanf(str, "%s %s %s", cmd, src, dest) != 3) {
 			ast_log(LOG_WARNING, "Unable to parse ident string %s\n", str);
 			return 0;
 		}
@@ -2216,9 +2220,10 @@ static int handle_remote_data(struct rpt *myrpt, char *str)
 		return 0;
 	}
 #endif
-	if (tmp[0] == 'L')
+	if (*str == 'L') {
 		return 0;
-	if (sscanf(tmp, "%s %s %s %d %c", cmd, dest, src, &seq, &c) != 5) {
+	}
+	if (sscanf(str, "%s %s %s %d %c", cmd, dest, src, &seq, &c) != 5) {
 		ast_log(LOG_WARNING, "Unable to parse link string %s\n", str);
 		return 0;
 	}
@@ -2234,7 +2239,7 @@ static int handle_remote_data(struct rpt *myrpt, char *str)
 	}
 	c = func_xlat(myrpt, c, &myrpt->p.outxlat);
 	if (!c)
-		return (0);
+		return 0;
 	res = handle_remote_dtmf_digit(myrpt, c, NULL, 0);
 	if (res != 1)
 		return res;
@@ -3200,7 +3205,6 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 		if ((!l->linklisttimer) && (l->name[0] != '0') && (!l->isremote)) {
 			struct ast_frame lf;
 			char lstr[MAXLINKLIST];
-
 			memset(&lf, 0, sizeof(lf));
 			lf.frametype = AST_FRAME_TEXT;
 			lf.subclass.format = ast_format_slin;
@@ -3209,7 +3213,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			lf.samples = 0;
 			l->linklisttimer = LINKLISTTIME;
 			strcpy(lstr, "L ");
-			__mklinklist(myrpt, l, lstr + 2, 0);
+			__mklinklist(myrpt, l, lstr + 2, sizeof(lstr) - 2, 0);
 			if (l->chan) {
 				lf.datalen = strlen(lstr) + 1;
 				lf.data.ptr = lstr;
@@ -4863,8 +4867,8 @@ static void *rpt(void *this)
 			myrpt->rpt_thread = AST_PTHREADT_STOP;
 			pthread_exit(NULL);
 		}
-		/* \todo At this point, we have a memory leak, because dsp needs to be freed. */
-		/* \todo Find out what the right place is to free dsp, i.e. when myrpt itself goes away. */
+		/*! \todo At this point, we have a memory leak, because dsp needs to be freed. */
+		/*! \todo Find out what the right place is to free dsp, i.e. when myrpt itself goes away. */
 		ast_dsp_set_features(myrpt->dsp, DSP_FEATURE_FREQ_DETECT);
 		ast_dsp_set_freqmode(myrpt->dsp, myrpt->p.rxburstfreq, myrpt->p.rxbursttime, myrpt->p.rxburstthreshold, 0);
 #else
@@ -4948,11 +4952,11 @@ static void *rpt(void *this)
 			char lat[25], lon[25], elev[25];
 
 			myrpt->lastgpstime = t_mono;
-			
+
 			/* If the app_gps custom function GPS_READ exists, read the GPS position */
 			if (!ast_custom_function_find("GPS_READ")) {
 				break;
-			}				
+			}
 			if (ast_func_read(NULL, "GPS_READ()", gps_data, sizeof(gps_data))) {
 				break;
 			}
@@ -4966,7 +4970,7 @@ static void *rpt(void *this)
 				break;
 			}
 			sprintf(tmpstr, "G %s %s %s %s", myrpt->name, lat, lon, elev);
-			
+
 			rpt_mutex_lock(&myrpt->lock);
 			l = myrpt->links.next;
 			myrpt->voteremrx = 0;	/* no voter remotes keyed */
--- a/apps/app_rpt/rpt_cli.c
+++ b/apps/app_rpt/rpt_cli.c
@@ -525,10 +525,11 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 				tel_mode = "3";
 			}
 
-//### GET CONNECTED NODE INFO ####################
-			// Traverse the list of connected nodes 
+			/* ### GET CONNECTED NODE INFO ####################
+			 * Traverse the list of connected nodes
+			 */
 
-			__mklinklist(myrpt, NULL, lbuf, 0);
+			__mklinklist(myrpt, NULL, lbuf, sizeof(lbuf), 0);
 
 			j = 0;
 			l = myrpt->links.next;
@@ -590,7 +591,7 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 
 //### GET ALL LINKED NODES INFO ####################
 			/* parse em */
-			ns = finddelim(lbuf, strs, MAXLINKLIST);
+			ns = finddelim(lbuf, strs, ARRAY_LEN(strs));
 			/* sort em */
 			if (ns)
 				qsort((void *) strs, ns, sizeof(char *), mycompar);
@@ -657,10 +658,10 @@ static int rpt_do_nodes(int fd, int argc, const char *const *argv)
 			/* Make a copy of all stat variables while locked */
 			myrpt = &rpt_vars[i];
 			rpt_mutex_lock(&myrpt->lock);	/* LOCK */
-			__mklinklist(myrpt, NULL, lbuf, 0);
+			__mklinklist(myrpt, NULL, lbuf, sizeof(lbuf), 0);
 			rpt_mutex_unlock(&myrpt->lock);	/* UNLOCK */
 			/* parse em */
-			ns = finddelim(lbuf, strs, MAXLINKLIST);
+			ns = finddelim(lbuf, strs, ARRAY_LEN(strs));
 			/* sort em */
 			if (ns)
 				qsort((void *) strs, ns, sizeof(char *), mycompar);
--- a/apps/app_rpt/rpt_config.c
+++ b/apps/app_rpt/rpt_config.c
@@ -14,6 +14,7 @@
 #include "asterisk/dns_srv.h"	/* use for srv dns lookup */
 #include "asterisk/dns_txt.h" /* user for dns lookup */
 #include "asterisk/vector.h" /* required for dns */
+#include "asterisk/utils.h" /* required for ARRAY_LEN */
 
 #include "app_rpt.h"
 #include <arpa/nameser.h> /* needed for dns - must be after app_rpt.h */
@@ -23,7 +24,7 @@
 #include "rpt_utils.h" /* use myatoi */
 #include "rpt_rig.h" /* use setrem */
 
-/*! \brief Echolink queryoption for retrieving call sign */ 
+/*! \brief Echolink queryoption for retrieving call sign */
 #define ECHOLINK_QUERY_CALLSIGN 2
 
 /*! \brief The Link Box queryoptions */
@@ -292,7 +293,7 @@ int elink_query_callsign(char *node, char *callsign, int callsignlen)
 		ast_log(LOG_WARNING, "chan_echolink not loaded.  Cannot query callsign.\n");
 		return res;
 	}
-	
+
 	/* data is passed to and from the query option using the callsign field */
 	ast_copy_string(callsign, node, callsignlen);
 
@@ -312,11 +313,11 @@ int tlb_query_node_exists(const char *node)
 		ast_debug(5, "chan_tlb not loaded.\n");
 		return res;
 	}
-	
+
 	if (!chan_tech->queryoption(NULL, TLB_QUERY_NODE_EXISTS, (void *) node, 0)) {
 		res = 1;
 	}
-	
+
 	return res;
 }
 
@@ -331,7 +332,7 @@ int tlb_query_callsign(const char *node, char *callsign, int callsignlen)
 		ast_debug(5, "chan_tlb not loaded. Cannot query callsign.\n");
 		return res;
 	}
-	
+
 	/* data is passed to and from the query option using the callsign field */
 	ast_copy_string(callsign, node, callsignlen);
 
@@ -343,12 +344,12 @@ int tlb_query_callsign(const char *node, char *callsign, int callsignlen)
 /*!
  * \brief AllStar Network node lookup by dns.
  * Calling routine should pass a buffer for nodedata and nodedatalength
- * of sufficient length. A typical response is 
+ * of sufficient length. A typical response is
  * "radio@123.123.123.123:4569/50000,123.123.123.123
  * This routine uses the SRV or TXT records provided by AllStarLink
  *
- * \note This routine can be called by app_rpt multiple times as 
- * it constructs the node number.  The routine will only perform a 
+ * \note This routine can be called by app_rpt multiple times as
+ * it constructs the node number.  The routine will only perform a
  * lookup after it receives 4 digits.  The actual node number may be
  * longer than 4 digits.
  *
@@ -385,7 +386,7 @@ static int node_lookup_bydns(const char *node, char *nodedata, size_t nodedatale
 		char *hostname;
 		const char *ipaddress;
 		unsigned short iaxport;
-		
+
 		/* setup the domain to lookup */
 		memset(domain,0, sizeof(domain));
 		res = snprintf(domain, sizeof(domain), "_iax._udp.%s.nodes.allstarlink.org", node);
@@ -406,7 +407,7 @@ static int node_lookup_bydns(const char *node, char *nodedata, size_t nodedatale
 
 		/* get the response */
 		record = ast_dns_result_get_records(result);
-	
+
 		if(!record) {
 			ast_debug(4, "No SRV records returned for %s\n", domain);
 			ast_dns_result_free(result);
@@ -464,7 +465,7 @@ static int node_lookup_bydns(const char *node, char *nodedata, size_t nodedatale
 		}
 
 		ast_debug(4, "Resolving DNS TXT records for: %s\n", domain);
-	
+
 		/* resolve the domain name */
 		if (ast_dns_resolve(domain, T_TXT, C_IN, &result)) {
 			ast_log(LOG_ERROR, "DNS request failed\n");
@@ -476,14 +477,14 @@ static int node_lookup_bydns(const char *node, char *nodedata, size_t nodedatale
 
 		/* get the response */
 		record = ast_dns_result_get_records(result);
-	
+
 		if(!record) {
 			ast_dns_result_free(result);
 			return -1;
 		}
 
-		/* process the text records 
-		text records are in the format 
+		/* process the text records
+		text records are in the format
 		"NN=2530" "RT=2023-02-21 17:33:07" "RB=0" "IP=104.153.109.212" "PIP=0" "PT=4569" "RH=register-west"
 		*/
 		txtrecords = ast_dns_txt_get_strings( record);
@@ -500,7 +501,7 @@ static int node_lookup_bydns(const char *node, char *nodedata, size_t nodedatale
 				ast_copy_string(iaxport,tmp + 3, sizeof(iaxport));
 			}
 		}
-	
+
 		/* format the response */
 		memset(nodedata, 0, nodedatalength);
 		snprintf(nodedata, nodedatalength, "radio@%s:%s/%s,%s", ipaddress, iaxport, actualnode, ipaddress);
@@ -558,7 +559,7 @@ int node_lookup(struct rpt *myrpt, char *digitbuf, char *nodedata, size_t nodeda
 
 	/* try to lookup using the external file(s) */
 	if(rpt_node_lookup_method == LOOKUP_BOTH || rpt_node_lookup_method == LOOKUP_FILE) {
-		
+
 		/* lock the node lookup */
 		ast_mutex_lock(&nodelookuplock);
 		if (!myrpt->p.extnodefilesn) {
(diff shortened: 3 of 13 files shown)


apps/app_rpt/rpt_link.c
```diff
@@ -23,6 +23,9 @@
 #include "rpt_link.h"
 #include "rpt_telemetry.h"
 
+#define OBUFSIZE(size) (size + 21)
```
Allan-N, line 26: We should say that the "21" is to ensure we have space for the leading count followed by a comma.  Something like the following would be even clearer :
```
#define OBUFSIZE(size) (size + sizeof"12345678,")
```
Also, I think the buffer size calculation includes the trailing NUL (so we don't need the "- 1"s)

apps/app_rpt/rpt_link.c
```diff
@@ -450,15 +453,42 @@ void rpt_link_remove(struct rpt *myrpt, struct rpt_link *l)
...
+	int buffer_size;
+	buffer_size = 0;
+	for (l = myrpt->links.next; l != &myrpt->links; l = l->next) {
+		/* if is not a real link, ignore it */
+		if (l->name[0] == '0')
+			continue;
+		if (l->mode > 1)
+			continue;			/* dont report local modes */
+		if (l->linklist[0]) {
+			buffer_size += (strlen(l->linklist) + strlen(l->name) + 3); /*+3: 2 for the commas, 1 for mode, 1 extra as the first pass has no comma*/
```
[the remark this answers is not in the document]
mkmer: There is a comma appended to the end of the full list except for the first pass:
T123,<stuff from the linklist>, T345,<stuff from the next link list> -> 2 commas if more than 1 link other wise only 1 comma.
Maybe my comment about 1 extra comma for the first pass is not clear enough :)

apps/app_rpt/rpt_link.c
```diff
@@ -482,9 +512,9 @@ void __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, int fla
 			strcat(buf, ",");
 			spos++;
 		}
```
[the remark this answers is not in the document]
mkmer: We are only counting ALINKS right here, the rest of the links are in l->linklist and we just copy them out.

apps/app_rpt/rpt_link.c
```diff
@@ -482,9 +512,9 @@ void __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, int fla
...
-		if (flag) {
+		if (flag) { /* RPT_ALINK format - only show adjacent nodes*/
 			snprintf(buf + spos, MAXLINKLIST - spos, "%s%c%c", l->name, mode, (l->lastrx1) ? 'K' : 'U');
-		} else {
+		} else { /* RPT_LINK format - show all nodes*/
 			/* add nodes into buffer */
 			if (l->linklist[0]) {
 				snprintf(buf + spos, MAXLINKLIST - spos, "%c%s,%s", mode, l->name, l->linklist);
```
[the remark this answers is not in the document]
mkmer: I have to loop through the l->linklist values to count here.  I don't think we save any CPU looping here vs where we are do we?

apps/app_rpt/rpt_link.c
```diff
@@ -503,7 +533,12 @@ void __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, int fla
 				buf[i] = mode;
 		}
 	}
-	return;
+	/*Afer building the string, count number of nodes (commas) in buffer string. The first
```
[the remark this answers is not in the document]
mkmer: If the for loop fires up, we have at least 1 link, otherwise all of this is skipped.

mkmer:
I will squash into a single commit once we are done reviewing it.

apps/app_rpt/rpt_link.c
```diff
@@ -523,40 +558,60 @@ void __kickshort(struct rpt *myrpt)
...
 
 	ast_mutex_lock(&myrpt->lock);
-	__mklinklist(myrpt, NULL, buf, 1);
+	n = __mklinklist(myrpt, NULL, buf, 1);
 	ast_mutex_unlock(&myrpt->lock);
 	/* parse em */
-	n = finddelim(ast_strdupa(buf), strs, MAXLINKLIST);
 	if (n) {
-		snprintf(obuf, sizeof(obuf) - 1, "%d,%s", n, buf);
+		snprintf(obuf, OBUFSIZE(buffer_size) - 1, "%d,%s", n, buf);
```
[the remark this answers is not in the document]
mkmer: changed to `calloc()`

apps/app_rpt/rpt_link.c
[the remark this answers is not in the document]
Allan-N: Counting here means we're only looking at the l->linklist nodes.  Counting with one loop below means that we're counting these AND the adjacent nodes.

apps/app_rpt/rpt_link.c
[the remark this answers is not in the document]
Allan-N: The `for (l = myrpt->links.next; ...` loop (above) can iterate and add NO links.

Then, the `for (link_count = 1; ...` loop (below) runs and we end up with a min link_count of 1.

apps/app_rpt/rpt_link.c
```diff
@@ -523,40 +558,57 @@ void __kickshort(struct rpt *myrpt)
...
+	}
+
+	obuf = ast_calloc(1, OBUFSIZE(buffer_size));
+	if (!obuf) {
+		return;
+	}
 
 	ast_mutex_lock(&myrpt->lock);
-	__mklinklist(myrpt, NULL, buf, 1);
+	n = __mklinklist(myrpt, NULL, buf, MAXLINKLIST, 1);
```
Allan-N, line 581: Can the list of linked nodes change between when the lock was acquired and we sized the buffer and now when we are putting together the list?  I'm thinking that we need to hold the lock across both operations.

Allan-N: Also, instead of MAXLINKLIST we should be passing in BUFSIZE(buffer_size).

and check the other places that __mklinklist() is called.

apps/app_rpt/rpt_link.c
```diff
@@ -523,40 +558,57 @@ void __kickshort(struct rpt *myrpt)
...
 	rpt_manager_trigger(myrpt, "RPT_ALINKS", obuf);
-	snprintf(obuf, sizeof(obuf) - 1, "%d", n);
+	snprintf(obuf, OBUFSIZE(buffer_size) - 1, "%d", n);
 	pbx_builtin_setvar_helper(myrpt->rxchannel, "RPT_NUMALINKS", obuf);
 	rpt_manager_trigger(myrpt, "RPT_NUMALINKS", obuf);
 	ast_mutex_lock(&myrpt->lock);
-	__mklinklist(myrpt, NULL, buf, 0);
+	n = __mklinklist(myrpt, NULL, buf, MAXLINKLIST, 0);
 	ast_mutex_unlock(&myrpt->lock);
 	/* parse em */
```
Allan-N, line 597: This is an old comment.

apps/app_rpt/rpt_link.c
```diff
@@ -523,40 +558,57 @@ void __kickshort(struct rpt *myrpt)
...
 	if (n) {
-		snprintf(obuf, sizeof(obuf) - 1, "%d,%s", n, buf);
+		snprintf(obuf, OBUFSIZE(buffer_size), "%d,%s", n, buf);
 	} else {
 		strcpy(obuf, "0");
 	}
 	pbx_builtin_setvar_helper(myrpt->rxchannel, "RPT_ALINKS", obuf);
 	rpt_manager_trigger(myrpt, "RPT_ALINKS", obuf);
-	snprintf(obuf, sizeof(obuf) - 1, "%d", n);
+	snprintf(obuf, OBUFSIZE(buffer_size) - 1, "%d", n);
```
Allan-N, line 591: Seriously, snprintf ensures that there's a NUL at the end of the buffer.  No need for "- 1"

apps/app_rpt/rpt_manager.c
```diff
@@ -207,7 +207,7 @@ static int rpt_manager_do_xstat(struct mansession *ses, const struct message *m,
 			/* Get connected node info */
 			/* Traverse the list of connected nodes */
 
-			__mklinklist(myrpt, NULL, lbuf, 0);
+			__mklinklist(myrpt, NULL, lbuf, MAXLINKLIST, 0);
```
Allan-N, line 210: Here, you can use sizeof(lbuf)

apps/app_rpt/rpt_link.c
```diff
@@ -650,7 +702,7 @@ int connect_link(struct rpt *myrpt, char *node, int mode, int perma)
 		l->disced = 2;
 		modechange = 1;
 	} else {
-		__mklinklist(myrpt, NULL, lstr, 0);
+		__mklinklist(myrpt, NULL, lstr, MAXLINKLIST, 0);
```
Allan-N, line 705: Here, you can use sizeof(lbuf)

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1249,7 +1249,7 @@ void *rpt_tele_thread(void *this)
 
 		rpt_mutex_lock(&myrpt->lock);
 		/* get all the nodes */
-		__mklinklist(myrpt, NULL, lbuf, 0);
+		__mklinklist(myrpt, NULL, lbuf,MAXLINKLIST, 0);
```
Allan-N, line 1252: Here, you can use sizeof(lbuf)

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2143,7 +2143,7 @@ void *rpt_tele_thread(void *this)
 	case FULLSTATUS:
 		rpt_mutex_lock(&myrpt->lock);
 		/* get all the nodes */
-		__mklinklist(myrpt, NULL, lbuf, 0);
+		__mklinklist(myrpt, NULL, lbuf, MAXLINKLIST, 0);
```
Allan-N, line 2146: Here, you can use sizeof(lbuf)

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2862,7 +2862,7 @@ void rpt_telemetry(struct rpt *myrpt, int mode, void *data)
 			rpt_mutex_lock(&myrpt->lock);
 			sprintf(mystr, "STATUS,%s,%d", myrpt->name, myrpt->callmode);
 			/* get all the nodes */
-			__mklinklist(myrpt, NULL, lbuf, 0);
+			__mklinklist(myrpt, NULL, lbuf, MAXLINKLIST, 0);
```
Allan-N, line 2865: Here, you can use sizeof(lbuf)

apps/app_rpt/rpt_link.h
```diff
@@ -59,7 +59,10 @@ void rpt_link_add(struct rpt *myrpt, struct rpt_link *l);
 void rpt_link_remove(struct rpt *myrpt, struct rpt_link *l);
 
 /*! \brief must be called locked */
-void __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, int flag);
+int __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, size_t *bufsize, int flag);
+
+/*! \brief must be called locked */
+int __get_buffer_size (struct rpt *myrpt);
```
Allan-N, line 65: __get_nodelist_size

apps/app_rpt.c
```diff
@@ -1660,86 +1660,85 @@ static inline void init_text_frame(struct ast_frame *wf)
...
+	if (*str == 'G') {		/* got GPS data */
 		/* re-distribute it to attached nodes */
 		distribute_to_all_links(myrpt, mylink, src, NULL, str, &wf);
 		return;
 	}
-	if (tmp[0] == 'L') {
+	if (*str == 'L') {
 		rpt_mutex_lock(&myrpt->lock);
-		strcpy(mylink->linklist, tmp + 2);
+		ast_copy_string(mylink->linklist, (str + 2), MAXLINKLIST);
```
Allan-N, line 1705: Instead of MAXLINKLIST, could we use sizeof(mylink->linklist) ?

apps/app_rpt/rpt_cli.c
```diff
@@ -528,7 +528,7 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 //### GET CONNECTED NODE INFO ####################
 			// Traverse the list of connected nodes 
 
-			__mklinklist(myrpt, NULL, lbuf, 0);
+			__mklinklist(myrpt, NULL, lbuf, MAXLINKLIST, 0);
```
Allan-N, line 531: sizeof(lbuf)

apps/app_rpt/rpt_cli.c
```diff
@@ -657,7 +657,7 @@ static int rpt_do_nodes(int fd, int argc, const char *const *argv)
 			/* Make a copy of all stat variables while locked */
 			myrpt = &rpt_vars[i];
 			rpt_mutex_lock(&myrpt->lock);	/* LOCK */
-			__mklinklist(myrpt, NULL, lbuf, 0);
+			__mklinklist(myrpt, NULL, lbuf, MAXLINKLIST, 0);
```
Allan-N, line 660: sizeof(lbuf)

apps/app_rpt/rpt_link.c
mkmer: Yah - got a bit sloppy. Thanks

apps/app_rpt/rpt_link.c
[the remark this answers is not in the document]
mkmer: That's what we want - total node count in `buff` - if `flag` is set, we are counting ALINK and generating the RPT_ALINK string - note the format is a bit different.  When we fall out the end we count the ALINK string.  If `flag` is flase, we are building a RPT_LINK string, 2 cases: 1 where a link has a l->linklist (aka it's connected to other nodes), or l->linklist[0]=0 where the node has no nodes linked to it.

apps/app_rpt/rpt_cli.c
```diff
@@ -657,10 +657,10 @@ static int rpt_do_nodes(int fd, int argc, const char *const *argv)
 			/* Make a copy of all stat variables while locked */
 			myrpt = &rpt_vars[i];
 			rpt_mutex_lock(&myrpt->lock);	/* LOCK */
-			__mklinklist(myrpt, NULL, lbuf, MAXLINKLIST, 0);
+			__mklinklist(myrpt, NULL, lbuf, sizeof(lbuf), 0);
 			rpt_mutex_unlock(&myrpt->lock);	/* UNLOCK */
 			/* parse em */
-			ns = finddelim(lbuf, strs, MAXLINKLIST);
+			ns = finddelim(lbuf, strs, sizeof(lbuf));
```
Allan-N, line 663: At first, I was going to write that `finddelim()` was different than `__mklinklist()` and that the size passed in should be the number of elements in the `strs` array (i.e. change `sizeof(lbuf)` --> `sizeof(strs)/sizeof(strs[0])`).  But, `explode_string()` is checking `(l < limit)` instead of `(i < limit)`.  I think `explode_string()` needs to be fixed so that it knows the size of `strp` and we should still be passing the size of the `strs` array.

apps/app_rpt/rpt_link.c
```diff
@@ -702,9 +701,9 @@ int connect_link(struct rpt *myrpt, char *node, int mode, int perma)
 		l->disced = 2;
 		modechange = 1;
 	} else {
-		__mklinklist(myrpt, NULL, lstr, MAXLINKLIST, 0);
+		__mklinklist(myrpt, NULL, lstr, sizeof(lstr), 0);
 		rpt_mutex_unlock(&myrpt->lock);
-		n = finddelim(lstr, strs, MAXLINKLIST);
+		n = finddelim(lstr, strs, sizeof(lstr));
```
Allan-N, line 706: I already commented on how finddelim() needs help.  IMHO, we also need a tweak here.

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1249,10 +1249,10 @@ void *rpt_tele_thread(void *this)
 
 		rpt_mutex_lock(&myrpt->lock);
 		/* get all the nodes */
-		__mklinklist(myrpt, NULL, lbuf,MAXLINKLIST, 0);
+		__mklinklist(myrpt, NULL, lbuf,sizeof(lbuf), 0);
 		rpt_mutex_unlock(&myrpt->lock);
 		/* parse em */
-		ns = finddelim(lbuf, strs, MAXLINKLIST);
+		ns = finddelim(lbuf, strs, sizeof(lbuf));
```
Allan-N, line 1255: Another finddelim()

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2143,10 +2143,10 @@ void *rpt_tele_thread(void *this)
 	case FULLSTATUS:
 		rpt_mutex_lock(&myrpt->lock);
 		/* get all the nodes */
-		__mklinklist(myrpt, NULL, lbuf, MAXLINKLIST, 0);
+		__mklinklist(myrpt, NULL, lbuf, sizeof(lbuf), 0);
 		rpt_mutex_unlock(&myrpt->lock);
 		/* parse em */
-		ns = finddelim(lbuf, strs, MAXLINKLIST);
+		ns = finddelim(lbuf, strs, sizeof(lbuf));
```
Allan-N, line 2149: finddelim()

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2862,10 +2862,10 @@ void rpt_telemetry(struct rpt *myrpt, int mode, void *data)
 			rpt_mutex_lock(&myrpt->lock);
 			sprintf(mystr, "STATUS,%s,%d", myrpt->name, myrpt->callmode);
 			/* get all the nodes */
-			__mklinklist(myrpt, NULL, lbuf, MAXLINKLIST, 0);
+			__mklinklist(myrpt, NULL, lbuf, sizeof(lbuf), 0);
 			rpt_mutex_unlock(&myrpt->lock);
 			/* parse em */
-			ns = finddelim(lbuf, strs, MAXLINKLIST);
+			ns = finddelim(lbuf, strs, sizeof(lbuf));
```
Allan-N, line 2868: finddelim()

apps/app_rpt/rpt_link.c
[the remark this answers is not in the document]
mkmer: Alright - I thought I had tested this. It is 0 until someone connects, then it is stuck at 1 when they disconnect... AND I though this was in the for loop.
I'm pickin up what your saying now...

apps/app_rpt/rpt_cli.c
mkmer: Should we have 2 limits? or just limit the thing write to, and trust that the thing we read from is null terminated. That's probably "fine" since all string reads work that way.

Allan-N: I think we're ok assuming that we have a NUL terminated string.

mkmer:
Fixing up finddelim may not be possible.  I think some of the instances are searching through non null terminated values (but I can't be sure right now).  With numbers like 3 in the limit while strs[100].
The finddelim is all in one commit we can easily back out if we think it should hold off for another PR.

Allan-N:
I'll take a look at the finddelim call sites tomorrow.

Allan-N:
FYI : in `.../asterisk/include/asterisk/utils.h` I just found :
```
#define ARRAY_LEN(a) (size_t) (sizeof(a) / sizeof(0[a]))
```
So, when needed, we can use `ARRAY_LEN(strs)`

apps/app_rpt/rpt_cli.c
```diff
@@ -590,7 +590,7 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 
 //### GET ALL LINKED NODES INFO ####################
 			/* parse em */
-			ns = finddelim(lbuf, strs, sizeof(lbuf));
+			ns = finddelim(lbuf, strs, sizeof(strs));
```
Allan-N, line 593: ARRAY_LEN(strs)

apps/app_rpt/rpt_cli.c
```diff
@@ -660,7 +660,7 @@ static int rpt_do_nodes(int fd, int argc, const char *const *argv)
 			__mklinklist(myrpt, NULL, lbuf, sizeof(lbuf), 0);
 			rpt_mutex_unlock(&myrpt->lock);	/* UNLOCK */
 			/* parse em */
-			ns = finddelim(lbuf, strs, sizeof(lbuf));
+			ns = finddelim(lbuf, strs, sizeof(strs));
```
Allan-N, line 663: ARRAY_LEN(strs)

apps/app_rpt/rpt_config.c
```diff
@@ -668,7 +668,7 @@ int forward_node_lookup(char *digitbuf, struct ast_config *cfg, char *nodedata,
 		}
 
 		/* parse the external node file name(s) - we allow for multiple files */
-		n = finddelim(efil, strs, 100);
+		n = finddelim(efil, strs, sizeof(strs));
```
Allan-N, line 671: ARRAY_LEN(strs)

apps/app_rpt/rpt_config.c
```diff
@@ -1072,7 +1072,7 @@ void load_rpt_vars(int n, int init)
 	val = ast_variable_retrieve(cfg, cat, "inxlat");
 	if (val) {
 		memset(&rpt_vars[n].p.inxlat, 0, sizeof(struct rpt_xlat));
-		i = finddelim((char*) val, strs, 3); /*! \todo Illegal cast */
+		i = finddelim((char*) val, strs, sizeof(strs)); /*! \todo Illegal cast */
```
Allan-N, line 1075: ARRAY_LEN(strs)

apps/app_rpt/rpt_config.c
```diff
@@ -1087,7 +1087,7 @@ void load_rpt_vars(int n, int init)
 	val = ast_variable_retrieve(cfg, cat, "outxlat");
 	if (val) {
 		memset(&rpt_vars[n].p.outxlat, 0, sizeof(struct rpt_xlat));
-		i = finddelim((char*) val, strs, 3); /*! \todo Illegal cast */
+		i = finddelim((char*) val, strs, sizeof(strs)); /*! \todo Illegal cast */
```
Allan-N, line 1090: ARRAY_LEN(strs)

apps/app_rpt/rpt_config.c
```diff
@@ -1225,7 +1225,7 @@ void load_rpt_vars(int n, int init)
 		int k, nukw, statenum;
 		statenum = atoi(vp->name);
 		ast_copy_string(s1, vp->value, sizeof(s1));
-		nukw = finddelim(s1, strs, 32);
+		nukw = finddelim(s1, strs, sizeof(strs));
```
Allan-N, line 1228: ARRAY_LEN(strs)

apps/app_rpt/rpt_functions.c
```diff
@@ -361,7 +361,7 @@ int function_ilink(struct rpt *myrpt, char *param, char *digits, int command_sou
 
 	case 16:					/* Restore links disconnected with "disconnect all links" command */
 		strcpy(tmp, myrpt->savednodes);	/* Make a copy */
-		finddelim(tmp, strs, MAXLINKLIST);	/* convert into substrings */
+		finddelim(tmp, strs, sizeof(strs));	/* convert into substrings */
```
Allan-N, line 364: ARRAY_LEN(strs)

apps/app_rpt/rpt_functions.c
```diff
@@ -920,7 +920,7 @@ int function_autopatchup(struct rpt *myrpt, char *param, char *digitbuf, int com
 		if (!lparam) {
 			return DC_ERROR;
 		}
-		paramlength = finddelim(lparam, paramlist, 20);
+		paramlength = finddelim(lparam, paramlist, sizeof(paramlist));
```
Allan-N, line 923: ARRAY_LEN(paramlist)

apps/app_rpt/rpt_link.c
```diff
@@ -702,7 +702,7 @@ int connect_link(struct rpt *myrpt, char *node, int mode, int perma)
 	} else {
 		__mklinklist(myrpt, NULL, lstr, sizeof(lstr), 0);
 		rpt_mutex_unlock(&myrpt->lock);
-		n = finddelim(lstr, strs, sizeof(lstr));
+		n = finddelim(lstr, strs, sizeof(strs));
```
Allan-N, line 705: ARRAY_LEN(strs)

apps/app_rpt/rpt_manager.c
```diff
@@ -271,7 +271,7 @@ static int rpt_manager_do_xstat(struct mansession *ses, const struct message *m,
 			/* Get all linked nodes info */
 
 			/* parse em */
-			ns = finddelim(lbuf, strs, MAXLINKLIST);
+			ns = finddelim(lbuf, strs, sizeof(strs));
```
Allan-N, line 274: ARRAY_LEN(strs)

apps/app_rpt/rpt_rig.c
```diff
@@ -2444,7 +2444,7 @@ char check_tx_freq(struct rpt *myrpt)
 	/* Parse the limits */
 
 	ast_copy_string(limits, limitlist->value, sizeof(limits));
-	finddelim(limits, limit_ranges, 40);
+	finddelim(limits, limit_ranges, sizeof(limit_ranges));
```
Allan-N, line 2447: ARRAY_LEN(limit_ranges)

apps/app_rpt/rpt_telemetry.c
```diff
@@ -544,7 +544,7 @@ void handle_varcmd_tele(struct rpt *myrpt, struct ast_channel *mychannel, char *
 	unsigned int t1;
 	struct ast_tm localtm;
 
-	n = finddelim(varcmd, strs, 100);
+	n = finddelim(varcmd, strs, sizeof(strs));
```
Allan-N, line 547: ARRAY_LEN(strs)

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1252,7 +1252,7 @@ void *rpt_tele_thread(void *this)
 		__mklinklist(myrpt, NULL, lbuf,sizeof(lbuf), 0);
 		rpt_mutex_unlock(&myrpt->lock);
 		/* parse em */
-		ns = finddelim(lbuf, strs, sizeof(lbuf));
+		ns = finddelim(lbuf, strs, sizeof(strs));
```
Allan-N, line 1255: ARRAY_LEN(strs)

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2146,7 +2146,7 @@ void *rpt_tele_thread(void *this)
 		__mklinklist(myrpt, NULL, lbuf, sizeof(lbuf), 0);
 		rpt_mutex_unlock(&myrpt->lock);
 		/* parse em */
-		ns = finddelim(lbuf, strs, sizeof(lbuf));
+		ns = finddelim(lbuf, strs, sizeof(strs));
```
Allan-N, line 2149: ARRAY_LEN(strs)

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2865,7 +2865,7 @@ void rpt_telemetry(struct rpt *myrpt, int mode, void *data)
 			__mklinklist(myrpt, NULL, lbuf, sizeof(lbuf), 0);
 			rpt_mutex_unlock(&myrpt->lock);
 			/* parse em */
-			ns = finddelim(lbuf, strs, sizeof(lbuf));
+			ns = finddelim(lbuf, strs, sizeof(strs));
```
Allan-N, line 2868: ARRAY_LEN(strs)

apps/app_rpt/rpt_utils.c
```diff
@@ -58,6 +58,9 @@ int explode_string(char *str, char *strp[], int limit, char delim, char quote)
 			l++;
 			strp[i++] = str + 1;
 		}
+		if (i > limit) {
```
Allan-N, line 61: We need to use `i >= (limit - 1)` because we can only store `strp[0] ... strp[limit - 2]`.  That way, if we've filled the list we know that the last entry, `strp[limit - 1]` will be NULL.

Allan-N:
Looks like chan_echolink.c, chan_tlb.c, and chan_voter have their own copy/version of `findelim()` that all need the same help.

mkmer:
> Looks like chan_echolink.c, chan_tlb.c, and chan_voter have their own copy/version of `findelim()` that all need the same help.

I saw that when I was making the move.  
I think (and will) do the channels in separate PR. 1 for each channel.
This PR is getting to be one of those threads you don't want to pull :)  Just keeps going and going.

apps/app_rpt/rpt_utils.c
mkmer: strp[i++] -> we put the value in strp[] then add to `i` right?
when we get to the if, we are looking at the NEXT pointer, so if it is `i = limit` we are pointing at the last value, we break out and set `strp[i] = 0`
I commented on this in one of the other PR's -> what is the value of 0 on the last array of pointers?  Will we get an assert if we try to use it?

mkmer: I was missing 0 based index (blah).  Fixed.

Allan-N: I added this comment in another reply :
>It's OK to fill the last slot of the array ... as long as the call sites are iterating with `for (i = 0; i < n; i++) {}`. But, we would need the last element to be NULL if the usage is `while (*ptr) {}`. Regardless, if we have an array of size 10 we don't want to write to array[10].

And, for the comment history, setting `strp[i] = 0 /* or NULL */;` is what you would use to ensure that `while (*ptr) {}` works.

apps/app_rpt.c
```diff
@@ -3165,7 +3167,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			lf.samples = 0;
 			l->linklisttimer = LINKLISTTIME;
 			strcpy(lstr, "L ");
-			__mklinklist(myrpt, l, lstr + 2, 0);
+			__mklinklist(myrpt, l, lstr + 2, sizeof(lstr), 0);
```
Allan-N, line 3170: Because we are copying into `lstr + 2` we should use `sizeof(lstr) - 2` as the size of the buffer.

apps/app_rpt.c
```diff
@@ -3156,7 +3159,6 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 		if ((!l->linklisttimer) && (l->name[0] != '0') && (!l->isremote)) {
 			struct ast_frame lf;
 			char lstr[MAXLINKLIST];
-
```
Allan-N, line 3159: Not sure why you removed this line

apps/app_rpt/rpt_link.c
```diff
@@ -523,40 +562,50 @@ void __kickshort(struct rpt *myrpt)
...
+	/* figure out the RPT_LINK string size - this will be the largest size
+	 * RPT_ALINK is always a subset of RPT_LINK
+	 */
 	ast_mutex_lock(&myrpt->lock);
-	__mklinklist(myrpt, NULL, buf, 1);
-	ast_mutex_unlock(&myrpt->lock);
+	buffer_size = __get_nodelist_size(myrpt);
+	buf = ast_calloc(1, BUFSIZE(buffer_size));
+	if (!buf) {
+		return;
```
Allan-N, line 574: Need to ast_mutex_unlock()

apps/app_rpt/rpt_link.c
```diff
@@ -523,40 +562,50 @@ void __kickshort(struct rpt *myrpt)
...
+	}
+	obuf = ast_calloc(1, OBUFSIZE(buffer_size));
+	if (!obuf) {
+		return;
```
Allan-N, line 578: Need to ast_mutex_unlock()

... and do you also need to ast_free(buf) ?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1249,10 +1249,10 @@ void *rpt_tele_thread(void *this)
 
 		rpt_mutex_lock(&myrpt->lock);
 		/* get all the nodes */
-		__mklinklist(myrpt, NULL, lbuf, 0);
+		__mklinklist(myrpt, NULL, lbuf,sizeof(lbuf), 0);
```
Allan-N, line 1252: Maybe a space after `lbuf,` ?

Allan-N (changes requested):

Allan-N (approved):

apps/app_rpt/rpt_link.c
```diff
@@ -450,15 +454,48 @@ void rpt_link_remove(struct rpt *myrpt, struct rpt_link *l)
 	check_link_list(myrpt);
 }
 
-void __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, int flag)
+/*!
+ * \brief Get required buffer size for link message. 
+ * \retval		buffer size.
+ */
+
+int __get_nodelist_size (struct rpt *myrpt)
```
mkmer, line 462: I feel like we can do something more resource friendly here - why do we need two buffers?  Maybe we can start filling the buffer at the offset for ```<node length>,``` and `snprintf` first?  Possibly get the node count in this function as well.

Allan-N: Yes, we could use one buffer with enough space for the formatted count followed by the string.  Then, we fill the string starting at an offset based on the length of the link count + the comma.

We could also change __get_nodelist_size() to __get_nodelist_count() and using the result to compute how much space is needed both for the count(s) and for the list of nodes right from the start.

Lastly, we could have __mklinklist fill two buffers at the same time (one with adjacent nodes and the other with non-adjacent nodes). That would have us making a single pass.

But, these are all just optimizations.

apps/app_rpt/rpt_link.c
```diff
@@ -450,15 +454,52 @@ void rpt_link_remove(struct rpt *myrpt, struct rpt_link *l)
...
+		 */
+		if (l->linklist[0]) {
+			buffer_size_links += (strlen(l->linklist) + 1); /* +1: 1 for comma */
+		}
+		link_size = strlen(l->name);
+		buffer_size_links += (link_size + 2);  /* +2: 1 for comma, 1 form mode (T/R) */
+		buffer_size_alinks += (link_size + 3); /* +3: 1 for comma, 2 for mode (TU/TK/etc) */
+	}
+
+	return (buffer_size_links > buffer_size_alinks) ? buffer_size_links : buffer_size_alinks;
```
Allan-N, line 489: I don't get the point of this.  Won't buffer_size_links always be >= buffer_size_alinks?

If this change was driven by my comment about returning both sizes then what you'd want is to change the function to :
```
void __get_nodelist_size (struct rpt *myrpt, int *buffer_size_links, int *buffer_size_alinks)
```
and then the function woudl be called with :
```
__get_nodelist_size(myrpt, &buffer_size_links, &buffer_size_alinks);
```

Either that, or pass in the "flag" indicating which size you want returned.

mkmer: There are 2 different sizes -> `RPT_ALINKS` is +3, `RPT_LINKS` is +2.  When a node only has ALINKS attached (no links on the adjacent links), the required ALINKS buffer is larger than LINKS buffer. (My node was actually truncating the RPT_ALINKS value).  If there is at least 1 link connected to an adjacent node, RPT_LINKS is larger than RPT_ALINKS buffer. - We only really need to malloc() one buffer large enough for everything.
Simply adding 3 to everything will make a much larger buffer than needed on large linked systems (say 300 nodes like the original issue connecting to).
If we don't care about making the buffer much larger than necessary, we can just +3 for buffer_size for every node and remove the extra counter.

My original thought was that RPT_LINKS was always larger than RPT_ALINKS... but that was an incorrect assumption.

Allan-N: >My original thought was that RPT_LINKS was always larger than RPT_ALINKS... but that was an incorrect assumption.

Thank you for the explanation.  It sounds like we both made the same incorrect assumption.

apps/app_rpt/rpt_link.c
```diff
@@ -482,14 +523,15 @@ void __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, int fla
 			strcat(buf, ",");
 			spos++;
 		}
-		if (flag) {
-			snprintf(buf + spos, MAXLINKLIST - spos, "%s%c%c", l->name, mode, (l->lastrx1) ? 'K' : 'U');
-		} else {
+	    one_link = 1;
```
InterLinked1, line 526: whitespace

apps/app_rpt/rpt_link.c
```diff
@@ -482,14 +523,15 @@ void __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, int fla
...
+		if (flag) { /* RPT_ALINK format - only show adjacent nodes*/
+			snprintf(buf + spos, bufsize - spos, "%s%c%c", l->name, mode, (l->lastrx1) ? 'K' : 'U');
+		} else { /* RPT_LINK format - show all nodes*/
 			/* add nodes into buffer */
 			if (l->linklist[0]) {
-				snprintf(buf + spos, MAXLINKLIST - spos, "%c%s,%s", mode, l->name, l->linklist);
+				snprintf(buf + spos, bufsize - spos, "%c%s,%s", mode, l->name, l->linklist);
```
InterLinked1, line 532: If this was previously calculated to be sufficient space for the linked list (and it's *proven* to be correct), snprintf isn't needed, sprintf can be used safely

apps/app_rpt/rpt_link.c
```diff
@@ -503,7 +545,15 @@ void __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, int fla
 				buf[i] = mode;
 		}
 	}
-	return;
+	/*Afer building the string, count number of nodes (commas) in buffer string. The first
```
InterLinked1, line 548: Space after /, and after is mispelled (typo)

apps/app_rpt/rpt_link.c
```diff
@@ -450,15 +454,52 @@ void rpt_link_remove(struct rpt *myrpt, struct rpt_link *l)
 	check_link_list(myrpt);
 }
 
-void __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, int flag)
+/*!
+ * \brief Get required buffer size for link message. 
+ * \retval		buffer size.
+ */
+
+int __get_nodelist_size (struct rpt *myrpt)
```
InterLinked1, line 462: Did you consider consolidating calculating the length and building the list into one function, that can just return the new list?

I really don't like the idea of doing two passes like this. It would be cleaner to use dynamically allocated strings like asprintf or the `ast_str` API which is intended for this purpose.

InterLinked1 (changes requested):

KB4MDD (approved):
Good - pending changes for @InterLinked1

apps/app_rpt/rpt_link.c
mkmer: Do we want to remove `snprintf`?  @Allan-N ?

apps/app_rpt/rpt_link.c
mkmer: I did consider looking at a single pass, but we need to know the size first with any of these (right?)  Because we are building the string on the fly adding to the buffer I don't think we can use those without knowing the size or allocating a TON of memory to hold the largest list?  
`l->linklist` is a variable up to 512 bytes long * number of connected nodes + 2 or 3 depending on RPT_ALINK or RPT_LINK.

Goal 1 -> make sure we have a variable buffer and big enough to support a large list (based on issue https://github.com/AllStarLink/app_rpt/issues/469).
Goal 2 -> be space efficient.  The list is usually small, but on bigger systems (like the one in the attached issue) can create very large lists.

I'd REALLY like to insert the count value rather than using 2 buffers, either by knowing the count ahead of time and filling the buffer x bytes forward, or adding the count first.  I suppose we could count the commas in the `l->linklist` as we calculate the size and return both the count and the size, then allocate the buffer, add the count, then start building the strings....

That all said - we still have a problem on the receiving end -> if we generate a RPT_LINK > 512 and send it upstream, the other end only has 512 bytes allocate for it's `l->linklist` thus truncating it.   What is "fixed" here is the RPT_LINK list will not be truncated locally when there is a large network.

While we "could" dynamically set the `l->linklist`, we would be constantly alloc/free every time we get a message -> not sure it's worth all that CPU effort, and probably should be addressed in a different PR.

Allan-N: >I really don't like the idea of doing two passes like this. It would be cleaner to use dynamically allocated strings like asprintf or the ast_str API which is intended for this purpose.

Agreed.  But, I think we should work on the optimizations with a followup PR.  There, we can figure out how to make a single pass over the node lists and build the string(s) on the fly using ast_str_create(), ast_str_set(), ast_str_append(), etc.

mkmer:
Go figure - a spelling error... I'll get after it.

apps/app_rpt/rpt_link.c
```diff
@@ -6,6 +6,7 @@
 
 #include "asterisk.h"
 
+#include <asterisk/lock.h>
```
Allan-N, line 9: "asterisk/lock.h"

and you can nix the blank line above

apps/app_rpt/rpt_config.c
```diff
@@ -1,6 +1,7 @@
 
 #include "asterisk.h"
 
+#include <asterisk/utils.h>
```
Allan-N, line 4: "asterisk/utils.h"

... and you can nix the blank line above

Allan-N (changes requested):

apps/app_rpt/rpt_link.c
InterLinked1: Actually, by convention, there is always a blank line after asterisk.h

apps/app_rpt/rpt_config.c
InterLinked1: Same here, there's always a blank line after asterisk.h

The typical format is:

````
include "asterisk.h"
<BLANK LINE>
... all the system includes
<BLANK LINE>
... any other asterisk includes
````

File: apps/app_rpt.c [hunk 53]
```diff
@@ -1701,86 +1701,89 @@ static inline void init_text_frame(struct ast_frame *wf)
 }
 
 static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *str)
-{
-	char tmp[512], tmp1[512], cmd[300] = "", dest[300], src[30], c;
+{ /* I think cmd[32] is big enough (same as others)? Why is remote_data src[300] in other locations?
```
InterLinked1, line 1704: Maybe prefix this comment with XXX, and start it on its own line

apps/app_rpt.c
```diff
@@ -1701,86 +1701,89 @@ static inline void init_text_frame(struct ast_frame *wf)
...
 		rpt_mutex_unlock(&myrpt->lock);
-		ast_debug(7, "@@@@ node %s received node list %s from node %s\n", myrpt->name, tmp, mylink->name);
+		ast_debug(7, "@@@@ node %s received node list %s from node %s\n", myrpt->name, str, mylink->name);
 		return;
 	}
-	if (tmp[0] == 'M') {
+	if (*str == 'M') {
 		rest = 0;
-		if (sscanf(tmp, "%s %s %s %n", cmd, src, dest, &rest) < 3) {
+		if (sscanf(str, "%s %s %s %n", cmd, src, dest, &rest) < 3) { /* TODO: We should limit to sizeof(cmd, src, dest) */
```
InterLinked1, line 1758: Use proper doxygen syntax for all the todo's: `/*! \todo We should do foobar... */`

apps/app_rpt.c
```diff
@@ -3209,7 +3212,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			lf.samples = 0;
 			l->linklisttimer = LINKLISTTIME;
 			strcpy(lstr, "L ");
-			__mklinklist(myrpt, l, lstr + 2, 0);
+			__mklinklist(myrpt, l, lstr + 2, (sizeof(lstr) - 2), 0);
```
InterLinked1, line 3215: No parentheses needed here

apps/app_rpt.c
```diff
@@ -1701,86 +1701,89 @@ static inline void init_text_frame(struct ast_frame *wf)
...
 	int i, seq, res, ts, rest;
 	struct ast_frame wf;
 
 	init_text_frame(&wf);
 	wf.datalen = strlen(str) + 1;
 	wf.src = "handle_link_data";
-	/* put string in our buffer */
-	ast_copy_string(tmp, str, sizeof(tmp) - 1);
-
+
```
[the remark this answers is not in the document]
mkmer: We only call it from here:
https://github.com/AllStarLink/app_rpt/blob/01b809921dbe5155722323f4f690148e69e9837a/apps/app_rpt.c#L4452-L4458
Is it possible to have an AST_FRAME_TEXT with nothing in it?

apps/app_rpt.c
```diff
@@ -1701,86 +1701,89 @@ static inline void init_text_frame(struct ast_frame *wf)
 }
 
 static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *str)
```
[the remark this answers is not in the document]
mkmer: There are a few changes to make this happen, but this
https://github.com/AllStarLink/app_rpt/blob/01b809921dbe5155722323f4f690148e69e9837a/apps/app_rpt.c#L1631
requires a cast to (char *) to make the whole thing work... is that still "ok"?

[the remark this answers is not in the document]
InterLinked1: Never mind in that case, was just wondering

apps/app_rpt.c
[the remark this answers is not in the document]
InterLinked1: It *looks* like tstr is always nonempty in this case to me...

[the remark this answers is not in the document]
mkmer: It for sure is null terminated...

apps/app_rpt.c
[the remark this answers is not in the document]
mkmer: We don't actually modify it... but I can't change the `struct ast_frame wf` to be constant - even though the way we use it, it is constant.  Is casting it bad form here (if so, I'll back the change out).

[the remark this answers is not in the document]
Allan-N: In general, it's a bad idea to cast a `const char *` to a `char *`.

Why?  Because you are saying "I know that I have a string that can't be modified but I'm telling you that it's OK for you to make changes to the string".  What happens when the code attempts (or succeeds) to change this read-only string that may be stored in write-protected memory?

apps/app_rpt/rpt_config.c
Allan-N: OK.  So the blank line needs to stay but the asterisk/utils.h include should be moved down with the others.

apps/app_rpt/rpt_link.c
Allan-N: OK. So the blank line needs to stay but the asterisk/lock.h include should be moved down with the others.

apps/app_rpt.c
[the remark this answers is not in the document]
mkmer: I get it... it's unfortunate we can't change the structure, as we DON'T change it.
I will unwind it.

Allan-N (approved):

apps/app_rpt.c
```diff
@@ -6837,8 +6841,8 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		rpt_mutex_unlock(&myrpt->lock);
 
 		//return AST_PBX_KEEPALIVE;
-		pthread_exit(NULL); // BUGBUG: For now, this emulates the behavior of KEEPALIVE, but this won't be a clean exit. Makes it work, but since the PBX doesn't clean up we'll leak memory. Either do what the PBX core does here or we need to somehow do KEEPALIVE handling in the core, possibly with a custom patch for now.
-		//return -1;				/*! \todo AST_PBX_KEEPALIVE doesn't exist anymore. Figure out what we should return here. */
+		pthread_exit(NULL); /* BUGBUG: For now, this emulates the behavior of KEEPALIVE, but this won't be a clean exit. Makes it work, but since the PBX doesn't clean up we'll leak memory. Either do what the PBX core does here or we need to somehow do KEEPALIVE handling in the core, possibly with a custom patch for now. */
```
InterLinked1, line 6844: I think you could probably leave these two lines alone for now. This code is a hack that shouldn't remain long term and needs replacement, so I don't know that it needs to be "legitimized". In any case, it's a formatting change well unrelated to the purpose of this PR

InterLinked1 (changes requested):

apps/app_rpt.c
```diff
@@ -1702,85 +1702,89 @@ static inline void init_text_frame(struct ast_frame *wf)
 
 static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *str)
 {
-	char tmp[512], tmp1[512], cmd[300] = "", dest[300], src[30], c;
+  /* XXX I think cmd[32] is big enough (same as others)? Why is remote_data src[300] in other locations?
```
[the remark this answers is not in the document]
mkmer: I have to dig further, at some point other calls had cmd[32] defined.  We also have src[30] defined here, but in other similar calls https://github.com/AllStarLink/app_rpt/blob/7b97e6c758a4a76b72c68811f65f9b53b85d9d73/apps/app_rpt.c#L2182-L2185 it is 300
I will change the comment about src[30] rather than cmd.

apps/app_rpt.c
mkmer: // are not allowed :)

apps/app_rpt.c
```diff
@@ -1702,85 +1702,89 @@ static inline void init_text_frame(struct ast_frame *wf)
...
 		/* re-distribute it to attached nodes */
 		distribute_to_all_links(myrpt, mylink, src, NULL, str, &wf);
 		return;
 	}
-	if (tmp[0] == 'L') {
+	if (*str == 'L') {
 		rpt_mutex_lock(&myrpt->lock);
-		strcpy(mylink->linklist, tmp + 2);
+		if (strlen(str + 2) > sizeof(mylink->linklist)) {
+			ast_log(LOG_WARNING, "Link list too long: %s\n", str + 2);
```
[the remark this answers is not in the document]
mkmer: I think the string is probably enough for debug.  The thought was to add a message if we exceed the buffer size so we can easily find the offending code that is truncating the list.   We will have the string, and the sizesof() is well known.
This is one of the locations that triggered me creating an issue related to the limit of the list size.

apps/app_rpt.c
InterLinked1: It isn't, but the change isn't related to this PR in any way that I can see.

Generally speaking, in Asterisk (and many other projects), formatting changes as part of a code change are frowned upon, particularly if they aren't related to the affected code. Everybody does fix formatting occasionally as part of other changes, but usually only in the area of the affected code.

I'm fine with the other formatting fixes, I just personally don't like this one because it's completely unrelated to the other code, and it's bad code that needs to be removed anyways. That's why I'd prefer it be left alone as it has been thus far. For example, the git blame for these lines goes back to a relevant commit that's useful in the history. Making this change will add no real value and just make it harder to trace in the future.

It is a good reminder that I should take care of this soon though :)

apps/app_rpt.c
```diff
@@ -6837,8 +6841,8 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		rpt_mutex_unlock(&myrpt->lock);
 
 		//return AST_PBX_KEEPALIVE;
-		pthread_exit(NULL); // BUGBUG: For now, this emulates the behavior of KEEPALIVE, but this won't be a clean exit. Makes it work, but since the PBX doesn't clean up we'll leak memory. Either do what the PBX core does here or we need to somehow do KEEPALIVE handling in the core, possibly with a custom patch for now.
-		//return -1;				/*! \todo AST_PBX_KEEPALIVE doesn't exist anymore. Figure out what we should return here. */
+		pthread_exit(NULL);	// BUGBUG: For now, this emulates the behavior of KEEPALIVE, but this won't be a clean exit. Makes it work, but since the PBX doesn't clean up we'll leak memory. Either do what the PBX core does here or we need to somehow do KEEPALIVE handling in the core, possibly with a custom patch for now. */
```
InterLinked1, line 6844: I see you tried to revert the change here, but it's still showing as a change because of the whitespace, would you be able to fix that?
Should be able to just copy the old over to make it detect no diff

apps/app_rpt/rpt_link.c
```diff
@@ -450,15 +454,52 @@ void rpt_link_remove(struct rpt *myrpt, struct rpt_link *l)
...
+/*!
+ * \brief Get required buffer size for link message. 
+ * \retval		buffer size.
+ */
+
+int __get_nodelist_size (struct rpt *myrpt)
+{
+	struct rpt_link *l;
+	int buffer_size_links = 0, buffer_size_alinks = 0, link_size;
+
```
InterLinked1, line 466: As long as you're making one more change here, could we also remove one of the two blank lines?

InterLinked1 (approved):

PR #476, opened 2025-01-30
chan_tlb.c: finddelim update limit to length of strp from str

mkmer:
Check the correct length to prevent over running the output pointer.

### DIFF
--- a/channels/chan_tlb.c
+++ b/channels/chan_tlb.c
@@ -493,9 +493,9 @@ static void strupr(char* restrict str)
  * \param limit		Maximum number of substrings to process.
  */
 
-static int finddelim(char *str, char *strp[], int limit)
+static int finddelim(char *str, char *strp[], size_t limit)
 {
-	int i, l, inquo;
+	int i, inquo;
 
 	inquo = 0;
 	i = 0;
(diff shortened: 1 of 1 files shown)


channels/chan_tlb.c
```diff
@@ -519,6 +519,9 @@ static int finddelim(char *str, char *strp[], int limit)
 			l++;
 			strp[i++] = str + 1;
 		}
+		if (i => limit) {
+			break;
+		}
 	}
 	strp[i] = 0;
```
mkmer, line 526: I'm not sure this makes sense to 0 the last pointer value in the array of pointers.
`i => limit` at least makes sure we are not writing outside of the limit

Allan-N: I agree.  It's OK to fill the last slot of the array ... as long as the call sites are iterating with `for (i = 0; i < n; i++) {}`.  But, we would need the last element to be NULL if the usage is `while (*ptr) {}`.  Regardless, if we have an array of size 10 we don't want to write to array[10].

mkmer: Ah - I see my problem. 0 based counts 0 - 9 = length of 10...

Allan-N (approved):

InterLinked1 (approved):

KB4MDD (approved):

PR #477, opened 2025-01-30
chan_voter.c: finddelim update to length of strp from str

mkmer:
Check the correct length to prevent over running the output pointer.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -788,9 +788,9 @@ static int16_t deemp1(int16_t input, int32_t * restrict state0)
  * \param limit		Maximum number of substrings to process.
  * \return			Count of strings.
  */
-static int finddelim(char *str, char *strp[], int limit)
+static int finddelim(char *str, char *strp[], size_t limit)
 {
-	int i, l, inquo;
+	int i, inquo;
 
 	inquo = 0;
 	i = 0;
@@ -799,7 +799,7 @@ static int finddelim(char *str, char *strp[], int limit)
 		strp[0] = 0;
 		return 0;
 	}
-	for (l = 0; *str && (l < limit); str++) {
+	for (; *str && (i < (limit - 1)); str++) {
 		if (*str == QUOTECHR) {
 			if (inquo) {
 				*str = 0;
@@ -811,7 +811,6 @@ static int finddelim(char *str, char *strp[], int limit)
 		}
 		if ((*str == DELIMCHR) && (!inquo)) {
 			*str = 0;
-			l++;
 			strp[i++] = str + 1;
 		}
 	}
@@ -2509,7 +2508,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 		val = (char *) ast_variable_retrieve(cfg, (char *) data, "streams");
 		if (val) {
 			cp = ast_strdup(val);
-			p->nstreams = finddelim(cp, p->streams, MAXSTREAMS);
+			p->nstreams = finddelim(cp, p->streams, ARRAY_LEN(p->streams));
 		}
 		val = (char *) ast_variable_retrieve(cfg, (char *) data, "txctcss");
 		if (val) {
@@ -2538,7 +2537,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 			if (!cp) {
 				return NULL;
 			}
-			j = finddelim(cp, strs, 2);
+			j = finddelim(cp, strs, ARRAY_LEN(strs));
 			if (j < 2) {
 				ast_log(LOG_ERROR, "Channel %s: primary not specified properly\n", ast_channel_name(tmp));
 			} else {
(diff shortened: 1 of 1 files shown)


channels/chan_voter.c
```diff
@@ -2538,7 +2541,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 			if (!cp) {
 				return NULL;
 			}
-			j = finddelim(cp, strs, 2);
```
mkmer, line 2541: This one is a bit questionable.  2 means we only look through the first 2 elements of `cp`.  With the change we will look through all of `cp`... maybe it right?

channels/chan_voter.c
```diff
@@ -788,7 +788,7 @@ static int16_t deemp1(int16_t input, int32_t * restrict state0)
  * \param limit		Maximum number of substrings to process.
  * \return			Count of strings.
  */
-static int finddelim(char *str, char *strp[], int limit)
+static int finddelim(char *str, char *strp[], type_t limit)
```
KB4MDD, line 791: Should this be type_t?  In other places you used size_t.

mkmer: yes it should... missed it.

channels/chan_voter.c
Allan-N: Wow, I just noticed something else about the `for` loop.  We are iterating using `l`, checking `(l < limit)`, and then using `i` as the index into `strp`.

As far as passing "2", the idea is to ensure that there is only one value.  The first value goes into slot 0.  When we find DELIMCHR we move to the next slot, stop the loop, zero out slot 1, and return "2".  The caller checks if the number of returned values is less than 2.

Also, in this case we don't care whether `strs[1]` points to the second value or NULL.

Allan-N: Wow, I just noticed something else about the `for` loop.  We are iterating using `l`, checking `(l < limit)`, and then using `i` as the index into `strp`.

As far as passing "2", the idea is to ensure that there is only one value.  The first value goes into slot 0.  When we find DELIMCHR we move to the next slot, stop the loop, zero out slot 1, and return "2".  The caller checks if the number of returned values is less than 2.

Also, in this case we don't care whether `strs[1]` points to the second value or NULL.

channels/chan_voter.c
```diff
@@ -5107,7 +5110,7 @@ static int reload(void)
 		val = (char *) ast_variable_retrieve(cfg, (char *) data, "thresholds");
 		if (val) {
 			cp = ast_strdup(val);
-			p->nthresholds = finddelim(cp, strs, MAXTHRESHOLDS);
```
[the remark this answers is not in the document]
Allan-N: The problem here is that we are expanding the values into `strs`, then manipulating each value a bit, and storing the results into `p->linger_thresh[i]`. `p->count_thresh[i]`, and `p->rssi_thresh[i]`.  Those all appear to be sized to MAXTHRESHOLDS.  The value we want to pass to finddelim should be `MIN(ARRAY_LEN(strs), MAXTHRESHOLDS)`.

channels/chan_voter.c
```diff
@@ -799,7 +799,7 @@ static int finddelim(char *str, char *strp[], int limit)
 		strp[0] = 0;
 		return 0;
 	}
-	for (l = 0; *str && (l < limit); str++) {
+	for (l = 0; *str; str++) {
```
Allan-N, line 802: You removed the ` && (l < limit)` here and replaced it with the `if (i >= limit - 1)` below.  That's fine but now the `l` variable is really not used.

channels/chan_voter.c
```diff
@@ -2565,7 +2568,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 		val = (char *) ast_variable_retrieve(cfg, (char *) data, "thresholds");
 		if (val) {
 			cp = ast_strdup(val);
-			p->nthresholds = finddelim(cp, strs, MAXTHRESHOLDS);
+			p->nthresholds = finddelim(cp, strs, ARRAY_LEN(strs));
```
Allan-N, line 2571: Need to use `MIN(ARRAY_LEN(strs), MAXTHRESHOLDS)` to ensure that we don't overwrite `strs` and that we don't overwrite `p->linger_threads`, ...

channels/chan_voter.c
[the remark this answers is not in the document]
Allan-N: MIN(ARRAY_LEN(strs), MAXTHRESHOLDS).

Allan-N (changes requested):

channels/chan_voter.c
mkmer: Should we do that here, or in the for loop that actually has to stop?
How about ARRAY_LEN(p->linger_thresh)

Allan-N (approved):

channels/chan_voter.c
```diff
@@ -2565,7 +2564,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 		val = (char *) ast_variable_retrieve(cfg, (char *) data, "thresholds");
 		if (val) {
 			cp = ast_strdup(val);
-			p->nthresholds = finddelim(cp, strs, MAXTHRESHOLDS);
+			p->nthresholds = finddelim(cp, strs, MIN(ARRAY_LEN(strs),ARRAY_LEN(p->linger_thresh)));
```
InterLinked1, line 2567: missing space

channels/chan_voter.c
```diff
@@ -5107,7 +5106,7 @@ static int reload(void)
...
```
InterLinked1, line 5109: missing space

InterLinked1 (changes requested):

InterLinked1 (approved):

tsawyer:
Yay! chan_voter gets some love. Anxious to see this in the next release.

PR #480, opened 2025-01-30
radio-tune-menu.c: update explode_string()

mkmer:
Update explode_string to limit strp length.

### DIFF
--- a/utils/radio-tune-menu.c
+++ b/utils/radio-tune-menu.c
@@ -70,6 +70,8 @@
 #include <errno.h>
 #include <sys/wait.h>
 
+#include "asterisk/utils.h"
+
 /*! \brief type of signal detection used for carrier (cd) or ctcss (sd) */
 static const char * const cd_signal_type[] = {"no", "dsp", "vox", "usb", "usbinvert", "pp", "ppinvert"};
 static const char * const sd_signal_type[] = {"no", "usb", "usbinvert", "dsp", "pp", "ppinvert"};
@@ -124,9 +126,9 @@ static void ourhandler(int sig)
  *
  * \retval 			Returns number of substrings found.
  */
-static int explode_string(char *str, char *strp[], int limit, char delim, char quote)
+static int explode_string(char *str, char *strp[], size_t limit, char delim, char quote)
 {
-	int i, l, inquo;
+	int i, inquo;
 
 	inquo = 0;
 	i = 0;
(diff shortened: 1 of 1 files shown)


Allan-N (approved):

utils/radio-tune-menu.c
```diff
@@ -151,7 +151,6 @@ static int explode_string(char *str, char *strp[], int limit, char delim, char q
 		}
 		if ((*str == delim) && (!inquo)) {
 			*str = 0;
-			l++;
 			strp[i++] = str + 1;
 		}
 	}
```
Allan-N, line 156: The `strp[i] = 0` line below needs to be removed.

Allan-N (changes requested):

utils/radio-tune-menu.c
mkmer: I thought we wanted to set the end to 0?

Allan-N: This goes back to the 2 usage cases (for loops vs. while loops).  If we go with the for loops then there's no need for reserving/zapping the last slot being NULL'd.  I was just going to write that we'd be OK with just the for loop style.  But, the comments describing the arguments says that NULL will be placed at the end of the list.  So, the NULL'ing needs to stay but the `(i < limit)` test needs to become `(i < limit - 1)`.

Allan-N (approved):

utils/radio-tune-menu.c
```diff
@@ -69,6 +69,7 @@
 #include <signal.h>
 #include <errno.h>
 #include <sys/wait.h>
+#include <asterisk/utils.h>
```
InterLinked1, line 72: I'd add a blank line before this too
Also, does "asterisk/utils.h" not work? I don't think it should need to be referenced as a system include file

utils/radio-tune-menu.c
```diff
@@ -59,7 +59,7 @@
  * Most of these commands take optional parameters to set values.
  *
  */
-
+#include "asterisk.h"
```
InterLinked1, line 62: Blank line after asterisk.h
Also, curious, why is it needed for an external program?

utils/radio-tune-menu.c
mkmer: utils.h provides `ARRAY_LEN` - is there another way to get it without copying the define into this file?

utils/radio-tune-menu.c
mkmer: I originally added this for `ARRAY_LEN` - it didn't work and I forgot to remove it.  Good call

utils/radio-tune-menu.c
InterLinked1: I mean using `"asterisk/utils.h"`, not `<asterisk/utils.h>`. I don't think the latter is needed

InterLinked1 (approved):

PR #488, opened 2025-02-13
Update logger.conf: Add debug and verbose to console

KB4MDD:
This updates logger.conf to add debug and verbose messages to the console.  This will help users experiencing problem to easily see debug and verbose messages.

### DIFF
--- a/configs/rpt/logger.conf
+++ b/configs/rpt/logger.conf
@@ -15,6 +15,6 @@ dateformat=%F %T.%3q   ; with milliseconds
 ; Special filename "console" represents the system console
 ;
 ;debug.log => debug
-console => notice,warning,error,dtmf
-;console => notice,warning,error,debug,verbose
+console => notice,warning,error,dtmf,debug,verbose
+;console => notice,warning,error,dtmf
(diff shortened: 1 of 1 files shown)


configs/rpt/logger.conf
```diff
@@ -15,6 +15,6 @@ dateformat=%F %T.%3q   ; with milliseconds
 ; Special filename "console" represents the system console
 ;
 ;debug.log => debug
-console => notice,warning,error,dtmf
+console => notice,warning,error,dtmf,debug,verbose
```
mkmer, line 18: should we just remove the commented line following this, or should we swap the comment `;` lines and leave the original commented out?

mkmer (approved):

InterLinked1 (commented):
I think this is fine if you don't want to change it, though my nitpick would be to remove "Update" from the commit message title (so it becomes `logger.conf: Add debug and verbose to console log`). All commits are updates, stating that a file is being updated isn't needed beyond the file/module itself. I think GitHub might add it by default if you do it online so that might be something to watch for. Otherwise, looks good!

PR #489, opened 2025-02-13
Do not default to inverted PTT

N4IRS:
Should this really be inverted? Unless I missed it, I see no way in radio-tune-menu to swap the PTT logic

### DIFF
--- a/configs/rpt/usbradio.conf
+++ b/configs/rpt/usbradio.conf
@@ -167,7 +167,7 @@ txhpf = 0                   ; Transmitter Audio High Pass Filter 0,1,2
                             ; 2 - 120 Hz cutoff for special applications requiring additional bass response in transmitted audio.
                             ; Not recommended due to the increased possibility of voice energy interfering with sub-audible signaling
 
-invertptt = yes             ; Invert PTT: no,yes
+invertptt = no              ; Invert PTT: no,yes
                             ; no  - ground to transmit
                             ; yes - open to transmit
 
--- a/configs/samples/usbradio.conf.sample
+++ b/configs/samples/usbradio.conf.sample
@@ -153,7 +153,7 @@
                             ; 2 - 120 Hz cutoff for special applications requiring additional bass response in transmitted audio.
                             ; Not recommended due to the increased possibility of voice energy interfering with sub-audible signaling
 
-;invertptt = yes            ; Invert PTT: no,yes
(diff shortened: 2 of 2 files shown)


Allan-N:
There's one issue here (the radio-tune-menu not having a way to change the setting) and one question (should we change the default value for "invertptt").

IMHO, we should fix the former.

For the latter, the default value if not specified in the .conf file would be 0 ("No").  Changing the .conf file to match makes some sense but I am concerned about changing the value we've been including in the .conf since long ago.  Why?  because new configurations with the same edits (per whatever documentation folks find) might yield unexpected results.

Allan-N:
>IMHO, we should fix the former.

Looking at the code a bit more and I'm wondering what to do here.  I found the USBRadio channel driver reads 57 variables from the configuration but will only update 20 of the values.  Do we limit the "tune" menu to only make "tuning" changes and leave the "configuration" to other tools (e.g. your favorite editor or some new/not-yet-written configuration utility)?

For reference: the SimpleUSB channel driver will read 17 values and write 13.

N4IRS:
> There's one issue here (the radio-tune-menu not having a way to change the setting) and one question (should we change the default value for "invertptt").
> 
> IMHO, we should fix the former.
> 
> For the latter, the default value if not specified in the .conf file would be 0 ("No"). Changing the .conf file to match makes some sense but I am concerned about changing the value we've been including in the .conf since long ago. Why? because new configurations with the same edits (per whatever documentation folks find) might yield unexpected results.

The change from no to yes is rather recent (at least post ASL2) I ran into this when I told asl-menu I wanted to use USBRadio. Since my PTT was active low, The radio was locked into transmit. Since the radio-tune-menu did not have the option to invert, I had to shutdown asterisk while I edited the config. Since USBRadio thought the transmitter was not on, none of the times would have save me.

Allan-N:
Looks like the change was made with [PR #73](https://github.com/AllStarLink/app_rpt/pull/73)

Allan-N (approved):

Allan-N (changes requested):
Can you also update .../configs/samples/usbradio.conf.sample

N4IRS:
> Can you also update .../configs/samples/usbradio.conf.sample

Sure

KB4MDD (approved):

Allan-N (approved):

mkmer (approved):
LGTM

Issue #485, opened 2025-02-11
Can we move away from .wav files?


Allan-N:
The app_rpt code creates/uses .wav files for stored audio.  This choice isn't very "Mac" friendly.  Can we look into using (or at least offering) another format?

InterLinked1:
I think there are a few things to unpack here.

The files generated by a telephony system (or consumed by one) typically aren't meant to be interacted with by end users directly. It sounds like the files you are talking about might be intended for that purpose though.

WAV files are pretty standard, and can be played by pretty much everything since it's just raw audio. If it isn't Mac friendly, that sounds like a Mac problem specifically and something that should be addressed there (I'm honestly surprised none of the Mac software would support it). It's possible maybe it's something else here in a wav container, but I don't know the specifics here.

If this is the default now, I don't think the format should be changed, but perhaps the user could be allowed to control it. Asterisk supports a variety of formats. Probably the most versatile solution is to allow the user to specify the filename, in a similar manner to usage of MixMonitor, where it will then use that format. You could then, say, save as mp3 if you wanted to (not that you would want to!)

Allan-N:
>It sounds like the files you are talking about might be intended for that purpose though.

The files I'm talking about are those associated with the "archivedir" setting.

>If it isn't Mac friendly, that sounds like a Mac problem

AFAIK, the issue has been a Microsoft licensing thing and it's been a problem for a very long time.  In short, you can't play .wav files on a Mac without first converting the file to a supported format.

>perhaps the user could be allowed to control it.

Yes, being able to specify an alternate format (e.g. .mp3) would be great.  Just need a new setting to specify the archive format :-)

W6HBR:
I play WAV files on my Mac all the time.  Both Apple Music and Quicktime player will play WAV files natively on a Mac and both apps are installed by default.  WAV files are probably the most universal format around that every platform supports.

Since the archivedir files are separated by keyup, I like to use VLC media player to play a series of them together to be able to listen to a full QSO or net.  VLC is available for Mac, Windows, Linux, et al.

InterLinked1:
> > It sounds like the files you are talking about might be intended for that purpose though.
> 
> The files I'm talking about are those associated with the "archivedir" setting.

What are these files actually intended for? Is it expected users would copy them over and listen to them?

> > If it isn't Mac friendly, that sounds like a Mac problem
> 
[...]

PR #492, opened 2025-02-15
Allow replacement of GSM/WAV49 audio archives with other formats

jxmx:
Addresses #485

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2959,6 +2959,7 @@ static inline int rpt_any_hangups(struct rpt *myrpt)
 static inline void log_keyed(struct rpt *myrpt)
 {
 	char mydate[100], myfname[512];
+	const char *myformat;
 	time_t myt;
 
 	if (myrpt->monstream) {
@@ -2971,7 +2972,9 @@ static inline void log_keyed(struct rpt *myrpt)
 		time(&myt);
 		strftime(mydate, sizeof(mydate) - 1, "%Y%m%d%H%M%S", localtime(&myt));
 		sprintf(myfname, "%s/%s/%s", myrpt->p.archivedir, myrpt->name, mydate);
-		myrpt->monstream = ast_writefile(myfname, "wav49", "app_rpt Air Archive", O_CREAT | O_APPEND, 0, 0644);
+		myformat = myrpt->p.archiveformat ? myrpt->p.archiveformat : "wav49";
+		myrpt->monstream =
+			ast_writefile(myfname, myformat, "app_rpt Air Archive", O_CREAT | O_APPEND, 0, 0644);
 		if (myrpt->p.monminblocks) {
 			blocksleft = diskavail(myrpt);
 			if (blocksleft >= myrpt->p.monminblocks) {
@@ -3792,18 +3795,19 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 			if (myrpt->p.archivedir) {
 				if (myrpt->p.duplex < 2) {
 					char myfname[512], mydate[100];
+					const char *myformat;
 					long blocksleft;
 					time_t myt;
 
 					time(&myt);
 					strftime(mydate, sizeof(mydate) - 1, "%Y%m%d%H%M%S", localtime(&myt));
 					sprintf(myfname, "%s/%s/%s", myrpt->p.archivedir, myrpt->name, mydate);
+					myformat = myrpt->p.archiveformat ? myrpt->p.archiveformat : "wav49";
 					if (myrpt->p.monminblocks) {
 						blocksleft = diskavail(myrpt);
 						if (blocksleft >= myrpt->p.monminblocks) {
 							myrpt->monstream =
-								ast_writefile(myfname, "wav49", "app_rpt Air Archive", O_CREAT | O_APPEND, 0,
-											  0644);
+								ast_writefile(myfname, myformat, "app_rpt Air Archive", O_CREAT | O_APPEND, 0, 0644);
 						}
 					}
 				}
@@ -7081,16 +7085,18 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		b1 = b;
 	if (myrpt->p.archivedir) {
 		char mycmd[512], mydate[100];
+		const char *myformat;
 		time_t myt;
 		long blocksleft;
 
 		mkdir(myrpt->p.archivedir, 0700);
-		sprintf(mycmd, "%s/%s", myrpt->p.archivedir, myrpt->name);
+		snprintf(mycmd, sizeof(mycmd), "%s/%s", myrpt->p.archivedir, myrpt->name);
 		mkdir(mycmd, 0775);
 		time(&myt);
-		strftime(mydate, sizeof(mydate) - 1, "%Y%m%d%H%M%S", localtime(&myt));
-		sprintf(mycmd, "mixmonitor start %s %s/%s/%s.wav49 a", ast_channel_name(chan), myrpt->p.archivedir, myrpt->name,
-				mydate);
+		strftime(mydate, sizeof(mydate), "%Y%m%d%H%M%S", localtime(&myt));
+		myformat = myrpt->p.archiveformat ? myrpt->p.archiveformat : "wav49";
+		snprintf(mycmd, sizeof(mycmd), "mixmonitor start %s %s/%s/%s.%s a",
+			ast_channel_name(chan), myrpt->p.archivedir, myrpt->name, myformat, mydate);
 		if (myrpt->p.monminblocks) {
 			blocksleft = diskavail(myrpt);
 			if (myrpt->p.remotetimeout) {
@@ -7100,7 +7106,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 				ast_cli_command(nullfd, mycmd);
 		} else
 			ast_cli_command(nullfd, mycmd);
-		sprintf(mycmd, "CONNECT,%s", b1);
+		snprintf(mycmd, sizeof(mycmd), "CONNECT,%s", b1);
 		donodelog(myrpt, mycmd);
 		rpt_update_links(myrpt);
 		doconpgm(myrpt, b1);
--- a/configs/rpt/rpt.conf
+++ b/configs/rpt/rpt.conf
@@ -184,8 +184,35 @@ parrottime = 1000                   ; Set the amount of time in milliseconds
 									; module app_gps.so will need to be enabled
 
 ; *** Status Reporting ***
-; Uncomment the statpost line to report the status of your node to stats.allstarlink.org
+;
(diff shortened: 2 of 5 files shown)


apps/app_rpt.c
```diff
@@ -2962,7 +2962,7 @@ static inline void log_keyed(struct rpt *myrpt)
 		time(&myt);
 		strftime(mydate, sizeof(mydate) - 1, "%Y%m%d%H%M%S", localtime(&myt));
 		sprintf(myfname, "%s/%s/%s", myrpt->p.archivedir, myrpt->name, mydate);
-		myrpt->monstream = ast_writefile(myfname, "wav49", "app_rpt Air Archive", O_CREAT | O_APPEND, 0, 0644);
+		myrpt->monstream = ast_writefile(myfname, "wav", "app_rpt Air Archive", O_CREAT | O_APPEND, 0, 0644);
```
Allan-N, line 2965: While I support this change I will ask whether this going to [significantly] change the size of the archived audio files?

InterLinked1 (changes requested):
Code wise this seems fine, but how about a more useful commit message?

```
app_rpt: Use wav for recordings instead of wav49

Resolves #485
```
or something like that?

apps/app_rpt.c
jxmx: Yes, but I'm not sure there's a better format to choose if we want "native" tools to work with the archive data. Ideally you could do an MP3 or OGG file but that isn't supported. From looking at the code, we can pick from what's in `core show file format`. Nothing else on that list would be generally readable by Windows and Mac commonly-installed software I don't believe.

InterLinked1: mp3 can be supported (I did some work to patch the mp3 support in Asterisk to allow this for a client), but the problem with any compressed formats is it increases the transcoding load, hence why raw formats like wav tend to get used.

Obviously the question there is whether to optimize for space or CPU, and I think the latter is more common on production systems but I could see the value in the other way as well. "Compatibility with consumer devices" is certainly another factor and not one I usually see come up. For that, I agree you can't go wrong with wav

InterLinked1:
> Code wise this seems fine, but how about a more useful commit message?
> 
> ```
> app_rpt: Use wav for recordings instead of wav49
> 
> Resolves #485
> ```
> 
> or something like that?

Slightly better but still not in the suggested format. The `app_rpt` prefix and the reference to the issue number are critical.
Also, it's GSM, not GMS.

apps/app_rpt.c
jxmx: If the size of WAV files become a real problem, we could create a systemd.timer that compresses any WAV files older than 5 minutes.

Allan-N: @jxmx `asterisk -x 'module load format_mps'` 😄 

One suggestion made in the issue discussion was to add a configuration file setting that would specify the file format.  By adding that we can keep the current .wav49 and, at the same time, allow other formats (.wav, .mp3, ...)

jxmx: Adding a configuration option to change the filetype seems unnecessary given that `archivedir` isn't a commonly-used command to begin with that I see. My understanding of why MP3 and OGG isn't loaded by default is there's problem with them, perhaps performance-related? If we're going to change this to mp3, then we'll also need to change `modules.conf` **and** we'll have to add logic to `asl3-asterisk.postinst` to edit customized `modules.conf` to load mp3.

InterLinked1: Allan brings up a good point. Now would be as easy a time as ever to just do away with a hardcoded format and turn it into a variable that can be set from the config. Defaulting to wav would make sense, but users could easily override it.

Allan-N: My $0.02 :
- make the format choice a configuration variable
- have the new "archiveformat" default value continue to be "wav49"

This way, all remains the same for running configurations yet folks will have the ability to change the format to one that works better for them.  And, if folks do change the format then they can deal with any increased size, CPU requirements, compressioin issues, etc.,

mkmer:
Maybe while tinkering right here we should fix the file name generation to `snprintf`?
https://github.com/AllStarLink/app_rpt/blob/b5b110a7bfbefe00037cb86eaf0376dcb1fe2cd1/apps/app_rpt.c#L2964

apps/app_rpt.c
InterLinked1: I think slin is technically the "raw format" but the PCM formats (ulaw/alaw) are pretty close I believe.
`core show translations` shows some of the appx. costs in converting between different formats

ulaw/alaw are raw telephony formats and wav isn't compressed so that's why these are usually good choices. gsm and mp3 are compressed and lossy, so they are more computationally intensive.

Allan-N:
For those following this PR, I had my hopes up that we might be able to create "mp3" audio archives.  Sadly :
```
[2025-02-19 16:42:48.386] ERROR[193754]: format_mp3.c:277 mp3_rewrite: I Can't write MP3 only read them.
[2025-02-19 16:42:48.386] WARNING[193754]: file.c:512 fn_wrapper: Unable to rewrite format mp3
[2025-02-19 16:42:48.386] WARNING[193754]: file.c:1514 ast_writefile: Unable to rewrite /var/spool/asterisk/monitor/491301/20250219164248.mp3
```

InterLinked1:
> For those following this PR, I had my hopes up that we might be able to create "mp3" audio archives. Sadly :
> 
> ```
> [2025-02-19 16:42:48.386] ERROR[193754]: format_mp3.c:277 mp3_rewrite: I Can't write MP3 only read them.
> [2025-02-19 16:42:48.386] WARNING[193754]: file.c:512 fn_wrapper: Unable to rewrite format mp3
> [2025-02-19 16:42:48.386] WARNING[193754]: file.c:1514 ast_writefile: Unable to rewrite /var/spool/asterisk/monitor/491301/20250219164248.mp3
> ```

The ability exists, but the MP3 write patch isn't merged upstream yet - see Asterisk PR 665 / 704

Allan-N:
I have updated the PR to allow you to specify the audio recording format using the new  `archiveformat` variable.  We still default to `wav49` but you now has options.

... and `archivedir` and `archiveformat` are now documented in `rpt.conf`

configs/rpt/rpt.conf
```diff
@@ -184,9 +184,36 @@ parrottime = 1000                   ; Set the amount of time in milliseconds
...
+; *** Audio Archiving ***
+;
+; The following "archivedir" line can be used to enable a simple log and
+; audio recorder of the activity on a node.  When enabled, a series of
+; recordings, one for each active COR on the node.  The file(s) will be
+; named with the date and time down to the 1/100th of a second.  This
+; logging can be useful in debugging, policing, or other creative things.
+;
+; The "archiveformat" line can be used to specify the format of the audio
+; recordings.  By default, the format will be "wav49" (GSM in a WAV|wav49
```
InterLinked1, line 204: Just say WAV instead of WAV|wav49 (wav49 is gsm in a wav wrapper)

apps/app_rpt.c
```diff
@@ -2971,7 +2971,12 @@ static inline void log_keyed(struct rpt *myrpt)
 		time(&myt);
 		strftime(mydate, sizeof(mydate) - 1, "%Y%m%d%H%M%S", localtime(&myt));
 		sprintf(myfname, "%s/%s/%s", myrpt->p.archivedir, myrpt->name, mydate);
-		myrpt->monstream = ast_writefile(myfname, "wav49", "app_rpt Air Archive", O_CREAT | O_APPEND, 0, 0644);
+		myrpt->monstream = ast_writefile(myfname,
+										 myrpt->p.archiveformat ? myrpt->p.archiveformat : "wav49",
```
InterLinked1, line 2975: These seem to be indented several levels, I think it should only be one level of indentation further?
Same with a few other areas

Allan-N: The line was getting way too long.  Now, I have all of the args lining up.

InterLinked1: I see that, but typically we just indent 1 tab further in that case. This seems to be indented 8 tabs further. Not sure if that is just an artifact of the GitHub code viewer but it seems to be way off to the right now.

Allan-N: I think this is [partially] an artifact of the GitHub viewer.  It looks clean in a text editor (vim) with tabstop=4.  One can argue about whether to indent with a single tab or have the args line up.  My goal is readability.

InterLinked1: I just checked out your PR and opened it in nano and vim and I see the same thing:

image

Lining up is fine but I don't see why every line except for the first is indented *waaaaaaayyyy* to the right. Was that intentional?

This would be the Asterisk style for longer lines like that (one of the two):

```
x = function(arg1, arg2,
<TAB>arg3, arg4, arg5...)

x = function(arg1,
<TAB>arg2,
<TAB>arg3,
etc.
```

mkmer (approved):

apps/app_rpt.c
Allan-N: I've changed (and reworked) the formatting.  I don't really like it but, hopefully, this time all will be OK.

apps/app_rpt.c
```diff
@@ -2971,7 +2972,9 @@ static inline void log_keyed(struct rpt *myrpt)
 		time(&myt);
 		strftime(mydate, sizeof(mydate) - 1, "%Y%m%d%H%M%S", localtime(&myt));
 		sprintf(myfname, "%s/%s/%s", myrpt->p.archivedir, myrpt->name, mydate);
-		myrpt->monstream = ast_writefile(myfname, "wav49", "app_rpt Air Archive", O_CREAT | O_APPEND, 0, 0644);
+		myformat = myrpt->p.archiveformat ? myrpt->p.archiveformat : "WAV";
```
InterLinked1, line 2975: I thought you were preserving wav49 as the default? (Personally, I wouldn't mind wav either, but just want to confirm if that's what you intended)

Also a handy macro you could use if you wanted: `myformat = S_OR(myrpt->p.archiveformat, "WAV");`

But you don't have to do that, those are just optional helper macros

Allan-N: Too much confusion about .wav49, .wav, .WAV.  I do want to preserve .wav49 ... but also know that the asterisk code changes .wav49 to .WAV.  I'll switch this back ... so we're sure what we're getting.

InterLinked1: Yes, wav49 indicates it will be gsm wrapped in wav, but the file extension is always wav. Providing wav means something differently.
See note number 2 here: https://docs.asterisk.org/Latest_API/API_Documentation/Dialplan_Applications/MixMonitor/

InterLinked1 (approved):

apps/app_rpt.c
Allan-N: Yup.  It was the _Asterisk will silently convert the extension to '.WAV'_ comment that had me confused (like was .wav49 the legacy name and, instead, one should now use .WAV)

Allan-N (approved):

Issue #481, opened 2025-02-01
cop,48 produces choppy and broken telemetry


AI3I:
**Description**
cop,48 in app_rpt does not produce consistent audio fidelity with tone playback, either with single tones, mixed tones/tone sequences, or pre-programmed tones, i.e. dual-tone multi-frequency (DTMF) sequences.  The bug seems to manifest as though it has a buffering issue; it may, for example, produce one or two DTMF digits from a named sequence and not produce the rest, or skip through sequences, or hold on and play some tone sequences endlessly.

**Reproduction**
Two examples that 'stress' cop,48 behavior:
`node504381*CLI> rpt cmd 504381 cop 48 !830+827/450,!0/250,!659+656/450,!0/250,!739+736/450,!0/250,!493+490/450,!0/750,!493+490/450,!0/250,!739+736/450,!0/250,!830+827/45 0,!0/250,!659+656/450`  ...should emulate a Computer Automated Technology CAT-1000 grandfather clock feature, but produces clipped audio and completes one or two tone sequences before playing what remains in its buffer endlessly.  Results vary.

`node504381*CLI> rpt cmd 504381 cop 48 6,5,4,5,6,6,6,5,5,5,6,#,#`  ...should whimsically play 'Mary Had a Little Lamb' in DTMF, but produces 6, 5, may skip other digits, perhaps #, and stops.  Results vary.

Playing multiple sequences may even return chan_simpleusb errors, with Asterisk abruptly restarting:
```
[2025-01-31 21:19:29.073] WARNING[1049]: chan_simpleusb.c:2073 simpleusb_read: Possibly stuck USB read channel. [504381]
[2025-01-31 21:19:29.135] WARNING[1049]: chan_simpleusb.c:2084 simpleusb_read: USB read channel [504381] was not stuck.
```

Additional logging during observed behavior (from /var/log/asterisk/messages.log):
```
[2025-01-31 20:36:07.311] Asterisk 20.11.0+asl3-3.2.0-2.deb12 built by builder @ allstarlink.org on a aarch64 running Linux on 2024-12-20 20:30:41 UTC
[2025-01-31 20:36:07.404] NOTICE[883] loader.c: 70 modules will be loaded.
[2025-01-31 20:36:07.930] NOTICE[883] cdr.c: CDR simple logging enabled.
[2025-01-31 20:36:07.937] NOTICE[883] dnsmgr.c: Managed DNS entries will be refreshed every 300 seconds.
[...]

PR #493, opened 2025-02-16
rpt_telemetry: Update PAGE command to use correct channel.

mkmer:
Addressing issue #481 
Looks like we are pointing to the wrong channel.

### DIFF
--- a/apps/app_rpt/rpt_functions.c
+++ b/apps/app_rpt/rpt_functions.c
@@ -1592,7 +1592,7 @@ int function_cop(struct rpt *myrpt, char *param, char *digitbuf, int command_sou
 				strcat(cp, ",");
 			strcat(cp, argv[i]);
 		}
-		rpt_telemetry(myrpt, PAGE, cp);
+		rpt_telemetry(myrpt, PAGE, cp); /* cp is passed to rpt_telem_thread where it is free'd after use */
 		return DC_COMPLETE;
 
 	case 49:					/* Disable Incoming connections */
@@ -1686,7 +1686,7 @@ int function_cop(struct rpt *myrpt, char *param, char *digitbuf, int command_sou
 		}
 		ast_copy_string(mdcp->type, argv[1], sizeof(mdcp->type));
 		mdcp->UnitID = (short) strtol(argv[2], NULL, 16);
-		rpt_telemetry(myrpt, MDC1200, (void *) mdcp);
+		rpt_telemetry(myrpt, MDC1200, (void *) mdcp); /* mdcp is passed to rpt_telem_thread where it is free'd after use */
 		return DC_COMPLETE;
 #endif
 	case 61:					/* send GPIO change */
--- a/apps/app_rpt/rpt_telemetry.c
+++ b/apps/app_rpt/rpt_telemetry.c
@@ -1057,13 +1057,20 @@ void *rpt_tele_thread(void *this)
 	/* make a conference for the tx */
 	/* If the telemetry is only intended for a local audience, only connect the ID audio to the local tx conference so linked systems can't hear it */
 	/* first put the channel on the conference in announce mode */
-	if (rpt_conf_add(mychannel,
-					 myrpt,
-					 mytele->mode == ID1 || mytele->mode == PLAYBACK || mytele->mode == TEST_TONE ||
-							 mytele->mode == STATS_GPS_LEGACY
-						 ? RPT_CONF
-						 : RPT_TELECONF,
-					 RPT_CONF_CONFANN)) {
+	int mode;
+
+	switch (mytele->mode) {
+		case ID1:
+		case PLAYBACK:
+		case TEST_TONE:
+		case STATS_GPS_LEGACY:
+			mode = RPT_CONF;
+			break;
+		default:
+			mode = RPT_TELECONF;
+			break;
+	}
+	if (rpt_conf_add(mychannel, myrpt, mode, RPT_CONF_CONFANN)) {
 		rpt_mutex_lock(&myrpt->lock);
 		goto abort;
 	}
--- a/apps/app_rpt/rpt_telemetry.c
+++ b/apps/app_rpt/rpt_telemetry.c
@@ -1057,9 +1057,13 @@ void *rpt_tele_thread(void *this)
 	/* make a conference for the tx */
 	/* If the telemetry is only intended for a local audience, only connect the ID audio to the local tx conference so linked systems can't hear it */
 	/* first put the channel on the conference in announce mode */
-	if (rpt_conf_add(mychannel, myrpt, mytele->mode == MDC1200 || mytele->mode == PAGE ||
-			mytele->mode == ID1 || mytele->mode == PLAYBACK || mytele->mode == TEST_TONE || 
-			mytele->mode == STATS_GPS_LEGACY ? RPT_CONF : RPT_TELECONF, RPT_CONF_CONFANN)) {
+	if (rpt_conf_add(mychannel,
+					 myrpt,
+					 mytele->mode == ID1 || mytele->mode == PLAYBACK || mytele->mode == TEST_TONE ||
+							 mytele->mode == STATS_GPS_LEGACY
+						 ? RPT_CONF
+						 : RPT_TELECONF,
+					 RPT_CONF_CONFANN)) {
 		rpt_mutex_lock(&myrpt->lock);
 		goto abort;
 	}
--- a/apps/app_rpt/rpt_telemetry.c
+++ b/apps/app_rpt/rpt_telemetry.c
@@ -1057,20 +1057,19 @@ void *rpt_tele_thread(void *this)
 	/* make a conference for the tx */
 	/* If the telemetry is only intended for a local audience, only connect the ID audio to the local tx conference so linked systems can't hear it */
 	/* first put the channel on the conference in announce mode */
-	int mode;
-
+	int type;
 	switch (mytele->mode) {
 		case ID1:
 		case PLAYBACK:
 		case TEST_TONE:
 		case STATS_GPS_LEGACY:
-			mode = RPT_CONF;
+			type = RPT_CONF;
 			break;
 		default:
-			mode = RPT_TELECONF;
+			type = RPT_TELECONF;
 			break;
 	}
-	if (rpt_conf_add(mychannel, myrpt, mode, RPT_CONF_CONFANN)) {
+	if (rpt_conf_add(mychannel, myrpt, type, RPT_CONF_CONFANN)) {
 		rpt_mutex_lock(&myrpt->lock);
 		goto abort;
 	}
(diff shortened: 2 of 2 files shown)


mkmer:
[debug.log](https://github.com/user-attachments/files/18815877/debug.log)
For history - debug log attached

mkmer:
BTW: I suspect this is wrong as well, but have no way to test.
https://github.com/AllStarLink/app_rpt/blob/b73bc4c630d4be301e4ff9875e6bf63d72fc5f82/apps/app_rpt/rpt_telemetry.c#L1171-L1174

KB4MDD:
I have to ask what is the difference between mychannel and txchannel?

Was paging broken?

mkmer:
COP 48 -> is paging. 
Best I can tell, `myrpt->txchannel` is not the conferenced channel this function is supposed to play in.  Every other instance in `rpt_tele_thread` is using `mychannel`.  This change DOES fix the problem (maybe not correctly?)

mkmer:
I'm not positive, but I don't think this is "right" - a bit out of my comfort zone here:
https://github.com/AllStarLink/app_rpt/blob/c857a1a0d88dcccdd3de2204ea2cbc2917ec8b2b/apps/app_rpt/rpt_telemetry.c#L1172

Is this a "hidden" malloc?
https://github.com/AllStarLink/app_rpt/blob/c857a1a0d88dcccdd3de2204ea2cbc2917ec8b2b/apps/app_rpt/rpt_telemetry.c#L1156

Same here:
https://github.com/AllStarLink/app_rpt/blob/c857a1a0d88dcccdd3de2204ea2cbc2917ec8b2b/apps/app_rpt/rpt_telemetry.c#L1149

Allan-N:
I gotta ask ... Without the proposed change, the PAGE code appears to be waiting on "mychannel" (an allocated pseudo-channel) and then playing the "page" to "myrpt->txchannel".  What's the difference between the two channels?  Is the "page" ONLY supposed to go to the "tx" channel?  and could we be waiting on the wrong channel?

mkmer:
The lock out is between `rpt_tele_thread` and `app_rpt` -> both trying to send "stuff" to the TX channel (I think anyway).
Since everything else in telemetry talks to the psudo channel, why wouldn't this? (As usual, I could easily be off mark)

Allan-N:
>Is this a "hidden" malloc?

No.  The allocation is here : https://github.com/AllStarLink/app_rpt/blob/master/apps/app_rpt/rpt_functions.c#L1595

mkmer:
> > Is this a "hidden" malloc?
> 
> No. The allocation is here :
> 
> https://github.com/AllStarLink/app_rpt/blob/c857a1a0d88dcccdd3de2204ea2cbc2917ec8b2b/apps/app_rpt/rpt_functions.c#L1584-L1595

Oh - so my proposed memory leak fix is hidden in the called function?  That seems a bit odd/difficult to follow.  
I think we should pull that out if this is true (go with my proposed free and remove the above ref one)
Thoughts?

Allan-N:
>so my proposed memory leak fix is hidden in the called function? That seems a bit odd/difficult to follow.
I think we should pull that out if this is true (go with my proposed free and remove the above ref one)
Thoughts?

If the intent is for the caller to say "here's everything you need to do what I want and I'm gonna move on" then it's essentially handing off the allocated memory and need not worry about it's disposal.  The alternative would be to assume that the rpt_telemetry() call either operates synchronously or that it will make a copy of any data it needs to retain.  I'd leave the code alone (having rpt_telemetry free any passed "data").  Of course, a few comments can't hurt 😄

KB4MDD:
So is mychannel associated with a conference that would cause this to go out to all linked repeaters?

If so, that could be the reason it is directed to the txchannel.

Allan-N:
>So is mychannel associated with a conference that would cause this to go out to all linked repeaters?

It looks like the conference (as currently being setup) is NOT limited to the local tx conference.

https://github.com/AllStarLink/app_rpt/blob/c857a1a0d88dcccdd3de2204ea2cbc2917ec8b2b/apps/app_rpt/rpt_telemetry.c#L1057-L1060

mkmer:
So - what's the right answer? Use `mychannel` and add it to the local conference or try to figure out the channel lock issue?

mkmer:
> > So is mychannel associated with a conference that would cause this to go out to all linked repeaters?
> 
> It looks like the conference (as currently being setup) is NOT limited to the local tx conference.
> 
> https://github.com/AllStarLink/app_rpt/blob/c857a1a0d88dcccdd3de2204ea2cbc2917ec8b2b/apps/app_rpt/rpt_telemetry.c#L1057-L1060

This code is a bit interesting (as it's not doing what we think?)
I generated a test where `rpt_conf_add(mychannel, myrpt,  RPT_TELECONF, RPT_CONF_CONFANN)` and one with `rpt_conf_add(mychannel, myrpt,  RPT_CONF , RPT_CONF_CONFANN)` -> neither send telemetry out links.

Debug shows DAHDI joining different conferences but not hearing a difference on the audio route.

Looking through conference setup...  I'm wondering if we should be using `myrpt->txpchannel` rather than `myrpt->txchannel`
https://github.com/AllStarLink/app_rpt/blob/c857a1a0d88dcccdd3de2204ea2cbc2917ec8b2b/apps/app_rpt.c#L2830

https://github.com/AllStarLink/app_rpt/blob/c857a1a0d88dcccdd3de2204ea2cbc2917ec8b2b/apps/app_rpt.c#L2777

mkmer:
My interpretation/understanding at this point :

- `myrpt->txchannel` -> the actual tx output channel.  In my case it's simpleusb on one system and usrp on the other.  Locked often by `app_rpt()`
- `myrpt->txpchannel' -> the psuedo channel used for tx?  Locked often by `app_rpt()`.
- `mychannel` is a psuedo channel created in `rpt_telem_thread()` and joined to a conference (Either `RPT_CONF` or  `RPT_TELECONF`.

In the first two cases (using given pointers), cop 48 1,2,3,4,5,6... will generate "already blocked by thread" debug messages and unreliably transmit tones.

apps/app_rpt/rpt_functions.c
```diff
@@ -1592,7 +1592,7 @@ int function_cop(struct rpt *myrpt, char *param, char *digitbuf, int command_sou
 				strcat(cp, ",");
 			strcat(cp, argv[i]);
 		}
-		rpt_telemetry(myrpt, PAGE, cp);
+		rpt_telemetry(myrpt, PAGE, cp); /* cp is eventually handed off rpt_telem_thread where it is freed after use */
```
Allan-N, line 1595: /* cp is passed to rpt_telem_thread where it is free'd after use */

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1057,7 +1057,9 @@ void *rpt_tele_thread(void *this)
 	/* make a conference for the tx */
 	/* If the telemetry is only intended for a local audience, only connect the ID audio to the local tx conference so linked systems can't hear it */
 	/* first put the channel on the conference in announce mode */
-	if (rpt_conf_add(mychannel, myrpt, mytele->mode == ID1 || mytele->mode == PLAYBACK || mytele->mode == TEST_TONE || mytele->mode == STATS_GPS_LEGACY ? RPT_CONF : RPT_TELECONF, RPT_CONF_CONFANN)) {
+	if (rpt_conf_add(mychannel, myrpt, mytele->mode == MDC1200 || 
+			mytele->mode == ID1 || mytele->mode == PLAYBACK || mytele->mode == TEST_TONE ||
```
Allan-N, line 1061: Why have you added `mytele->mode == MDC1200` here?  and not added `mytele->mode == PAGE`?

mkmer: Well - page was in there :(

apps/app_rpt/rpt_functions.c
mkmer: Codespell likes it both ways :)

Allan-N: Can you at least make it "... handed off TO ..."

mkmer: Sorry - I missed most of that.  Should be good now.

InterLinked1 (commented):
I don't really have the expertise to comment here, but I think the most important thing though here is testing. It sounds like mkmer has already tested it, have multiple people validated the old way didn't work properly and the new way works on their systems?

Allan-N:
Can you comment on `myrpt->txchannel` vs. `mychannel`? and `RPT_CONF` vs. `RPT_TELECONF`?  and whether/how to control whether the transmitted tones are sent only locally vs. locally + downstream?

Allan-N:
And here is what I believe to be the answer to one of my questions :

https://github.com/AllStarLink/app_rpt/blob/ce12077a9c0e6da973fdb13f26afb43678d6d7b8/apps/app_rpt.c#L2773-L2777
https://github.com/AllStarLink/app_rpt/blob/ce12077a9c0e6da973fdb13f26afb43678d6d7b8/apps/app_rpt.c#L2808-L2809

I do think we want to be using the RPT_TELECONF conference

Allan-N (approved):

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1057,7 +1057,9 @@ void *rpt_tele_thread(void *this)
 	/* make a conference for the tx */
 	/* If the telemetry is only intended for a local audience, only connect the ID audio to the local tx conference so linked systems can't hear it */
 	/* first put the channel on the conference in announce mode */
-	if (rpt_conf_add(mychannel, myrpt, mytele->mode == ID1 || mytele->mode == PLAYBACK || mytele->mode == TEST_TONE || mytele->mode == STATS_GPS_LEGACY ? RPT_CONF : RPT_TELECONF, RPT_CONF_CONFANN)) {
+	if (rpt_conf_add(mychannel, myrpt, mytele->mode == MDC1200 || mytele->mode == PAGE ||
```
Allan-N, line 1060: Not sure about adding MDC120 and PAGE here.  We want to be treated like LOCALPLAY, *not* PLAYBACK

Allan-N (changes requested):

apps/app_rpt/rpt_telemetry.c
mkmer: Seems odd that we would send a TEST_TONE or ID1 out links... Is there a LOCALPLAY defined or is PLAYBACK actually `rpt xxx localplay`?

mkmer:
Bit of off line conversation: 
It's possible `myrpt->txchannel` was an incorrect "hack" to get MDC1200 and PAGE (cop 48) to go out the TX in any mode.  Using  RPT_TELECONF will (as it appears anyway) not send tones out if we are in duplex mode = 0...

Allan-N (approved):

InterLinked1 (changes requested):

InterLinked1 (approved):

KB4MDD:
Does paging work with all duplex settings?

PR #496, opened 2025-02-18
rpt_install.sh should fail on "make" errors

Allan-N:


InterLinked1 (changes requested):
I don't think this is quite the right approach. For one, everything in the CI already fails if it exits nonzero, so set -e there doesn't doing anything. But I see what you're trying to do.

What I would like to do is add set -e to the top of rpt_install.sh, and then all the checks for nonzero return can be removed. I don't know if it would run successfully to completion the way it is now but that is something we could fine tune easily.

Allan-N:
OK, adding `set -e` at the top of `rpt_install.sh` is too early (we get a complaint about not being on a git branch).   It's possible that we could move the `set -e` later in the script but I opted to fix the error handling just for the `make` commands (checking for errors and **correctly** exiting with a non-zero error code on errors).

InterLinked1:
> OK, adding `set -e` at the top of `rpt_install.sh` is too early (we get a complaint about not being on a git branch). It's possible that we could move the `set -e` later in the script but I opted to fix the error handling just for the `make` commands (checking for errors and **correctly** exiting with a non-zero error code on errors).

That's sort of what I was alluding too. We can add `set -e` to the script, but for any commands that may return nonzero, like that, we now need to explicitly handle them.

In the long run, it would probably be good to do that at some point to tighten things up, but it's definitely not a high priority. This works fine for now.

mkmer (approved):

InterLinked1 (approved):

PR #498, opened 2025-02-18
radio-tune-menu.c: need include "asterisk.h"

mkmer:
simpleusb-tune-menu.c: need include "asterisk.h"

As asterisk developer site says, you need to start with #include "asterisk.h" for the others to work.

### DIFF
--- a/utils/radio-tune-menu.c
+++ b/utils/radio-tune-menu.c
@@ -59,6 +59,7 @@
  * Most of these commands take optional parameters to set values.
  *
  */
+#include "asterisk.h"
 
 #include <stdio.h>
 #include <stdlib.h>
--- a/utils/simpleusb-tune-menu.c
+++ b/utils/simpleusb-tune-menu.c
@@ -52,6 +52,8 @@
  * Most of these commands take optional parameters to set values.
  *
  */
+#include "asterisk.h"
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>


InterLinked1:
That may be true for stuff that is part of Asterisk, but stuff in the utils directory is not part of Asterisk.
What's the actual issue here this fixes? Was there a compilation error?

Allan-N:
>That may be true for stuff that is part of Asterisk, but stuff in the utils directory is not part of Asterisk.
What's the actual issue here this fixes? Was there a compilation error?

You can currently see the error @ https://github.com/AllStarLink/app_rpt/actions/runs/13401225310/job/37432303106?pr=497

I think the problem was the `#include "asterisk/utils.h"` without first having `#include "asterisk.h"`

mkmer:
YES, `#include asterisk.h` allows for the `#include asterisk/utils.h`  to work. Unfortunately, without it we get compilation errors .

Allan-N (approved):

Allan-N (approved):

InterLinked1:
> > That may be true for stuff that is part of Asterisk, but stuff in the utils directory is not part of Asterisk.
> > What's the actual issue here this fixes? Was there a compilation error?
> 
> You can currently see the error @ https://github.com/AllStarLink/app_rpt/actions/runs/13401225310/job/37432303106?pr=497
> 
> I think the problem was the `#include "asterisk/utils.h"` without first having `#include "asterisk.h"`

Ah, okay, I got thrown off again by that build failing. It seems like it fails on the `make install` target which should have caused an abort, I think. Is that why you change `$?` to 1 Allan? The build should've been failing before I feel like but it wasn't for some reason.

Allan-N:
>Is that why you change $? to 1 Allan? The build should've been failing before I feel like but it wasn't for some reason

Yes, we had :
```
make ...
if [ $? -ne 0 ]; then
    exit $?    # Here, $? is no longer the exit code, it's the status of the "if"
fi
```
We now use `exit 1` to ensure we are returning an error.

InterLinked1:
> > Is that why you change $? to 1 Allan? The build should've been failing before I feel like but it wasn't for some reason
> 
> Yes, we had :
> 
> ```
> make ...
> if [ $? -ne 0 ]; then
>     exit $?    # Here, $? is no longer the exit code, it's the status of the "if"
> fi
> ```
> 
> We now use `exit 1` to ensure we are returning an error.

Gotcha! Learned something today :)

Allan-N:
>Gotcha! Learned something today :)

Even though this has bit me before it still took me some time before the light bulb lit

Issue #490, opened 2025-02-14
The perform_statpost function needs better explanation of errors


jxmx:
For example:

```
[2025-02-14 14:46:37.060] WARNING[1333]: app_rpt.c:1009 perform_statpost: statpost to URL ‘http://stats.allstarlink.org/uhandler?node=46287&time=1739508396&seqno=3&nodes=T3428425&apprptvers=3.2.0&apprptuptime=14&totalkerchunks=0&totalkeyups=0&totaltxtime=0&timeouts=0&totalexecdcommands=0’ failed with code 401
```

should say something like "failed with HTTP Code 401 Unauthorized: Wrong password?"

InterLinked1:
> For example:
> 
> ```
> [2025-02-14 14:46:37.060] WARNING[1333]: app_rpt.c:1009 perform_statpost: statpost to URL ‘http://stats.allstarlink.org/uhandler?node=46287&time=1739508396&seqno=3&nodes=T3428425&apprptvers=3.2.0&apprptuptime=14&totalkerchunks=0&totalkeyups=0&totaltxtime=0&timeouts=0&totalexecdcommands=0’ failed with code 401
> ```
> 
> should say something like "failed with HTTP Code 401 Unauthorized: Wrong password?"

It does say "failed with code 401" at the end.

jxmx:
Yeah, but 99% of users have no idea what "failed with code 401" means. We answer A LOT of questions about that in the user forum.

PR #499, opened 2025-02-19
app_rpt.c: perform_statpost - add verbose text to errors.

mkmer:
Provide text error descriptions from the curl library.
Closes issue #490

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -980,10 +980,12 @@ static void doconpgm(struct rpt *myrpt, char *them)
 	return;
 }
 
-/*! \brief Waste the output of libcurl (the OK is sent to stdout) */
-static size_t writefunction(void *ptr, size_t size, size_t nmemb, void *stream)
+/*! \brief Store the output of libcurl (the OK is sent to stdout) */
+static size_t writefunction(char *contents, size_t size, size_t nmemb, void *userdata)
 {
-	return (nmemb*size);
+	struct ast_str **buffer = userdata;
+
+	return  ast_str_append(buffer, 0, "%.*s", (int) (size * nmemb), contents);
 }
 
 static void *perform_statpost(void *stats_url)
@@ -991,24 +993,54 @@ static void *perform_statpost(void *stats_url)
 	char *str;
 	long rescode = 0;
 	CURL *curl = curl_easy_init();
+	CURLcode res;
+	char error_buffer[CURL_ERROR_SIZE];
+	struct ast_str *response_msg;
 
 	if (!curl) {
+		ast_free(stats_url);
+		return NULL;
+	}
+
+	response_msg = ast_str_create(50);
+	if (!response_msg) {
+		ast_free(stats_url);
+		curl_easy_cleanup(curl);
 		return NULL;
 	}
 
 	str = (char *) stats_url;
 	curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writefunction);
+	curl_easy_setopt(curl, CURLOPT_WRITEDATA, (void *) &response_msg);
 	curl_easy_setopt(curl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
 	curl_easy_setopt(curl, CURLOPT_URL, str);
 	curl_easy_setopt(curl, CURLOPT_USERAGENT, AST_CURL_USER_AGENT);
-	curl_easy_perform(curl);
-	curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &rescode);
-	curl_easy_cleanup(curl);
+	curl_easy_setopt(curl, CURLOPT_ERRORBUFFER, error_buffer);
 
-	if (rescode != 200) {
-		ast_log(LOG_WARNING, "statpost to URL '%s' failed with code %ld\n", (char *) stats_url, rescode);
+	res = curl_easy_perform(curl);
+	if (res != CURLE_OK) {
+		if (*error_buffer) { /* Anything in the error buffer? */
+			ast_log(LOG_WARNING, "statpost to URL '%s' failed with error: %s\n", (char *) stats_url, error_buffer);
+		} else {
+			ast_log(LOG_WARNING,
+					"statpost to URL '%s' failed with error: %s\n",
+					(char *) stats_url,
+					curl_easy_strerror(res));
+		}
+	} else {
+		curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &rescode);
+		if (rescode != 200) {
+			ast_log(LOG_WARNING,
+					"statpost to URL '%s' failed with code %ld : %s\n",
+					(char *) stats_url,
+					rescode,
+					ast_str_buffer(response_msg));
+		}
 	}
+	ast_debug(3, "Response: %s\n", ast_str_buffer(response_msg));
 	ast_free(stats_url); /* Free here since parent is not responsible for it. */
+	ast_free(response_msg);
+	curl_easy_cleanup(curl);
 	return NULL;
 }
 


Allan-N (approved):

mkmer:
> Does the error from the library include the number in it? If not, I think we should print both the number and the error name

Actually, there are library errors and web link errors -> that's why I changed it to draft.
We have to capture the response message text in the `writefunction` to get the web error text.

For web errors it will look something like this:
`<http link> failed with code 401 : Unauthorized`

For library errors it will look like this:

`<http link> failed with error: Could not resolve host: tats.allstarlink.org`

I don't believe the library error values make much sense to share (do they?)

apps/app_rpt.c
```diff
@@ -980,35 +980,73 @@ static void doconpgm(struct rpt *myrpt, char *them)
 	return;
 }
 
-/*! \brief Waste the output of libcurl (the OK is sent to stdout) */
-static size_t writefunction(void *ptr, size_t size, size_t nmemb, void *stream)
+struct curl_memory {
+	char *response;
+	size_t size;
```
Allan-N, line 985: Where do you actually use "size"?

apps/app_rpt.c
```diff
@@ -980,35 +980,73 @@ static void doconpgm(struct rpt *myrpt, char *them)
...
+};
+
+/*! \brief Store the output of libcurl (the OK is sent to stdout) */
+static size_t writefunction(void *data, size_t size, size_t nmemb, void *stream)
 {
-	return (nmemb*size);
+	size_t realsize = nmemb * size;
+	struct curl_memory *mem = (struct curl_memory *) stream;
+	char *ptr = ast_realloc(mem->response, mem->size + realsize + 1);
```
Allan-N, line 993: This code seems overly complicated.  Is the goal of this simply to capture the output?  If so, can this be abbreviated to:
```
if (mem->respose) {
    free(mem->response);
}
mem->response = strdup(data);

return realsize;
```

apps/app_rpt.c
mkmer: Not sure, it's part of the structure that is passed in best I can tell.

apps/app_rpt.c
mkmer: They say it can be called multiple times and build a massive message (it could be a web page).
All we want is to see what it has if there is an error, but we don't know in here if it's an error.

apps/app_rpt.c
mkmer: Oh - they use it to expand pointer for multiple calls...

Allan-N: So, the passed structure is actually the one that YOU defined/declared/passed in :
```
curl_easy_setopt(curl, CURLOPT_WRITEDATA, (void *) &response_msg);
```
with the last argument being passed to `writefunction()` as the last `void *stream` argument.  That argument is essentially a pointer to anything you want.

apps/app_rpt.c
Allan-N: >They say it can be called multiple times and build a massive message

In that case, this is a bit more complicated than needing to retain a simple string.

Allan-N (approved):

apps/app_rpt.c
mkmer: OK - I found a "better" version of this in `res_rpt_http_registrations` using `ast_str*` -> refactored to that.
Also added the debug string.

apps/app_rpt.c
```diff
@@ -980,35 +980,71 @@ static void doconpgm(struct rpt *myrpt, char *them)
...
+	size_t realsize = size * nmemb;
+	char *rawdata;
+
+	rawdata = ast_malloc(realsize + 1);
+	if (!rawdata) {
+		return 0;
+	}
+	memcpy(rawdata, contents, realsize);
+	rawdata[realsize] = 0;
+	ast_str_append(buffer, 0, "%s", rawdata);
```
InterLinked1, line 996: This function can be simplified a lot. There's no need to malloc, memcpy, and then append to an ast_str, only to free the originally duplicated content.

I only know this because I did a similar PR at one point and Sean pointed this out. Do something like this instead: https://github.com/InterLinked1/phreakscript/blob/175b2a09d8f8a8cf0efc03fac9265be9c7195a3a/res/res_phreaknet.c#L295

apps/app_rpt.c
```diff
@@ -980,35 +980,73 @@ static void doconpgm(struct rpt *myrpt, char *them)
...
-	curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &rescode);
-	curl_easy_cleanup(curl);
+	curl_easy_setopt(curl, CURLOPT_ERRORBUFFER, error_buffer);
 
-	if (rescode != 200) {
-		ast_log(LOG_WARNING, "statpost to URL '%s' failed with code %ld\n", (char *) stats_url, rescode);
+	res = curl_easy_perform(curl);
+	if (res != CURLE_OK) {
+		if (error_buffer[0]) { /* Anything in the error buffer? */
+			ast_log(LOG_WARNING, "statpost to URL '%s' failed with error: %s\n", (char *) stats_url, error_buffer);
```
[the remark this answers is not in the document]
mkmer: The error buffer contains a string based on the API error number (not a web error). This is now the same output as `res_rpt_http_registrations`, except I've added the `curl_easy_strerror(res)` as the API recommends (sometimes `error_buffer` can be blank).

apps/app_rpt.c
```diff
@@ -980,35 +980,71 @@ static void doconpgm(struct rpt *myrpt, char *them)
...
+	CURLcode res;
+	char error_buffer[CURL_ERROR_SIZE];
+	struct ast_str *response_msg;
 
 	if (!curl) {
+		ast_free(stats_url);
 		return NULL;
 	}
-
+	response_msg = ast_str_create(50);
```
InterLinked1, line 1015: ast_str_create can fail

apps/app_rpt.c
```diff
@@ -980,35 +980,71 @@ static void doconpgm(struct rpt *myrpt, char *them)
...
+		curl_easy_getinfo(curl, CURLINFO_RESPONSE_CODE, &rescode);
+		if (rescode != 200) {
+			ast_log(LOG_WARNING,
+					"statpost to URL '%s' failed with code %ld : %s\n",
+					(char *) stats_url,
+					rescode,
+					ast_str_buffer(response_msg));
+		}
 	}
+	ast_debug(3, "Response: %s\n", ast_str_buffer(str));
```
InterLinked1, line 1044: Should str here be response_msg?

InterLinked1 (changes requested):

apps/app_rpt.c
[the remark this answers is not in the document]
mkmer: https://curl.se/libcurl/c/CURLOPT_ERRORBUFFER.html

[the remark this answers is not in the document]
InterLinked1: The phrasing in the man page suggests the error doesn't include the code (I wasn't expecting it did)
That's why I was wondering about it. Previously, we printed out the error code, I think we should print out both in this case, e.g. `%d: %s", code, error_message`

apps/app_rpt.c
mkmer: Sure should, I was just looking at why the debug was wacky...

apps/app_rpt.c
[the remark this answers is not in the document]
mkmer: We still print a web error code, but these are new codes we previously would log nothing for.  Does an API error code make any sense to share?

apps/app_rpt.c
mkmer: That's awsome!

apps/app_rpt.c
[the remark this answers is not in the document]
InterLinked1: Oh, I see, this isn't the HTTP response code in this case. Ignore me, this is fine!

InterLinked1 (approved):

PR #501, opened 2025-02-20
app_rpt: Don't emit warnings about <sys/io.h> being unavailable

Allan-N:
While some may argue that a few known/understood compilation warnings are OK we get into trouble when stop looking for "new" warnings in build output that includes a number of "old" ones. The simple solution is to banish ALL warnings from the project and then treat any warnings as errors.

This PR addresses the `sys.io is not available on this architecture` warnings reported on `arm64` builds.

### DIFF
--- a/apps/app_rpt/app_rpt.h
+++ b/apps/app_rpt/app_rpt.h
@@ -22,13 +22,8 @@
 #endif
 
 /*! \note <sys/io.h> is not portable to all architectures, so don't call non-portable functions if we don't have them */
-#if defined(__alpha__) || defined(__x86_64__) || defined(__ia64__)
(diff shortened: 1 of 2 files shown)


InterLinked1 (commented):
The commit message is not really correct, as all warnings are treated as errors already, and this warning was explicitly not in the past (that's what the -Wcpp did).

Nonetheless I don't think the warning served any real purpose so might as well get rid of it. That said, I think this would be a more useful commit message title:

`app_rpt: Don't emit warnings about <sys/io.h> being unavailable.`

mkmer (approved):

InterLinked1 (approved):

Issue #506, opened 2025-02-21
app_rpt: Don't use mixmonitor CLI commands


InterLinked1:
Instead of using the MixMonitor CLI commands, which is a hacky kludge that should be avoided at all costs, we should use native APIs instead.

PR #507, opened 2025-02-21
app_rpt: Start/stop recordings using APIs instead of CLI commands.

InterLinked1:
Use the public MixMonitor APIs to start and stop archive recordings rather than hackily using a CLI command to do so.

Resolves: #506


KB4MDD (approved):

InterLinked1:
Hold off on merge just yet, there is a small issue here I need to fix.
(Side note: We should probably make it so a PR can't be merged when the build is failing)

apps/app_rpt.c
```diff
@@ -7127,17 +7127,18 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		time(&myt);
 		strftime(mydate, sizeof(mydate), "%Y%m%d%H%M%S", localtime(&myt));
 		myformat = myrpt->p.archiveformat ? myrpt->p.archiveformat : "wav49";
-		snprintf(mycmd, sizeof(mycmd), "mixmonitor start %s %s/%s/%s.%s a",
-			ast_channel_name(chan), myrpt->p.archivedir, myrpt->name, myformat, mydate);
```
InterLinked1, line 7131: I'm preserving the way it was before, but it seems odd to me that the format has extension.date, rather than date.extension

Is this desired for a particular reason?

mkmer:
I don't think we want to "Force" builds to pass to merge.  The merger should make sure there is no failure, or if there is a failure to confirm it's not related to the PR.  It's possible to screw up the tree, or maybe you like a spelling error or code format is OK to over ride... etc

Allan-N:
>Hold off on merge just yet
I converted the PR to a draft.

apps/app_rpt.c
Allan-N: That does seem backwards ... and looks to be a bug that I introduced.

Allan-N (changes requested):

apps/app_rpt.c
```diff
@@ -7126,18 +7126,18 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		mkdir(mycmd, 0775);
 		time(&myt);
 		strftime(mydate, sizeof(mydate), "%Y%m%d%H%M%S", localtime(&myt));
-		myformat = myrpt->p.archiveformat ? myrpt->p.archiveformat : "wav49";
-		snprintf(mycmd, sizeof(mycmd), "mixmonitor start %s %s/%s/%s.%s a",
-			ast_channel_name(chan), myrpt->p.archivedir, myrpt->name, myformat, mydate);
+		snprintf(mycmd, sizeof(mycmd), "%s/%s/%s.%s", myrpt->p.archivedir, myrpt->name, mydate, S_OR(myrpt->p.archiveformat, "wav49"));
```
Allan-N, line 7129: The string you are now creating is no longer the "command".  It is now the "filename".  Let's give the variable a more proper name (e..g "filename")

Allan-N (changes requested):

Allan-N (approved):

mkmer (approved):
It looks like the right thing to do here.

apps/app_rpt.c
```diff
@@ -7117,7 +7118,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		b1 = b;
 	if (myrpt->p.archivedir) {
 		char mycmd[512], mydate[100];
-		const char *myformat;
+		char filename[256];
```
mkmer, line 7121: I don't know - should we alloc for this rather than limiting the path to 256 chars?

InterLinked1: I *personally* don't think allocation is necessary here, but it's a fair point about the buffer size. I think somehow I thought it was being put into another buffer (the one of size 512) but it isn't so there's no need to limit.

The proper fix technically might be to use `PATH_MAX`, but interestingly I just noticed nowhere in `app_rpt` is `PATH_MAX` currently used at all.

Allan-N: I concur that no allocation is needed.  Using a bit of memory on the stack is quick/easy.

As for PATH_MAX, some use MAXPATHLEN (from <sys/param.h>

mkmer: Sure - so just "right size it"? Is there really a max path on linux (unix)?

InterLinked1: I checked and asterisk.h always ensures PATH_MAX is available so I went with that as it's already standard for the project.
Typically the value is 4096.

PR #512, opened 2025-02-24
app_rpt.c, rpt_cli.c, rpt_manager.c: Add missing mutex_lock()

mkmer:
It appears these calls should be mutex_locked.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3214,7 +3214,9 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			lf.samples = 0;
 			l->linklisttimer = LINKLISTTIME;
 			strcpy(lstr, "L ");
+			rpt_mutex_lock(&myrpt->lock);
 			__mklinklist(myrpt, l, lstr + 2, sizeof(lstr) - 2, 0);
+			rpt_mutex_unlock(&myrpt->lock);
 			if (l->chan) {
 				lf.datalen = strlen(lstr) + 1;
 				lf.data.ptr = lstr;
--- a/apps/app_rpt/rpt_cli.c
+++ b/apps/app_rpt/rpt_cli.c
@@ -529,8 +529,9 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 			 * Traverse the list of connected nodes
 			 */
 
+			rpt_mutex_lock(&myrpt->lock);
 			__mklinklist(myrpt, NULL, lbuf, sizeof(lbuf), 0);
-
+			rpt_mutex_unlock(&myrpt->lock);
 			j = 0;
 			l = myrpt->links.next;
 			while (l && (l != &myrpt->links)) {


mkmer:
While reviewing code that generated this PR, I notice this:
https://github.com/AllStarLink/app_rpt/blob/7b97e6c758a4a76b72c68811f65f9b53b85d9d73/apps/app_rpt/rpt_link.c#L175-L185

I'm not positive, but I think we should be locked when adding to the list? Or maybe when calling `rpt_qwrite()`?

apps/app_rpt/rpt_cli.c
```diff
@@ -526,10 +526,10 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 			}
 
 //### GET CONNECTED NODE INFO ####################
-			// Traverse the list of connected nodes 
-
+			// Traverse the list of connected nodes
+			rpt_mutex_lock(&myrpt->lock); // LOCK
```
InterLinked1, line 530: Comments LOCK and UNLOCK are unnecessary

apps/app_rpt.c
```diff
@@ -3210,13 +3210,15 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			lf.samples = 0;
 			l->linklisttimer = LINKLISTTIME;
 			strcpy(lstr, "L ");
+			rpt_mutex_lock(&myrpt->lock);
 			__mklinklist(myrpt, l, lstr + 2, 0);
 			if (l->chan) {
 				lf.datalen = strlen(lstr) + 1;
 				lf.data.ptr = lstr;
 				rpt_qwrite(l, &lf);
```
InterLinked1, line 3218: Is the lock supposed to be held while calling rpt_rqwrite? At least based on a quick look at other code, it seems like we should unlock right after __mklinklist?

InterLinked1:
> While reviewing code that generated this PR, I notice this:
> 
> https://github.com/AllStarLink/app_rpt/blob/7b97e6c758a4a76b72c68811f65f9b53b85d9d73/apps/app_rpt/rpt_link.c#L175-L185
> 
> I'm not positive, but I think we should be locked when adding to the list? Or maybe when calling `rpt_qwrite()`?

Ah, that's a good point. It looks like it's using a linked list without a lock so some kind of locking does seem needed.

apps/app_rpt.c
mkmer: Probably not.  For now I will move it back.  If we determine `rpt_qwrite()` needs a lock we can fix it in that PR.

apps/app_rpt/rpt_cli.c
```diff
@@ -525,12 +525,11 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 				tel_mode = "3";
 			}
 
-			/* ### GET CONNECTED NODE INFO ####################
-			 * Traverse the list of connected nodes
-			 */
-
-			__mklinklist(myrpt, NULL, lbuf, sizeof(lbuf), 0);
-
+//### GET CONNECTED NODE INFO ####################
```
InterLinked1, line 528: Looks like the comments were (mostly) formatted properly before, but now they're not? (using C99 comments)

Also, to be clear, I don't think the qwrite thing was wrong. But since there are more instances of that, maybe doing another pass addressing that specifically would make more sense.

mkmer: Not really sure how that happened.

mkmer: Understood.  It's up for further review.  I wanted to pose the question becuase it is possibly wrong in so many places.

mkmer: On the format: I did a rebase on master... I suspect it got in from that.

InterLinked1 (approved):

Allan-N (approved):
I'm approving this change as an incremental improvement.  But, we really need to take a step back and review/analyze/update all of the logic around `myrpt` and the associated links.

Why?  If you look at the call to `__mklnklist` inside the `periodic_process_links()` function you will see we are passing `myrpt` and `l`.  Even if we lock `myrpt` one can/should ask whether `l` is still valid as it is captured from `myrpt->links.next` at the top of the function and advanced at the end of the while() loop without holding any locks.

PR #513, opened 2025-02-24
app_rpt.c: ast_frdup() test for failure to alloc

mkmer:
Noticed `ast_frdup()` can fail.  Add test for failure.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4026,6 +4026,9 @@ static inline int dahditxchannel_read(struct rpt *myrpt, char *restrict myfirst)
 					AST_LIST_TRAVERSE(&myrpt->txq, f1, frame_list) x++;
 					for (; x < myrpt->p.simplexpatchdelay; x++) {
 						f1 = ast_frdup(f);
+						if (!f1) {
+							return 0;
+						}
 						RPT_MUTE_FRAME(f1);
 						memset(&f1->frame_list, 0, sizeof(f1->frame_list));
 						AST_LIST_INSERT_TAIL(&myrpt->txq, f1, frame_list);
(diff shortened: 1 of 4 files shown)


InterLinked1 (commented):
ast_frisolate can behave the same as ast_frdup in the case of a non-alloc'd frame, but I seem to recall there being some cases where we duplicate an already allocated frame for something, and in that case, this changes the behavior. I would be very careful with this, especially given there are known memory leaks with this code as it is. This has a high probability of introducing regressions otherwise.

That would require tracing through the usage more extensively, which if it's not already been done, I don't think there is much benefit to changing it and a lot of potential downside.

Of course, checking for failure is always a good thing, but the log messages on failure are redundant.

mkmer:
> ast_frisolate can behave the same as ast_frdup in the case of a non-alloc'd frame, but I seem to recall there being some cases where we duplicate an already allocated frame for something, and in that case, this changes the behavior. I would be very careful with this, especially given there are known memory leaks with this code as it is. This has a high probability of introducing regressions otherwise.
> 
> That would require tracing through the usage more extensively, which if it's not already been done, I don't think there is much benefit to changing it and a lot of potential downside.
> 
> Of course, checking for failure is always a good thing, but the log messages on failure are redundant.

After I switch to frisloate caused by the note, I reviewed frisolate() and also determined there are cases where we could/would get a pointer to the original frame, which is not what we wanted best I could tell.

Thanks for the confirmation!

Allan-N (approved):

InterLinked1 (commented):
Change looks good. As with the other change, I would avoid `Multiple: ` in a commit message title. `apps, channels: ` would be better as it's more specific.

PR #516, opened 2025-02-26
chan_simpleusb.c: Add COR timout timer

mkmer:
Addressing https://github.com/AllStarLink/ASL3/issues/134

Adding first to simpleusb.  Once we all agree on names, function, etc I will follow up with chan_usbradio.

### DIFF
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -125,8 +125,11 @@ static struct ast_jb_conf global_jbconf;
 #define	DIVSAMP (DIVLCM / SAMPRATE)
 
 #define	QUEUE_SIZE	5			/* 100 milliseconds of sound card output buffer */
-
-#define CONFIG	"simpleusb.conf"				/* default config file */
+#define COR_TIMEOUT                                                                      \
+	50 * 60 * 5 /* 50 - 20ms count per second * 60 seconds * 5 minutes  Initial timeout   \
+				 * with no config file setting.                                          \
+				 */
+#define CONFIG "simpleusb.conf" /* default config file */
 
 /* file handles for writing debug audio packets */
 static FILE *frxcapraw = NULL;
@@ -241,6 +244,8 @@ struct chan_simpleusb_pvt {
 	int txoffcnt;				/* Counts the number of 20 ms intervals after TX unkey */
 	int rxondelay;				/* This is the value which RX is ignored after RX activity */
 	int txoffdelay;				/* This is the value which RX is ignored after TX unkey */
+	int cor_timeout;			/* This is the 20ms setpoint value for COR timeout. */
+	int cor_tot; /* Counts the down the number of 20 ms intervals after COR is active. */
 
 	int pager;
 	int waspager;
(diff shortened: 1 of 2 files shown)


KB4MDD:
Does the timeout timer in app_rpt not work in this user's case?

Allan-N:
Seems like the functionality should be one level up from the channel drivers.  That way, we don't need to touch SimpleUSB, USBRadio, Voter, EchoLink, ...  In addition, making the change in the channel driver doesn't give the hub operator the ability to ignore an adjacent node that is not dropping its PTT.  How can we do this for all app_rpt inputs?

Allan-N:
Also, is the current timeout timer attempting to stop us (our node) from transmitting too long?  and does that apply to both what we send to the channel drivers AND downstream nodes?  Hopefully so.  But, maybe we should also apply the same timer to the incoming streams.

KB4MDD:
@Allan-N I agree, I think this should be one level up.  We need to be able to announce the timeout.

mkmer:
The used case is described in the issue: https://github.com/AllStarLink/ASL3/issues/134.  
They want to have connected nodes "unkey" when they timeout on the hardware input.

Allan-N:
>They want to have connected nodes "unkey" when they timeout on the hardware input.

We can't control the connected node(s).  All we can do is discard their input if we feel that they've been talking too long.

>Maybe the local "repeater" should timeout and stop sending (K) upstream.

If a local node is setup with a time-out timer then it should no longer be transmitting nor should it be sending data upstream (that would, in essence, still be transmitting).

mkmer:
As I was testing, I was thinking the same thing: Default is disabled (aka no change to existing configs).  If a repeater group wants something like this they could enable it by activating the parameter.

As far as the totimer -> we STOP local TX on timeout (according to the comments.)  
There is a comment here that looks a bit suspicious -> https://github.com/AllStarLink/app_rpt/blob/7b926425281203552b48840f9c5092af669263e9/apps/app_rpt.c#L3071-L3077
Maybe we just aren't updating the links when we drop the tx.  Still researching,

mkmer:
I can confirm: Local node times out (and announces it) - the alink you are connected to still stays up.
I think that should not behave that way (right?)  If the local node times out it should stop transmitting on all outputs.

Allan-N:
>I think that should not behave that way (right?) If the local node times out it should stop transmitting on all outputs.

My $0.02 :

In general, I think that if we're going to stop transmitting to our own node that it would not be correct for us to keep transmitting (sending) the same data to our adjacent nodes.

But, we might need to think a bit more about streams we would otherwise be forwarding.  I think the reported issue is that OUR transmitters should not be locked up because someone 4 nodes away is stuck with key down.  But, at the same time, if we stop forwarding streams and if our timeout interval is really short then we might end up cutting off the audio to our adjacent nodes.

So, maybe there are two changes to make.  First, for streams that we are originating and we are timing out I think we should also stop sending to our adjacent nodes.  Then, for streams that we are receiving/forwarding we might want different rules (and timers).

mkmer:
I guess the question (in my mind anyway) is what IS "originating"?
audio coming in from voter, simpleusb, etc would timeout and not send upstream (maybe this is not IAX),
Anything IAX would continue?

Allan-N:
>I guess the question (in my mind anyway) is what IS "originating"?

To me, "originating" would be directly attached channel drivers (SimpleUSB, USBRadio, Voter, probably SIP/Autopatch, ...).  Here, "we" control the PTT/KEY ... and if we've been keyed up for too long (time out) then we should stop keying (locally and to anyone downstream).

The "non-originating" sources would be those streams we are receiving from adjacent nodes (IAX, EchoLink, TLB, ...).  Here, we don't control the PTT but we do pass it on to our transmitters and downstream nodes.  In these cases I think passing the PTT to our transmitters  should be under our control (i.e. we can/should enforce a local timeout) but I'm not sure that we should be interrupting the flow to other connected nodes.

PR #518, opened 2025-02-27
app_rpt.c: address memory leaks at thread_exit.

mkmer:
These are all before `l` is added to the linked list and would be left as a memory leak.


Allan-N (approved):

apps/app_rpt.c
```diff
@@ -6742,6 +6742,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		cap = ast_format_cap_alloc(AST_FORMAT_CAP_FLAG_DEFAULT);
 		if (!cap) {
 			ast_log(LOG_ERROR, "Failed to alloc cap\n");
+			ast_free(l);
 			pthread_exit(NULL); /*! \todo This and all subsequent pthread_exit's do not clean up properly */
```
InterLinked1, line 6746: Is this comment outdated now or is there still stuff that is leaked at exit?

mkmer: There may be some other leaks, I'm not positive enough to remove the note.
`ao2_ref(cap, -1);` does this cleanup after itself by removing a ref (aka when ref goes to 0 it is freed)?
What other things need to be cleaned up when the thread is exited vs `return`?

InterLinked1: That cleans up `cap`
I didn't take a detailed look just now, but basically any of the allocations prior to the `pthread_exit`. Clean up wise it's the same as return though personally I don't really like the `pthread_exit` and using return is probably a bit cleaner. But that's its own issue.

mkmer: The I think we have them all. Note removed

InterLinked1 (approved):

PR #520, opened 2025-02-27
app_rpt.c: Fix use after free in rxchannel_read()

mkmer:
Shouldn't free the frame before we are done using it.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3760,8 +3760,6 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 		f1 = myrpt->lastf2;
 		if (f1) {
 			ast_write(myrpt->localoverride ? myrpt->txpchannel : myrpt->pchannel, f1);
-			ast_frfree(f1);
-			myrpt->lastf2 = NULL; /* Now invalid since aliased with f1 */
 			if ((myrpt->p.duplex < 2) && myrpt->monstream && (!myrpt->txkeyed) && myrpt->keyed) {
 				ast_writestream(myrpt->monstream, f1);
 			}
@@ -3769,6 +3767,8 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 				(myrpt->outstreampipe[1] != -1)) {
 				outstream_write(myrpt, f1);
 			}
+			myrpt->lastf2 = NULL; /* Aliased with f1, so set to NULL since this reference is no longer valid */
+			ast_frfree(f1);
 		}
 	} else if (f->frametype == AST_FRAME_DTMF_BEGIN) {
 		if (myrpt->lastf1)
@@ -5948,8 +5948,8 @@ static inline int exec_chan_read(struct rpt *myrpt, struct ast_channel *chan, ch
 				else
 					ast_write(myrpt->txchannel, f);
 			}
-			ast_frfree(f1);
 			myrpt->lastf2 = NULL; /* Aliased with f1, so set to NULL since this reference is no longer valid */
+			ast_frfree(f1);
 		}
 	} else if (f->frametype == AST_FRAME_DTMF_BEGIN) {
 		if (myrpt->lastf1)


Allan-N (approved):

Allan-N (approved):

apps/app_rpt.c
```diff
@@ -3760,7 +3760,6 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 		f1 = myrpt->lastf2;
 		if (f1) {
 			ast_write(myrpt->localoverride ? myrpt->txpchannel : myrpt->pchannel, f1);
-			ast_frfree(f1);
 			myrpt->lastf2 = NULL; /* Now invalid since aliased with f1 */
```
InterLinked1, line 3763: This comment is now inaccurate so something should be done about that.

We should probably leave this statement where it is, but might be worth moving the comment to the new location where f1 is freed, saying `myrpt->lastf2 is now invalid, since it's aliased with f1`? What do you think?

mkmer: I had looked at this... I don't think it matters if we move `myrpt->lastf2` as it's not referenced anywhere in the code that follows.  I haven't dug into the threads to see if someone else is potentially accessing it concurrently. Either way probably should `NULL` it BEFORE we free in `f1`.

apps/app_rpt.c
```diff
@@ -3759,16 +3759,16 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 		mute_frame_helper(myrpt, f, ismuted);
 		f1 = myrpt->lastf2;
 		if (f1) {
+			myrpt->lastf2 = NULL; /* We are done with this reference and free them memory after we finish using f1 */
```
InterLinked1, line 3762: free the
I think we should keep the comment about aliasing with f1, I forget why I added that but it came up for some reason so might be good to point out in the future

apps/app_rpt.c
```diff
@@ -3759,16 +3759,16 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 		mute_frame_helper(myrpt, f, ismuted);
 		f1 = myrpt->lastf2;
 		if (f1) {
+			myrpt->lastf2 = NULL; /* We are done with this reference and free the memory after we finish using the alias in f1 */
```
InterLinked1, line 3762: The meaning seems to different to me... we don't finish using the alias, rather the alias just becomes invalid.
How about
"This is aliased with f1, so we set to NULL now since we'll momentarily free f1" or something like that?

mkmer: If that makes sense to you.  I mean f1 set one line above this comment - it's not to hard to see.
By setting `myrpt->lastf2` NULL, no other threads can try to use it from that reference, that's really all we are doing here (right?)

Allan-N: "I" think you can simply drop the comment.  And, as to other threads using the pointer, now you're crossing over to multi-threading, locking, taking references to the data, ...

Allan-N: I take that back.  Here's what "I" think you need.
1. as you are doing now, move the ast_frfree(f1)
2. move the `myrpt->lastf2 = NULL;` with the ast_frfree() since it's the free that makes the reference no longer valid.
3. and the comment can simply be /* remove the no-longer-valid reference */

apps/app_rpt.c
```diff
@@ -3759,16 +3759,16 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 		mute_frame_helper(myrpt, f, ismuted);
 		f1 = myrpt->lastf2;
 		if (f1) {
+			myrpt->lastf2 = NULL; /* This is aliased with f1, so we set to NULL now since we'll momentarily free f1" or something like that? */
```
InterLinked1, line 3762: " or something like that? :)

apps/app_rpt.c
```diff
@@ -3760,15 +3760,15 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
...
-			ast_frfree(f1);
-			myrpt->lastf2 = NULL; /* Now invalid since aliased with f1 */
 			if ((myrpt->p.duplex < 2) && myrpt->monstream && (!myrpt->txkeyed) && myrpt->keyed) {
 				ast_writestream(myrpt->monstream, f1);
 			}
 			if ((myrpt->p.duplex < 2) && myrpt->keyed && myrpt->p.outstreamcmd &&
 				(myrpt->outstreampipe[1] != -1)) {
 				outstream_write(myrpt, f1);
 			}
+			myrpt->lastf2 = NULL; /* remove the no-longer-valid reference */
```
InterLinked1, line 3770: I'd just use the same comment as on line 5951 here for consistency. With that I think this is good, the ordering is also consistent now

mkmer: I suppose a better question at this point (for me anyway) is what is the point of making a copy of the pointer in `myrpt->lastf2`.  `f1` is still pointing at the same mutable data, so why not just use `myrpt->lastf2`?
The only advantage to copying it would be to allow us to set `myrpt->lastf2` to NULL BEFORE freeing it (which is how I have it now).

InterLinked1: It's certainly worth asking or investigating. Full disclaimer, I didn't write the code, I think I just added the comment. I don't remember the context anymore. I think there was some code somewhere using it and at the time the pointer was invalid or something like that.
It's probably not fully redundant or Jim wouldn't have added it, but it may be in certain circumstances.

InterLinked1 (approved):

KB4MDD (approved):

PR #523, opened 2025-02-28
rpt_config.c: Add missing votermode load parameter

mkmer:
Found another missing initialization. The clue was that we were initializing votertype 2x.
Also, ASL2 inits these with default 0


Allan-N:
>Also, ASL2 inits these with default 0

I believe that all of the variables start off with zero'd out values.

Allan-N (approved):

PR #524, opened 2025-02-28
app_rpt.c: mutex around link activity macro

mkmer:
At a minimum, we should be locking the write to the buffer just like every other one.
Without it, we may have flaky behavior executing the macro.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2592,8 +2592,10 @@ static void do_scheduler(struct rpt *myrpt)
 				ast_log(LOG_WARNING, "Link Activity timer could not execute macro %s: Macro buffer full\n", myrpt->p.lnkactmacro);
 			} else {
 				ast_debug(5, "Executing link activity timer macro %s\n", myrpt->p.lnkactmacro);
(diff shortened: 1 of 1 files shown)


InterLinked1 (commented):
Minor nitpick: "flakey" is misspelled in the commit message

I do wonder if the codespell stuff would pick up. GitHub is not very good about ease of reviewing the commit message itself. Gerrit was much better for this.

mkmer:
https://www.merriam-webster.com/dictionary/flakey
Spell check thinks it's ok because it is....

InterLinked1:
> https://www.merriam-webster.com/dictionary/flakey Spell check thinks it's ok because it is....

But activity wasn't. Is there a way to catch that?
Now that I think about it, even in the Asterisk project, a lot of commit messages have typos. I imagine it's more common for people to hastily throw that together at the end, even if they spend a lot of time on the code.

mkmer:
Codespell is only running against the source.  My browser gives me squiggles in the PR text but nothing is checking the commit messages.

Allan-N (approved):

InterLinked1 (approved):

PR #525, opened 2025-03-02
app_rpt.c: Move link list addition in rpt_exec()

mkmer:
Moving the addition of the link to the linked list prevents `ast_waitfor_n` from adding an unfinished link to the list.  If we `return -1` in `rpt_exec()` AFTER adding a link to the `ast_waitfor_n`, the channel pointer is dumped causing a crash.

The main repeater thread is essentially checking every 20ms (max), rebuilding the list of channels and restarting `ast_waitfor_n`.  This can happen much faster if there are messages on the channels.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -6816,7 +6816,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 
 		/* insert at end of queue */
 		rpt_mutex_lock(&myrpt->lock);
-		rpt_link_add(myrpt, l);
+		rpt_link_add(myrpt, l); /* After putting the link on the linked list, ast_waitfor_n can and will start referencing it */
 		__kickshort(myrpt);
 		gettimeofday(&myrpt->lastlinktime, NULL);
 		rpt_mutex_unlock(&myrpt->lock);
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -6773,20 +6773,17 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 			ast_free(l);
 			pthread_exit(NULL);
 		}
-
-		rpt_mutex_lock(&myrpt->lock);
+		
 		if ((phone_mode == 2) && (!phone_vox))
 			l->lastrealrx = 1;
 		l->max_retries = MAX_RETRIES;
-		rpt_mutex_lock(&myrpt->blocklock);
 
 		if (ast_channel_state(chan) != AST_STATE_UP) {
 			ast_answer(chan);
 			if (l->name[0] > '9') {
 				if (ast_safe_sleep(chan, 500) == -1) {
 					ast_debug(3, "Channel %s hung up\n", ast_channel_name(chan));
-					rpt_mutex_unlock(&myrpt->blocklock);
-					rpt_mutex_unlock(&myrpt->lock);
+					ast_free(l);
 					return -1;
 				}
 			} else {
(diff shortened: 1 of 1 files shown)


mkmer:
Unfortunately,  now if we trigger the `return -1` something is happening to the IAX channel.
```[2025-03-02 10:54:10.342] WARNING[1597877]: chan_iax2.c:3619 __attempt_transmit: Max retries exceeded to host 149.154.11.243 on IAX2/149.154.11.243:24533-5248 (type = 6, subclass = 2, ts=63004, seqno=9)```
 I `return -1` not finishing cleanup?

**Comment by mkmer** [comment]
Now that I'm "looking" this is happening with the original base code too.
Maybe this is because we have sent a text message on the channel with `send_newkey()` and it's stuck on the queue?  Shouldn't a `hangup` or `return -1` clear that?

It's NOT a crash, but it will eventually fill up some log files :(

**Comment by InterLinked1** [comment]
> Now that I'm "looking" this is happening with the original base code too. Maybe this is because we have sent a text message on the channel with `send_newkey()` and it's stuck on the queue? Shouldn't a `hangup` or `return -1` clear that?
> 
> It's NOT a crash, but it will eventually fill up some log files :(

Yeah, sometimes that can happen if something gets lost or the two sides get out of sync. I think it can happen for a few different reasons, but it does usually mean something is wrong.

**Comment by mkmer** [comment]
So will this eventually "stop" -> it is spamming the log file.

**Comment by InterLinked1** [comment]
> So will this eventually "stop" -> it is spamming the log file.

It will, but I think this is likely a bug somewhere (as are most things that spam the log file)

**Comment by mkmer** [comment]
This is 100% related to the `softhangup()`.  I believe we are in the midst of building out the linklist at the end of the `rpt_exec()` function when the hangup rolls in.  The new "safe_hangup" function calls `softhangup` and wammo...
As a test, I removed softhangup (only hardhangup in the `safe_hangup` routine) and this goes away.
I'm thinking we are safe to hardhangup from the polling thread because we are 100% finished looking at the `chan` reference in `rpt_exec()` once we put the link on the linklist.

**Comment by mkmer** [comment]
Sounds like something wrong in the IAX2 channel code - really shouldn't keep retrying forever...

**Comment by mkmer** [comment]
Looks like Asterisk 22.2.0 fixes this problem!

**Comment by InterLinked1** [comment]
> Looks like Asterisk 22.2.0 fixes this problem!

Oh, yeah that reminds me now, I'm guessing you didn't have this fix on your system yet? https://github.com/asterisk/asterisk/commit/f6e1c163bb136bf43b889c3b310ed97b2e40e112

I think that resulted in quite a few of these scenarios. That's only been in the last version or two I think so always good to have the latest source!

**Comment by mkmer** [comment]
Correct, I was sticking with 20.x in order to minimize configuration changes while trying to isolate the crash issue.
Now that I think we have an answer to the crash, it makes sense to move up to 22.x.  Glad it was an easy fix :)

**Review by Allan-N — APPROVED** [review]

File: apps/app_rpt.c [hunk 1]
```diff
@@ -6804,39 +6792,19 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
...
 		if (l->chan) {
-			rpt_mutex_unlock(&myrpt->lock);
 			if (!strcasecmp(ast_channel_tech(chan)->type, "echolink") || !strcasecmp(ast_channel_tech(chan)->type, "tlb") || (l->name[0] > '9')) {
 				rpt_telemetry(myrpt, CONNECTED, l);
 			}
-		} else {
-			rpt_mutex_unlock(&myrpt->lock);
 		}
 
 		/* In theory, both these conditions should be true if one is, since the node thread will queue a soft hangup on this channel and then set l->chan to NULL */
```
InterLinked1, line 6810: With these changes, I don't think we need to check for `!l->chan` anymore. since nobody else can touch it. It could go back to just `if (ast_check_hangup(chan))`

apps/app_rpt.c
```diff
@@ -6846,10 +6814,19 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
...
+		__kickshort(myrpt);
+		gettimeofday(&myrpt->lastlinktime, NULL);
+		rpt_mutex_unlock(&myrpt->lock);
+		rpt_update_links(myrpt);
+
 		/* Set the PBX to NULL to avoid a warning in channel.c about a PBX remaining on the channel when it gets destroyed.
 		 * This goes hand in hand with mirroring the old "KEEPALIVE" behavior. Past this point, there is no PBX for this channel.
 		 * We don't do this until the very end, because the node thread will check if this channel has a PBX to determine
 		 * if it's still "owned" by the PBX thread, as opposed to by an app_rpt thread. */
+
```
InterLinked1, line 6829: Extra blank line?

InterLinked1 (approved):

Issue #527, opened 2025-03-02
rpt_link.c: High fence violation following attempt to calloc 0 bytes in rpt_update_links


InterLinked1:
This code will occasionally attempt to allocate 0 bytes: https://github.com/AllStarLink/app_rpt/blob/master/apps/app_rpt/rpt_link.c#L582

That, in turn, triggers a memory warning in Asterisk, which could cause it to crash:

```
WARNING[432247]: asterisk.asterisk:249 errReceived: Asterisk 127.0.0.1 received error: WARNING: High fence violation of 0x7ff61c029dc0 allocated at app_rpt/rpt_link.c rpt_update_links() line 582
```

I'm not sure if attempts to allocate 0 bytes are actually illegal and that's the direct cause of this, there may be another issue in play somewhere here. However, other allocations don't trigger this in testing, just those with an argument of 0. Probably makes sense to just return early if size is 0.

mkmer:
That's a good find.  I'm surprised I didn't see this error on either of my machines. 
FWIW, I'm looking at moving this whole thing to `ast_str* ` - maybe by the April release.  This should simplify everything.

mkmer:
535 fixes this

PR #528, opened 2025-03-02
rpt_link.c: Avoid allocations of size 0.

InterLinked1:
Abort early if the allocation size will be 0. This is possibly due to changes in commit 7b926425281203552b48840f9c5092af669263e9. Apart from being pointless, such an allocation also triggers a high fence violation in astmm, which can cause a crash.

Resolves: #527


mkmer (approved):

mkmer:
ON second thought, I don't think returning is the right answer for 0.  We still need to generate a link list and ALINK list with the obuf data.  Just because the list is empty, we still need to do the rest.  Maybe +1 or just build the obuf

mkmer:
I think we are missing a +1 for the null here -> we should never get a size = 0, minimum would be 1.

PR #529, opened 2025-03-02
rpt_channel.c: Remove noisy dtmf_local_timer debug message.

InterLinked1:
Remove debug message about dtmf_local_timer reaching 1. This is obvious from the code structure and is incredibly noisy during test executions.

### DIFF
--- a/apps/app_rpt/rpt_channel.c
+++ b/apps/app_rpt/rpt_channel.c
@@ -177,8 +177,6 @@ void do_dtmf_local(struct rpt *myrpt, char c)
 	}
 	/* if at timeout */
 	if (myrpt->dtmf_local_timer == 1) {
-		ast_debug(7, "time out dtmf_local_timer=%i\n", myrpt->dtmf_local_timer);
-
 		/* if anything in the string */
 		if (myrpt->dtmf_local_str[0]) {
 			digit = myrpt->dtmf_local_str[0];


mkmer (approved):

Allan-N (approved):

mkmer:
I'm thinking this may be a pointer to a problem rather than just a noisy debug message.
I will do a bit of research to be sure it's not a bug/problem.

mkmer:
Found it:
https://github.com/AllStarLink/app_rpt/blob/f99565842cc60dc71736c127621c64423d88b6bb/apps/app_rpt/rpt_channel.c#L179
is supposed to act like a oneshot -> after it triggers 1 time it sets `dtmf_local_timer = 0`
https://github.com/AllStarLink/app_rpt/blob/f99565842cc60dc71736c127621c64423d88b6bb/apps/app_rpt/rpt_channel.c#L210

BUT the timer logic https://github.com/AllStarLink/app_rpt/blob/f99565842cc60dc71736c127621c64423d88b6bb/apps/app_rpt.c#L3091-L3103
Forces it to the end value.

I will submit a different PR to fix this.  I think we should leave this debug message in tact.

mkmer:
https://github.com/AllStarLink/app_rpt/pull/539 Addresses the cause of the noise.

PR #530, opened 2025-03-02
app_rpt: Shut down repeaters if Asterisk is shutting down.

InterLinked1:
* Abort repeater loop if Asterisk is shutting down.
* Don't try restarting repeaters if Asterisk is shutting down.

This prevents tests from hanging forever and timing out when the test suite attempts to shut Asterisk down.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4952,7 +4952,7 @@ static void *rpt(void *this)
 			break;
 		}
 
-		if (rpt_any_hangups(myrpt)) {
+		if (ast_shutting_down() || rpt_any_hangups(myrpt)) {
 			break;
 		}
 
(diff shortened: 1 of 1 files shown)


Allan-N (commented):
Looks like you have removed the only reference to `rpt_any_hangups()`.  If true, shouldn't the function also be removed?

But, looking at `rpt_any_hangups()` I'm not sure this is a wise idea.  Don't we want to encourage a graceful shutdown of the channels and give them a chance to send any goodbyes, quiese, etc?

InterLinked1:
> Looks like you have removed the only reference to `rpt_any_hangups()`. If true, shouldn't the function also be removed?
> 
> But, looking at `rpt_any_hangups()` I'm not sure this is a wise idea. Don't we want to encourage a graceful shutdown of the channels and give them a chance to send any goodbyes, quiese, etc?

I didn't remove the reference, I just added to the condition

Allan-N:
> didn't remove the reference, I just added to the condition

Yup.  I see that know.  My eyes focused on the GitHub highlighted differences.

Allan-N (approved):

mkmer (approved):

PR #532, opened 2025-03-02
app_rpt: Eliminate START_DELAY and use per-node logic instead.

InterLinked1:
Currently, Rpt() rejects all calls received within the first two seconds after the repeater's thread is launched, to give the thread time to start. This is an arbitrary wait time that unnecessarily delays node connects in tests and doesn't guarantee anything is actually started properly. Instead of using an arbitary number for all nodes, look at whether the specific node is actually ready or not yet.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -282,8 +282,6 @@
 
 #include "asterisk.h"
 
-#define	START_DELAY 2
-
 #include <signal.h>
 #include <stdio.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


InterLinked1:
While I think this improves the logic and it seems to work fine in my tests, if others could do some tests with their configurations and make sure this doesn't have any unintended consequences, that would be great!

Allan-N (approved):

Allan-N (approved):

mkmer (commented):
Appears to be ok with my configs.

mkmer (approved):

KB4MDD (approved):

PR #535, opened 2025-03-02
rpt_link.c: __get_nodelist_size minimum string size is 1 byte

mkmer:
Missing the null terminator in the buffer size calculation.
Elminatates the need for PR #528 and resloves issue #527


Allan-N:
>Missing the null terminator in the buffer size calculation.
We should have been OK as long as we had 1 link as each added link included either a "," or trailing NULL.  The only time you might have had a zero size would be if you had no links.

Allan-N (approved):

mkmer:
Our problem was when 0 nodes are linked.  The way the code is now, we need that null string (1 byte) to build the obuf.

InterLinked1 (approved):

PR #537, opened 2025-03-03
app_rpt.c: fix memory leak around mute_frame_helper()

mkmer:
Add docs to the `mute_frame_helper()`
Copy lastf2 BEFORE shifting frames

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3583,7 +3583,7 @@ static inline void mute_frame_helper(struct rpt *myrpt, struct ast_frame *f, int
 static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 {
 	int ismuted;
-	struct ast_frame *f, *f1;
+	struct ast_frame *f, *lastf2_old;
 	int i, dtmfed = 0;
 
 	rpt_mutex_lock(&myrpt->blocklock);
@@ -3785,18 +3785,18 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 		if (myrpt->p.votertype == 1 && myrpt->voted_link != NULL) {
 			ismuted = 1;
 		}
-		f1 = myrpt->lastf2;
+		lastf2_old = myrpt->lastf2;
 		mute_frame_helper(myrpt, f, ismuted);
-		if (f1) {
-			ast_write(myrpt->localoverride ? myrpt->txpchannel : myrpt->pchannel, f1);
+		if (lastf2_old) {
+			ast_write(myrpt->localoverride ? myrpt->txpchannel : myrpt->pchannel, lastf2_old);
 			if ((myrpt->p.duplex < 2) && myrpt->monstream && (!myrpt->txkeyed) && myrpt->keyed) {
-				ast_writestream(myrpt->monstream, f1);
+				ast_writestream(myrpt->monstream, lastf2_old);
 			}
 			if ((myrpt->p.duplex < 2) && myrpt->keyed && myrpt->p.outstreamcmd &&
 				(myrpt->outstreampipe[1] != -1)) {
-				outstream_write(myrpt, f1);
+				outstream_write(myrpt, lastf2_old);
 			}
-			ast_frfree(f1);
+			ast_frfree(lastf2_old);
 		}
 	} else if (f->frametype == AST_FRAME_DTMF_BEGIN) {
 		if (myrpt->lastf1)
@@ -5911,7 +5911,7 @@ done:
 
 static inline int exec_chan_read(struct rpt *myrpt, struct ast_channel *chan, char *restrict keyed, const int phone_mode, const int phone_vox, char *restrict myfirst, int *restrict dtmfed)
 {
-	struct ast_frame *f = ast_read(chan);
+	struct ast_frame *f = ast_read(chan), *lastf2_old;
 	if (!f) {
 		ast_debug(1, "@@@@ link:Hung Up\n");
 		return -1;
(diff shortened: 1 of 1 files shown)


InterLinked1 (commented):
I haven't followed all the pointers closely, but while this is brought up, any reason we can't ditch this code and use the Asterisk MUTEAUDIO function instead?
https://github.com/asterisk/asterisk/blob/2d7948fa483c8703cf9948811fd67147f4d560d4/res/res_mutestream.c

This seems to be a good example of duplicating logic that already exists in Asterisk, and I think we should continue removing redundant code where we find it, although to be fair to Jim, I don't think this existed at the time he wrote it

InterLinked1:
> There is a bit of a bucket brigade going on here. It looks like we shift 2 frames around for a total of 2 frames of delay: f -> lastf1, lastf1 -> lastf2, lastf2 is the audio we send out in most cases. I'm thinking if we use the built in mute channel, we will hear part of the DTMF tones as it doesn't fill the lastf1/lastf2 frames.

Okay, that could be a consideration. I imagine they addressed that somehow but I don't know exactly. This is the implementation: https://github.com/asterisk/asterisk/blob/master/main/channel.c#L10743

apps/app_rpt.c
```diff
@@ -3541,25 +3541,30 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
...
+ * \param f - the frame to be stored in lastf1
+ * \param ismuted - if true, the frame is muted by filling f, lastf1 and lastf2 with zeros
+ */
+static inline void mute_frame_helper(struct rpt *myrpt, struct ast_frame *f, int ismuted) {
 	struct ast_frame *f2;
+
 	if (ismuted) {
-		memset(f->data.ptr, 0, f->datalen);
-		if (myrpt->lastf1)
+		if (f) {
```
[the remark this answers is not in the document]
InterLinked1: If it is NULL, is that a bug? Or is that possible in correct code?
If it shouldn't happen, it should probably be an assertion, rather than checking and tolerating it

apps/app_rpt.c
```diff
@@ -3541,31 +3541,36 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
...
 
-static inline void mute_frame_helper(struct rpt *myrpt, struct ast_frame *f, int ismuted)
-{
+/*! \brief Shifts frames: myrpt->lastf1 -> myrpt->lastf2, f -> myrpt->lastf1.
+ * If muted, old lastf2, lastf1, lastf2 and f are filled with zeros.
+ * \param myrpt - the rpt structure
+ * \param f - the frame to be stored in lastf1
+ * \param ismuted - if true, the frame is muted by filling f, lastf1 and lastf2 with zeros
+ */
+static inline void mute_frame_helper(struct rpt *myrpt, struct ast_frame *f, int ismuted) {
```
InterLinked1, line 3550: { should be on separate line

InterLinked1:
> I'm not sure where the best place for a discussion on this would be. Let's start here (we can move it later if we need to): In `process_link_channels`, we use a 2 frame delay process when we are in tlb, echolink, or phone/vox:
> 
> https://github.com/AllStarLink/app_rpt/blob/9db0ce2f3ca6a6edd73c1b313ec2ddfe3d62d8b4/apps/app_rpt.c#L4381
> 
> 
> via some "magic" we determine if we want to mute the channel with `ismuted` where we then `memset` zeros into the 2 delayed frames:
> https://github.com/AllStarLink/app_rpt/blob/9db0ce2f3ca6a6edd73c1b313ec2ddfe3d62d8b4/apps/app_rpt.c#L4437-L4443
> 
> 
> The write the oldest frame to the `pchan`
> https://github.com/AllStarLink/app_rpt/blob/9db0ce2f3ca6a6edd73c1b313ec2ddfe3d62d8b4/apps/app_rpt.c#L4458
> 
> Next we look at voter traffic (if it's a voter channel) and mute audio that's NOT winning the voter logic, but here it's the actual frame, not the delay frames:
> 
> https://github.com/AllStarLink/app_rpt/blob/9db0ce2f3ca6a6edd73c1b313ec2ddfe3d62d8b4/apps/app_rpt.c#L4463-L4469
> 
> https://github.com/AllStarLink/app_rpt/blob/9db0ce2f3ca6a6edd73c1b313ec2ddfe3d62d8b4/apps/app_rpt.c#L4470
> 
> At this point of the code, we have already received an audio frame and we are now trying to decide what to do with it: Clear it, delay it, or just write it to the `pchan`. I think it's too late to ask "asterisk" to mute the channel at this point, plus if we mute it, we will stop receiving voice frames that essentially trigger this logic (right?)
> 
> It may be possible to mute the talker in the `pchan` conference based on the evaluation of `ismuted` -> then we could mute/unmute in this location based on `ismuted`? I'm sure there is a command to mute a talker (right??) I don't see where it is used in app_rpt.

I agree in the voter case, using the MUTEAUDIO function doesn't make sense. It would if we wanted to mute all audio from the channel for an extended period of time, but if we're making that decision frame by frame, embedding the logic is better. It's not a big deal since muting is basically just zeroing out the frame.

mkmer:
I'm closing this in favor of https://github.com/AllStarLink/app_rpt/pull/538 -> combining everything into one PR

mkmer:
In the end, this PR will fix the leak and add docs. As we moved the rest to PR# 538

Allan-N (approved):
I'm going to approve this PR but will also share some (random?) thoughts.
- would it make sense to have mute_frame_helper() return the lastf2 that we're dropping?
- if we are calling mute_frame_helper() multiple times in a row with ismuted==1 then are we are clearing the new frame (good).  But, are we not re-clearing lastf1 and lastf2?  Or maybe that's being cleaned up in 538?

mkmer:
I was just thinking the same thing about mute frame helper.  I'll look at doing that in pr 538.  
This one (537) may make sense to put in this release.

apps/app_rpt.c
```diff
@@ -3575,7 +3583,7 @@ static inline void mute_frame_helper(struct rpt *myrpt, struct ast_frame *f, int
 static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 {
 	int ismuted;
-	struct ast_frame *f, *f1;
+	struct ast_frame *f, *lastf2_old;
```
Allan-N, line 3586: How about naming it just `last_frame`?

Allan-N:
>This one (537) may make sense to put in this release.
As in, re-roll the current beta packages?

mkmer:
Everything is in PR 538 now.  Closing this one

PR #538, opened 2025-03-03
app_rpt.c: add rpt_frame_helper and refactor mute_frame_helper

mkmer:
Noticed this chunk of code reused a few times.
Add a new `mute_frame_helper()`
Remove unnecessary `memset()` from `mute_frame_helper()`
Rename original `mute_frame_helper()` to `rpt_frame_helper()`
Add macro for muting frames.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4441,11 +4441,7 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 					}
 				} else {
 					/* if a voting rx link and not the winner, mute audio */
-					if (myrpt->p.votertype == 1 && l->voterlink && myrpt->voted_link != l)
-						ismuted = 1;
-					else
-						ismuted = 0;
-
+					ismuted = (myrpt->p.votertype == 1) && l->voterlink && (myrpt->voted_link != l);
 					if (!l->lastrx || ismuted)
 						RPT_MUTE_FRAME(f);
 					ast_write(l->pchan, f);
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3790,8 +3790,8 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 		if (myrpt->p.votertype == 1 && myrpt->voted_link != NULL) {
 			ismuted = 1;
 		}
-		mute_frame_helper(myrpt, f, ismuted);
 		f1 = myrpt->lastf2;
+		mute_frame_helper(myrpt, f, ismuted);
 		if (f1) {
 			ast_write(myrpt->localoverride ? myrpt->txpchannel : myrpt->pchannel, f1);
 			if ((myrpt->p.duplex < 2) && myrpt->monstream && (!myrpt->txkeyed) && myrpt->keyed) {
@@ -3801,7 +3801,6 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 				(myrpt->outstreampipe[1] != -1)) {
 				outstream_write(myrpt, f1);
 			}
-			myrpt->lastf2 = NULL; /* Aliased with f1, so set to NULL since this reference is no longer valid */
 			ast_frfree(f1);
 		}
 	} else if (f->frametype == AST_FRAME_DTMF_BEGIN) {
@@ -5970,8 +5969,8 @@ static inline int exec_chan_read(struct rpt *myrpt, struct ast_channel *chan, ch
 		if (*dtmfed && phone_mode)
 			ismuted = 1;
 		*dtmfed = 0;
-		mute_frame_helper(myrpt, f, ismuted);
 		f1 = myrpt->lastf2;
+		mute_frame_helper(myrpt, f, ismuted);
 		if (f1) {
 			if (!myrpt->remstopgen) {
 				if (phone_mode)
@@ -5979,7 +5978,6 @@ static inline int exec_chan_read(struct rpt *myrpt, struct ast_channel *chan, ch
 				else
 					ast_write(myrpt->txchannel, f);
 			}
-			myrpt->lastf2 = NULL; /* Aliased with f1, so set to NULL since this reference is no longer valid */
 			ast_frfree(f1);
 		}
 	} else if (f->frametype == AST_FRAME_DTMF_BEGIN) {
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3570,7 +3570,7 @@ static inline void dtmf_mute_frame_helper(struct rpt *myrpt)
  * \param f - the frame to be stored in lastf1
  * \param ismuted - if true, the frame is muted by filling f, lastf1 and lastf2 with zeros
  */
-static inline ast_frame *rpt_mute_frame_helper(struct rpt *myrpt, struct ast_frame *f, int ismuted)
+static inline struct ast_frame *rpt_frame_helper(struct rpt *myrpt, struct ast_frame *f, int ismuted)
 {
 	struct ast_frame *f2, *last_frame;
 
@@ -3791,7 +3791,7 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 		if (myrpt->p.votertype == 1 && myrpt->voted_link != NULL) {
 			ismuted = 1;
 		}
-		f1 = rpt_mute_frame_helper(myrpt, f, ismuted);
+		f1 = rpt_frame_helper(myrpt, f, ismuted);
 		if (f1) {
 			ast_write(myrpt->localoverride ? myrpt->txpchannel : myrpt->pchannel, f1);
 			if ((myrpt->p.duplex < 2) && myrpt->monstream && (!myrpt->txkeyed) && myrpt->keyed) {
@@ -5969,7 +5969,7 @@ static inline int exec_chan_read(struct rpt *myrpt, struct ast_channel *chan, ch
 		if (*dtmfed && phone_mode)
 			ismuted = 1;
 		*dtmfed = 0;
-		f1 = rpt_mute_frame_helper(myrpt, f, ismuted);
+		f1 = rpt_frame_helper(myrpt, f, ismuted);
 		if (f1) {
 			if (!myrpt->remstopgen) {
 				if (phone_mode)
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3564,13 +3564,13 @@ static inline void mute_frame_helper(struct rpt *myrpt)
  * If muted, old lastf2, lastf1, lastf2 and f are filled with zeros.
  * \param myrpt - the rpt structure
  * \param f - the frame to be stored in lastf1
- * \param ismuted - if true, the frame is muted by filling f, lastf1 and lastf2 with zeros
+ * \param mute - if true, the frame is muted by filling last_frame, f, lastf1 and lastf2 with zeros
  */
-static inline struct ast_frame *rpt_frame_helper(struct rpt *myrpt, struct ast_frame *f, int ismuted)
+static inline struct ast_frame *rpt_frame_helper(struct rpt *myrpt, struct ast_frame *f, int mute)
 {
 	struct ast_frame *f2, *last_frame;
 
-	if (ismuted) {
+	if (mute) {
 		RPT_MUTE_FRAME_IE(f);
 		mute_frame_helper(myrpt);
 	}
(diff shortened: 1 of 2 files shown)


apps/app_rpt.c
```diff
@@ -3541,6 +3541,19 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
 	}
 }
 
+/*! \brief Zero data in myrpt->lastf1 and lastf2 registers (muting audio)
+ * \param myrpt The rpt structure to mute
+ * \retval void
+*/
+static inline void dtmf_mute_frame_helper (struct rpt *myrpt){
```
InterLinked1, line 3548: Opening brace goes on its own line
`\retval void` isn't needed since the function is void. The parameter is pretty straightforward to, at most a single line brief is probably all we need

apps/app_rpt.c
```diff
@@ -3806,10 +3816,7 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 
 			return 0;
 		}
-		if (myrpt->lastf1)
-			memset(myrpt->lastf1->data.ptr, 0, myrpt->lastf1->datalen);
-		if (myrpt->lastf2)
-			memset(myrpt->lastf2->data.ptr, 0, myrpt->lastf2->datalen);
+		dtmf_mute_frame_helper(myrpt;
```
InterLinked1, line 3819: close paren?

InterLinked1 (changes requested):

apps/app_rpt.c
```diff
@@ -3541,6 +3541,19 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
 	}
 }
 
+/*! \brief Zero data in myrpt->lastf1 and lastf2 registers (muting audio)
+ * \param myrpt The rpt structure to mute
+ */
+static inline void dtmf_mute_frame_helper (struct rpt *myrpt)
```
InterLinked1, line 3547: no space needed after helper

InterLinked1 (changes requested):

InterLinked1 (approved):

apps/app_rpt.c
```diff
@@ -3541,27 +3541,40 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
...
-		if (myrpt->lastf2)
-			memset(myrpt->lastf2->data.ptr, 0, myrpt->lastf2->datalen);
+		if (f) {
+			memset(f->data.ptr, 0, f->datalen);
+		}
+		dtmf_mute_frame_helper(myrpt);
 	}
+
 	f2 = f ? ast_frdup(f) : NULL;
 	myrpt->lastf2 = myrpt->lastf1;
```
Allan-N, line 3575: I think we've talked about this before but I [still] gotta ask.  If we start off with a frame `f`, create a dup of it with `ast_frdup()`, and assign the dup to `->lastf1` then who's responsible for calling `ast_frfree()` on that frame?  

Asking because in the non-muted case we call `ast_write()` and `outstream_write()` on `->lastf2` followed by a call to `ast_frfree()`.  Here, where we are pushing on a new frame it looks like we may be leaking `->lastf2`.

Before we make this last assignment do we need to add :
```
if (myrpt->lastf2) {
    ast_frfree(myrpt->lastf2);
}
```

mkmer: Assuming I understand the question: 
the caller frees them: This was that convoluted assignment / free logic that I have addressed in PR #537  
https://github.com/AllStarLink/app_rpt/blob/5fd107e9232c34c799e0c4ce49b9732bd6b5f4bf/apps/app_rpt.c#L3783-L3787
https://github.com/AllStarLink/app_rpt/blob/5fd107e9232c34c799e0c4ce49b9732bd6b5f4bf/apps/app_rpt.c#L5960-L5970
f is freed at the end of the function it was called to.

Allan-N: Shuffling back and forth between https://github.com/AllStarLink/app_rpt/pull/537 and https://github.com/AllStarLink/app_rpt/pull/538 ...

Let's say we have 3 frames, "oldest" (in lastf2), "old" (in lastf1), and "new" (in f).

The code in 538 has :
```
                mute_frame_helper(myrpt, f, ismuted);
                f1 = myrpt->lastf2;
```
 In 537, the mute_frame_helper() drops "oldest" (the current f2), makes "old" become the new "oldest", and the "new" frame becomes "old".  Then, after the shift we are picking up "lastf2" (what was "old") into the "f1" variable, write it out, and then drop "lastf2" as it's now been sent.  I'm thinking that the order needs to be :
```
                f1 = myrpt->lastf2;
                mute_frame_helper(myrpt, f, ismuted);
```
This way, we're going to send out the "oldest" frame and keep the "old" and "new" frames for the next round.

mkmer: Let's look at "original" base code here:
https://github.com/AllStarLink/app_rpt/blob/9db0ce2f3ca6a6edd73c1b313ec2ddfe3d62d8b4/apps/app_rpt.c#L3780-L3782
https://github.com/AllStarLink/app_rpt/blob/9db0ce2f3ca6a6edd73c1b313ec2ddfe3d62d8b4/apps/app_rpt.c#L5965-L5968
Both copy f1 AFTER the `mute_frame_helper`

That's why the `f1` thing didn't make sense, just use `myrpt->lastf2`

I think you're pointing out a bug in the original base?  We should be copying `myrpt->lastf2` before the helper... Right?
So we are currently leaking the frame we dumped in the shift of `myrpt->lastf2 = myrpt->lastf1`???

mkmer: Also , we probably shouldn't be NULLing `myrpt->lastf2` because they are NOT the same if we move the assignment of f1

mkmer: Something like this: https://github.com/AllStarLink/app_rpt/tree/mute_frame_leak

apps/app_rpt.c
```diff
@@ -3541,27 +3541,40 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
...
+	if (myrpt->lastf1) {
+		memset(myrpt->lastf1->data.ptr, 0, myrpt->lastf1->datalen);
+	}
+	if (myrpt->lastf2) {
+		memset(myrpt->lastf2->data.ptr, 0, myrpt->lastf2->datalen);
+	}
+}
+
+/*! \brief Shifts frames: myrpt->lastf1 -> myrpt->lastf2, f -> myrpt->lastf1.
+ * If muted, old lastf2, lastf1, lastf2 and f are filled with zeros.
```
Allan-N, line 3558: This comment and #537 may be conflicting with each other.  Regardless, shouldn't this comment be :
```
 * If muted, old lastf2, old lastf1, and f are filled with zeros.
```

Allan-N: and I also wrote this in 537 but it might make sense to have mute_frame_helper() return the lastf2 that we're dropping?

Allan-N: and is it mute_frame_helper?  or rpt_frame_helper?

apps/app_rpt.c
```diff
@@ -3541,27 +3541,41 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
...
+ * \param myrpt - the rpt structure
+ * \param f - the frame to be stored in lastf1
+ * \param ismuted - if true, the frame is muted by filling f, lastf1 and lastf2 with zeros
+ */
+static inline ast_frame *rpt_mute_frame_helper(struct rpt *myrpt, struct ast_frame *f, int ismuted)
```
Allan-N, line 3563: My suggestion was actually to drop "mute" fro the function name.  I see this function as an rpt helper for dealing with frames.  The "ismuted" argument just has an impact on the rpt frames.  If ismuted==0 then we're not muting, we're just passing frames.

mkmer: Gotcha - makes sense to me. Sometimes one is just fixing "stuff" and not looking at the big picture.

Allan-N (approved):

apps/app_rpt.c
```diff
@@ -3551,27 +3551,41 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
...
+	}
+}
+
+/*! \brief Shifts frames: myrpt->lastf1 -> myrpt->lastf2, f -> myrpt->lastf1.
+ * If muted, old lastf2, lastf1, lastf2 and f are filled with zeros.
+ * \param myrpt - the rpt structure
+ * \param f - the frame to be stored in lastf1
+ * \param ismuted - if true, the frame is muted by filling f, lastf1 and lastf2 with zeros
+ */
+static inline struct ast_frame *rpt_frame_helper(struct rpt *myrpt, struct ast_frame *f, int ismuted)
```
InterLinked1, line 3573: The name is a bit confusing, especially since we also have dtmf_mute_frame_helper. Any more descriptive name we could give it? even something as simple as rpt_shift_mute_frames or something

Allan-N: I see two things happening.  The primary action is to delay our sending frames out for a short time so that we can, if need, make a change before they actually get used.  The second action is to mute frames the DTMF frames.  While I'm OK with your MUTE_FRAME() that only operates on the one frame and, at least in these cases, we're muting the new frame and possibly the 2 saved frames.

For the main function, my suggestion was `rpt_frame_helper` (it's helping rpt process the audio frames).  But, if you don't like "helper" then maybe something like `rpt_process_frame` (we use this to process the frames).  What I don't like is saying that this is specifically for DTMF.

and one might also consider changing the `ismuted` arg to be `mute` or `mute_audio`

apps/app_rpt.c
Allan-N: The comments still include `If muted, old lastf2, lastf1, lastf2 and f are filled with zeros`.  That's 4 frames being zero'd out.

apps/app_rpt/app_rpt.h
```diff
@@ -928,3 +928,5 @@ void __donodelog_fmt(struct rpt *myrpt, const char *file, int lineno, const char
 
 void rpt_event_process(struct rpt *myrpt);
 void *rpt_call(void *this);
+
+#define RPT_MUTE_FRAME_IE(f) if (f) memset(f->data.ptr, 0, f->datalen)
```
Allan-N, line 932: I'm not convinced that the _IE helps.  Would RPT_MUTE_FRAME() ... that just handles a null frame ... be sufficient?

mkmer:
https://github.com/AllStarLink/app_rpt/pull/538#discussion_r1989334435

We are dealing with 4 cleared frames.... f, lastf1, lastf2 -> then we put lastf2 -> last_frame, lastf1 -> lastf2, and f -> lastf1.
I suppose we are not actually filling old lastf2....

InterLinked1 (approved):

apps/app_rpt.c
```diff
@@ -3551,27 +3551,34 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
 	}
 }
 
-static inline void mute_frame_helper(struct rpt *myrpt, struct ast_frame *f, int ismuted)
+/*! \brief Zero data in myrpt->lastf1 and lastf2 registers (muting audio)
+ * \param myrpt The rpt structure to mute
+ */
+static inline void mute_frame_helper(struct rpt *myrpt)
```
InterLinked1, line 3558: Now that we have a macro for this, do you think it makes sense to just use these macros directly instead of having a mute_frame_helper? It seems like it's only used a few times anyways. Might increase clarity a bit in places like line 3575

mkmer: I think it may make more sense - Let's see what Allan thinks.

apps/app_rpt.c
```diff
@@ -3551,27 +3551,35 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
...
-	}
+	RPT_MUTE_FRAME(myrpt->lastf1);
+	RPT_MUTE_FRAME(myrpt->lastf2);
 }
 
+/*! \brief Shifts frames: myrpt->lastf1 -> myrpt->lastf2, f -> myrpt->lastf1.
+ * If muted, lastf1, lastf2 and f are replaced with &ast_null_frame pointer and original frame is freed.
+ * \param myrpt - the rpt structure
+ * \param f - the frame to be stored in lastf1
+ * \param mute - if true, the frame is muted by filling last_frame, f, lastf1 and lastf2 with zeros
```
Allan-N, line 3568: The "If muted, ..." comment above should replace this "\param mute" comment (because you don't need to say the same thing twice, because this comment still talks about zero'ing, AND because this comment suggests that we are muting 4 frames when we only touch 3.

apps/app_rpt.c
Allan-N: I think we should keep `mute_frame_helper()`.  Why?  Right now we are keeping 2 frames (lastf1, lastf2).  What happens when we want to add another.  Then, we have to sprinkle extra RPT_MUTE_FRAME's in lots of places vs. the much smaller handful of functions that know how many frames we are keeping.  In fact, I might suggest we abstract all of the places that reference lastf1 & lastf2 ... but maybe do that later :)

PR #540, opened 2025-03-04
pre-commit.yml: Use specific version of codespell.

mkmer:
I should have locked this to a version -> using master is not a good idea.


Allan-N (approved):

mkmer:
Addressed this in clang-format PR

PR #542, opened 2025-03-06
Multiple: Add string length protection to all sscanf() calls

mkmer:
Add buffer overflow protection to sscanf() string reads.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1687,9 +1687,9 @@ static inline void init_text_frame(struct ast_frame *wf)
 static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *str)
 {
 	/* XXX cmd, dst, and src should be validated. Why is remote_data src[300] in other locations?
-	* Is this a typo here?  Why would dest be any bigger than src?
-	*/
-	char tmp1[512], cmd[300] = "", dest[300], src[30], c;
+	 * Is this a typo here?  Why would dest be any bigger than src?
+	 */
+	char tmp1[RPT_TMP_SZ + 1], cmd[RPT_CMD_SZ + 1] = "", dest[RPT_DEST_SZ + 1], src[RPT_SRC_SZ + 1], c;
 	int i, seq, res, ts, rest;
 	struct ast_frame wf;
 
@@ -1737,7 +1737,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 	if (*str == 'M') {
 		rest = 0;
-		if (sscanf(str, "%s %s %s %n", cmd, src, dest, &rest) < 3) { /*! \todo We should limit to sizeof(cmd, src, dest) */
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_SRC_SZ) S_FMT(RPT_DEST_SZ) "%n", cmd, src, dest, &rest) < 3) {
 			ast_log(LOG_WARNING, "Unable to parse message string %s\n", str);
 			return;
 		}
@@ -1765,7 +1765,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		return;
 	}
 	if (*str == 'T') {
-		if (sscanf(str, "%s %s %s", cmd, src, dest) != 3) { /*! \todo We should limit to sizeof(cmd, src, dest) */
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_SRC_SZ) S_FMT(RPT_DEST_SZ), cmd, src, dest) != 3) {
 			ast_log(LOG_WARNING, "Unable to parse telem string %s\n", str);
 			return;
 		}
@@ -1797,7 +1797,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 
 	if (*str == 'C') {
-		if (sscanf(str, "%s %s %s %s", cmd, src, tmp1, dest) != 4) { /*! \todo We should limit to sizeof(cmd, src, tmp1, dest) */
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_SRC_SZ) S_FMT(RPT_TMP_SZ) S_FMT(RPT_DEST_SZ), cmd, src, tmp1, dest) != 4) {
 			ast_log(LOG_WARNING, "Unable to parse ctcss string %s\n", str);
 			return;
 		}
@@ -1816,7 +1816,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 
 	if (*str == 'K') {
-		if (sscanf(str, "%s %s %s %d %d", cmd, dest, src, &seq, &ts) != 5) { /*! \todo We should limit to sizeof(cmd, src, dest) */
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_DEST_SZ) S_FMT(RPT_SRC_SZ) N_FMT(d) N_FMT(d), cmd, dest, src, &seq, &ts) != 5) {
 			ast_log(LOG_WARNING, "Unable to parse keying string %s\n", str);
 			return;
 		}
@@ -1873,14 +1873,14 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		return;
 	}
 	if (*str == 'I') {
-		if (sscanf(str, "%s %s %s", cmd, src, dest) != 3) { /*! \todo We should limit to sizeof(cmd, src, dest) */
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_SRC_SZ) S_FMT(RPT_DEST_SZ), cmd, src, dest) != 3) {
 			ast_log(LOG_WARNING, "Unable to parse ident string %s\n", str);
 			return;
 		}
 		mdc1200_notify(myrpt, src, dest);
 		strcpy(dest, "*");
 	} else {
-		if (sscanf(str, "%s %s %s %d %c", cmd, dest, src, &seq, &c) != 5) {/*! \todo We should limit to sizeof(cmd, src, dest) */
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_DEST_SZ) S_FMT(RPT_SRC_SZ) N_FMT(d) " %c", cmd, dest, src, &seq, &c) != 5) {
 			ast_log(LOG_WARNING, "Unable to parse link string %s\n", str);
 			return;
 		}
@@ -2166,7 +2166,7 @@ static int handle_remote_dtmf_digit(struct rpt *myrpt, char c, char *keyed, int 
 static int handle_remote_data(struct rpt *myrpt, const char *str)
 {
 	/* Should src[300] be src[30] as in handle_link_data?*/
-	char cmd[300], dest[300], src[300], c;
+	char cmd[RPT_CMD_SZ + 1], dest[RPT_DEST_SZ + 1], src[RPT_SRC_SZ + 1], c;
 	int seq, res;
 
 	/* put string in our buffer */
@@ -2191,9 +2191,9 @@ static int handle_remote_data(struct rpt *myrpt, const char *str)
 	if (*str == 'T')
 		return 0;
 
-#ifndef	DO_NOT_NOTIFY_MDC1200_ON_REMOTE_BASES
+#ifndef DO_NOT_NOTIFY_MDC1200_ON_REMOTE_BASES
 	if (*str == 'I') {
-		if (sscanf(str, "%s %s %s", cmd, src, dest) != 3) {
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_SRC_SZ) S_FMT(RPT_DEST_SZ), cmd, src, dest) != 3) {
 			ast_log(LOG_WARNING, "Unable to parse ident string %s\n", str);
 			return 0;
 		}
@@ -2204,7 +2204,7 @@ static int handle_remote_data(struct rpt *myrpt, const char *str)
 	if (*str == 'L') {
 		return 0;
 	}
-	if (sscanf(str, "%s %s %s %d %c", cmd, dest, src, &seq, &c) != 5) {
+	if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_DEST_SZ) S_FMT(RPT_SRC_SZ) N_FMT(d) " %c", cmd, dest, src, &seq, &c) != 5) {
 		ast_log(LOG_WARNING, "Unable to parse link string %s\n", str);
 		return 0;
 	}
@@ -3551,9 +3551,8 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 	}
 
 	if (f->frametype == AST_FRAME_TEXT && myrpt->rxchankeyed) {
-		char myrxrssi[32];
-
-		if (sscanf((char *) f->data.ptr, "R %s", myrxrssi) == 1) {
+		char myrxrssi[RSSI_SZ + 1];
+		if (sscanf((char *) f->data.ptr, "R " S_FMT(RSSI_SZ), myrxrssi) == 1) {
 			myrpt->rxrssi = atoi(myrxrssi);
 			ast_debug(8, "[%s] rxchannel rssi=%i\n", myrpt->name, myrpt->rxrssi);
 			if (myrpt->p.votertype == 2)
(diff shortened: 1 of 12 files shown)


mkmer:
Well, this is "right" but the compiler doesn't like it (even though my local build has no warning).
Not sure what to do about that.

mkmer:
I'm a bit confused by that statement:
```extern int sscanf (const char *__restrict __s,
		   const char *__restrict __format, ...) __THROW;```
It's looking for the same restrictions on the input string which is also passed as a `char *`

Why doesn't my local compile throw the same warning?  I thought we had matched the compile flags?!?

**Comment by InterLinked1** [comment]
> I'm a bit confused by that statement:
> 
> ```
> extern int sscanf (const char *__restrict __s,
> 		   const char *__restrict __format, ...) __THROW;
> ```
> 
> It's looking for the same restrictions on the input string which is also passed as a `char *`

Generally, it's best for format strings to be macros and not variables at all. This is how most if not all stuff in Asterisk is

> Why doesn't my local compile throw the same warning? I thought we had matched the compile flags?!?

Are you compiling in devmode?
Do you have the same version of gcc?

**Comment by Allan-N** [comment]
>I'm a bit confused by that statement:

You are correct.  I was thinking about some [Apple] optimizations that added extra constraints on some format strings.

**Comment by mkmer** [comment]
Well, I don't see a way to get `sizeof()` into a string, macro or otherwise. The only way I see right now is to hard code the sizes in each `sscanf()`.  I hope that's not the answer....

**Comment by InterLinked1** [comment]
> Well, I don't see a way to get `sizeof()` into a string, macro or otherwise. The only way I see right now is to hard code the sizes in each `sscanf()`. I hope that's not the answer....

What are you actually trying to do? Limit the sscanf read sizes to the size of the buffer it's reading into? What's the goal, to prevent truncation?

You could maybe do something like

```
#define SIZE_A 30
#define SIZE_B 40

char buf1[SIZE_A], buf2[SIZE_B];

sscanf(str, "%" SIZE_A "s %" SIZE_B "s", buf1, buf2);
```

File: apps/app_rpt.c [hunk 1]
```diff
@@ -1764,7 +1764,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 	if (*str == 'M') {
 		rest = 0;
-		if (sscanf(str, "%s %s %s %n", cmd, src, dest, &rest) < 3) { /*! \todo We should limit to sizeof(cmd, src, dest) */
+		if (sscanf(str, "%299s %29s %299s %n", cmd, src, dest, &rest) < 3) {
```
InterLinked1, line 1767: Now there's magic numbers everywhere... it would be nice if they could just be defined in one place

mkmer: Me to (I hate the magic numbers).  This "works" without compiler errors. 
I've can't seem to find a macro or other way to generate a fixed format string that is dynamically created at compile time.  If you find one let me know.

InterLinked1: Why does it need to be dynamic, exactly?
In the example I showed, the number only appears once in a #define. I've seen this done elsewhere in the Asterisk codebase

mkmer: I think it is complaining that it can not check the vars against the format string because it's not static.
What I mean by dynamic is that the compiler puts the right number is (like sizeof()).  The string really doesn't need to be "dynamic" at run time, just compile time

InterLinked1: I think I'm still missing something... don't we know what the max size of the buffer should be? So we can just define that size as a macro?

mkmer: This is probably my inexperience speaking.  `sscanf()` is parsing the buffer into 3 or 4 strings which are smaller than the buffer.  The memory safe thing to do is make sure we only put `sizeof(x)-1` into the parsed variable. I suppose we COULD make all of the commands >= the buffer thus allowing a whole buffer to go into any parsed var.
So - ideally, we build a format string with `%(sizeof(x)s` such that the compiler makes sure we have it right (Except I can't figure out a way to create the format string in a way the compiler likes).

InterLinked1: I think you're missing what I'm saying. You don't need to use sizeof even, we can define the buffer sizes using `#define`, and use that in the sscanf. That is what the example shows. All of this can be resolved at compile time to a constant format string.

mkmer: Let say #define SIZE 10.
How do we get `"%" SIZE "s"` in the formatted string?

InterLinked1: I don't get it... that is the format string. The compiler will concatenate them together because they're all constant.

mkmer: I think the problem is that `#define SIZE 10` (for the size of the array) IS NOT a string, so `"%" SIZE "s"` does not work.
if `#define SIZE "10"` then the format string is fine, but we can't use `x[SIZE]` to define it.

InterLinked1: You can turn a number into a string by using `#`, e.g. `#SIZE_A`.
Or it could be provided as an argument for size like with printf.

apps/app_rpt.c
```diff
@@ -2233,7 +2233,7 @@ static int handle_remote_data(struct rpt *myrpt, const char *str)
 	if (*str == 'L') {
 		return 0;
 	}
-	if (sscanf(str, "%s %s %s %d %c", cmd, dest, src, &seq, &c) != 5) {
+	if (sscanf(str, "%" RPT_STRING(RPT_CMD_SZ) "s %" RPT_STRING(RPT_DEST_SZ) "s %" RPT_STRING(RPT_SRC_SZ) "s %d %c", cmd, dest, src, &seq, &c) != 5) {
```
mkmer, line 2236: We could do something like this, but now we need to +1 when we define the var.
(I only did these two for review)

apps/app_rpt.c
```diff
@@ -2195,7 +2195,7 @@ static int handle_remote_dtmf_digit(struct rpt *myrpt, char c, char *keyed, int
 static int handle_remote_data(struct rpt *myrpt, const char *str)
 {
 	/* Should src[300] be src[30] as in handle_link_data?*/
-	char cmd[300], dest[300], src[300], c;
+	char cmd[RPT_CMD_SZ + 1], dest[RPT_DEST_SZ + 1], src[RPT_SRC_SZ + 1], c;
```
mkmer, line 2198: + 1 for the size

apps/app_rpt.c
InterLinked1: This looks fine to me, it's not the most elegant format string but it works and avoids duplication

InterLinked1: I can't think of a more elegant way off the top of my head that doesn't sacrifice one of the other aspects...

I actually did this just an hour ago for a similar task, limiting the printf width of a variable based on a constant:

```
#define WS_FRAME_PREVIEW_LEN 250
if (strstr(wss_frame_payload(frame), str)) {
	if (wss_frame_payload_length(frame) > WS_FRAME_PREVIEW_LEN) {
		bbs_debug(3, "Received expected payload for line %d: '%." XSTR(WS_FRAME_PREVIEW_LEN) "s'...\n", line, wss_frame_payload(frame));
```

Granted, in this case, it can actually be "prettified" by just making it an argument:
```
#define WS_FRAME_PREVIEW_LEN 250
if (strstr(wss_frame_payload(frame), str)) {
	if (wss_frame_payload_length(frame) > WS_FRAME_PREVIEW_LEN) {
		bbs_debug(3, "Received expected payload for line %d: '%.*s'...\n", line, WS_FRAME_PREVIEW_LEN, wss_frame_payload(frame));
```

I hardly ever use `fscanf` so I'm not sure if the same idea works here but have you tried that?

mkmer: Cool - so I don't need to make my own string converter.  The rest will be the same (+1 in the var define)

mkmer: I don't see XSTR() in the base code... is this something you defined or am I missing it?

InterLinked1: That is a custom define, same as you did. I don't think it's defined in an Asterisk header file already.

Allan-N (commented):
Two comments :

First, I think I found a minor variation to your #define that may help a bit.  What do you think of the following :
```
#include <stdio.h>
#include <stdlib.h>

#define BUF_SZ 5
#define RPT_TO_STRING(x) #x
#define XSTR(x) "%" RPT_TO_STRING(x) "s"

int
main(int argc, char **argv)
{
    int n;
    char str1[BUF_SZ + 1];
    char str2[BUF_SZ + 1];

    n = sscanf("1234567890 abcdefghij",
 		XSTR(BUF_SZ) XSTR(BUF_SZ),
 		str1, str2);
    printf("n = %d, str1 = %s, str2 = %s\n", n, str1, str2);
}
```
The change here is that XSTR() builds the complete sscanf %123s format string :)


The other comment I will make is that I'm not wild about having all of the size #define's all bunched together at the top of the files.  I think the sizes each need some context and would be better placed near where there are allocated/used, possibly with a comment about why a given length was selected (e.g. what a sample string might look like).  Using the chan_echolink changes as an example, you would keep the `#define RPT_TO_STRING` and `#define XSTR` at the top of the file but, down below, you would have :
```
#define SNAPSHOT_ID_SZ 49
static char snapshot_id[SNAPSHOT_ID_SZ + 1] = { '0', 0 };
```

apps/app_rpt.c
```diff
@@ -1764,7 +1764,12 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 	if (*str == 'M') {
 		rest = 0;
-		if (sscanf(str, "%s %s %s %n", cmd, src, dest, &rest) < 3) { /*! \todo We should limit to sizeof(cmd, src, dest) */
+		if (sscanf(str,
```
InterLinked1, line 1767: Seem to be two tabs here instead of one
Also, I don't think every single argument needs to be on its own line, just the big one can have its own

apps/app_rpt/app_rpt.h
```diff
@@ -103,6 +103,16 @@ typedef struct {
 #define	PATCH_DIALPLAN_TIMEOUT 1500
 
 #define	RPT_LOCKOUT_SECS 10
+#define RPT_CMD_SZ 299
+#define RPT_SRC_SZ 299
+#define RPT_DEST_SZ 299
+#define RPT_TMP_SZ 511
+#define RSSI_SZ 31
+#define RPT_TO_STRING(x) #x
+#define XSTR(x) RPT_TO_STRING(x)
```
InterLinked1, line 112: The macro was named `XSTR` in my project, but it doesn't have to be here if you have a better name for it, that wasn't an Asterisk thing or anything.

channels/chan_usbradio.c
```diff
@@ -85,6 +85,9 @@
 #define	PP_MASK 0xbffc
 #define	PP_PORT "/dev/parport0"
 #define	PP_IOPORT 0x378
+#define BUF_SZ 15
```
InterLinked1, line 88: buf size for what?

InterLinked1 (changes requested):
I've left a few comments here once each, but they apply to multiple instances in the code

apps/app_rpt/app_rpt.h
mkmer: XSTR is nice and small, seems to make sense.  Don't see why not keep it.

channels/chan_usbradio.c
mkmer: How about STR_SZ?  It's the string size

apps/app_rpt/app_rpt.h
```diff
@@ -109,7 +109,7 @@ typedef struct {
 #define RPT_TMP_SZ 511
 #define RSSI_SZ 31
 #define RPT_TO_STRING(x) #x
-#define XSTR(x) RPT_TO_STRING(x)
+#define XSTR(x) "%" RPT_TO_STRING(x) "s "
```
InterLinked1, line 112: Instead of XSTR, how about BUF_FMT or something like that? Usually macro names ending in FMT are pretty common in Asterisk for format strings

Allan-N: SCAN_FMT() ?

apps/app_rpt.c
```diff
@@ -1792,7 +1792,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		return;
 	}
 	if (*str == 'T') {
-		if (sscanf(str, "%s %s %s", cmd, src, dest) != 3) { /*! \todo We should limit to sizeof(cmd, src, dest) */
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_SRC_SZ) S_FMT(RPT_DEST_SZ), cmd, src, dest) != 3) {
```
Allan-N, line 1795: I suspected that we might run into an issue here ... and we have.  The `S_FMT()` define maps to a format string like `%299s ` WITH A TRAILING SPACE.  In this case, the original format string `"%s %s %s"` now expects a space at the end :-(.  You may need two variants of `S_FMT()`, one with and one without the trailing space.

apps/app_rpt.c
```diff
@@ -1824,7 +1824,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 
 	if (*str == 'C') {
-		if (sscanf(str, "%s %s %s %s", cmd, src, tmp1, dest) != 4) { /*! \todo We should limit to sizeof(cmd, src, tmp1, dest) */
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_SRC_SZ) S_FMT(RPT_TMP_SZ) S_FMT(RPT_DEST_SZ), cmd, src, tmp1, dest) != 4) {
```
Allan-N, line 1827: The new format string should not include a trailing space.

apps/app_rpt.c
```diff
@@ -1900,14 +1900,14 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		return;
 	}
 	if (*str == 'I') {
-		if (sscanf(str, "%s %s %s", cmd, src, dest) != 3) { /*! \todo We should limit to sizeof(cmd, src, dest) */
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_SRC_SZ) S_FMT(RPT_DEST_SZ), cmd, src, dest) != 3) {
```
Allan-N, line 1903: The new format string should not include a trailing space.

apps/app_rpt.c
```diff
@@ -2222,7 +2222,7 @@ static int handle_remote_data(struct rpt *myrpt, const char *str)
 
 #ifndef	DO_NOT_NOTIFY_MDC1200_ON_REMOTE_BASES
 	if (*str == 'I') {
-		if (sscanf(str, "%s %s %s", cmd, src, dest) != 3) {
+		if (sscanf(str, S_FMT(RPT_CMD_SZ) S_FMT(RPT_SRC_SZ) S_FMT(RPT_DEST_SZ), cmd, src, dest) != 3) {
```
Allan-N, line 2225: The new format string should not include a trailing space.

apps/app_rpt.c
```diff
@@ -3587,9 +3587,8 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 	}
 
 	if (f->frametype == AST_FRAME_TEXT && myrpt->rxchankeyed) {
-		char myrxrssi[32];
-
-		if (sscanf((char *) f->data.ptr, "R %s", myrxrssi) == 1) {
+		char myrxrssi[RSSI_SZ + 1];
+		if (sscanf((char *) f->data.ptr, "R " S_FMT(RSSI_SZ), myrxrssi) == 1) {
```
Allan-N, line 3591: The new format string should not include a trailing space.

apps/app_rpt.c
```diff
@@ -4987,7 +4986,7 @@ static void *rpt(void *this)
 			}
 
 			/* gps_data format monotonic time, epoch, latitude, longitude, elevation */
-			if (sscanf(gps_data, "%llu %*u %s %s %s", &u_mono, lat, lon, elev) != 4) {
+			if (sscanf(gps_data, "%llu %*u " S_FMT(LAT_SZ) S_FMT(LON_SZ) S_FMT(ELEV_SZ), &u_mono, lat, lon, elev) != 4) {
```
Allan-N, line 4989: The new format string should not include a trailing space.

Allan-N (changes requested):

apps/app_rpt.c
mkmer: I don't think this matters??
I've tested `%s%s%s`, `%s %s %s `, and `%s %s %s` and all appear to resolve the same.
The most generic has the `"s "` so that we can use it with other "%d"... but it appears the space doesn't matter?
Looks like we can go with no space as well...

Allan-N: Interesting.  And the man page actually (kinda) explains this.  The %s format matches non-white-space characters.

InterLinked1 (approved):

Allan-N (approved):

PR #543, opened 2025-03-06
rpt_cli: Fix rpt cmd <node> cmd. 

mkmer:
Parameter size is too small for actual command line parameters.
Index was including a `,` producing a malformed command line.

### DIFF
--- a/apps/app_rpt/app_rpt.h
+++ b/apps/app_rpt/app_rpt.h
@@ -550,11 +550,11 @@ struct sysstate {
 #define CMD_STATE_EXECUTING 3
 
 struct rpt_cmd_struct {
-    int state;
-    int functionNumber;
-    char param[MAXDTMF];
-    char digits[MAXDTMF];
-    int command_source;
+	int state;
+	int functionNumber;
+	char param[MAXMACRO];
+	char digits[MAXDTMF];
+	int command_source;
 };
 
 enum {TOP_TOP,TOP_WON,WON_BEFREAD,BEFREAD_AFTERREAD};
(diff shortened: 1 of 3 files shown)


apps/app_rpt/app_rpt.h
```diff
@@ -552,8 +552,8 @@ struct sysstate {
 struct rpt_cmd_struct {
     int state;
     int functionNumber;
-    char param[MAXDTMF];
-    char digits[MAXDTMF];
+	char param[MAXMACRO];
```
InterLinked1, line 555: If you're fixing formatting, could you fix the alignment for the whole struct at once? Otherwise it's not aligned with the rest of the struct

apps/app_rpt/rpt_functions.c
```diff
@@ -1800,7 +1800,7 @@ int function_cmd(struct rpt *myrpt, char *param, char *digitbuf, int command_sou
 
 	if (param) {
 		if (*param == '#') {	/* to execute asterisk cli command */
-			ast_cli_command(rpt_nullfd(), param + 1);
+			ast_cli_command(rpt_nullfd(), param + 2);
```
InterLinked1, line 1803: Why + 2? Just curious, it's not obvious to me where that came from

mkmer: The paramter is passed with a `,` -> so we send `#,rpt cmd <blah> <blah>`, we get `,rpt cmd <blah> <blah>`
Clearly no one was using it, there aren't may command line commands less than 32 bytes long...

mkmer: Full command is `rpt cmd <node> cmd # <your command line command>`

apps/app_rpt/app_rpt.h
mkmer: Strange -> the original stuff is spaces instead of tabs.

PR #544, opened 2025-03-07
app_rpt.c: use ast_str for macrobuf

mkmer:
Refactor `myrpt->macrobuf` to use `ast_str*`. 
Create function `macro_append()` for securely adding commands to the buffer.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -834,17 +834,10 @@ void rpt_event_process(struct rpt *myrpt)
 		if (!cmd) {
 			continue;
 		}
-		if (action == 'F') {	/* execute a function */
-			rpt_mutex_lock(&myrpt->lock);
-			if ((sizeof(myrpt->macrobuf) - strlen(myrpt->macrobuf)) <= strlen(cmd)) { /* Make sure we have 1 extra char for null */
-				ast_log(LOG_WARNING, "Could not execute event %s for %s: Macro buffer overflow\n", cmd, argv[1]);
-			} else {
-				ast_verb(3, "Event on node %s doing macro %s for condition %s\n", myrpt->name, cmd, v->value);
-				myrpt->macrotimer = MACROTIME;
-				strncat(myrpt->macrobuf, cmd, sizeof(myrpt->macrobuf) - strlen(myrpt->macrobuf)); /* no -1, want null fill from strncat */
-			}
-			rpt_mutex_unlock(&myrpt->lock);
-		} else if (action == 'C') {	/* execute a command */
+		if (action == 'F') { /* execute a function */
+			ast_verb(3, "Event on node %s doing macro %s for condition %s\n", myrpt->name, cmd, v->value);
+			macro_append(myrpt, cmd);
+		} else if (action == 'C') { /* execute a command */
 			/* make a local copy of the value of this entry */
 			myval = ast_strdupa(cmd);
 			/* separate out specification into comma-delimited fields */
@@ -886,7 +879,7 @@ void rpt_event_process(struct rpt *myrpt)
 				ast_log(LOG_WARNING, "Could not execute event %s for %s: Command buffer in use\n", cmd, argv[1]);
 			}
 			rpt_mutex_unlock(&myrpt->lock);
-		} else if (action == 'S') {	/* execute a shell command */
+		} else if (action == 'S') { /* execute a shell command */
 			char *cp;
 
 			ast_verb(3, "Event on node %s doing shell command %s for condition %s\n", myrpt->name, cmd, v->value);
@@ -2568,17 +2561,8 @@ static void do_scheduler(struct rpt *myrpt)
 		}
 		if (myrpt->linkactivitytimer >= myrpt->p.lnkacttime) {
 			/* Execute lnkactmacro */
-			rpt_mutex_lock(&myrpt->lock);
-			if ((sizeof(myrpt->macrobuf) - strlen(myrpt->macrobuf)) <= strlen(myrpt->p.lnkactmacro)) { /* Make sure we have 1 extra char for null */
-				ast_log(LOG_WARNING, "Link Activity timer could not execute macro %s: Macro buffer full\n", myrpt->p.lnkactmacro);
-			} else {
-				ast_debug(5, "Executing link activity timer macro %s\n", myrpt->p.lnkactmacro);
-				rpt_mutex_lock(&myrpt->lock);
-				myrpt->macrotimer = MACROTIME;
-				strncat(myrpt->macrobuf, myrpt->p.lnkactmacro, sizeof(myrpt->macrobuf) - strlen(myrpt->macrobuf));
-				rpt_mutex_unlock(&myrpt->lock);
-			}
-			rpt_mutex_unlock(&myrpt->lock);
+			ast_debug(5, "Executing link activity timer macro %s\n", myrpt->p.lnkactmacro);
+			macro_append(myrpt, myrpt->p.lnkactmacro);
 			myrpt->linkactivitytimer = 0;
 			myrpt->linkactivityflag = 0;
 		}
@@ -2590,17 +2574,8 @@ static void do_scheduler(struct rpt *myrpt)
 		else {
 			myrpt->rptinacttimer = 0;
 			myrpt->rptinactwaskeyedflag = 0;
-			rpt_mutex_unlock(&myrpt->lock);
-			if ((sizeof(myrpt->macrobuf) - strlen(myrpt->macrobuf)) <= strlen(myrpt->p.rptinactmacro)) { /* Make sure we have 1 extra char for null */
-				ast_log(LOG_WARNING, "Rpt inactivity timer could not execute macro %s: Macro buffer full\n", myrpt->p.rptinactmacro);
-			} else {
-				ast_debug(5, "Executing rpt inactivity timer macro %s\n", myrpt->p.rptinactmacro);
-				if ((sizeof(myrpt->macrobuf) - strlen(myrpt->macrobuf)) > strlen(myrpt->p.rptinactmacro)) {
-					myrpt->macrotimer = MACROTIME;
-					strncat(myrpt->macrobuf, myrpt->p.rptinactmacro, sizeof(myrpt->macrobuf) - strlen(myrpt->macrobuf));
-				}
-			}
-			rpt_mutex_unlock(&myrpt->lock);
+			ast_debug(5, "Executing rpt inactivity timer macro %s\n", myrpt->p.rptinactmacro);
+			macro_append(myrpt, myrpt->p.rptinactmacro);
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


apps/app_rpt.c
```diff
@@ -4769,6 +4722,12 @@ static void *rpt(void *this)
 	sprintf(tmpstr, "%s/%s", myrpt->p.archivedir, myrpt->name);
 	mkdir(tmpstr, 0775);
 	myrpt->ready = 0;
+	myrpt->macrobuf = ast_str_create(MAXMACRO);
+	if (!myrpt->macrobuf) {
+		ast_log(LOG_ERROR, "Failed to allocate macro buffer\n");
```
InterLinked1, line 4727: no error message needed

apps/app_rpt/rpt_config.c
```diff
@@ -1318,29 +1318,17 @@ void load_rpt_vars(int n, int init)
...
 
-	if (busy) {
-		ast_log(LOG_WARNING, "Function decoder busy on app_rpt command macro.\n");
+	ast_debug(1, "rpt_push_alt_macro %s\n", sptr);
+	for (x = 0; *(sptr + x) && (x < sizeof(tmp) - 1); x++) {
+		tmp[x] = *(sptr + x) | 0x80;
 	}
-
-	return busy;
+	*(sptr + x) = 0; /* Do we really need to do this? it was 0 to end loop*/
```
InterLinked1, line 1328: The condition is `*(sptr + x) && (x < sizeof(tmp) - 1)` so if the latter were not true, I don't think this is necessarily true

apps/app_rpt/rpt_utils.c
```diff
@@ -228,3 +228,12 @@ time_t rpt_time_monotonic(void)
 	
 	return ts.tv_sec;
 }
+
+int macro_append(struct rpt *myrpt, const char *cmd) {
```
InterLinked1, line 232: Opening { should be on its line

apps/app_rpt/rpt_config.c
mkmer: Well, I added the size verification after I added the comment.  Either way, sptr is the input, tmp is the output.  I don't think we should be changing it at all.

InterLinked1: Yeah, it's also a little unclear since that variable is a char* rather than const char*. Probably because we are modifying it here, but if we don't need to for the caller, would make sense to remove this line and make it a const char*

Allan-N: I think you've made this too complicated.  We have a string passed in, sptr.  We want to take each character of the string and pick the high bit.  Isn't this as simple as :
```
char *altstr, *cp;
altstr = strdup(sptr);
for (cp = altstr; *cp; cp++) {
    *cp |= 0x80;        
}
macro_append(myrpt, altstr);
free(altstr);
```

Alternatively, you can play games with the ast_str buffer by capturing the length before appending the string, append the string, and then pick the high bit for every character appended.  That would eliminate the need to strdup/free.

apps/app_rpt.c
```diff
@@ -4769,6 +4722,11 @@ static void *rpt(void *this)
 	sprintf(tmpstr, "%s/%s", myrpt->p.archivedir, myrpt->name);
 	mkdir(tmpstr, 0775);
 	myrpt->ready = 0;
+	myrpt->macrobuf = ast_str_create(MAXMACRO);
```
Allan-N, line 4725: Do we need to add a call to `ast_free(myrpt->macrobuf)` in the cleanup code?

Allan-N (changes requested):

apps/app_rpt.c
mkmer: Well... I don't see where we free the myrpt-> structure....

mkmer: It looks like myrpt is static -> so let's just create one if we don't have one.

Allan-N: Yes, they are statically allocated :
```
struct rpt rpt_vars[MAXRPTS];
```

apps/app_rpt.c
```diff
@@ -4901,7 +4861,7 @@ static void *rpt(void *this)
 #endif
 	}
 	if (myrpt->p.startupmacro) {
-		snprintf(myrpt->macrobuf, MAXMACRO - 1, "PPPP%s", myrpt->p.startupmacro);
+		ast_str_append(&myrpt->macrobuf, 0, "PPPP%s", myrpt->p.startupmacro);
```
Allan-N, line 4864: OK.  Here, we are initializing macrobuf.  So, instead of ast_str_append() I think we should be using ast_str_set()

apps/app_rpt.c
```diff
@@ -7068,7 +7031,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 	myrpt->hfscanmode = 0;
 	myrpt->hfscanstatus = 0;
 	if (myrpt->p.startupmacro) {
-		snprintf(myrpt->macrobuf, MAXMACRO - 1, "PPPP%s", myrpt->p.startupmacro);
+		ast_str_append(&myrpt->macrobuf, 0, "PPPP%s", myrpt->p.startupmacro);
```
Allan-N, line 7034: Again, we are initializing macrobuf. So, instead of ast_str_append() I think we should be using ast_str_set()

Allan-N (changes requested):

apps/app_rpt.c
mkmer: This is a bit of an interesting "catch" -> Why are we initializing the buffer in two places?
Seems like we should only init `myrpt` in ONE place.
This could also mean I need to create the buffer in two places!

apps/app_rpt.c
mkmer: Well... it looks like we clear the whole structure on reload/load:
https://github.com/AllStarLink/app_rpt/blob/95166c29f49344f2fe5afb6220674ea590c857bd/apps/app_rpt.c#L5611

So I need to find a spot to free this where we won't crash while reloading (after the repeater "stops" and before we `memset`).

apps/app_rpt.c
```diff
@@ -4769,6 +4722,13 @@ static void *rpt(void *this)
 	sprintf(tmpstr, "%s/%s", myrpt->p.archivedir, myrpt->name);
 	mkdir(tmpstr, 0775);
 	myrpt->ready = 0;
+	if (!myrpt->macrobuf) {
+		myrpt->macrobuf = ast_str_create(MAXMACRO);
+	}
+	if (!myrpt->macrobuf) {
```
InterLinked1, line 4728: This branch can be moved inside the previous one, right?
Both are correct but that would be slightly more optimized

apps/app_rpt.c
```diff
@@ -7081,6 +7049,17 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 	myrpt->lastitx = !myrpt->lastitx;
 	myrpt->tunerequest = 0;
 	myrpt->tunetx = 0;
+	if (!myrpt->macrobuf) {
+		myrpt->macrobuf = ast_str_create(MAXMACRO);
+	}
+	if (!myrpt->macrobuf) {
```
InterLinked1, line 7055: Same here, this can only be true if the previous branch wasn't

apps/app_rpt.c
```diff
@@ -7401,10 +7380,13 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
...
 		}
 		rpt_mutex_lock(&myrpt->lock);
-		c = myrpt->macrobuf[0];
+		str = ast_str_buffer(myrpt->macrobuf);
+		len = ast_str_strlen(myrpt->macrobuf);
+		c = str[0];
 		if (c && !myrpt->macrotimer) {
 			myrpt->macrotimer = MACROTIME;
-			memmove(myrpt->macrobuf, myrpt->macrobuf + 1, MAXMACRO - 1);
+			ast_copy_string(str, str + 1, len - 1);
```
InterLinked1, line 7388: I'm not sure if ast_copy_string is safe to use for using memory in an overlapping buffer? Or in any case, I'm not sure why memmove wouldn't work here, could probably keep that and truncate on the following line like you are now if that just zeros out the last char

apps/app_rpt/rpt_config.c
```diff
@@ -1318,29 +1318,16 @@ void load_rpt_vars(int n, int init)
...
-		}
-		*(sptr + x) = 0;
-	}
-	rpt_mutex_unlock(&myrpt->lock);
+	char *altstr, *cp;
 
-	if (busy) {
-		ast_log(LOG_WARNING, "Function decoder busy on app_rpt command macro.\n");
+	ast_debug(1, "rpt_push_alt_macro %s\n", sptr);
+	altstr = ast_strdup(sptr);
```
InterLinked1, line 1324: This can fail

apps/app_rpt.c
mkmer: true

apps/app_rpt.c
mkmer: Looks safe as memcopy :
https://github.com/asterisk/asterisk/blob/c7db162844a53be33f8cbc82f0723c3c016860b2/include/asterisk/strings.h#L413-L425

apps/app_rpt.c
```diff
@@ -5605,6 +5569,10 @@ static int load_config(int reload)
 		if (reload) {
 			for (n = 0; n < nrpts; n++) {
 				if (!strcmp(this, rpt_vars[n].name)) {
+					if (rpt_vars[n].macrobuf) {
```
mkmer, line 5572: I'm not 100% sure this is the right place to free `macrobuf` (more like 99.9%).
It looks like on a reload, we match here then `memset()` the structure to 0 
https://github.com/AllStarLink/app_rpt/blob/9db0ce2f3ca6a6edd73c1b313ec2ddfe3d62d8b4/apps/app_rpt.c#L5626
This should be a safe spot as we should be stopped at this point, no longer accessing the buffer.

apps/app_rpt.c
```diff
@@ -3895,35 +3863,19 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
...
-							myrpt->macrotimer = MACROTIME;
-							strncat(myrpt->macrobuf, val, MAXMACRO - 1);
-						}
-						rpt_mutex_unlock(&myrpt->lock);
+						macro_append(myrpt, val);
 					}
-					if (!busy)
-						strcpy(myrpt->lasttone, (char *) f->data.ptr);
+					rpt_mutex_lock(&myrpt->lock);
+					ast_copy_string(myrpt->lasttone, (char *) f->data.ptr, sizeof(myrpt->lasttone) - 1);
```
Allan-N, line 3877: You do not need/want the "- 1".  Just pass the size of the buffer (myrpt->lasttone) and let ast_copy_string ensure the trailing NUL.

apps/app_rpt.c
Allan-N: Or do you move the check to just before the memset() ?

Allan-N (changes requested):

apps/app_rpt.c
mkmer: We also get called for "normal" startup, so then the memory is not initialized and we can't be sure !0 is actually a pointer (at least that's my theory).

Allan-N: In app_rpt.c, we have :
```
struct rpt rpt_vars[MAXRPTS]; 
```
This variable, not only the stack, should start off zero'd ... right?

Allan-N (approved):

apps/app_rpt.c
mkmer: I believe this is the code that zeroes it.  I don't know what the compiler does when it allocates static memory.  Is there a definitive operation for compilers?

InterLinked1: static variables are zero initialized in the C standard, so if the intent is for them to default to 0, it doesn't *need* to be specified. Sometimes it's done just for clarity.

mkmer: So we can move it to just before the `memset()` -> I think that will be more clear.

InterLinked1:
> # Cpp-Linter Report ⚠️
> Some files did not pass the configured checks!
> 
> clang-format (v14.0.6) reports: **3 file(s) not formatted**
> * apps/app_rpt.c
> * apps/app_rpt/app_rpt.h
> * apps/app_rpt/rpt_mdc1200.c
> 
> Have any feedback or feature suggestions? [Share it here.](https://github.com/cpp-linter/cpp-linter-action/issues)

@mkmer How does one see the actual lines causing it to fail in the CI logs?
If it just says the file failed, that doesn't seem very useful.

PR #545, opened 2025-03-07
app_rpt: Use ast_str for l->linklist

mkmer:
Refactor `l->linklist` to use `ast_str`.  This eliminates the limits on connected nodes related to the fixed size of `l->linklist`

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3274,7 +3274,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			}
 			/* hang-up on call to device */
 			ast_hangup(l->pchan);
-			rpt_free_link_helper(l);
+			rpt_link_free(l);
 			rpt_mutex_lock(&myrpt->lock);
 			break;
 		}
@@ -3297,7 +3297,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			dodispgm(myrpt, l->name);
 			/* hang-up on call to device */
 			ast_hangup(l->pchan);
-			rpt_free_link_helper(l);
+			rpt_link_free(l);
 
 			rpt_mutex_lock(&myrpt->lock);
 			break;
@@ -4178,7 +4178,7 @@ static void remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 	rpt_mutex_unlock(&myrpt->lock);
 
 	ast_hangup(l->pchan);
-	rpt_free_link_helper(l);
+	rpt_link_free(l);
 }
 
 static inline void fac_frame(struct ast_frame *restrict f, float fac)
@@ -5225,7 +5225,7 @@ static void *rpt(void *this)
 				if (l->chan)
 					ast_hangup(l->chan);
 				ast_hangup(l->pchan);
-				rpt_free_link_helper(l);
+				rpt_link_free(l);
 				rpt_mutex_lock(&myrpt->lock);
 				/* re-start link traversal */
 				l = myrpt->links.next;
@@ -5434,7 +5434,7 @@ static void *rpt(void *this)
 			ast_hangup(l->chan);
 		ast_hangup(l->pchan);
 		l = l->next;
-		rpt_free_link_helper(ll);
+		rpt_link_free(ll);
 	}
 	if (myrpt->xlink == 1)
 		myrpt->xlink = 2;
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1729,10 +1729,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 	if (*str == 'L') {
 		rpt_mutex_lock(&myrpt->lock);
-		if (strlen(str + 2) > sizeof(mylink->linklist) - 1) {
-			ast_log(LOG_WARNING, "Link list too long: buffer size: %ld, link size: %ld linklist: %s\n", sizeof(mylink->linklist) - 1, strlen(str + 2), str + 2);
-		}
-		ast_copy_string(mylink->linklist, str + 2, sizeof(mylink->linklist));
+		ast_str_set(&mylink->linklist, 0, "%s", str + 2); /* Dropping the "L " of the message */
 		time(&mylink->linklistreceived);
 		rpt_mutex_unlock(&myrpt->lock);
 		ast_debug(7, "@@@@ node %s received node list %s from node %s\n", myrpt->name, str, mylink->name);
@@ -3051,7 +3048,6 @@ static inline void update_timer(int *timer_ptr, int elap, int end_val)
 	}
 }
 
-
 static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 {
 	struct ast_frame *f;
@@ -3174,7 +3170,10 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 
 		if ((!l->linklisttimer) && (l->name[0] != '0') && (!l->isremote)) {
 			struct ast_frame lf;
-			char lstr[MAXLINKLIST];
+			struct ast_str *lstr = ast_str_create(RPT_AST_STR_INIT_SIZE);
+			if (!lstr) {
+				return;
+			}
 			memset(&lf, 0, sizeof(lf));
 			lf.frametype = AST_FRAME_TEXT;
 			lf.subclass.format = ast_format_slin;
@@ -3182,16 +3181,18 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			lf.mallocd = 0;
 			lf.samples = 0;
 			l->linklisttimer = LINKLISTTIME;
-			strcpy(lstr, "L ");
+			ast_str_set(&lstr, 0, "%s", "L ");
 			rpt_mutex_lock(&myrpt->lock);
-			__mklinklist(myrpt, l, lstr + 2, sizeof(lstr) - 2, 0);
+			__mklinklist(myrpt, l, &lstr, 0);
 			rpt_mutex_unlock(&myrpt->lock);
 			if (l->chan) {
-				lf.datalen = strlen(lstr) + 1;
-				lf.data.ptr = lstr;
+				lf.datalen = ast_str_strlen(lstr) + 1;
+				lf.data.ptr = ast_str_buffer(lstr);
 				rpt_qwrite(l, &lf);
-				ast_debug(7, "@@@@ node %s sent node string %s to node %s\n", myrpt->name, lstr, l->name);
+				ast_debug(
+					7, "@@@@ node %s sent node string %s to node %s\n", myrpt->name, ast_str_buffer(lstr), l->name);
 			}
+			ast_free(lstr);
 		}
 		if (l->link_newkey == RADIO_KEY_ALLOWED_REDUNDANT) {
 			if ((l->retxtimer += elap) >= REDUNDANT_TX_TIME) {
@@ -3273,7 +3274,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			}
 			/* hang-up on call to device */
 			ast_hangup(l->pchan);
-			ast_free(l);
+			rpt_free_link_helper(l);
 			rpt_mutex_lock(&myrpt->lock);
 			break;
 		}
@@ -3296,7 +3297,8 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 			dodispgm(myrpt, l->name);
 			/* hang-up on call to device */
 			ast_hangup(l->pchan);
-			ast_free(l);
+			rpt_free_link_helper(l);
+
 			rpt_mutex_lock(&myrpt->lock);
 			break;
 		}
@@ -4176,7 +4178,7 @@ static void remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 	rpt_mutex_unlock(&myrpt->lock);
 
 	ast_hangup(l->pchan);
-	ast_free(l);
+	rpt_free_link_helper(l);
 }
 
 static inline void fac_frame(struct ast_frame *restrict f, float fac)
@@ -5223,7 +5225,7 @@ static void *rpt(void *this)
 				if (l->chan)
 					ast_hangup(l->chan);
 				ast_hangup(l->pchan);
-				ast_free(l);
+				rpt_free_link_helper(l);
 				rpt_mutex_lock(&myrpt->lock);
 				/* re-start link traversal */
 				l = myrpt->links.next;
@@ -5432,7 +5434,7 @@ static void *rpt(void *this)
 			ast_hangup(l->chan);
 		ast_hangup(l->pchan);
 		l = l->next;
-		ast_free(ll);
+		rpt_free_link_helper(ll);
 	}
 	if (myrpt->xlink == 1)
 		myrpt->xlink = 2;
@@ -6612,6 +6614,11 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		l = ast_calloc(1, sizeof(struct rpt_link));
 		if (!l) {
 			return -1;
+		}
+		l->linklist = ast_str_create(RPT_AST_STR_INIT_SIZE);
+		if (!l->linklist) {
+			ast_free(l);
+			pthread_exit(NULL);
 		}
 		l->mode = 1;
 		ast_copy_string(l->name, b1, MAXNODESTR);
--- a/apps/app_rpt/app_rpt.h
+++ b/apps/app_rpt/app_rpt.h
@@ -1,4 +1,3 @@
-
 #define VERSION_MAJOR 3
 #define VERSION_MINOR 3
 #define VERSION_PATCH 0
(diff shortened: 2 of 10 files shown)


apps/app_rpt.c
```diff
@@ -3224,24 +3220,32 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
...
 			rpt_mutex_unlock(&myrpt->lock);
 			if (l->chan) {
-				lf.datalen = strlen(lstr) + 1;
-				lf.data.ptr = lstr;
+				lf.datalen = ast_str_strlen(lstr) + 1;
+				lf.data.ptr = ast_str_buffer(lstr);
 				rpt_qwrite(l, &lf);
-				ast_debug(7, "@@@@ node %s sent node string %s to node %s\n", myrpt->name, lstr, l->name);
+				ast_debug(7,
+						  "@@@@ node %s sent node string %s to node %s\n",
```
InterLinked1, line 3243: ident seems to be two tabs + one space, should just be one tab

apps/app_rpt/app_rpt.h
```diff
@@ -71,7 +70,8 @@ typedef struct {
 /* maximum digits in DTMF buffer, and seconds after * for DTMF command timeout */
 #define	MAXDTMF 32
 #define	MAXMACRO 2048
-#define	MAXLINKLIST 5120
+#define MAXNODES 500		  /* Maximum number of nodes allowed in the link list*/
+#define AST_STR_INIT_SIZE 500 /* initial guess for ast_str size*/
```
InterLinked1, line 74: Don't define prefixes starting with `AST_` if it's not in an Asterisk header file. Maybe RPT_STR_INIT_SIZE, RPT_AST_STR_INIT_SIZE, or something like that

apps/app_rpt/app_rpt.h
```diff
@@ -71,7 +70,8 @@ typedef struct {
 /* maximum digits in DTMF buffer, and seconds after * for DTMF command timeout */
 #define	MAXDTMF 32
 #define	MAXMACRO 2048
-#define	MAXLINKLIST 5120
+#define MAXNODES 500		  /* Maximum number of nodes allowed in the link list*/
```
InterLinked1, line 73: Should there be a limit at all? What if someone wants 501 nodes?
It looks like there are three places where MAXNODES is used, where we have an array to pass to qsort to sort them. We could use a dynamic array (count # nodes then allocate) in those places and then get rid of any hardcoded limit altogether.

mkmer: Unfortunately, we still have a array of pointers  that comes out of `finddelim()`, The most we can handle is the size of this pointer array.  Expanding this list is the next "choke" point, but in reality I don't think a 500 node network is going to be very usable. 
We may be able to use the return value from __mklinklist to dynamically allocate the `strs` array - I think that should be in another PR if we do it.

apps/app_rpt.c
mkmer: I think something is odd with how GITHUB is displaying this.  It lines up perfectly in the editors.

apps/app_rpt.c
```diff
@@ -3224,24 +3220,32 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
...
 			rpt_mutex_unlock(&myrpt->lock);
 			if (l->chan) {
-				lf.datalen = strlen(lstr) + 1;
-				lf.data.ptr = lstr;
+				lf.datalen = ast_str_strlen(lstr) + 1;
+				lf.data.ptr = ast_str_buffer(lstr);
 				rpt_qwrite(l, &lf);
-				ast_debug(7, "@@@@ node %s sent node string %s to node %s\n", myrpt->name, lstr, l->name);
+				ast_debug(7,
+						"@@@@ node %s sent node string %s to node %s\n",
```
InterLinked1, line 3243: still 2 indents here instead of 1

InterLinked1 (approved):

apps/app_rpt/rpt_utils.c
```diff
@@ -84,12 +85,20 @@ char *string_toupper(char *str)
...
 }
-
-int finddelim(char *str, char *strp[], size_t limit)
-{
-	return explode_string(str, strp, limit, DELIMCHR, QUOTECHR);
-}
-
+/*!
+ * \brief Find the delimiter in a string and return a pointer array to the start of each token.
+ * Note: This modifies the string str, be suer to save an intact copy if you need it later.
```
Allan-N, line 90: Be sure

apps/app_rpt/rpt_link.c
```diff
@@ -120,6 +120,17 @@ static void check_tlink_list(struct rpt *myrpt)
 	}
 }
 
+/*! \brief free a link structure and it's ast_str linklist if it exists
+ *  \param link the link to free
+ */
+void rpt_free_link_helper(struct rpt_link *link) {
+	if (link->linklist) {
+		link->linklist = NULL;
```
Allan-N, line 128: Are these 2 lines in the reverse order?  or did you really want to NULL out the pointer you will soon free?

apps/app_rpt/rpt_link.c
```diff
@@ -518,37 +489,43 @@ int __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, size_t b
...
+				links_buf[i] = mode;
 		}
 	}
 	/* After building the string, count number of nodes (commas) in buffer string. The first
 	 * node doesn't have a comma, so we need to add 1 if there is at least one_link.  
 	 */
-	for (link_count = 0; buf[link_count]; buf[link_count]==',' ? link_count++: *buf++);
+	links_buf = ast_str_buffer(*buf);
+	buf_end = links_buf + ast_str_strlen(*buf);
+	for (link_count = 0; links_buf[link_count] && (&links_buf[link_count] < buf_end);
```
Allan-N, line 525: There's something about this line ... and the next 2 ... that just doesn't read well.

Allan-N (changes requested):

apps/app_rpt/rpt_link.c
```diff
@@ -120,6 +120,17 @@ static void check_tlink_list(struct rpt *myrpt)
 	}
 }
 
+/*! \brief free a link structure and it's ast_str linklist if it exists
+ *  \param link the link to free
+ */
+void rpt_free_link_helper(struct rpt_link *link) {
```
InterLinked1, line 126: Also { should be on its own line

apps/app_rpt/rpt_link.c
```diff
@@ -626,10 +600,11 @@ void rpt_update_links(struct rpt *myrpt)
 int connect_link(struct rpt *myrpt, char *node, int mode, int perma)
 {
 	char *s, *s1, *tele, *cp;
-	char lstr[MAXLINKLIST], *strs[MAXLINKLIST];
 	char tmp[300], deststr[325] = "", modechange = 0;
 	char sx[320], *sy;
+	char *strs[MAXNODES]; /* List of pointers to links in link list string*/
```
InterLinked1, line 605: space before end comment

InterLinked1 (changes requested):

apps/app_rpt/rpt_utils.c
mkmer: Wonder why codespell isn't finding this stuff.  I'll review the settings.

apps/app_rpt/rpt_link.c
mkmer: Do we want to stop people from referencing the memory BEFORE we free it (I think so).  Unfortunately my execution bad :)  
Probably doesn't matter, or we could lock the link... we are actually done, so no one should try to access it in the last moments of existence.

apps/app_rpt/rpt_link.c
mkmer: The goal was to limit the length we loop through in case we are missing a null terminator.  Reality is that ast_str_buffer() probably manages this good enough?  We can probably just look for the `null`

apps/app_rpt/rpt_link.c
```diff
@@ -521,11 +522,8 @@ int __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, struct ast_str **bu
 	 * node doesn't have a comma, so we need to add 1 if there is at least one_link.  
 	 */
 	links_buf = ast_str_buffer(*buf);
-	buf_end = links_buf + ast_str_strlen(*buf);
-	for (link_count = 0; links_buf[link_count] && (&links_buf[link_count] < buf_end);
-		 links_buf[link_count] == ',' ? link_count++ : *links_buf++)
-		 ;
-
+	for (link_count = 0; links_buf[link_count]; links_buf[link_count] == ',' ? link_count++ : *links_buf++)
```
Allan-N, line 525: This still doesn't look right.  Try this:
```
links_count = 0;
links_buf = ast_str_buffer(*buf);
for (i = 0; i < ast_str_strlen(*buf); i++) {
        if (links_buf[i] == ',') {
                links_count++;
        }
}
```

Allan-N (changes requested):

apps/app_rpt/rpt_link.c
mkmer: Are you saying you don't like how it works currently (or how it looks/reads)?  It is doing the same thing as it did before.
It's a bit weird how it uses `link_count` and `links_buf++` to move the pointer forward, but that's how they got it into one line.

Allan-N: I don't think what you had will work.  What is `links_buf[link_count] == ',' ? link_count++ : *links_buf++` doing?

FWIW, trying to squeeze lots of logic into one line does not help with readability.

mkmer: It's moving the pointer in 2 ways.  If we found a `,` we add to the count, if we didn't we add to the pointer.  Both counters are advancing what `links_buf[link_count]` looks at.  If links_buf[link_count] is null, we end it.

I like your more "readable" version for sure.  This one takes too much "noodling" to understand what it does.

apps/app_rpt/rpt_link.c
```diff
@@ -518,42 +490,49 @@ int __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, size_t b
...
 			continue;
 		/* downgrade everyone on this node if appropriate */
-		for (i = spos; buf[i]; i++) {
-			if (buf[i] == 'T')
-				buf[i] = mode;
-			if ((buf[i] == 'R') && (mode == 'C'))
-				buf[i] = mode;
+		links_buf = ast_str_buffer(*buf);
+		len = ast_str_strlen(*buf);
+		for (i = spos; links_buf[i] && i < len; i++) {
```
Allan-N, line 514: Since we know the length of the string, can this be simplified to `for (i = spos; i < len; i++) {` ?

Also, the other day I was connected to the East Coast Reflector and  I exec'd `rpt nodes <mynode>`.  What I noticed were non-numeric nodes.  Might this mess with the following checks for 'T', 'R', and 'C' ?

apps/app_rpt.c
```diff
@@ -3224,24 +3220,32 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
...
+			}
 			memset(&lf, 0, sizeof(lf));
 			lf.frametype = AST_FRAME_TEXT;
 			lf.subclass.format = ast_format_slin;
 			lf.offset = 0;
 			lf.mallocd = 0;
 			lf.samples = 0;
 			l->linklisttimer = LINKLISTTIME;
-			strcpy(lstr, "L ");
+			ast_str_set(&lstr, 2, "%s", "L ");
```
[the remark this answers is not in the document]
mkmer: I was thinking the size of the input is only 2, `L `, but 0 is fine too.
I think we're ok:
```
* \param max_len The maximum allowed capacity of the ast_str. Note that
 *  if the value of max_len is less than the current capacity of the
 *  ast_str (as returned by ast_str_size), then the parameter is effectively
 *  ignored.
 ```

File: apps/app_rpt/rpt_link.c (hunk 11 above)
Comment (mkmer) [reply apps/app_rpt/rpt_link.c]: Non numeric nodes: This probably will cause a problem IF there is a capitol T, we'll change it to an R or C if we are not int T mode.  
Since they are not supposed to be non numeric, and this function handles it (safely), I don't think we really need to do anything. 
I don't see a way around it either.  
Maybe look at the char just before the T and see if it's a number?  This has it's own set of challenges too.

Comment (Allan-N) [reply apps/app_rpt/rpt_link.c]: I asked the question.  I have an answer.  Move along :)

**Review by Allan-N — APPROVED** [review]

File: apps/app_rpt.c [hunk 13]
```diff
@@ -3224,24 +3220,32 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
...
+			__mklinklist(myrpt, l, &lstr, 0);
 			rpt_mutex_unlock(&myrpt->lock);
 			if (l->chan) {
-				lf.datalen = strlen(lstr) + 1;
-				lf.data.ptr = lstr;
+				lf.datalen = ast_str_strlen(lstr) + 1;
+				lf.data.ptr = ast_str_buffer(lstr);
 				rpt_qwrite(l, &lf);
-				ast_debug(7, "@@@@ node %s sent node string %s to node %s\n", myrpt->name, lstr, l->name);
+				ast_debug(7,
```
InterLinked1, line 3242: I'm not personally a fan of breaking this log message up over 5 lines. It makes sense for longer messages but this doesn't seem *that* long to me... (it was 91 columns fully formatted, previously, which seems more than reasonable to me)

If either of you disagree, I won't hold it against it, but app_rpt.c is long enough so those seem like easy lines not to add.

apps/app_rpt/app_rpt.h
InterLinked1, line 73: space before `*/` at EOL

apps/app_rpt/app_rpt.h
```diff
@@ -71,7 +70,8 @@ typedef struct {
...
+#define RPT_AST_STR_INIT_SIZE 500 /* initial guess for ast_str size*/
```
InterLinked1, line 74: space before `*/` at EOL

apps/app_rpt/rpt_link.c
```diff
@@ -712,24 +690,36 @@ int connect_link(struct rpt *myrpt, char *node, int mode, int perma)
 		l->retries = l->max_retries + 1;
 		l->disced = 2;
 		modechange = 1;
-	} else {
-		__mklinklist(myrpt, NULL, lstr, sizeof(lstr), 0);
+	} else { /* Check to see if this noed is already linked */
```
InterLinked1, line 693: node

apps/app_rpt/rpt_utils.c
```diff
@@ -84,12 +85,20 @@ char *string_toupper(char *str)
...
 	return str;
 }
+/*!
+ * \brief Find the delimiter in a string and return a pointer array to the start of each token.
+ * Note: This modifies the string str, be sure to save an intact copy if you need it later.
+ * \param str The string to search
+ * \param strp An array of pointers to the start of each token
+ * \param limit The maximum number of tokens to find
+ * \return The number of tokens found
+ */
```
InterLinked1, line 96: Don't need blank line here

apps/app_rpt/rpt_utils.c
```diff
@@ -84,12 +85,20 @@ char *string_toupper(char *str)
...
 
 int finddelim(char *str, char *strp[], size_t limit)
 {
 	return explode_string(str, strp, limit, DELIMCHR, QUOTECHR);
-}
-
+ }
```
InterLinked1, line 100: extraneous whitespace on these 2 lines (what we used to call "redness" on Gerrit... but it shows up as a different shade of green here)

apps/app_rpt/rpt_utils.h
```diff
@@ -10,18 +10,14 @@
...
-* strp- list of pointers to substrings (this is built by this function), NULL will be placed at end of list
-* limit- maximum number of substrings to process
-* delim- user specified delimiter
-* quote- user specified quote for escaping a substring. Set to zero to escape nothing.
-*
-* Note: This modifies the string str, be suer to save an intact copy if you need it later.
-*
-* Returns number of substrings found.
-*/
+ *
```
InterLinked1, line 13: This one extra line here can be removed

apps/app_rpt/rpt_utils.h
```diff
@@ -10,18 +10,14 @@
...
+ * \brief Explode a string into an array of pointers to the start of each token.
+ * \param str The string to explode (will be modified)
+ * \param strp An array of pointers to the start of each token
+ * \param delim The delimiter to use
+ * \param quote The quote character to use
+ * \return The number of substrings found.
+ */
```
InterLinked1, line 21: Remove this line too

InterLinked1 (changes requested):

apps/app_rpt.c
mkmer: The line size is set to 120 in clang-format.  This line is 123, so clang-format is suggesting breaking it up.
The recommended width on the ASTERISK indent tool is actually 90.
`# indent -i4 -ts4 -br -brs -cdw -lp -ce -nbfda -npcs -nprs -npsl -nbbo -saf -sai -saw -cs -l90 foo.c`
What should we set max length to then?

InterLinked1: I think 90 is pretty outdated based on what I've seen over the years in the Asterisk repo (I'm sure that indent command hasn't been updated in a while), I'm not sure if there's a definite magic number. But when lines do get broken up, every argument isn't put on its own line, only the long one typically gets its own. In these examples, everything else can still be sharing a line.

apps/app_rpt/rpt_cli.c
```diff
@@ -528,10 +532,10 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 			/* ### GET CONNECTED NODE INFO ####################
 			 * Traverse the list of connected nodes
 			 */
+			rpt_mutex_lock(&myrpt->lock); /* LOCK */
```
Allan-N, line 535: Are we locking again?  or are we missing an unlock?

apps/app_rpt/rpt_cli.c
```diff
@@ -541,7 +545,8 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 				}
 				if ((s = (struct rpt_lstat *) ast_malloc(sizeof(struct rpt_lstat))) == NULL) {
 					ast_log(LOG_ERROR, "Malloc failed in rpt_do_lstats\n");
-					rpt_mutex_unlock(&myrpt->lock);	// UNLOCK 
+					rpt_mutex_unlock(&myrpt->lock);	// UNLOCK
```
Allan-N, line 548: Unlocking when not locked?

apps/app_rpt/rpt_functions.c
```diff
@@ -363,7 +363,7 @@ int function_ilink(struct rpt *myrpt, char *param, char *digits, int command_sou
 	case 16:					/* Restore links disconnected with "disconnect all links" command */
 		strcpy(tmp, myrpt->savednodes);	/* Make a copy */
 		finddelim(tmp, strs, ARRAY_LEN(strs));	/* convert into substrings */
-		for (i = 0; tmp[0] && strs[i] != NULL && i < MAXLINKLIST; i++) {
+		for (i = 0; tmp[0] && strs[i] != NULL && i < ARRAY_LEN(strs); i++) {
```
Allan-N, line 366: Need to check `i < ARRAY_LEN(strs)` before referencing `strs[i]`

apps/app_rpt/rpt_link.c
```diff
@@ -518,42 +490,49 @@ int __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, char *buf, size_t b
...
 			mode = 'C';			/* indicate connecting */
-		spos = strlen(buf);		/* current buf size (b4 we add our stuff) */
-		if (spos) {
-			strcat(buf, ",");
-			spos++;
+		spos = ast_str_strlen(*buf); /* current buf size (b4 we add our stuff) */
+		if (spos > 2) {
+			ast_str_append(buf, 1, "%s", ",");
 		}
 	    one_link = 1;
```
Allan-N, line 497: The spacing on this line appears to be off

Allan-N (changes requested):

apps/app_rpt/rpt_cli.c
mkmer: Well.. it won't hurt, but it doesn't belong.  Out she goes :)

apps/app_rpt/rpt_cli.c
mkmer: We just established it was locked (twice) -> This should stay from what I can tell.

Allan-N (approved):

InterLinked1 (approved):

PR #547, opened 2025-03-07
app_rpt.c: Incorrect unlock on failure return.  

mkmer:
Function is called locked, should return locked

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3353,6 +3353,11 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 		l = l->next;
 	}
 }
+/*! \brief Post link data to Allstar. 
+ * Must be called locked.
+ * \param myrpt The rpt structure
+ * \retval 0 on success, -1 on failure
+ */	
 
 static inline int do_link_post(struct rpt *myrpt)
 {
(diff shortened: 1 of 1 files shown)


InterLinked1 (commented):
Is this actually how it's used? The bottom of the function locks before returning.
It seems strange but seems like it's supposed to be called locked and returns locked.

Allan-N (approved):

Allan-N:
>> It seems strange but seems like it's supposed to be called locked and returns locked.
>
> It looks like it. It's called locked to my eyes.

Optional : maybe add a comment at the top of the function that says "must be called locked"

apps/app_rpt.c
```diff
@@ -3353,6 +3353,11 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 		l = l->next;
 	}
 }
+/*! \brief Post link data to Allstar.
```
InterLinked1, line 3356: We do want a blank line before the comments, but not after :)

For the brief, technically it may not be Allstar since I think that link is configurable, so maybe just "Post link data"?

Also, if you want to be fancy, the note about locking can be a `\note` in doxygen

And just a general comment, for doxygen, you don't always have to include every component, for example if the param is obvious or you just want to add a note, etc.

mkmer: So, copilot generates the whole section when I start to build `/* \brief`.  I could delete it, but for the most part it's correct.
I agree, `myrpt` is obvious, the `return` values are not unless you hunt in the code, so I think they are nice to have.
I'm not very familiar with doxygen and "doc stuff".  I suppose that will be next thing to noodle out.

InterLinked1 (approved):

PR #548, opened 2025-03-07
app_rpt.c: rxchannel_read remove incorrect unlocks.

mkmer:
Couple of double unlocks.


InterLinked1 (approved):

Allan-N (changes requested):
Does line 3916 also need to be removed?

mkmer:
> Does line 3916 also need to be removed?

Looks like it...

Allan-N (approved):

PR #552, opened 2025-03-08
app_rpt.c: Use ast_asprintf instead of malloc/memset.

mkmer:
Use `ast_asprintf()` reducing the `malloc`/`memset` repeated code


InterLinked1 (changes requested):
You need to check for ast_asprintf failing (returning -1). The buffer is undefined if it fails.
Also I already have 2 out of 3 of these in that other PR, so I'll just amend mine to include the 3rd one I missed

mkmer:
Perfect.  I thought these were not in your PR.

PR #554, opened 2025-03-10
app_gps.c: Use ARRAY_LEN for strs

mkmer:
Use ARRAY_LEN for strs                  
Match explode_string to rpt_utils.c version
limit checksum calc to sizeof buffer

### DIFF
--- a/apps/app_gps.c
+++ b/apps/app_gps.c
@@ -351,7 +351,7 @@ static time_t time_monotonic(void)
  */
 static int explode_string(char *str, char *strp[], int limit, const char delim, const char quote)
 {
-	int i, l, inquo;
+	int i, inquo;
 
 	inquo = 0;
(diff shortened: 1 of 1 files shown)


apps/app_gps.c
```diff
@@ -995,7 +994,7 @@ static void *gps_reader(void *data)
 			}
 			/* Calculate the check sum */
 			c = 0;
-			for (i = 1; buf[i]; i++) {
+			for (i = 1; buf[i] && i < sizeof(buf); i++) {
```
Allan-N, line 997: You should check `i < sizeof(buf)` before trying to acces `buf[i]`

Allan-N (changes requested):

Allan-N (approved):

InterLinked1 (approved):

PR #555, opened 2025-03-10
Multiple: Eliminate MAXNODES and limits imposed by MAXNODES

mkmer:
Based on PR https://github.com/AllStarLink/app_rpt/pull/545
Use `ast_malloc()` rather than static sizes for `*strs[]`
Eliminate limits imposed by MAXNODES.
Remove MAXNODES define.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3189,8 +3189,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 				lf.datalen = ast_str_strlen(lstr) + 1;
 				lf.data.ptr = ast_str_buffer(lstr);
 				rpt_qwrite(l, &lf);
-				ast_debug(
-					7, "@@@@ node %s sent node string %s to node %s\n", myrpt->name, ast_str_buffer(lstr), l->name);
+				ast_debug(7, "@@@@ node %s sent node string %s to node %s\n", myrpt->name, ast_str_buffer(lstr), l->name);
 			}
 			ast_free(lstr);
 		}
@@ -6618,7 +6617,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		l->linklist = ast_str_create(RPT_AST_STR_INIT_SIZE);
 		if (!l->linklist) {
 			ast_free(l);
-			pthread_exit(NULL);
+			return -1;
 		}
 		l->mode = 1;
 		ast_copy_string(l->name, b1, MAXNODESTR);
--- a/apps/app_rpt/rpt_cli.c
+++ b/apps/app_rpt/rpt_cli.c
@@ -406,9 +406,9 @@ static int rpt_do_lstats(int fd, int argc, const char *const *argv)
 
 static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 {
-	int i, j;
+	int i, j, n = 1;
 	unsigned int ns;
-	char *strs[MAXNODES];
+	char **strs;
 	struct rpt *myrpt;
 	struct ast_var_t *newvariable;
 	char *connstate;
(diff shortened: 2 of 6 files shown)


apps/app_rpt/rpt_cli.c
```diff
@@ -596,8 +596,13 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
...
 
 //### GET ALL LINKED NODES INFO ####################
+			strs = ast_calloc(n, sizeof(char *));
+			if (!strs) {
+				ast_free(lbuf);
+				return RESULT_FAILURE;
+			}
 			/* parse em */
-			ns = finddelim(ast_str_buffer(lbuf), strs, ARRAY_LEN(strs));
+			ns = finddelim(ast_str_buffer(lbuf), strs, n * sizeof(char *));
```
InterLinked1, line 605: I could be wrong here... but should this just be n? At least if I understand correctly, ARRAY_LEN was getting how many entries were in the array strs previously, which now would be equal to n?

apps/app_rpt/rpt_cli.c
```diff
@@ -596,8 +596,13 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 			}
 
 //### GET ALL LINKED NODES INFO ####################
+			strs = ast_calloc(n, sizeof(char *));
```
InterLinked1, line 599: I don't think you need calloc, malloc should be sufficient

apps/app_rpt/rpt_cli.c
mkmer: Yes, you are correct.   Made sense when I was doing it, but it's not right :(

apps/app_rpt.c
```diff
@@ -6713,6 +6718,11 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		if (!l) {
 			pthread_exit(NULL);
 		}
+		l->linklist = ast_str_create(RPT_AST_STR_INIT_SIZE);
+		if (!l->linklist) {
+			ast_free(l);
+			pthread_exit(NULL);
```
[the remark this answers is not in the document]
mkmer: Yes

Allan-N (approved):

apps/app_rpt/rpt_cli.c
```diff
@@ -540,8 +542,9 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 					l = l->next;
 					continue;
 				}
-				if ((s = ast_malloc(sizeof(struct rpt_lstat))) == NULL) {
-					rpt_mutex_unlock(&myrpt->lock);	// UNLOCK 
+				if ((s = (struct rpt_lstat *) ast_malloc(sizeof(struct rpt_lstat))) == NULL) {
```
InterLinked1, line 545: Can't we use calloc here, since we're memset'ing it after?
Casts from void are also never necessary in C.
Preferred way to check for NULL would be if (!(s = ... rather than == NULL

InterLinked1 (changes requested):

apps/app_rpt/rpt_cli.c
```diff
@@ -540,11 +542,11 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
 					l = l->next;
 					continue;
 				}
-				if ((s = ast_malloc(sizeof(struct rpt_lstat))) == NULL) {
-					rpt_mutex_unlock(&myrpt->lock);	// UNLOCK 
+				if (!(s = (struct rpt_lstat *) ast_calloc(1, sizeof(struct rpt_lstat)))) {
```
InterLinked1, line 545: The cast still needs to be removed

InterLinked1 (changes requested):

InterLinked1 (approved):

PR #556, opened 2025-03-11
chan_voter.c: Incorrectly handing stack pointer in a passed data frame.

mkmer:
We are passing a stack pointer in a frame to asterisk.
Let's `malloc()` some memory to pass a good pointer.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -1122,7 +1122,7 @@ static int voter_text(struct ast_channel *ast, const char *text)
 	int cnt, i, j, audio_samples, divcnt, divdiv, audio_ptr, baud;
 	struct pocsag_batch *batch, *b;
 	short *audio;
-	char *cmd, audio1[AST_FRIENDLY_OFFSET + (FRAME_SIZE * sizeof(short))];
+	char *cmd, *cmdp, *audio1;
 	struct ast_frame wf, *f1;
 
 	cmd = ast_alloca(strlen(text) + 10);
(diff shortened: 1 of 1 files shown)


channels/chan_voter.c
```diff
@@ -1167,11 +1167,15 @@ static int voter_text(struct ast_channel *ast, const char *text)
 			AST_LIST_TRAVERSE(&o->txq, f1, frame_list) if (f1->src && (!strcmp(f1->src, PAGER_SRC)))
 				i++;
 			ast_mutex_unlock(&o->txqlock);
-			cmd = (i) ? "PAGES" : "NOPAGES";
+			cmdp = ast_malloc(sizeof("NOPAGES"));
+			if (!cmdp) {
+				return 0;
+			}
+			cmdp = (i) ? "PAGES" : "NOPAGES";
```
Allan-N, line 1174: Huh?  Why are you allocating memory and the setting the same memory pointer back to the same constant strings you were using before.  Does wf.data.ptr get free'd later?  If so, then sure you can allocate memory and COPY the string into that newly allocated buffer.  But, if wf.data.ptr does not get passed to free() then the existing code was fine.

Note: the code was not passing the address of `cmd` (on the stack), just the address of what it points to (the constant strings)

Allan-N (changes requested):

channels/chan_voter.c
mkmer: The pointer was a alloca() => This is a stack pointer which dies when the function ends (right)?  
If we are puting a pointer in a frame and expecting it to live on - we need something that out lives the function.  Either way we shouldn't be passing stack pointers around like this should we?

mkmer: And yes, frame_free() frees the pointer (which was a stack pointer) - I'm fairly sure this is wrong.
https://github.com/asterisk/asterisk/blob/f0f220b36944dbedc713b1205c9d7d48968a5b36/main/frame.c#L152-L154

Allan-N: OK.  I just looked at the complete function.   First, we have the declaration : `char *cmd`.  Then, we use sscanf() as a place to stash some of the passed in "text".  It looks like this text is never used.  Then, later, we use the same `char *cmd` as a pointer to a string (either PAGES or NOPAGES).  That string and it's length is passed to ast_queue_frame().  If that function makes a copy of the "data" then when it returns then we don't care where it was allocated (ast_alloca'd, calloc'd, malloc'd, or a pointer to constant strings).

I don't think this change is needed at all (because this frame was not marked with `f->mallocd = AST_MALLOCD_DATA`)

Allan-N (changes requested):

KB4MDD:
If you look at ast_queue_frame, you will find that it performs ast_frdup on the incoming frame.  This should duplicate everything.

channels/chan_voter.c
mkmer: Actually, the whole wf1 frame is not right either... we need to alloc that too

Allan-N: I don't think so.  ast_queue_frame() is going to copy what needs to be copied.  On return, the frame we created/passed will no longer be used/needed.

mkmer: Alright - I see the description says same thing.... 
Man I thought I found a big ol bobo :)

mkmer:
Nothing to see here

Issue #412, opened 2024-09-28
app_rpt: Be able to execute dialplan on nodes instead of "localplay"


InterLinked1:
The `rpt localplay` command allows for playing audio files onto a node. This is used by functionality such as: https://github.com/AllStarLink/ASL3/pull/101/files

More flexible would be the ability to execute arbitrary dialplan on a node in the same contexts (e.g. `localexec`), which could play audio or do other stuff. This would eliminate a lot of duplicated functionality in Asterisk in the ASL repos, such as constructing and playing the time, when there are Asterisk applications that already do the same things.

mkmer:
I'm thinking something like this could work (few options or all of them):

```
[command_process]
; ---- MJK test extension ---
exten=> _C1,1,rpt(${NODE4}|localplay|rpt/goodafternoon)
same => n, Hangup()

exten=> _C2,1,rpt(${NODE4}|playback|rpt/goodafternoon)
same => n, Hangup()

exten=> _C3,1,rpt(${NODE4}|cli|"rpt localplay ${NODE4} rpt/goodafternoon")
same => n, Hangup()
```

The most generic version would be `|cli|` -> one could put ANY command line string in there and execute it.

This is a "mockup" of a possible solution: https://github.com/AllStarLink/app_rpt/tree/Allow-rpt-xxx-from-dialplan

InterLinked1:
> I'm thinking something like this could work (few options or all of them):
> 
> ```
> [command_process]
> ; ---- MJK test extension ---
> exten=> _C1,1,rpt(${NODE4}|localplay|rpt/goodafternoon)
> same => n, Hangup()
> 
> exten=> _C2,1,rpt(${NODE4}|playback|rpt/goodafternoon)
> same => n, Hangup()
> 
> exten=> _C3,1,rpt(${NODE4}|cli|"rpt localplay ${NODE4} rpt/goodafternoon")
> same => n, Hangup()
> ```
> 
> The most generic version would be `|cli|` -> one could put ANY command line string in there and execute it.
> 
> This is a "mockup" of a possible solution: https://github.com/AllStarLink/app_rpt/tree/Allow-rpt-xxx-from-dialplan

That's not really what I was talking about, I was referring to executing dialplan withing the repeater channel, and eliminating the use of special CLI commands for that purpose altogether. There wouldn't be any modification to the Rpt() application usage itself.

PR #557, opened 2025-03-12
rpt_telemetry.c: Add dialplan capabilities to LOCALPLAY and PLAYBACK.

mkmer:
Potentially covers issue https://github.com/AllStarLink/app_rpt/issues/412

Add the ability to activate dial plan via telemetry playback via dialplan context [telemetry].

`rpt localplay <node> *123` will execute a dialplan on the local telemetry channel.
`rpt playback <node> *123` will execute a dialplan on all repeater channels.

This example dialplan would say 1.2.3 when using the above calls:
```
[telemetry]

exten=> _1.,1,SayAlpha(${EXTEN})
same => n, Hangup()
```

### DIFF
--- a/apps/app_rpt/rpt_telemetry.c
+++ b/apps/app_rpt/rpt_telemetry.c
@@ -966,7 +966,7 @@ void *rpt_tele_thread(void *this)
 	struct rpt_tele *tlist;
 	struct rpt *myrpt;
 	struct rpt_link *l, *l1, linkbase;
-	struct ast_channel *mychannel = NULL;
+	struct ast_channel *mychannel = NULL, *dpchannel = NULL;
 	int id_malloc = 0, m;
 	char *p, *ct, *ct_copy, *ident, *nodename;
 	time_t t, t1, t_mono, was_mono;
@@ -1018,7 +1018,6 @@ void *rpt_tele_thread(void *this)
 
 	/* allocate a pseudo-channel thru asterisk */
 	mychannel = ast_request("DAHDI", cap, NULL, NULL, "pseudo", NULL);
-	ao2_ref(cap, -1);
 
 	if (!mychannel) {
 		ast_log(LOG_WARNING, "Unable to obtain pseudo channel (mode: %d)\n", mytele->mode);
@@ -1649,9 +1648,34 @@ treataslocal:
 		if (wait_interval(myrpt, DLY_TELEM, mychannel) == -1) {
 			break;
 		}
-		res = ast_stream_and_wait(mychannel, mytele->param, "");
+		if (mytele->param[0] != '*') {
+			res = ast_stream_and_wait(mychannel, mytele->param, "");
+			imdone = 1;
+			break;
+		}
+		/* If parameter starts with a *, use dialplan to translate */
+		/* allocate a pseudo-channel thru asterisk */
+		dpchannel = ast_request("DAHDI", cap, NULL, NULL, "pseudo", NULL);
+		if (!dpchannel) {
+			break;
+		}
+		rpt_disable_cdr(dpchannel);
+		if (rpt_conf_add(dpchannel, myrpt, type, RPT_CONF_CONFANN)) {
+			ast_hangup(dpchannel);
+			break;
+		}
+		ast_channel_context_set(dpchannel, "telemetry");
+		ast_channel_exten_set(dpchannel, mytele->param + 1);
+		ast_channel_priority_set(dpchannel, 1);
+		ast_pbx_start(dpchannel);
+		ast_debug(5, "Playback dialplan extension %s\n", mytele->param);
+		while (!ast_check_hangup(dpchannel)) {
+			usleep(MSWAIT * 1000); /* Wait for PBX thread to hangup */
+		}
+		ast_debug(5, "PBX has finished on telemetry\n");
 		imdone = 1;
 		break;
+
 	case TOPKEY:
 		/* wait a little bit */
 		if (wait_interval(myrpt, DLY_TELEM, mychannel) == -1)
@@ -2577,6 +2601,7 @@ treataslocal:
 		ast_free(ident);
 	}
 	ast_free(mytele);
+	ao2_ref(cap, -1);
 	ast_hangup(mychannel);
 #ifdef  APP_RPT_LOCK_DEBUG
 	{
@@ -2606,6 +2631,7 @@ abort3:
 	if (mychannel) {
 		ast_hangup(mychannel);
 	}
+	ao2_ref(cap, -1);
 	pthread_exit(NULL);
 }
 


apps/app_rpt/rpt_telemetry.c
```diff
@@ -1649,9 +1648,34 @@ void *rpt_tele_thread(void *this)
...
 		}
-		res = ast_stream_and_wait(mychannel, mytele->param, "");
+		if (mytele->param[0] != '*') {
+			res = ast_stream_and_wait(mychannel, mytele->param, "");
+			imdone = 1;
+			break;
+		}
+		/* If parameter starts with a *, use dialplan to translate */
+		/* allocate a pseudo-channel thru asterisk */
+		dpchannel = ast_request("DAHDI", cap, NULL, NULL, "pseudo", NULL);
```
InterLinked1, line 1658: Please use the rpt_request request wrapper, in case we move away from DAHDI bridging in the future. We shouldn't be using ast_request directly for any DAHDI channels

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1649,9 +1648,34 @@ void *rpt_tele_thread(void *this)
...
+		if (!dpchannel) {
+			break;
+		}
+		rpt_disable_cdr(dpchannel);
+		if (rpt_conf_add(dpchannel, myrpt, type, RPT_CONF_CONFANN)) {
+			ast_hangup(dpchannel);
+			break;
+		}
+		ast_channel_context_set(dpchannel, "telemetry");
```
InterLinked1, line 1667: My conception of this is any arbitrary context,exten,priority should be able to be executed.
Certainly, the context to execute should not be hardcoded, at the very least that would need to be an option, since that could already exist in the user's dialplan for something else

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1649,9 +1648,34 @@ void *rpt_tele_thread(void *this)
 		if (wait_interval(myrpt, DLY_TELEM, mychannel) == -1) {
 			break;
 		}
-		res = ast_stream_and_wait(mychannel, mytele->param, "");
+		if (mytele->param[0] != '*') {
+			res = ast_stream_and_wait(mychannel, mytele->param, "");
+			imdone = 1;
+			break;
+		}
+		/* If parameter starts with a *, use dialplan to translate */
```
InterLinked1, line 1656: I think we should have a separate parameter for this, personally, since there's a lot to unpack here

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1649,9 +1648,34 @@ void *rpt_tele_thread(void *this)
...
+			ast_hangup(dpchannel);
+			break;
+		}
+		ast_channel_context_set(dpchannel, "telemetry");
+		ast_channel_exten_set(dpchannel, mytele->param + 1);
+		ast_channel_priority_set(dpchannel, 1);
+		ast_pbx_start(dpchannel);
+		ast_debug(5, "Playback dialplan extension %s\n", mytele->param);
+		while (!ast_check_hangup(dpchannel)) {
+			usleep(MSWAIT * 1000); /* Wait for PBX thread to hangup */
```
InterLinked1, line 1673: Sleeping 1 second at a time is probably too long... not sure if there is a better way to do this though off the top of my head. Maybe just reduce to 100?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1649,9 +1648,34 @@ void *rpt_tele_thread(void *this)
...
+		}
+		ast_debug(5, "PBX has finished on telemetry\n");
```
InterLinked1, line 1675: Where does this channel get hung up?

InterLinked1 (commented):
Thanks for taking a stab at this! It was in my backlog for "eventually" but sooner is nicer than later:)

My overarching question would be, are there other circumstances where executing dialplan would be useful? I see here you've done it for telemetry playback, it looks like, which makes sense, but are there are other things where users would like to do something "configurable"? I don't know enough yet about the user side of this to know where all we would need hooks, but I wouldn't be surprised if there are more.

Some specific comments for what you have here so far though as well:

apps/app_rpt/rpt_telemetry.c
mkmer: I think this is 20ms (MSWAIT = 20 * 1000us) right?

apps/app_rpt/rpt_telemetry.c
mkmer: I suppose we could.  I was thinking `*` will never lead in a valid file name, it should be "safe" to pass `*<blah> ` .
This is all just my first "thoughts" on could it be done and could "I" do it ;)
Getting closer to understanding HOW things work.

apps/app_rpt/rpt_telemetry.c
mkmer: I tested this in several ways (but still could be wrong).
1 - Missing dial plan : PBX fails the plan and hangs up the channel
2 - Good dial plan : Last step in the plan is Hangup()
3 - Good dial plan without Hangup() - appears the PBX hangs up anyway.

OR are you saying we need to hangup on this side of the channel?  It doesn't appear to be hanging around when I do `core show channels` -> it's there while the dialplan is running and goes away when it's finished.

apps/app_rpt/rpt_telemetry.c
mkmer: Same here - I think this is 20ms (MSWAIT = 20ms * 1000ms/us)

apps/app_rpt/rpt_telemetry.c
mkmer: Maybe we just use the `p.telemetry` config value?  It would force the same context in both rpt.conf and extension.conf which may or may not be good.

apps/app_rpt/rpt_telemetry.c
InterLinked1: It's not that it wouldn't work, it's just adding modifiers to existing options to make it do other things is not a best practice.

apps/app_rpt/rpt_telemetry.c
InterLinked1: Oh, okay, I assumed MSWAIT was 1 ms and thus it was a second. If it's 20ms, that's fine then

apps/app_rpt/rpt_telemetry.c
InterLinked1: In that case, we're all good! I was just wondering if we needed to, if we didn't clean up, the channel would linger around with core show channels so that seems like it's sufficient then...

The 3 cases you tested are as expected as well. Number 3 should log a warning or notice about missing priority or something, but it will get hung up anyways as the default action.

apps/app_rpt/rpt_telemetry.c
mkmer: At the moment, I can't see a good reason to have a different context for "telemetry stuff", but maybe there is a good case somewhere?

InterLinked1: > At the moment, I can't see a good reason to have a different context for "telemetry stuff", but maybe there is a good case somewhere?

I think as long as the name of the context is adjustable and other settings can be modified without impacting this one, then it works for me.

apps/app_rpt/rpt_telemetry.c
mkmer: When you say "parameter" are you thinking somthing like:
`rpt dialplan <node> <localplay/playback> <dialplan extension>`
Essentially adding a new command or something different?
We could then launch this from a macro similar to `autopatchup` or `localplay`.

apps/app_rpt/rpt_telemetry.c
mkmer: Figured out a "better" way around this hangup test -> using `ast_pbx_run()` does not launch a thread like `ast_pbx_start()` , so we just wait here for it to finish.

apps/app_rpt/rpt_telemetry.c
InterLinked1: The way I was thinking about it, it wouldn't be a CLI command at all (since these are not best practice, CLI commands are intended for humans to type manually, not for automation). It would be its own setting where you execute dialplan, instead of running a CLI command.

InterLinked1: To be clear, I'm saying it wouldn't be a command - thus, there is no command name, we're trying to allow executing dialplan to allow for custom logic, as part of a config option. There are already CLI commands to originate dialplan and that's not what we're trying to do here.

mkmer: Yes - I read "wouldn't" as "would", my mistake.  
This would still be a command (just not a CLI command) that we would be able to macro with right?   
For example:
We use this macro today:
`89=autopatchup,context=command_process,noct=1,farenddisconnect=1,dialtime=7000,quiet=1; Command Processing`
This "fires up" autopatch to essentially execute a dialplan based on the numbers that follow `*89`.  Then the dialplan looks like this:
```
; ---- Command Process extentions ------  See /usr/local/bin/command_process
[command_process]
exten => _x.,1,System(/usr/local/bin/command_process ${EXTEN})
 same => n,Hangup()
```
The our scripts look at the extension arguments passed in to do things.

I "thought" the idea here is to create a way to run the dialplan without calling `autopatchup`. 
Macro might look like: `89=dialplan,context=xxx,localplay=1`

InterLinked1: > Yes - I read "wouldn't" as "would", my mistake. This would still be a command (just not a CLI command) that we would be able to macro with right? For example: We use this macro today: `89=autopatchup,context=command_process,noct=1,farenddisconnect=1,dialtime=7000,quiet=1; Command Processing` This "fires up" autopatch to essentially execute a dialplan based on the numbers that follow `*89`. Then the dialplan looks like this:
> 
> ```
> ; ---- Command Process extentions ------  See /usr/local/bin/command_process
> [command_process]
> exten => _x.,1,System(/usr/local/bin/command_process ${EXTEN})
>  same => n,Hangup()
> ```
> 
> The our scripts look at the extension arguments passed in to do things.
> 
> I "thought" the idea here is to create a way to run the dialplan without calling `autopatchup`. Macro might look like: `89=dialplan,context=xxx,localplay=1`

I don't think focusing on `System()` is a good example as ultimately, that's still shelling out. If we can already do that without the dialplan option, there's not much value in doing it that way and it's clunkier.

The real value is when doing something custom, e.g.:

```
exten => s,1,Playback(hello-there)
   same => n,Foobar()
   same => n,PlayTones(400)
   same => n,Wait(2)
   same => n,Hangup()
```

mkmer: I was sharing how we get to a dialplan today.  
With the "new" command I was thinking we would get to the dialplan avoiding autopatch.... otherwise we can do exactly what you have shared using the *89 macro I shared -> it executes the matching number in the chosen context.

InterLinked1: > I was sharing how we get to a dialplan today. With the "new" command I was thinking we would get to the dialplan avoiding autopatch.... otherwise we can do exactly what you have shared using the *89 macro I shared -> it executes the matching number in the chosen context.

As long as via the macro, it is possible to specify a full context,exten,priority, that sounds good. I don't like the idea personally of only being able to specify exten,priority and the context being either hardcoded or only set to 1 context globally. There may be the same extension in multiple different contexts that a user would want to use.

A related example I think is in app_confbridge, it is possible to execute certain dialplan subroutines using a feature code. There are some limitations with bridging that would probably apply here too, so there may be some overlap.

InterLinked1:
> I have a "few" of the messages converted to use dialplan. Some are a bit more complicated, so I wanted to see everyone's thoughts on where I'm at and how we may want to change things. One of the "problem" areas is around multiple calls to `telem_lookup()` -> once we fire off the pbx on the channel, it hangs up the channel when it's finish, not allowing for further messages on the channel. I'm sure there's a way to have the PBX not hangup at the end of it's "work", but it's not obvious to me at the moment.

Yes, there is, that is the whole thing with keepalive and masquerading the channel.
But why do we want to keep it alive instead of hanging it up and using a new channel is needed? There are very few legitimate reasons for wanting to keep a channel around after PBX has finished.

> The dialplan that goes with this update:
> 
> ```
> [telemetry]
> 
> ; Say the time of day
> exten => TOD,1,Gotoif($[${DAY}] = "MORNING"?morning:other)
> same  => n(morning),playback(rpt/goodmorning)
> same  => n, Goto(time)
> same  => n(other), Gotoif($[${DAY}] = "AFTERNOON"?afternoon:evening)
> same  => n(afternoon), 1, playback(rpt/goodafternoon)
> same  => n, Goto(time)
> same  => n(evening), playback(rpt/goodevening)
> same  => n(time), playback(rpt/thetimeis)
> same  => n, SayUnixTime(,EST,IMP)
> same  => n, Hangup()
> 
> ;Say Version information
> exten => VERSION,1,playback(rpt/version)
> same => n, SayNumber($[${MAJOR}])
> same => n, SayAlpha(.)
> same => n, SayNumber($[${MINOR}])
> same  => n, Hangup()
> 
> ; Say message file
> exten => FILE,1,playback($[${FILE}])
> same  => Hangup()
> ```

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1117,56 +1185,57 @@ void *rpt_tele_thread(void *this)
...
 		}
 		imdone = 1;
 		break;
 	case TAILMSG:
 		/* wait a little bit longer */
-		if (!wait_interval(myrpt, DLY_TELEM, mychannel))
-			res =
-				ast_streamfile(mychannel, myrpt->p.tailmessages[myrpt->tailmessagen], ast_channel_language(mychannel));
+		if (!wait_interval(myrpt, DLY_TELEM, mychannel)) {
+			pbx_builtin_setvar_helper(mychannel, "FILE", myrpt->p.tailmessages[myrpt->tailmessagen]);
```
InterLinked1, line 1203: In this case, a variable might make sense, but "FILE" is too short, if the user already is using that name there is high chance of a conflict. Any dialplan variables should probably have a common prefix like `RPT_`. Maybe `RPT_PLAYBACK_FILE` or something like that for this one.

apps/app_rpt/rpt_telemetry.c
```diff
@@ -37,6 +37,12 @@
 
 extern struct rpt rpt_vars[MAXRPTS];
 
+#define TELEM_STAT_TIME_EXTN "TOD"
```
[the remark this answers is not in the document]
mkmer: The "should be"... but why would they ever change?  They are configurable in code via defines (so you don't need to hunt).  I originally wanted to tie them directly to the enum, but it doesn't work well :(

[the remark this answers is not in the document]
InterLinked1: Because the user may want to configure them to something else. That is the whole idea of dialplan. I can't think of many things if any in Asterisk that assume a specific dialplan location - these are always configurable options.

mkmer:
[extensions.txt](https://github.com/user-attachments/files/19653874/extensions.txt)
Several opportunities to deal with repetition, just not sure the best way to do it.

InterLinked1:
> [extensions.txt](https://github.com/user-attachments/files/19653874/extensions.txt) Several opportunities to deal with repetition, just not sure the best way to do it.

I'm not sure what you're trying to do here, but this looks like a big no no:
```
exten=> _C2,1,rpt(${NODE4}|playback|rpt/goodafternoon)
same => n, Hangup()

exten=> _C3,1,rpt(${NODE4}|cli|"rpt localplay ${NODE4} rpt/goodafternoon")
same => n, Hangup()
```

I think these are supposed to play files on a node, but with these changes, the idea would be that you have normal dialplan, e.g. a call to Playback(), that the node is running for that channel. So no |playback| or localplay, etc.

The other  dialplan could be cleaned up but that's basically the right idea!

apps/app_rpt/rpt_telemetry.c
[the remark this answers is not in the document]
InterLinked1: After looking at this further, if you want to use hardcoded extensions, I am fine with that, as long as the context is configurable. This way, the user can have it go somewhere else if desired.

However, the extensions need to be documented - ideally in the sample config, although I don't think there is one currently.

Issue #15, opened 2022-07-21
Cleanly eliminate AST_PBX_KEEPALIVE


InterLinked1:
Cleanly eliminate AST_PBX_KEEPALIVE by converting this to something that works nicely with the present core and isn't a total and complete hack.

See also: https://github.com/asterisk/asterisk/commit/50a25ac8474d7900ba59a68ed4fd942074082435

mkmer:
Should we be doing it like this (not sure):
https://www.asteriskguru.com/archives/image-vp339222.html

https://github.com/asterisk/asterisk/blob/e904ae5df0b8ee52e8893b4bf6ecdc985140849d/res/res_stasis.c#L1634-L1656

InterLinked1:
> Should we be doing it like this (not sure): https://www.asteriskguru.com/archives/image-vp339222.html
> 
> https://github.com/asterisk/asterisk/blob/e904ae5df0b8ee52e8893b4bf6ecdc985140849d/res/res_stasis.c#L1634-L1656

Possibly, we kind of punted this down the road, but thanks for giving us the nudge to look into this again. I was having a conversation with the Asterisk team at one point and maybe I should see this through now. But yes, those are relevant links.

PR #558, opened 2025-03-12
app_rpt: Fix incorrect cleanup code.

InterLinked1:
* Don't use pthread_exit in rpt_exec. This isn't our thread, it's the PBX thread, and thus we need to return -1 to allow it to clean up properly.
* Remove unneeded check for l->chan being NULL. It can't be NULL at this point.
* Remove unneeded hangup check, which could have been removed as part of commit 78845176f3c725fb4da5fc5a8d5ab1bf67dbc768.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -6711,7 +6711,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		/* establish call in tranceive mode */
 		l = ast_calloc(1, sizeof(struct rpt_link));
 		if (!l) {
-			pthread_exit(NULL);
+			return -1;
 		}
 		l->mode = 1;
 		ast_copy_string(l->name, b1, MAXNODESTR);
@@ -6756,7 +6756,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		if (!cap) {
 			ast_log(LOG_ERROR, "Failed to alloc cap\n");
 			ast_free(l);
-			pthread_exit(NULL);
+			return -1;
 		}
 
 		ast_format_cap_append(cap, ast_format_slin, 0);
@@ -6765,7 +6765,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		if (__rpt_request_pseudo(l, cap, RPT_PCHAN, RPT_LINK_CHAN)) {
 			ao2_ref(cap, -1);
 			ast_free(l);
-			pthread_exit(NULL);
+			return -1;
 		}
 
 		ao2_ref(cap, -1);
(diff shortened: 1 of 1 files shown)


mkmer (approved):
You have tested this right?
I had a few `retrurn -1` cause crashes BUT I believe they are/were related to putting the link on the list https://github.com/AllStarLink/app_rpt/blob/eafeb5f4febc321a7f15b86ccf1ae196533c2498/apps/app_rpt.c#L6812
Essentially, once we put the link on the list, the other thread HAS to be responsible for the channel and all cleanup that follows

InterLinked1:
> You have tested this right? I had a few `retrurn -1` cause crashes BUT I believe they are/were related to putting the link on the list
> 
> https://github.com/AllStarLink/app_rpt/blob/eafeb5f4febc321a7f15b86ccf1ae196533c2498/apps/app_rpt.c#L6812
> 
> 
> Essentially, once we put the link on the list, the other thread HAS to be responsible for the channel and all cleanup that follows

No, I haven't. These are all off nominal paths (setup failed and we're cleaning up). Successful paths should not be affected by this change.

I have a follow-up PR to address something related (more in the area you're referring to), but I'll wait until after this is finalized since it would create a merge conflict.

mkmer:
> > You have tested this right? I had a few `retrurn -1` cause crashes BUT I believe they are/were related to putting the link on the list
> > https://github.com/AllStarLink/app_rpt/blob/eafeb5f4febc321a7f15b86ccf1ae196533c2498/apps/app_rpt.c#L6812
> > 
> > Essentially, once we put the link on the list, the other thread HAS to be responsible for the channel and all cleanup that follows
> 
> No, I haven't. These are all off nominal paths (setup failed and we're cleaning up). Successful paths should not be affected by this change.
> 
> I have a follow-up PR to address something related (more in the area you're referring to), but I'll wait until after this is finalized since it would create a merge conflict.

Might be worth making it a "sunny day" path (as a test) and confirm `return -1` is "ok" any place before adding it to the `l->linklist` .

InterLinked1:
> > > You have tested this right? I had a few `retrurn -1` cause crashes BUT I believe they are/were related to putting the link on the list
> > > https://github.com/AllStarLink/app_rpt/blob/eafeb5f4febc321a7f15b86ccf1ae196533c2498/apps/app_rpt.c#L6812
> > > 
> > > Essentially, once we put the link on the list, the other thread HAS to be responsible for the channel and all cleanup that follows
> > 
> > 
> > No, I haven't. These are all off nominal paths (setup failed and we're cleaning up). Successful paths should not be affected by this change.
> > I have a follow-up PR to address something related (more in the area you're referring to), but I'll wait until after this is finalized since it would create a merge conflict.
> 
> Might be worth making it a "sunny day" path (as a test) and confirm `return -1` is "ok" any place before adding it to the `l->linklist` .

We already have `return -1` in other places, these were a few that were erroneously using `pthread_exit()`. Applications *must* return, terminating the thread is "illegal" for lack of a better way to describe it. We are not supposed to be doing that for any reason.

I will modify the sunny day path in the next PR to fix #15 finally, since that entails some slightly different logic. Ultimately that will return -1 as well though but it'll be a separate change.

mkmer (approved):
Ran the test - As long as `return -1` is before adding the `l-linklist` all is good.

Allan-N (approved):

PR #560, opened 2025-03-15
app_rpt: Clean up a couple debug messages.

InterLinked1:
* Remove testing debug message that ran once per frame.
* Clean up debug messages for if scheduler is disabled.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2629,25 +2629,16 @@ static void do_scheduler(struct rpt *myrpt)
 		myrpt->dailyexecdcommands = 0;
 	}
 
-	if (tmnow.tm_sec != 0)
-		return;
-
-	/* Code below only executes once per minute */
-
-	/* Don't schedule if remote */
-
-	if (myrpt->remote)
-		return;
-
-	/* Don't schedule if disabled */
-
-	if (myrpt->p.s[myrpt->p.sysstate_cur].schedulerdisable) {
-		ast_debug(7, "Scheduler disabled\n");
+	if (tmnow.tm_sec != 0) {
 		return;
 	}
 
-	if (ast_strlen_zero(myrpt->p.skedstanzaname)) {	/* No stanza means we do nothing */
-		ast_debug(7, "No stanza for scheduler in rpt.conf\n");
+	/* Code below only executes once per minute */
+	if (myrpt->remote) { /* Don't schedule if remote */
+		return;
+	} else if (myrpt->p.s[myrpt->p.sysstate_cur].schedulerdisable) { /* Don't schedule if disabled */
+		return;
+	} else if (ast_strlen_zero(myrpt->p.skedstanzaname)) {	/* No stanza means we do nothing */
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


mkmer (commented):
If there is a return, does `else if` make any difference at all?
Is there no need for debug here? Maybe just a higher level?
For example; Is there some other indication the scheduler is disabled or the stanza is missing?

InterLinked1:
> If there is a return, does `else if` make any difference at all?

Nope! It just allowed condensing that code a little bit

> Is there no need for debug here? Maybe just a higher level? For example; Is there some other indication the scheduler is disabled or the stanza is missing?

I'm not sure, but this was currently logging every minute, which is unnecessary. We don't need debug messages of the various config values. If it is important for something, we could add a warning message when the config is parsed but the previous way didn't really make sense.

mkmer:
They are "debug" messages - only log if you turn it on (right?).  
If you are using the standard rpt.conf file, these will NOT be triggering every minute (or at all).
I was thinking "maybe" they should remain and be at a higher level (like 10 ,11, 100?).  It is useful to get some feedback when things are configured incorrectly - like a typo in the context configuration.

On that comment: I don't think we have a "standard" around what these levels SHOULD be.  I assume as the level increases, the messages are more "noisy".  Is there something hiding over in Asterisk dev? (I don't see it right off hand).

InterLinked1:
> They are "debug" messages - only log if you turn it on (right?). If you are using the standard rpt.conf file, these will NOT be triggering every minute (or at all)

It would be logger.conf, not rpt.conf, but "standard" file is irrelevant, really. The code needs to account for all valid configurations.

> I was thinking "maybe" they should remain and be at a higher level (like 10 ,11, 100?).

There isn't really any higher level than 10. There isn't anything in the code that caps it at 10, but that is the highest debug level used.

> It is useful to get some feedback when things are configured incorrectly - like a typo in the context configuration.

As this exists today, I don't personally think this is useful at all. I don't think a debug message that every 60 seconds logs that something isn't happening is the right way to do that. We also don't do it on all branches, so it's not consistent anyways.

If we want to alert the user, that's great, I agree with that actually. But the warning should be emitted when app_rpt loads, not at periodic intervals where there's no guarantee the user will see it. And user warnings should also be warnings, not debug messages. I don't know what the circumstances are under which we would want to warn a user about the scheduler being disabled, do you have any insight on that?

(FWIW, the reason I noticed this is because I often run systems with full debug all the time. These were 2 useless messages that were polluting the log output for no reason. I'm not proposing their removal because they're noisy, per se, but more because they're pointless)

> On that comment: I don't think we have a "standard" around what these levels SHOULD be. I assume as the level increases, the messages are more "noisy". Is there something hiding over in Asterisk dev? (I don't see it right off hand).

That is correct. There is no exact "science", but you often get an intuitive feel, especially comparing messages against each other. Periodically levels have been refined as needed if they are deemed to be inappropriate by someone, I've even submitted a few changes myself. Sometimes, people disagree too, so discussion is always good.

Allan-N (approved):

mkmer (approved):

PR #563, opened 2025-03-16
Multiple: Remove RPT_TELECONF and Use ast_audiohook_volume* for ducking

mkmer:
Use ast_audiohook_volume*
Remove telemchannel and btelemchannel.
Adjusts volume for ALL telemetry messages.


InterLinked1 (commented):
First of all, I'm really excited about changes like this! The more code we can remove, the better! (without changing functionality, of course)

I think the commit message needs much more detail. It should explain why the old way was unnecessary and we can use audiohooks instead. For example, I thought these channels had something to do with telemetry, not adjusting the volume. Was that a misnomer? Or something else?

Also, I think for a change of this magnitude (removing channels), we should insist on tests. I would like to see a relevant test that passes prior to this change, and continues to pass afterwards. Once #533 merges, the tests should all be run by the CI.

mkmer:
All that happens between `telechannel` and `btelechannel` is the ducking/volume adjustment code... That's it.
Every time we spin up a telemetry thread, we create a new DAHDI pseudo channel and link it either to the `RPT_TELECONF` or the `RPT_CONF`.  `RPT_CONF` is connected to all audio moving around the system.

Also in the `RPT_TELECONF` is the `telechannel`.  We simply copy `telechannel` to `bchannel` and in the middle we apply one of the two volume adjustments.  Finally, the `bchannel` is connected to the `RPT_TXCONF` (which is local audio out)

So - all this does is move local audio connection in the telemetry thread FROM `RPT_TELECONF` to `RPT_TXCONF`.

Additionally, telemetry audio that was NOT connect to `RPT_TELECONF` was not receiving the `p.telemduckgain` or `p.telemnomgain)` adjustments.

InterLinked1:
> All that happens between `telechannel` and `btelechannel` is the ducking/volume adjustment code... That's it. Every time we spin up a telemetry thread, we create a new DAHDI pseudo channel and link it either to the `RPT_TELECONF` or the `RPT_CONF`. `RPT_CONF` is connected to all audio moving around the system.
> 
> Also in the `RPT_TELECONF` is the `telechannel`. We simply copy `telechannel` to `bchannel` and in the middle we apply one of the two volume adjustments. Finally, the `bchannel` is connected to the `RPT_TXCONF` (which is local audio out)
> 
> So - all this does is move local audio connection in the telemetry thread FROM `RPT_TELECONF` to `RPT_TXCONF`.
> 
> Additionally, telemetry audio that was NOT connect to `RPT_TELECONF` was not receiving the `p.telemduckgain` or `p.telemnomgain)` adjustments.

Thanks for the explanation! If you can capture the gist of that in the commit message for posterity, I think that would be good.

I've seen some strange things in this codebase, but yeah, that sounds awfully inefficient. My guess, without confirming, is that audiohooks didn't exist in Asterisk when Jim wrote the code initially, and he may have deemed that an easy way to do that. But now that there are a lot of tools in the Asterisk core at our disposal, it makes sense to leverage those as much as we can.

apps/app_rpt/app_rpt.h
```diff
@@ -272,7 +272,7 @@ enum  rpt_dns_method {
 };
 
 #define DEFAULT_NODE_LOOKUP_METHOD LOOKUP_BOTH
-#define DEFAULT_TELEMDUCKDB "-9"
+#define DEFAULT_TELEMDUCKDB -9
```
Allan-N, line 275: Our `rpt.conf` file has :
```
telemduckdb = -15                   ; Telemetry Ducking in dB when local or link voice tx in progress
```
It's OK for the two values to differ but it would be good for `rpt.conf` to note the default

apps/app_rpt/rpt_config.c
```diff
@@ -1031,12 +1031,8 @@ void load_rpt_vars(int n, int init)
 	RPT_CONFIG_VAR_INT(votertype, "votertype");
 	RPT_CONFIG_VAR_INT(votermode, "votermode");
 	RPT_CONFIG_VAR_INT_DEFAULT(votermargin, "votermargin", 10);
-
-	val = ast_variable_retrieve(cfg, cat, "telemnomdb");
-	rpt_vars[n].p.telemnomgain = pow(10.0, atof(S_OR(val, "0")) / 20.0);
-
-	val = ast_variable_retrieve(cfg, cat, "telemduckdb");
-	rpt_vars[n].p.telemduckgain = pow(10.0, atof(S_OR(val, DEFAULT_TELEMDUCKDB)) / 20.0);
+	RPT_CONFIG_VAR_INT_DEFAULT(telemnomgain, "telemnomdb", 1);
```
Allan-N, line 1034: Changing a configuration file "float" value to an integer value?

... and our current `rpt.conf` file has :
```
telemnomdb = -3                     ; Telemetry Nominal Amplitude reference in dB
```
Should this default also be the same?
and, it's OK for the two values to differ but it would be good for `rpt.conf` to note the default

apps/app_rpt/rpt_config.c
```diff
@@ -1031,12 +1031,8 @@ void load_rpt_vars(int n, int init)
...
+	RPT_CONFIG_VAR_INT_DEFAULT(telemduckgain, "telemduckgain", DEFAULT_TELEMDUCKDB);
```
Allan-N, line 1035: Changing the configuration file "telemduckdb" variable to "telemduckgain" :-(

Is there a way to retain the current variable and convert it to the other?

apps/app_rpt/app_rpt.h
```diff
@@ -677,8 +677,8 @@ struct rpt {
 		const char *locallist[16];
 		int nlocallist;
 		char ctgroup[16];
-		float telemnomgain;		/*!< \brief nominal gain adjust for telemetry */
-		float telemduckgain;	/*!< \brief duck on busy gain adjust for telemetry */
+		int telemnomgain;  /*!< \brief nominal gain adjust for telemetry */
```
Allan-N, line 680: We are changing the variable types from "float" to "int" ?

apps/app_rpt.c
```diff
@@ -3659,7 +3613,7 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 				myrpt->linkactivitytimer = 0;
 				myrpt->keyed = 1;
 				time(&myrpt->lastkeyedtime);
-				myrpt->keyposttimer = KEYPOSTSHORTTIME;
+				myrpt->keyposttimer = KEYPOSTSHORTTIME;
```
Allan-N, line 3616: Looks like we gained a trailing space

Allan-N (changes requested):

apps/app_rpt/app_rpt.h
mkmer: We don't really need a float here... all the examples are int's and what kind of resolution do we really need on the gain for telemetry? (I get radio gains having decimal values to get things "perfect", but telemetry?)

apps/app_rpt/rpt_config.c
mkmer: Allegedly, the asterisk volume want's dB, so we don't need to convert to the decimal value, just use dB.  Of course if it's NOT dB on the volume, then we need to do something different.

apps/app_rpt/rpt_config.c
mkmer: I don't see why not

apps/app_rpt/app_rpt.h
mkmer: -15 is a LONG way down... but I don't see why the shouldn't match

apps/app_rpt/rpt_config.c
mkmer: Let me elaborate a bit on this statement:
The current telemetry gain code: 
https://github.com/AllStarLink/app_rpt/blob/9db0ce2f3ca6a6edd73c1b313ec2ddfe3d62d8b4/apps/app_rpt.c#L4710-L4723
The asterisk version of the gain code is here:
https://github.com/asterisk/asterisk/blob/606d5d054afc7c79268a54d27256a443bf6cba8d/main/frame.c#L787-L808

They appear to do the exact same math except asterisk claims the `int` value is in dB, while app_rpt is converting dB into a scaling multiplier (typical dB math).

The `ast_audiohook_volume_set()` function eventually calls `audiohook_volume_callback()` which uses `ast_frame_adjust_volume()` gain value which is an int.

IF we want to do something other than an `int` for the adjustment, we will need to make our own `audiohook` implementation that uses a float.

My thoughts on this are: Telemetry ducking  is not a calibrated value (aka we're not trying to match radio gains and modulation levels, etc), it's really just a "volume" level adjustment and if it changes a bit, it won't hurt anything.

Allan-N: My concern is about handling existing configurations / formats. Folks have existing configuration files (or make updates based on web content) that, ideally, we don't want to break.  If we already have a "db" variable ... and if we can convert it to a gain value ... then why not leave the variable as-is (and perform any needed math conversion when reading the configuration)?  Else, we need to add a configuration file updater to the mix ... and restores of older configurations break unless one remembers to run the updater.

apps/app_rpt/app_rpt.h
Allan-N: If we no longer have the ability to support a fractional gain then making the switch (here) to "int" values is OK.  But, in the config code, I do think we want to accept any float values found in [older] configuration files.

apps/app_rpt/rpt_config.c
Allan-N: My concern is that someone already has a configuration file with `telemduckdb = 3.2`.  I'm good as long as we accept the "3.2" value.  If we round it down/up, fine.  Bonus points for logging a message when the configuration file is processed  ("telemduckdb must now be an integer value, rounded to 3").

InterLinked1: > My concern is about handling existing configurations / formats. Folks have existing configuration files (or make updates based on web content) that, ideally, we don't want to break. If we already have a "db" variable ... and if we can convert it to a gain value ... then why not leave the variable as-is (and perform any needed math conversion when reading the configuration)? Else, we need to add a configuration file updater to the mix ... and restores of older configurations break unless one remembers to run the updater.
> 
> As for ints/floats. Can we accept floats, convert them to ints, and maybe log a message that we've done so? That way, we maintain backward compatibility.

In Asterisk, it's okay to have breaking modifications across standard releases. So for Asterisk 22, we would want to keep the config compatible, but for 23/24, we can drop the old setting. It's a little bit weird since we're not maintaining for multiple versions simultaneously like the Asterisk tree is.

Allan-N: > In Asterisk, it's okay to have breaking modifications across standard releases. So for Asterisk 22, we would want to keep the config compatible, but for 23/24, we can drop the old setting.

Certainly.  When can break things (and remove backward compatibility support) when we release ASL4 :-)

mkmer: To close the loop on this one: the `atoi` function happily converts floating type strings to integers, so an existing config file will still work.  
As i have mentioned in other forums, the code that is adjusting the volume between the `telechannel` and `btelecahnnel` is not correct.  Today it is `int = float * short` which results in a VERY limited adjustment for negative gain values due to truncating math in the conversion.  The more correct way is to `float = float * short` then convert back to `short`.
That being said, if we move to this asterisk hook version, we can get to the same volume levels but they will follow large steps:  say in dB 6.0, 9.5, 12.0, 14.0, 15.5, 16.9, 18.0 would be the step adjustments.   This is all we really have today as well (any choice up to -6.0db is getting ~ -6db) 
```
Volume #	dB Gain
-1	        0
-2	        -6.020599913
-3	        -9.542425094
-4	        -12.04119983
-5	        -13.97940009
-6	        -15.56302501
-7	        -16.9019608
-8	        -18.06179974
```

apps/app_rpt/rpt_config.c
```diff
@@ -795,6 +795,14 @@ void load_rpt_vars(int n, int init)
 		rpt_vars[n].p.var = atoi(val); \
 	}
 
+#define RPT_CONFIG_VAR_FLOAT_DB_DEFAULT(var, name, default) \
+	val = ast_variable_retrieve(cfg, cat, name); \
+	if (!ast_strlen_zero(val)) { \
+		rpt_vars[n].p.var = pow(10.0, atof(val)) / 20.0; \
+	} else { \
+		rpt_vars[n].p.var = default; \
```
Allan-N, line 803: Has the value "default" already received the `pow(10.0, atof(val)) / 20.0` computation?

apps/app_rpt.c
```diff
@@ -5341,13 +5241,26 @@ static void *rpt(void *this)
 				}
 			}
 		}
+		/* If we have a new telemetry or we haven't adjusted ducking and we are keyed up */
+		if (((myrpt->tele.next != last_tele) || !lastduck) && myrpt->tele.next->chan && (myrpt->keyed || myrpt->remrx)) {
+			if (ast_audiohook_volume_set(myrpt->tele.next->chan, AST_AUDIOHOOK_DIRECTION_WRITE, (int) myrpt->p.telemduckgain)) {
+				ast_debug(7, "I've set the volume on channel %s to %2.2f", ast_channel_name(myrpt->tele.next->chan), myrpt->p.telemduckgain);
```
Allan-N, line 5247: Maybe "Setting the volume on channel ..."

apps/app_rpt/rpt_config.c
```diff
@@ -1030,14 +1031,19 @@ void load_rpt_vars(int n, int init)
...
-
-	val = ast_variable_retrieve(cfg, cat, "telemnomdb");
-	rpt_vars[n].p.telemnomgain = pow(10.0, atof(S_OR(val, "0")) / 20.0);
-
-	val = ast_variable_retrieve(cfg, cat, "telemduckdb");
-	rpt_vars[n].p.telemduckgain = pow(10.0, atof(S_OR(val, DEFAULT_TELEMDUCKDB)) / 20.0);
-
+	RPT_CONFIG_VAR_INT_DEFAULT(votermargin, "votermargin", DEFAULT_VOTERGAIN);
+	RPT_CONFIG_VAR_FLOAT_DB_DEFAULT(telemnomgain, "telemnomdb", DEFAULT_TELEMNOMDB);
+	/* Because asterisk usese an int for gain adjustment and divides when negative
```
Allan-N, line 1036: usese?

apps/app_rpt/rpt_config.c
```diff
@@ -1030,14 +1031,19 @@ void load_rpt_vars(int n, int init)
...
+	 * numbers used, we need the gain to be a non decimal number.  Note: negative gain
```
Allan-N, line 1037: non-decimal / fractional ?

Also, should we be using 1.0 in the below divides ?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1066,9 +1067,13 @@ void *rpt_tele_thread(void *this)
 			type = RPT_CONF;
 			break;
 		default:
-			type = RPT_TELECONF;
+			type = RPT_TXCONF;
 			break;
 	}
+	if (ast_audiohook_volume_set(mychannel, AST_AUDIOHOOK_DIRECTION_WRITE, (int) myrpt->p.telemnomgain)) {
+		ast_debug(7, "I've set the volume on channel %s to %2.2f", ast_channel_name(mychannel), myrpt->p.telemnomgain);
```
Allan-N, line 1074: Setting the volume on channel ...

apps/app_rpt/rpt_config.c
mkmer: It doesn't appear to make a difference, but it's best to be explicit.

apps/app_rpt/app_rpt.h
mkmer: Here is where we are at with this:  These values are back to type `float` and we store them in dB of gain.
If the gain is negative, we convert it to `1 / gain` and change the sign.  With this change, when we convert the gain value to an `int` we get a non fractional number for negative gain with a negative sign. `ast_audiohook*` then uses the sign to determine it needs to divide or multiply gain values.
This now will "function" as we expect, EXCEPT we have "large" steps in the gain where anything <6dB results in a "no change" gain value.   

This table shows the steps:
```
Volume #	dB Gain
-1	        0
-2	        -6.020599913
-3	        -9.542425094
-4	        -12.04119983
-5	        -13.97940009
-6	        -15.56302501
-7	        -16.9019608
-8	        -18.06179974
```

Allan-N: >This now will "function" as we expect, EXCEPT we have "large" steps in the gain where anything <6dB results in a "no change" gain value.

Just asking ... can we do a bit of rounding so that if one asks for 0.1...6.0 we get 6?  That way, we get "some" change?  Or will it be simpler to just document requested vs. actual?

mkmer:
I have submitted a PR for the changes included in the `audiohook.c` and `audiohook.h` files.  If we go this route, we can simply change the `include` back to `"asterisck\"`  and remove the files from `app_rpt` when/if they accept the PR and mainline it.

mkmer:
Waiting on https://github.com/asterisk/asterisk/pull/1168
OR a patch in our CI to pass

apps/app_rpt/rpt_telemetry.c
```diff
@@ -15,6 +15,7 @@
 #include "asterisk/format_cache.h" /* use ast_format_slin */
 
 #include "app_rpt.h"
+#include "asterisk/audiohook.h"
```
InterLinked1, line 18: Asterisk header files should go with the rest of them, before any module specific ones

rpt_install.sh
```diff
@@ -123,6 +123,8 @@ rpt_add "apps/app_rpt/rpt_translate.c"
 rpt_add "apps/app_rpt/rpt_translate.h"
 rpt_add "apps/app_rpt/rpt_rig.c"
 rpt_add "apps/app_rpt/rpt_rig.h"
+rpt_add "apps/app_rpt/audiohook.h"
```
InterLinked1, line 126: I think you forgot to remove these two

apps/app_rpt/rpt_telemetry.c
mkmer: Yes, a hangover from my previous attempt.  Thanks

Allan-N (approved):

apps/app_rpt/rpt_config.c
```diff
@@ -1029,14 +1030,19 @@ void load_rpt_vars(int n, int init)
...
-
-	val = ast_variable_retrieve(cfg, cat, "telemnomdb");
-	rpt_vars[n].p.telemnomgain = pow(10.0, atof(S_OR(val, "0")) / 20.0);
-
-	val = ast_variable_retrieve(cfg, cat, "telemduckdb");
-	rpt_vars[n].p.telemduckgain = pow(10.0, atof(S_OR(val, DEFAULT_TELEMDUCKDB)) / 20.0);
-
+	RPT_CONFIG_VAR_INT_DEFAULT(votermargin, "votermargin", DEFAULT_VOTERGAIN);
+	RPT_CONFIG_VAR_FLOAT_DB_DEFAULT(telemnomgain, "telemnomdb", DEFAULT_TELEMNOMDB);
+	/* Because asterisk uses an int for gain adjustment and divides when negative
```
InterLinked1, line 1035: Asterisk is a proper noun so it should be capitalized in comments

mkmer:
We still don't have the upstream code for this to pass or be merged :(

mkmer:
Should hold off on merging until https://github.com/asterisk/asterisk/issues/1230 is addressed as well.

mkmer:
> Should hold off on merging until [asterisk/asterisk#1230](https://github.com/asterisk/asterisk/issues/1230) is addressed as well.

Actually, this will probably NOT be an issue because the telemetry path is all file sources and should not create the "bad" frame due to failed interpolation.

apps/app_rpt.c
```diff
@@ -4708,7 +4611,7 @@ static void *rpt(void *this)
 	int ms = MSWAIT, lasttx = 0, lastexttx = 0, lastpatchup = 0, val, identqueued, othertelemqueued;
 	int tailmessagequeued, ctqueued, lastmyrx, localmsgqueued;
 	struct rpt_link *l;
-	struct rpt_tele *telem;
+	struct rpt_tele *telem, *last_tele = NULL;
```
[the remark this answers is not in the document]
mkmer: `last_telem` certainly makes more sense...

PR #565, opened 2025-03-18
app_rpt.c: fix non channel type "irlp"

mkmer:
Original code was looking for strings in the channel name such that IRLP and Echolink ~~were both in the name~~.  As written, gain applied to `fs` was not making it to the channel because "echolink" != "irlp".  If it's "irlp", no gain is applied, if it's "echolink", gain is applied but not used...

In the end, "irlp" was apparently removed years ago and appears to be a hangover.  All code related to gain and `fs` can simply be removed.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4478,9 +4478,7 @@ static inline int monchannel_read(struct rpt *myrpt)
 		return -1;
 	}
 	if (f->frametype == AST_FRAME_VOICE) {
-		struct ast_frame *fs;
-		float fac;
-		struct rpt_link *l = &myrpt->links;
+		struct rpt_link *l;
 
 		if ((myrpt->p.duplex > 1) || (myrpt->txkeyed)) {
 			if (myrpt->monstream)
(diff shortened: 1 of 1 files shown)


mkmer:
I'm wanted to wait on this volume adjust stuff.  I think we can use the hooks on the specific channel type and set the gain when we create it, then remove all of the channel specific gain code.

InterLinked1:
I'm a bit confused after reading the commit message - where was there an echolink channel name check in the original code? I don't see it?

Allan-N (approved):

mkmer:
I was a bit "aggressive" on the removal of the frame duplication.  We actually only apply the gain when writing to the echolink channel. Applying the gain to `f` would apply the echolink txgain to all channels (oops).
I believe this is now the most "efficient" way - only dup if there IS and echolink channel.
We also needed to deal with a failure to duplicate - I say it's "ok" to just send a frame with no gain applied if we fail.  If we were to just "exit" no audio would flow (I think that's "bad").  Other thoughts?

mkmer:
Looking at this again.. I think something is just "wrong" with this chunk of code as it was originally.
It looks like we only apply the gain calc if `myrpt->links;` (aka the first link in the list) is echolink,  Then we go through all of the links and write the `fs` frame it that link is echo link, otherwise we write the original frame (assuming I reading the code right).

I think something is generally wrong here....

mkmer:
What a can of worms this one is :(  
This is how I'm seeing this chunk of code:

Anything that comes IN from and echolink channel gets `p.erxgain` applied here (before putting it on the `l->pchan`
https://github.com/AllStarLink/app_rpt/blob/e554ae424e653f28ce468571faf05ab1e953f318/apps/app_rpt.c#L4270-L4278

Anything heading out the door on an echolink channel get `p.etxgain` applied here:
https://github.com/AllStarLink/app_rpt/blob/e554ae424e653f28ce468571faf05ab1e953f318/apps/app_rpt.c#L4426-L4437

EXCEPT if it's `altlink()`.  If it's and `altlink()` then we use the code found in this PR.

I thing is should be: when we receive "audio" from the monochannel we need to apply the `etxgain` to ONLY the echolink channel we are about to copy to... so as we loop through the links we need to see if it's an echolink channel and apply the gain ONLY to that link. (right?)

mkmer:
Oh boy - this was way harder than it need to be.

After reviewing the base code again with @Allan-N , it never can find a link with "irlp" and we NEVER send `fs`.
This boils down to completely removing the code operating on `fs` and the "irlp" test.

Allan-N (approved):

Allan-N (approved):

PR #566, opened 2025-03-19
app_rpt: Properly work around removal of AST_PBX_KEEPALIVE.

InterLinked1:
In the original refactor of app_rpt for current versions of Asterisk, we hacked around the fact that AST_PBX_KEEPALIVE was removed back in https://github.com/asterisk/asterisk/commit/50a25ac8474d7900ba59a68ed4fd942074082435. At the time, there was no correct way to emulate what we were trying to do, as ast_pbx_run_args could only be used when the caller knew the channel wouldn't hang up in advance, and the PBX doesn't do that.

The hack at the time to compile was simply to call pthread_exit() to abort the PBX thread and prevent it from hanging up the channel as it would have when it cleaned up after dialplan finished. This was far from ideal as it was an unclean and improper PBX termination.

Since the PBX args interface doesn't allow setting an arg after the PBX has been started, we instead masquerade the channel so that the original channel is killed off by the PBX thread (which is fine) and the masqueraded channel is what gets used by the repeater threads.

Resolves: #15

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -6780,26 +6780,37 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 			rpt_telemetry(myrpt, CONNECTED, l);
 		}
 
+		/* The way things work here is that other threads in app_rpt service the channel,
+		 * and the PBX thread which initially gave us this channel is going to exit momentarily.
+		 * Originally, we would return AST_PBX_KEEPALIVE to tell the PBX not to hangup
+		 * the channel when terminating the PBX.
+		 * This was removed in Asterisk commit 50a25ac8474d7900ba59a68ed4fd942074082435
+		 *
+		 * The new way things are done does not work for us out of the box, because the PBX
+		 * needs to be told from the get-go not to hangup the channel, and by the time
+		 * dialplan is running, it's already too late.
+		 *
+		 * Instead, we masquerade the channel here to force the old pointer to the channel
+		 * to become invalid. The old channel, now dead, can then get hung up by the
+		 * PBX thread as normal, while the new channel is what we insert into the list. */
+		chan = ast_channel_yank(chan);
+		if (!chan) {
+			/* l->chan still points to the original chan */
+			ast_log(LOG_ERROR, "Failed to masquerade channel %s\n", ast_channel_name(l->chan));
+			ast_free(l);
+			return -1;
+		}
+		l->chan = chan; /* Update pointer to the masqueraded channel. The original channel is dead. */
+
 		/* insert at end of queue */
 		rpt_mutex_lock(&myrpt->lock);
-		rpt_link_add(myrpt, l); /* After putting the link on the linked list, ast_waitfor_n can and will start referencing it */
+		rpt_link_add(myrpt, l); /* After putting the link in the link list, other threads can start using it */
 		__kickshort(myrpt);
 		gettimeofday(&myrpt->lastlinktime, NULL);
 		rpt_mutex_unlock(&myrpt->lock);
 		rpt_update_links(myrpt);
 
-		/* Set the PBX to NULL to avoid a warning in channel.c about a PBX remaining on the channel when it gets destroyed.
-		 * This goes hand in hand with mirroring the old "KEEPALIVE" behavior. Past this point, there is no PBX for this channel.
-		 * We don't do this until the very end, because the node thread will check if this channel has a PBX to determine
-		 * if it's still "owned" by the PBX thread, as opposed to by an app_rpt thread. */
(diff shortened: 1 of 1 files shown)


InterLinked1:
I think we should come up with a good test for this as well, though I haven't worked out what the best things to "look for" are exactly in that.

apps/app_rpt.c
```diff
@@ -6780,6 +6780,28 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
...
+		 * Originally, we would return AST_PBX_KEEPALIVE to tell the PBX not to hangup
+		 * the channel when terminating the PBX.
+		 * This was removed in Asterisk commit 50a25ac8474d7900ba59a68ed4fd942074082435
+		 *
+		 * The new way things are done does not work for us out of the box, because the PBX
+		 * needs to be told from the get-go not to hangup the channel, and by the time
+		 * dialplan is running, it's already too late.
+		 *
+		 * Instead, we masquerade the channel here to force the old pointer to the channel
+		 * to become invalid. The old channel, now dead, can then get hung up by the PBX thread as normal,
```
Allan-N, line 6794: This like is a bit long.

apps/app_rpt.c
```diff
@@ -6788,18 +6810,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
...
-		 * if it's still "owned" by the PBX thread, as opposed to by an app_rpt thread. */
-		rpt_mutex_lock(&myrpt->lock);
-		ast_debug(1, "Stopping PBX on %s\n", ast_channel_name(chan));
-		ast_channel_pbx_set(chan, NULL);
-		rpt_mutex_unlock(&myrpt->lock);
-
-		//return AST_PBX_KEEPALIVE;
-		pthread_exit(NULL); // BUGBUG: For now, this emulates the behavior of KEEPALIVE, but this won't be a clean exit. Makes it work, but since the PBX doesn't clean up we'll leak memory. Either do what the PBX core does here or we need to somehow do KEEPALIVE handling in the core, possibly with a custom patch for now.
-		//return -1;				/*! \todo AST_PBX_KEEPALIVE doesn't exist anymore. Figure out what we should return here. */
+		return -1; /* We can now safely return -1 to the PBX, as the old channel prior to the masquerade is what will get killed off */
```
Allan-N, line 6813: The comment is really long.  Can it be reformatted / broken out into multiple lines?

apps/app_rpt.c
```diff
@@ -6780,6 +6780,28 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
...
+			/* l->chan still points to the original chan */
+			ast_log(LOG_ERROR, "Failed to masquerade channel %s\n", ast_channel_name(l->chan));
+			ast_free(l);
+			return -1;
+		}
+		l->chan = chan; /* Update pointer to the masqueraded channel. The original channel is dead. */
+
 		/* insert at end of queue */
 		rpt_mutex_lock(&myrpt->lock);
 		rpt_link_add(myrpt, l); /* After putting the link on the linked list, ast_waitfor_n can and will start referencing it */
```
Allan-N, line 6807: The comment is really long. Can it be reformatted / broken out into multiple lines?

mkmer (approved):
Did a bit of testing - appears to work fine

apps/app_rpt.c
InterLinked1: The lines only go to 107, 130, and 97 columns respectively when I check, are they showing as longer for you?
This last one in particular I don't think is long at all... obviously this is just my opinion :)

I can do some more wrapping on the multiline comment but I'd prefer to leave the single line comments inline on the same line.

Allan-N: The l.6794 comment I flagged  (w/114 chars) is 16 characters longer than the surrounding comments.  The l.6807 comment is 136 chars.  The l.6813 comment is 143 chars.

Do I really need to have 2 27"+ monitors with windows at full width to do a side-by-side comparison of the old vs new changes?

InterLinked1: > The l.6794 comment I flagged (w/114 chars) is 16 characters longer than the surrounding comments. The l.6807 comment is 136 chars. The l.6813 comment is 143 chars.
> 
> Do I really need to have 2 27"+ monitors with windows at full width to do a side-by-side comparison of the old vs new changes?

I think code length is driven mainly be looking at the code itself (only), but doing a side by side like a review. It could be the older/longer code is reducing the real estate for showing the new stuff.

I'll trim some of it down later today... I usually look at code on a 24" screen and the editor doesn't even take up the full space so I'll double check it.

Allan-N (approved):

PR #568, opened 2025-03-19
app_rpt: Simplify sayfile logic.

InterLinked1:
ast_stream_and_wait does everything that this function was previously doing manually, except for emitting a warning, which isn't necessary since one will have been logged already in ast_streamfile if needed.


Allan-N (approved):

mkmer (approved):

mkmer:
I "think" there are a pile of these in the telemetry thread.  A mix of both ways + an imdone bit to manage it all.

PR #570, opened 2025-03-20
app_rpt.c: Remove unnecessary duplicate send_newkey()

mkmer:
As suspected, the `send_newkey()` only needs to happen once per connection.
Testing confirmed.


InterLinked1 (approved):
If "testing confirmed", any way we can capture the test you did in the test suite for the future?

Allan-N (approved):

PR #571, opened 2025-03-20
app_rpt.c: Use ast_frame_adjust_volume_float()

mkmer:
Essentially the "same" math.  No need to roll our own.
Once Asterisk accepts our PR, we will be able to adjust the gains when the specific channel type is created.  This will 100% eliminate the code in calling `ast_frame_adjust_volume_float()`.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4181,24 +4181,6 @@ static void remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 	rpt_link_free(l);
 }
 
-static inline void fac_frame(struct ast_frame *restrict f, float fac)
-{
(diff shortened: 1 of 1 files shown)


Allan-N (approved):

InterLinked1 (approved):
Looks good to me. But this would only work after that other PR is merged, how is this building successfully right now?

mkmer:
This replacement here is the `ast_frame_adjust_volume_float()` that already exists (and does the "same" math with the possibility of feeding it a (-) value instead of multiplying by a fractional value or just multiply by a fractional value. 
When the PR is mainlined, we can refactor this code (removing it from here, and setting up an audiohook volume when the channel is created)

PR #573, opened 2025-03-20
app_rpt.c: use snprintf for all sprintf

mkmer:


apps/app_rpt.c
```diff
@@ -5189,10 +5193,10 @@ static void *rpt(void *this)
 				break;
 			}
 
-			sprintf(myfname, PARROTFILE, myrpt->name, myrpt->parrotcnt);
-			strcat(myfname, ".wav");
+			snprintf(myfname, sizeof(myfname), PARROTFILE, myrpt->name, myrpt->parrotcnt);
+			strncat(myfname, ".wav", sizeof(myfname) - 1);
```
InterLinked1, line 5197: Couldn't this strncat be removed and just be part of the snprintf on the previous line?
Also, we overwrite the buffer on the following line - so aren't these 2 lines useless?

apps/app_rpt.c
```diff
@@ -2346,7 +2346,7 @@ static void local_dtmf_helper(struct rpt *myrpt, char c_in)
 
 	c = c_in & 0x7f;
 
-	sprintf(tone, "%c", c);
+	snprintf(tone, sizeof(tone), "%c", c);
```
InterLinked1, line 2349: In this particular case, we could set the buffer size to 2, and safely use sprintf. This will never need more than 2 bytes.

Allan-N: Yes, you could do that. But, at some future date someone might say we need to add a longer string and the miss the fact that the buffer is not large enough.  Simply put, we should not use sprintf!

InterLinked1: Yes, that risk exists... It didn't seem like something that would be likely to change. A comment explaining why it's safe would probably handle that. 

I'm not saying it has to be changed, but I do want to push back on the idea these functions should *never* be used. It isn't in the same class as gets. As an example, `strcpy` is used extensively in Asterisk. It's only used when it's 100% clear that it's safe to do so. Often times, the comment `/* Safe */` appears on the same line so that it's clear at a glance that it was verified that the usage is safe.

sprintf has far fewer use cases than strcpy, so I'm okay with eliminating it here if that's the preference.

mkmer: My thoughts are similar to Allan's - I suppose in this case it's fairly "sure" that it will fit.  
Most (if not all) of the cases in this PR are inputs coming from configuration files into fixed buffers.  These could be any size and I think we should protect them.

InterLinked1: > Most (if not all) of the cases in this PR are inputs coming from configuration files into fixed buffers. These could be any size and I think we should protect them.

Oh yes, of course, absolutely! Anything the user can do should have bounds checks. I just pointed this out as a case where it was known at compile time it would be fine. But that's the exception and not the norm. Thank you for fixing all these!

Allan-N: \<soapbox>

>I'm not saying it has to be changed, but I do want to push back on the idea these functions should never be used.

"Never" is a strong word ... and I said "we should not use".  Given that the number of buffer overflow exploits has been rapidly increasing, especially in code that was written long ago by those who had not contemplated this type of attack vector, I think that we should make any/every effort to minimize the risks.

> It isn't in the same class as gets. As an example, strcpy is used extensively in Asterisk. It's only used when it's 100% clear that it's safe to do so. Often times, the comment /* Safe */ appears on the same line so that it's clear at a glance that it was verified that the usage is safe.

and if "I" was asked to put on my "secure coding" hat and perform a complete review of the Asterisk code I would question/review each and every place that strcpy/strcat is used and each place where the code was tagged with /* Safe */.  The tag may have been added because, at the time the dev knew the input buffer being passed in was limited to 80 bytes.  But, if the input buffer size was (or is ever) increased then every called function may now be the source of an overflow.

\</soapbox>

InterLinked1: > and if "I" was asked to put on my "secure coding" hat and perform a complete review of the Asterisk code I would question/review each and every place that strcpy/strcat is used and each place where the code was tagged with /* Safe */. The tag may have been added because, at the time the dev knew the input buffer being passed in was limited to 80 bytes. But, if the input buffer size was (or is ever) increased then every called function may now be the source of an overflow.
> 
> </soapbox>

Those aren't the scenarios that I'm talking about. I typically see it when it's guaranteed that the copy will be safe. It doesn't matter how big it is since it's not possible to overflow.

Here's an example of this sort of thing. No matter what the lengths are, overflow is not possible since it's guaranteed the buffer is big enough: https://github.com/asterisk/asterisk/blob/f6d296256612d40ef787753a04040a68264aa16c/main/pbx_sw.c#L89

Allan-N: >Those aren't the scenarios that I'm talking about. I typically see it when it's guaranteed that the copy will be safe. It doesn't matter how big it is since it's not possible to overflow.

When you are computing the length of every string involved, managing the allocation, remembering all of the +1's, etc then, yes, all should be good ... until someone changes the code and forgets to update the size/pointer calculations.

InterLinked1: > > Those aren't the scenarios that I'm talking about. I typically see it when it's guaranteed that the copy will be safe. It doesn't matter how big it is since it's not possible to overflow.
> 
> When you are computing the length of every string involved, managing the allocation, remembering all of the +1's, etc then, yes, all should be good ... until someone changes the code and forgets to update the size/pointer calculations.

In that case, snprintf still fails, since the size of the buffer was probably calculated wrong. Just using the "safe" function won't guarantee safety in that case.

Allan-N (approved):

Issue #511, opened 2025-02-23
Permanent Connections Are Not Necessarily Permanent


jxmx:
https://community.allstarlink.org/t/creating-a-requested-feature-list/22699/7?u=n8ei

Make “permanent” connections actually permanent. e.g. reconnect after network interruptions.

PR #574, opened 2025-03-20
app_rpt.c: Address issue 511 eliminate retry timeout on permanent connections

mkmer:
indefinite retries for permanent connections
Closing #511


InterLinked1 (commented):
Commit message should:
- not contain "Address issue 511" in the title - just "app_rpt: Eliminate retry timeout on permanent connections" is good for the first line
- should end with `Resolves: #511` so it actually closes the issue

InterLinked1 (approved):

PR #576, opened 2025-03-21
app_rpt.c: Remove blocklock

mkmer:
In my quest to figure out what is causing the USB buffer under run, I noticed the commit message for adding the `blocklock` mentioned audio challenges.  I have been running a couple of nodes with the `blocklock` removed with no adverse effects.  On the positive side, it greatly reduced the audio artifacts (nearly none) when the USB is set up for low latency. 

I suspect there were other issues/bugs/mistakes that have now been addressed and we no longer need the `blocklock`.
I think we should test/verify this on a larger group of nodes.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3533,9 +3533,7 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 	struct ast_frame *f, *f1;
 	int i, dtmfed = 0;
 
-	rpt_mutex_lock(&myrpt->blocklock);
 	f = ast_read(myrpt->rxchannel);
-	rpt_mutex_unlock(&myrpt->blocklock);
 	if (!f) {
 		ast_debug(1, "@@@@ rpt:Hung Up\n");
 		return -1;
@@ -4050,9 +4048,7 @@ static inline int dahditxchannel_read(struct rpt *myrpt, char *restrict myfirst)
 			while ((f1 = AST_LIST_REMOVE_HEAD(&myrpt->txq, frame_list)))
 				ast_frfree(f1);
 		}
-		rpt_mutex_lock(&myrpt->blocklock);
 		ast_write(myrpt->txchannel, f);
-		rpt_mutex_unlock(&myrpt->blocklock);
 	}
 	return hangup_frame_helper(myrpt->dahditxchannel, "dahditxchannel", f);
 }
@@ -4376,9 +4372,7 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 					l->thisconnected = 1;
 					l->elaptime = -1;
 					if (!l->phonemode) {
-						rpt_mutex_lock(&myrpt->blocklock);
 						send_newkey(l->chan);
-						rpt_mutex_unlock(&myrpt->blocklock);
 					}
 					if (!l->isremote)
 						l->retries = 0;
@@ -4781,11 +4775,9 @@ static void *rpt(void *this)
 		setrem(myrpt);
 	/* wait for telem to be done */
 	while ((ms >= 0) && (myrpt->tele.next != &myrpt->tele)) {
-		rpt_mutex_lock(&myrpt->blocklock);
 		if (ast_safe_sleep(myrpt->rxchannel, 50) == -1) {
 			ms = -1;
 		}
-		rpt_mutex_unlock(&myrpt->blocklock);
 	}
 	lastmyrx = 0;
 	myfirst = 0;
@@ -5248,9 +5240,7 @@ static void *rpt(void *this)
 			cs1[x] = cs[s];
 		}
 		myrpt->scram++;
-		rpt_mutex_lock(&myrpt->blocklock);
 		who = ast_waitfor_n(cs1, n, &ms);
-		rpt_mutex_unlock(&myrpt->blocklock);
 		if (who == NULL) {
 			ms = 0;
 		}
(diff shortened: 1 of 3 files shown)


mkmer:
> I think you're probably right to some extent and it would be nice if we could remove. One reason it was added initially was multiple threads were simultaneously servicing the same channel, which is not allowed in Asterisk. This doesn't happen consistently, only occasionally. We would want to ensure that we prevent that somehow and that we test it out.

Based on the architecture I was mapping out (diagram I shared), the design is such that no code is accessing channels (for voice frames) EVER at the same time.   If we find a place where this is an issue, it's probably a mistake in what channel is being referenced.

The issue I found in the paging logic was a "mistake" causing a double access.  

I'm not sure how to "prove" this other than running it IRL for on systems.
Obviously, some tests would be great, but exercising the whole system via tests won't be ready any time "soon".

All that being said, removing the additional lock does effect audio for the better, so if we can get more comfortable with it out that will be good.  It's possible holding the lock as we are is part (or all) of the "twang"/ digital noise problem.

One question related:  Are we NOT allowed to send text frames and voice frames at the same time? (Seems like that shouldn't be a problem).

InterLinked1:
> > I think you're probably right to some extent and it would be nice if we could remove. One reason it was added initially was multiple threads were simultaneously servicing the same channel, which is not allowed in Asterisk. This doesn't happen consistently, only occasionally. We would want to ensure that we prevent that somehow and that we test it out.
> 
> Based on the architecture I was mapping out (diagram I shared), the design is such that no code is accessing channels (for voice frames) EVER at the same time. If we find a place where this is an issue, it's probably a mistake in what channel is being referenced.
> 
> The issue I found in the paging logic was a "mistake" causing a double access.

Thanks, that's good to know!

> I'm not sure how to "prove" this other than running it IRL for on systems. Obviously, some tests would be great, but exercising the whole system via tests won't be ready any time "soon".

Of course, it will probably take a long time before that happens - but even if we add one test a week, we will slowly build more coverage and that is a lot better than nothing.

> All that being said, removing the additional lock does effect audio for the better, so if we can get more comfortable with it out that will be good. It's possible holding the lock as we are is part (or all) of the "twang"/ digital noise problem.

I wouldn't be opposed to removing it sort of proboationary if we wanted to explore that. But I think we'd want to test it extensively on beta systems for some time before rolling that out more broadly. Previously, there were a lot of systems crashing due to the simultaneous servicing issue.

> One question related: Are we NOT allowed to send text frames and voice frames at the same time? (Seems like that shouldn't be a problem).

Not at the same exact time, no (as with any simultaneous servicing), but if it's serialized in some (any) way, I don't see how it would be an issue.

InterLinked1:
> We must have two threads somewhere or we wouldn't get the error (right?)

Yeah, likely two different threads simultaneously servicing the same channels (e.g. `ast_wait_for_n` or a related function called in 2 different places on the same channel)

> On that note, I was looking over the ways we send text frames... Some go in a queue, some go with ast_text_send, and some go right to the channel... Should these all be ast_send_text, or use the frame construction method? Should the all go to the queue?

I think `ast_send_text` is just a helper function to make it easier to send text. I don't know if the other ways are "wrong" but it's possible they could be simplified. There may have been something the helper function does that didn't work in those cases, I haven't investigated.

mkmer:
It looks like we have text frames that need to go to multiple links (distribute to all links) -> We put those messages on a queue and "peal" them off one at a time here:
https://github.com/AllStarLink/app_rpt/blob/e554ae424e653f28ce468571faf05ab1e953f318/apps/app_rpt.c#L3045-L3048

Everything in the `rpt()` thread is (I think) single threaded so we should be able to write to channels without multiple access...
The `rpt_exec()` function IS a new thread, but `rpt()` shouldn't know about it until we add it to the linked list... 

I'm 99.99% sure this has been addressed in this commit [78845176f3c725fb4da5fc5a8d5ab1bf67dbc768](https://github.com/AllStarLink/app_rpt/commit/78845176f3c725fb4da5fc5a8d5ab1bf67dbc768) by moving `rpt_link_add()` to a location after we `send_newkey()`.

Allan-N (approved):

Allan-N:
>I wouldn't be opposed to removing it sort of proboationary if we wanted to explore that. But I think we'd want to test it extensively on beta systems for some time before rolling that out more broadly.

I also think we should move forward on getting a few others involved in testing a set of packages with these changes.  We can start off with a limited set of developers/users, then expand to our #beta testers, and keep expanding the availability.

InterLinked1 (approved):

channels/chan_simpleusb.c
```diff
@@ -73,6 +73,7 @@
 #define PP_PORT "/dev/parport0"
 #define PP_IOPORT 0x378
 #define N_FMT(duf) "%30" #duf /* Maximum sscanf conversion to numeric strings */
+#define HID_POLL_RATE 50
```
InterLinked1, line 76: This change seems unrelated to the rest of the change...?

Allan-N: I think that change was to keep the CI happy

Allan-N: Nope.  Looks like @mkmer made the change (to better clarify the previously hard-coded "50" ... or maybe to see how the behavior would change if modified ... or ...).

mkmer: Yes, that was a change that snuck in here. It should be in a different pr.

PR #578, opened 2025-03-22
app_rpt: Use init_text_frame

mkmer:
Use `init_text_frame()` for all instances of text frames.
send text messages with rpt_qwrite() when writing to links

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1588,7 +1588,7 @@ static void do_aprstt(struct rpt *myrpt)
 	}
 }
 
-static int distribute_to_all_links(struct rpt *myrpt, struct rpt_link *mylink, const char *src, const char *dest, char *str, struct ast_frame *wf)
+static int distribute_to_all_links(struct rpt *myrpt, struct rpt_link *mylink, const char *src, const char *dest, struct ast_frame *wf)
 {
 	struct rpt_link *l = myrpt->links.next;
 	/* see if this is one in list */
@@ -1605,7 +1605,6 @@ static int distribute_to_all_links(struct rpt *myrpt, struct rpt_link *mylink, c
 		if (!dest || !strcmp(l->name, dest)) {
 			/* send, but not to src */
 			if (strcmp(l->name, src)) {
-				wf->data.ptr = str;
 				if (l->chan) {
 					rpt_qwrite(l, wf);
 				}
@@ -1675,15 +1674,6 @@ static int funcchar_common(struct rpt *myrpt, char c)
 	return 0;
 }
 
-static inline void init_text_frame(struct ast_frame *wf)
-{
-	wf->frametype = AST_FRAME_TEXT;
-	wf->subclass.format = ast_format_slin;
-	wf->offset = 0;
-	wf->mallocd = 0;
-	wf->samples = 0;
-}
-
 static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *str)
 {
 	/* XXX cmd, dst, and src should be validated. Why is remote_data src[300] in other locations?
@@ -1693,10 +1683,9 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	int i, seq, res, ts, rest;
 	struct ast_frame wf;
 
-	init_text_frame(&wf);
+	init_text_frame(&wf, "handle_link_data");
 	wf.datalen = strlen(str) + 1;
-	wf.src = "handle_link_data";
-
+	wf.data.ptr = str;
 	ast_debug(5, "Received text over link: '%s'\n", str);
 
 	if (!strcmp(str, DISCSTR)) {
@@ -1724,7 +1713,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 	if (*str == 'G') {		/* got GPS data */
 		/* re-distribute it to attached nodes */
-		distribute_to_all_links(myrpt, mylink, src, NULL, str, &wf);
+		distribute_to_all_links(myrpt, mylink, src, NULL, &wf);
 		return;
 	}
 	if (*str == 'L') {
@@ -1761,7 +1750,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 			ast_verb(3, "Text Message From %s: %s\n", src, str + rest);
 			ast_debug(1, "Node %s Got Text Message From Node %s: %s\n", myrpt->name, src, str + rest);
 		}
-		distribute_to_all_links(myrpt, mylink, src, NULL, str, &wf);
+		distribute_to_all_links(myrpt, mylink, src, NULL, &wf);
 		return;
 	}
 	if (*str == 'T') {
@@ -1770,7 +1759,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 			return;
 		}
 		/* otherwise, send it to all of em */
-		distribute_to_all_links(myrpt, mylink, src, NULL, str, &wf);
+		distribute_to_all_links(myrpt, mylink, src, NULL, &wf);
 		/* if is from me, ignore */
 		if (!strcmp(src, myrpt->name))
 			return;
@@ -1805,7 +1794,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 			return;
 		if (strcasecmp(myrpt->p.ctgroup, tmp1))
 			return;
-		distribute_to_all_links(myrpt, mylink, src, NULL, str, &wf);
+		distribute_to_all_links(myrpt, mylink, src, NULL, &wf);
 		/* if is from me, ignore */
 		if (!strcmp(src, myrpt->name))
 			return;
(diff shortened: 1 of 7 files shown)


InterLinked1 (commented):
Looks good! Only change I would do is change `Multiple` in the commit message title to `app_rpt` since this is all for one module

mkmer:
It looks like we have text frames that need to go to multiple links (distribute to all links) -> We put those messages on a queue and "peal" them off one at a time here:
https://github.com/AllStarLink/app_rpt/blob/e554ae424e653f28ce468571faf05ab1e953f318/apps/app_rpt.c#L3045-L3048

Everything in the `rpt()` thread is (I think) single threaded so we should be able to write to channels without multiple access...
The `rpt_exec()` function IS a new thread, but `rpt()` shouldn't know about it until we add it to the linked list... 

I'm 99.99% sure this has been addressed in this commit [78845176f3c725fb4da5fc5a8d5ab1bf67dbc768](https://github.com/AllStarLink/app_rpt/commit/78845176f3c725fb4da5fc5a8d5ab1bf67dbc768) by moving `rpt_link_add()` to a location after we `send_newkey()`.

apps/app_rpt/rpt_functions.c
```diff
@@ -298,8 +295,7 @@ int function_ilink(struct rpt *myrpt, char *param, char *digits, int command_sou
 				rpt_mutex_unlock(&myrpt->lock);
 				ast_debug(5, "dumping link %s\n",l->name);
 
-				memset(&wf, 0, sizeof(wf));
-				wf.frametype = AST_FRAME_TEXT;
+				init_text_frame(&wf);
 				wf.datalen = strlen(DISCSTR) + 1;
 				wf.data.ptr = DISCSTR;
 				wf.src = "function_ilink:6";
```
Allan-N, line 301: This should really be "function_ilink:10"

apps/app_rpt/rpt_utils.c
```diff
@@ -259,3 +260,10 @@ void update_timer(int *timer_ptr, int elap, int end_val)
 		*timer_ptr = end_val;
 	}
 }
+
+void init_text_frame(struct ast_frame *wf)
```
Allan-N, line 264: Here's a thought.  Maybe change the function prototype to :
```
void init_text_frame(struct ast_frame *wf, const char *src)
```
and then add `wf->src = src;`

and then the callers would use : `init_text_frame(&wf, "function_ilink:1");`

apps/app_rpt/rpt_functions.c
mkmer: Nice....

mkmer:
> Looks good! Only change I would do is change `Multiple` in the commit message title to `app_rpt` since this is all for one module

There are multiple files modified here - have I miss understood?

InterLinked1:
> > Looks good! Only change I would do is change `Multiple` in the commit message title to `app_rpt` since this is all for one module
> 
> There are multiple files modified here - have I miss understood?

It's all part of `app_rpt`, which is the same module.

The rule of thumb I typically use is:
- If all modifications are part of the same module, use the module name, e.g. `app_rpt`
- If the modifications are part of the core (not a module), use the filename, e.g. `frame.c` (doesn't apply to this repo really)
- If they affect multiple modules of the same type, use the prefix, e.g. `apps` or `channels`

So the appropriate thing would be `app_rpt` here, you're not modifying multiple modules.

apps/app_rpt/rpt_utils.c
InterLinked1: Even better, we can set up a macro: `#define init_text_frame(fr) __init_text_frame(fr, __func__)`

Then, just call `init_text_frame(&wf)` everywhere. The correct function name is then automatically passed to the function.

To be honest, I don't know why `frame.h` doesn't already have a macro that does this, if the location is typically just the function name. But this seems like the cleanest way to do that to avoid copy/paste errors.

apps/app_rpt/rpt_link.c
```diff
@@ -337,13 +332,8 @@ void send_link_dtmf(struct rpt *myrpt, char c)
 	struct rpt_link *l;
 
 	snprintf(str, sizeof(str), "D %s %s %d %c", myrpt->cmdnode, myrpt->name, ++(myrpt->dtmfidx), c);
-	wf.frametype = AST_FRAME_TEXT;
-	wf.subclass.format = ast_format_slin;
-	wf.offset = 0;
-	wf.mallocd = 0;
+	init_text_frame(&wf, "send_link_dtmf");
 	wf.datalen = strlen(str) + 1;
```
Allan-N, line 336: Why do we set wf.datalen here but wf.data.ptr below (inside the while loop) ?

apps/app_rpt.c
```diff
@@ -1693,9 +1684,8 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	int i, seq, res, ts, rest;
 	struct ast_frame wf;
 
-	init_text_frame(&wf);
+	init_text_frame(&wf, "handle_link_data");
 	wf.datalen = strlen(str) + 1;
```
Allan-N, line 1688: Seems odd that we are setting wf.datalen here but not wf.data.ptr

apps/app_rpt/rpt_utils.c
InterLinked1: I see. In that case, maybe a macro we can use when it's the function name and in other cases, we can override directly?

apps/app_rpt.c
mkmer: I think your right - makes sense to set the value here, then stop passing it around to `distribute_to_all_links`...

apps/app_rpt/rpt_link.c
mkmer: This is one of those "ask Jim" questions... 
I think we should move it.

mkmer:
Looking the whole flow over here, I think we can completely eliminate the `rpt_qwrite` and the queue an simply write to the channels at each `rpt_qwrite` location. No need to do all that messing around.

apps/app_rpt/rpt_link.c
```diff
@@ -382,13 +372,8 @@ void send_link_keyquery(struct rpt *myrpt)
 	time(&myrpt->topkeytime);
 	rpt_mutex_unlock(&myrpt->lock);
 	snprintf(str, sizeof(str), "K? * %s 0 0", myrpt->name);
-	wf.frametype = AST_FRAME_TEXT;
-	wf.subclass.format = ast_format_slin;
-	wf.offset = 0;
-	wf.mallocd = 0;
+	init_text_frame(&wf, "send_link_keyquery");
 	wf.datalen = strlen(str) + 1;
```
Allan-N, line 376: Missed this one earlier.  We set wf.datalen her and wf.data.ptr inside the while loop (below)

apps/app_rpt/rpt_link.c
```diff
@@ -406,13 +391,8 @@ void send_tele_link(struct rpt *myrpt, char *cmd)
 	struct rpt_link *l;
 
 	snprintf(str, sizeof(str) - 1, "T %s %s", myrpt->name, cmd);
-	wf.frametype = AST_FRAME_TEXT;
-	wf.subclass.format = ast_format_slin;
-	wf.offset = 0;
-	wf.mallocd = 0;
+	init_text_frame(&wf, "send_tele_link");
 	wf.datalen = strlen(str) + 1;
```
Allan-N, line 395: Missed this one earlier. We set wf.datalen her and wf.data.ptr inside the while loop (below)

apps/app_rpt/rpt_link.c
```diff
@@ -362,8 +352,8 @@ void send_link_dtmf(struct rpt *myrpt, char c)
 	}
 	l = myrpt->links.next;
 	/* if not, give it to everyone */
+	wf.data.ptr = str;
```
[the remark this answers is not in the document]
mkmer: Sure doesn't look it to me.

apps/app_rpt.c
```diff
@@ -3933,9 +3917,8 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 				myrpt->paging.tv_usec = 0;
 			} else {
 				snprintf(str, sizeof(str), "V %s %s", myrpt->name, (char *) f->data.ptr);
-				init_text_frame(&wf);
+				init_text_frame(&wf, "voter_text_send");
 				wf.datalen = strlen(str) + 1;
```
Allan-N, line 3921: Move wf.data.ptr from below to here :)

apps/app_rpt/rpt_link.c
```diff
@@ -307,13 +307,8 @@ void rssi_send(struct rpt *myrpt)
 	struct ast_frame wf;
 	char str[200];
 	sprintf(str, "R %i", myrpt->rxrssi);
-	wf.frametype = AST_FRAME_TEXT;
-	wf.subclass.format = ast_format_slin;
-	wf.offset = 0;
-	wf.mallocd = 0;
+	init_text_frame(&wf, "rssi_send");
 	wf.datalen = strlen(str) + 1;
```
Allan-N, line 311: Move wf.data.ptr from below to here :)

Allan-N (commented):
Apologies, again, for not noticing these earlier (the GitHub diffs, without expanding sections, don't always provide enough context)

mkmer:
Better hold off on this one - something in here is crashing Asterisk

Allan-N (approved):

Allan-N:
>Better hold off on this one - something in here is crashing Asterisk

Crash found.  Fix in https://github.com/AllStarLink/app_rpt/pull/579

mkmer:
PR#579 addresses the issue.  Nothing wrong with this PR

Issue #536, opened 2025-03-03
app_rpt: Channel already being serviced when calling send_newkey/ast_sendtext


InterLinked1:
Assertion from a test execution yesterday:

I seem to recall seeing this before occasionally as well. Probably only happens under certain conditions.
This is unrelated to #252 

```
[2025-03-01 23:13:23.930] ERROR[387445][C-00000013] : Got 14 backtrace records
# 0: [0x55f367b531a6] asterisk utils.c:2835 __ast_assert_failed()
# 1: [0x55f3679d463a] asterisk utils.h:737 _ast_assert()
# 2: [0x55f3679e3502] asterisk channel.c:4786 ast_sendtext_data()
# 3: [0x55f3679e3b26] asterisk channel.c:4851 ast_sendtext()
# 4: [0x7fe4a722c394] app_rpt.so rpt_channel.c:473 send_newkey()
# 5: [0x7fe4a72242d3] app_rpt.so app_rpt.c:6822 rpt_exec()
# 6: [0x55f367a9d3aa] asterisk pbx_app.c:492 pbx_exec()
# 7: [0x55f367a86b79] asterisk pbx.c:2980 pbx_extension_helper()
# 8: [0x55f367a8aea9] asterisk pbx.c:4235 ast_spawn_extension()
# 9: [0x55f367a8bc96] asterisk pbx.c:4409 __ast_pbx_run()
#10: [0x55f367a8d733] asterisk pbx.c:4734 pbx_thread()
#11: [0x55f367b4ff48] asterisk utils.c:1607 dummy_start()
#12: [0x7fe4e70a81c4] libc.so.6 pthread_create.c:442 start_thread()
#13: [0x7fe4e712885c] libc.so.6 clone3.S:83 clone3()
```

mkmer:
I'm curious if this too can be caused by adding the link to the linked list, allowing waitfor_n or other link processing to begin a bit early... Have to keep an eye out now the code is changed a bit.
ref PR https://github.com/AllStarLink/app_rpt/pull/525

mkmer:
Could this be the result of calling `send_newkey()` at the same time (or at least overlapping in some part)?
We call it a couple of times when we receive a new call (from dialplan):
https://github.com/AllStarLink/app_rpt/blob/9db0ce2f3ca6a6edd73c1b313ec2ddfe3d62d8b4/apps/app_rpt.c#L6791-L6796
and here
[...]

PR #582, opened 2025-03-23
app_rpt: Remove textq

mkmer:
There is no need to queue text message frames on links.  Every "access" to the channel is located in a single thread (`rpt()`) such that simply writing to the channel is all the is needed.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1619,7 +1619,7 @@ static int distribute_to_all_links(struct rpt *myrpt, struct rpt_link *mylink, c
 			/* send, but not to src */
 			if (strcmp(l->name, src)) {
 				if (l->chan) {
-					rpt_qwrite(l, wf);
+					ast_write(l->chan, wf);
 				}
 			}
 			if (dest) {
@@ -1846,7 +1846,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 			wf.data.ptr = tmp1;
 			wf.datalen = strlen(tmp1) + 1;
 			if (mylink->chan)
-				rpt_qwrite(mylink, &wf);
+				ast_write(mylink->chan, &wf);
 			return;
 		}
 		if (myrpt->topkeystate != 1)
@@ -2287,7 +2287,6 @@ static int attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	char *s1, *tele;
 	char tmp[300], deststr[325] = "";
 	char sx[320];
-	struct ast_frame *f1;
 	struct ast_format_cap *cap;
 
 	if (node_lookup(myrpt, l->name, tmp, sizeof(tmp) - 1, 1)) {
@@ -2331,8 +2330,6 @@ static int attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	l->rxlingertimer = RX_LINGER_TIME;
 	l->newkeytimer = NEWKEYTIME;
 	l->link_newkey = RADIO_KEY_NOT_ALLOWED;
-	while ((f1 = AST_LIST_REMOVE_HEAD(&l->textq, frame_list)))
-		ast_frfree(f1);
 	if (l->chan) {
 		rpt_make_call(l->chan, tele, 999, deststr, "(Remote Rx)", "attempt_reconnect", myrpt->name);
 	} else {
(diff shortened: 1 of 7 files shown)


Allan-N (commented):
At first look, these changes look to be clean.  But, I've gotta ask a few questions.
- Prior to this change we were putting the message (non-voice) frames on a separate queue, correct? ... and managing them separately?
- Now, we're writing the message (non-voice) frames on the channel along with any voice frames, correct?
- Do we have any code that is acting on frames thinking that they will be voice frames?  Asking because I saw one place that looked to be calling RPT_MUTE_FRAME() and I want to know that we're not clearing an important message.
- Do we have any code that is emptying out the frame queue that may also be removing message frames that really need to be processed (e.g. an UNKEY)?

InterLinked1:
> At first look, these changes look to be clean. But, I've gotta ask a few questions.
> 
> * Prior to this change we were putting the message (non-voice) frames on a separate queue, correct? ... and managing them separately?
> * Now, we're writing the message (non-voice) frames on the channel along with any voice frames, correct?
> * Do we have any code that is acting on frames thinking that they will be voice frames?  Asking because I saw one place that looked to be calling RPT_MUTE_FRAME() and I want to know that we're not clearing an important message.
> * Do we have any code that is emptying out the frame queue that may also be removing message frames that really need to be processed (e.g. an UNKEY)?

To add onto this, it would be nice if we could identify the functionality affected by this and add some simple tests capturing that behavior. Both for this change, and for future ones, that would give us more confidence that nothing has broken.

mkmer:
> At first look, these changes look to be clean. But, I've gotta ask a few questions.
> 
> * Prior to this change we were putting the message (non-voice) frames on a separate queue, correct? ... and managing them separately?
Yes, we were putting messages on a textq, then sending them 1 per loop in the `periodic_process_links()` here: https://github.com/AllStarLink/app_rpt/blob/eaeb0a1457320b023023b0c9fb893d2fbc8d9b6b/apps/app_rpt.c#L3034-L3038

This whole thing involves allocating a duplicate frame, putting on a queue, then writing it to the channel.

> * Now, we're writing the message (non-voice) frames on the channel along with any voice frames, correct?

We are doing exactly the same write as before, just at the place where we were writing to the queue.

> * Do we have any code that is acting on frames thinking that they will be voice frames?  Asking because I saw one place that looked to be calling RPT_MUTE_FRAME() and I want to know that we're not clearing an important message.

This should not be affected by any other code.  We generate the frame and send it before any other logic.  These are going directly out the `l->chan` which is the physical connection port.  No internal copying happens on the `l->chan` output.

> * Do we have any code that is emptying out the frame queue that may also be removing message frames that really need to be processed (e.g. an UNKEY)?

These messages always go out the door.  Using the new method, hanging up on a channel will cleanup anything that didn't make it out the door. (aka on the channel queue)

Generally, we are eliminating an extra frame copy, queue and unqueue intermediate process that has zero purpose except to consume CPU.

mkmer:
> To add onto this, it would be nice if we could identify the functionality affected by this and add some simple tests capturing that behavior. Both for this change, and for future ones, that would give us more confidence that nothing has broken.

We definitely need to build tests that verify triggers create the text messages and they arrive on the other end...   They are essentially out of band control messages that are necessary (like "NEWKEY" and "L  T123").
It was the NEWKEY logic that brought this to my attention -> we were already sending those outside of the textq in multiple places.

mkmer:
Another way to do this COULD be to use the `ast_sendtext()` and not build frames at all - just send the payloads. 
After a bit of exploration, it looked like we would lose some functionality (like setting the `src`), but it may still be possible and I just didn't see how.

Allan-N:
>Another way to do this COULD be to use the ast_sendtext() and not build frames at all - just send the payloads.

Would that work for non-current ASL3 (and other flavors of) nodes?

mkmer:
> > Another way to do this COULD be to use the ast_sendtext() and not build frames at all - just send the payloads.
> 
> Would that work for non-current ASL3 (and other flavors of) nodes?

Looking at the code, the `frame->src` looks to contain "dialplan" when we use `ast_sendtext()`.  Otherwise the payload `frame.data.ptr` is populated by the string passed in as an argument.   
The thing with `ast_sendtext()`; there is a ton more code to go through BEFORE it gets to the `ast_write()` (which it eventually does).   The current way we are generating text frames is lighter weight (good or bad).

My thought here was to reduce the hoops a text message jumps through to get out the door (that appear to have no "good" reason to jump through).

There is also a bit of discussion here: https://github.com/AllStarLink/app_rpt/issues/536

IF we really need a queue, we probably should use `ast_queue_frame()` in place of `rpt_qwrite()` or `ast_write()` rather than roll our own...  BUT this wouldn't put the `ast_sendtext()` on the queue, they essentially to an `ast_write()`

PR #583, opened 2025-03-23
simpleusb, usbradio: skip categories with no "devstr"

Allan-N:
While testing some new menu / "customization" changes I found that the initial "active" device configuration was set to a category that did not have a "devstr" variable setting.  In this case, the category was intended to augment a templated configuration.  To ensure that these categories are not being considered we now look for the presence of a "devstr" variable.

### DIFF
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -3692,12 +3692,12 @@ static void mixer_write(struct chan_simpleusb_pvt *o)
 
 /*!
  * \brief Store configuration.
- *	Initializes chan_usbradio and loads it with the configuration data.
+ *	Initializes chan_simpleusb and loads it with the configuration data.
  * \param cfg			ast_config structure.
  * \param ctg			Category.
- * \return				chan_usbradio_pvt.
+ * \return				chan_simpleusb_pvt.
  */
-static struct chan_simpleusb_pvt *store_config(const struct ast_config *cfg, const char *ctg)
+static struct chan_simpleusb_pvt *store_config(struct ast_config *cfg, const char *ctg)
 {
 	const struct ast_variable *v;
 	struct chan_simpleusb_pvt *o;
@@ -3712,6 +3712,9 @@ static struct chan_simpleusb_pvt *store_config(const struct ast_config *cfg, con
 		if (strcmp(ctg, "general") == 0) {
 			o = &simpleusb_default;
 		} else {
+			if (!ast_variable_retrieve(cfg, ctg, "devstr")) {
+				return NULL;
+			}
 			if (!(o = ast_calloc(1, sizeof(*o)))) {
 				return NULL;
 			}
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -4879,7 +4879,7 @@ static int xpmr_config(struct chan_usbradio_pvt *o)
  * \param ctg			Category.
  * \return				chan_usbradio_pvt.
  */
-static struct chan_usbradio_pvt *store_config(const struct ast_config *cfg, const char *ctg)
+static struct chan_usbradio_pvt *store_config(struct ast_config *cfg, const char *ctg)
 {
 	const struct ast_variable *v;
 	struct chan_usbradio_pvt *o;
@@ -4894,6 +4894,9 @@ static struct chan_usbradio_pvt *store_config(const struct ast_config *cfg, cons
 		if (strcmp(ctg, "general") == 0) {
 			o = &usbradio_default;
 		} else {
+			if (!ast_variable_retrieve(cfg, ctg, "devstr")) {
+				return NULL;
+			}
 			if (!(o = ast_calloc(1, sizeof(*o)))) {
 				return NULL;
 			}


InterLinked1:
What config actually prompted this change? Template sections should be ignored by the config walker APIs.

Allan-N:
>What config actually prompted this change? Template sections should be ignored by the config walker APIs.

The "customization" configurations being added with https://github.com/AllStarLink/asl3-menu/pull/79

InterLinked1:
> > What config actually prompted this change? Template sections should be ignored by the config walker APIs.
> 
> The "customization" configurations being added with [AllStarLink/asl3-menu#79](https://github.com/AllStarLink/asl3-menu/pull/79)

That's a lot to sift through... could you provide a minimally reproducible example of a `simpleusb.conf` or `usbradio.conf` that has an issue?

Allan-N:
> That's a lot to sift through... could you provide a minimally reproducible example of a `simpleusb.conf` or `usbradio.conf` that has an issue?

From the asl3-menu PR, you just need to drop in the config files in .../asl3-menu/etc/asterisk/custom.  The non-"devstr" categories are pulled in by the [new] #tryinclude statements and show up with `susb tune menu-support 1` .. and `susb active` returns one of the non-"devstr" category names.

InterLinked1:
> > That's a lot to sift through... could you provide a minimally reproducible example of a `simpleusb.conf` or `usbradio.conf` that has an issue?
> 
> From the asl3-menu PR, you just need to drop in the config files in .../asl3-menu/etc/asterisk/custom. The non-"devstr" categories are pulled in by the [new] #tryinclude statements and show up with `susb tune menu-support 1` and `susb active` returns one of the non-"devstr" category names.

Okay, so is this an example? https://github.com/AllStarLink/asl3-menu/pull/79/files#diff-ab6704e58b4ed567524d97afbe64856963bc4b5a2bead47fecd2ccb4c4d04d1c

Is that supposed to be a template or that supposed to instantiate something?

I guess I'm wondering if there are a valid configurations for stuff that wouldn't have a `devstr` option, since you said "the category was intended to augment a templated configuration". Why is this change needed now but wasn't before?

Allan-N:
>Okay, so is this an example? https://github.com/AllStarLink/asl3-menu/pull/79/files#diff-ab6704e58b4ed567524d97afbe64856963bc4b5a2bead47fecd2ccb4c4d04d1c

Yes.

> Is that supposed to be a template or that supposed to instantiate something?

Yes.  The customization menu would change the node configuration from:
```
[12345](node-main)
```
to :
```
[12345](node-main,allscan-uci-120)
```
so we get all of the settings from [node-main], then the settings from [allscan-uci-120], and lastly any settings from [12345].

Note: category templates do not need `(!)` ... and know that I started down that path but ran into issues with the "config" APIs.

> I guess I'm wondering if there are a valid configurations for stuff that wouldn't have a devstr option, since you said "the category was intended to augment a templated configuration". Why is this change needed now but wasn't before?

For simpleusb and usbradio we need a "devstr" variable to interact with the USB device (it's value can be "" where we'll auto-detach but it needs to be present)

Allan-N:
For more info on the "customization" have a look at https://github.com/AllStarLink/asl3-menu/pull/79/files#diff-9505a482027014414ad42a1d5790c98299b33c64e0f9028fb9f358c7987a0a3b

InterLinked1:
> > Okay, so is this an example? https://github.com/AllStarLink/asl3-menu/pull/79/files#diff-ab6704e58b4ed567524d97afbe64856963bc4b5a2bead47fecd2ccb4c4d04d1c
> 
> Yes.
> 
> > Is that supposed to be a template or that supposed to instantiate something?
> 
> Yes. The customization menu would change the node configuration from:
> 
> ```
> [12345](node-main)
> ```
> 
> to :
> 
> ```
> [12345](node-main,allscan-uci-120)
> ```
> 
> so we get all of the settings from [node-main], then the settings from [allscan-uci-120], and lastly any settings from [12345].
> 
> Note: category templates do not need `(!)` ... and know that I started down that path but ran into issues with the "config" APIs.

Interesting, where did you find that? I was wondering why that was absent.

I think `(!)` should indicate the config section isn't really a "real" section and is only used for inheritance.
I suppose you could inherit from a regular section as well in which case you wouldn't need that.

But is this supposed to be a standalone object? What is the reason for not using `(!)`? Would that fix this?

> > I guess I'm wondering if there are a valid configurations for stuff that wouldn't have a devstr option, since you said "the category was intended to augment a templated configuration". Why is this change needed now but wasn't before?
> 
> For simpleusb and usbradio we need a "devstr" variable to interact with the USB device (it's value can be "" where we'll auto-detach but it needs to be present)

I realize that, but is it valid to have a config object without a devstr? What other objects are possible?

For example, in most Asterisk configs, each section has a "type". Most of the radio modules seem to lack this.

The config parser knows what to parse it as, based on the type. In simpler config files with only one type, "type" might be omitted since it's not really needed.

Is there more than one "type" of config object here?

Allan-N:
>Interesting, where did you find that? I was wondering why that was absent.

https://docs.asterisk.org/Fundamentals/Asterisk-Configuration/Asterisk-Configuration-Files/Templates/Using-Templates/?h=template

_This works even when the section name referenced in parentheses is not defined as a template as in the case of the "not-a-template" section._

>I think (!) should indicate the config section isn't really a "real" section and is only used for inheritance.
I suppose you could inherit from a regular section as well in which case you wouldn't need that.
>
>But is this supposed to be a standalone object? What is the reason for not using (!)? Would that fix this?

The short answer is that one needs to switch from `ast_category_browse()` to `ast_category_browse_filtered()` and from `ast_variable_browse()` to `ast_variable_browse_filtered()`.  The gotcha is that this last function, while prototyped in `include/asterisk/config.h` is NOT implemented.

InterLinked1:
> > Interesting, where did you find that? I was wondering why that was absent.
> 
> https://docs.asterisk.org/Fundamentals/Asterisk-Configuration/Asterisk-Configuration-Files/Templates/Using-Templates/?h=template
> 
> _This works even when the section name referenced in parentheses is not defined as a template as in the case of the "not-a-template" section._

Yes, that's technically valid, but there's generally no good reason to NOT define such sections as templates. What is the reason for not doing so here? That seems to be the cause of the issue. If it's not a valid config object on its own, then it should use the `(!)`. I can't think of very many cases where it would be useful to not have that.

> > I think (!) should indicate the config section isn't really a "real" section and is only used for inheritance.
> > I suppose you could inherit from a regular section as well in which case you wouldn't need that.
> > But is this supposed to be a standalone object? What is the reason for not using (!)? Would that fix this?
> 
> The short answer is that one needs to switch from `ast_category_browse()` to `ast_category_browse_filtered()` and from `ast_variable_browse()` to `ast_variable_browse_filtered()`. The gotcha is that this last function, while prototyped in `include/asterisk/config.h` is NOT implemented.

Hmm, I never noticed that, very interesting!

But given that any template with `(!)` doesn't get included in a category traversal anyways, I don't know that really creates an issue for us here.

Allan-N:
> realize that, but is it valid to have a config object without a devstr? What other objects are possible?
>
>For example, in most Asterisk configs, each section has a "type". Most of the radio modules seem to lack this.

For SimpleUSB/USBRadio I think we only have one type of object (for each).  But, we can have more than one instance with the differences being what USB device we are associated with and the per-device configuration options.

>The config parser knows what to parse it as, based on the type. In simpler config files with only one type, "type" might be omitted since it's not really needed.
>
>Is there more than one "type" of config object here?

Here, a node in `rpt.conf` specifies it's `rxchannel` (e.g. SimpleUSB/12345) so the channel driver knows what to look for / instantiate.

The wrinkle I ran into, and what's driving this change, is that you can interact with the channel driver using `susb ...` or `radio ...`.  That interaction goes to the driver without specifying the device instance.  You can change what instance you want to interact with but, before doing so, you get whatever the driver selects as it's default ... and it was choosing one of the non-"devstr" categories.  So, I'm taking the non-"devstr" categories out of the list of possibilities.

InterLinked1:
> > realize that, but is it valid to have a config object without a devstr? What other objects are possible?
> > For example, in most Asterisk configs, each section has a "type". Most of the radio modules seem to lack this.
> 
> For SimpleUSB/USBRadio I think we only have one type of object (for each). But, we can have more than one instance with the differences being what USB device we are associated with and the per-device configuration options.

Okay, that answers my question. In that case, I would argue that the configs you have in the menu repo are invalid. Those templates should have a `(!)` if they are not valid config objects on their own.

> > The config parser knows what to parse it as, based on the type. In simpler config files with only one type, "type" might be omitted since it's not really needed.
> > Is there more than one "type" of config object here?
> 
> Here, a node in `rpt.conf` specifies it's `rxchannel` (e.g. SimpleUSB/12345) so the channel driver knows what to look for / instantiate.
> 
> The wrinkle I ran into, and what's driving this change, is that you can interact with the channel driver using `susb ...` or `radio ...`. That interaction goes to the driver without specifying the device instance. You can change what instance you want to interact with but, before doing so, you get whatever the driver selects as it's default ... and it was choosing one of the non-"devstr" categories. So, I'm taking the non-"devstr" categories out of the list of possibilities.

Have you confirmed this still happens when using `(!)` on those other sections?

Allan-N:
>But given that any template with (!) doesn't get included in a category traversal anyways, I don't know that really creates an issue for us here.

I hear ya ... and I started going down that path ... but I got tripped up on a one variant.  In the per-node configuration you specify the category to use for functions, events, etc.  Right now, you would have :
```
[functions]
...

[12345](node-main)
functions = functions
...

[67890](node-main)
functions = functions
...
```
If I want to have an enhanced version of of "functions" I would like to have :
 ```
[functions](!)
...

[functions-12345](functions)
...my extras...

[12345](node-main)
functions = functions-12345   <--- [functions] + extras
...

[67890](node-main)
functions = functions         <--- just [functions], no extras
...
```

Here, the first case [12345] is fine but the second case [67890] fails because it can't find [functions] because that's now just a template ... and I can't browse for it because I don't have the config API.

**Comment by Allan-N** [comment]
Hmmm... I supposed we could move to :
```
[functions-main](!)
...our current content...

[functions](functions-main)

[functions-12345](functions-main)
...my extras...

[12345](node-main)
functions = functions-12345   <--- [functions-main] + extras
...

[67890](node-main)
functions = functions         <--- [functions-main], no extras
...
```

and this would work ... but it's more change to the base .conf files

mkmer:
Shouldn't the main template have the "standard" mappings for theses???
Such that `(node-main)` has `functions = functions`, then only change `functions=different` in [12345]?

InterLinked1:
> [functions-12345](functions)

Is this for a simpleusb/usbradio config? (It looks more like rpt.conf stuff)

Allan-N:
>Shouldn't the main template have the "standard" mappings for theses???
Such that (node-main) has functions = functions, then only change functions=different in [12345]?

Yes, that's the idea.  As I wrote, I ran into an issue when changing rpt.conf's `[functions]` to `[functions](!)`.  But, also as I just wrote, I can work around that issue by changing `[functions]` to `[functions-main]` and adding a new (empty) `[functions](functions-main)`.

>Is this for a simpleusb/usbradio config? (It looks more like rpt.conf stuff)

This PR was to address simpleusb/usbradio's assigning it's default "active" device to be one of the non-device/non-template categories.  If I change all of the "custom" non-template categories to be templates than this issue goes away.  Also, I went with the non-template categories here to have some parity with the model I was using for rpt.conf.  If I switch the model used in rpt.conf then I can do the same here.

Moving this PR to "Draft" while I explore if I can make the asl3-menu PR work nicely.

InterLinked1:
> > Shouldn't the main template have the "standard" mappings for theses???
> > Such that (node-main) has functions = functions, then only change functions=different in [12345]?
> 
> Yes, that's the idea. As I wrote, I ran into an issue when changing rpt.conf's `[functions]` to `[functions](!)`. But, also as I just wrote, I can work around that issue by changing `[functions]` to `[functions-main]` and adding a new (empty) `[functions](functions-main)`.
> 
> > Is this for a simpleusb/usbradio config? (It looks more like rpt.conf stuff)
> 
> This PR was to address simpleusb/usbradio's assigning it's default "active" device to be one of the non-device/non-template categories. If I change all of the "custom" non-template categories to be templates than this issue goes away. Also, I went with the non-template categories here to have some parity with the model I was using for rpt.conf. If I switch the model used in rpt.conf then I can do the same here.
> 
> Moving this PR to "Draft" while I explore if I can make the asl3-menu PR work nicely.

I just wanted to understand in the context of the original issue. Is there a "function" object in simpleusb.conf? If not, could we use an example using simple.conf to stay on-topic, instead of rpt.conf to avoid conflating things?

> Also, I went with the non-template categories here to have some parity with the model I was using for rpt.conf. If I switch the model used in rpt.conf then I can do the same here.

I'm not entirely sure what you mean here, but each config file can be set up entirely differently, and there's no expectation that the config files would work the same way. For example, I believe rpt.conf involves multiple types of config objects but it sounds like simpleusb/usbradio.conf only have one.

Based on my current understanding, it sounds like the problem is on the config side and there's no problem with the current config parsing, so I agree we should hold off on this change for now.

Allan-N:
>I just wanted to understand in the context of the original issue. Is there a "function" object in simpleusb.conf? If not, could we use an example using simple.conf to stay on-topic, instead of rpt.conf to avoid conflating things?

No, simpleusb.conf does not have any variables that work like rpt.conf's `functions = <functions-category-name>`.  I was just trying to use the same template convention across all of the configs ('cause explaining how this all works once in our documentation is must easier than explaining that we use X in one file and Y in another).

Allan-N:
and the conversation we've been having here has a positive.

InterLinked1:
> > I just wanted to understand in the context of the original issue. Is there a "function" object in simpleusb.conf? If not, could we use an example using simple.conf to stay on-topic, instead of rpt.conf to avoid conflating things?
> 
> No, simpleusb.conf does not have any variables that work like rpt.conf's `functions = <functions-category-name>`. I was just trying to use the same template convention across all of the configs ('cause explaining how this all works once in our documentation is must easier than explaining that we use X in one file and Y in another).

Yes, that's fine for the documentation, I was just confused why we were bringing up rpt.conf when talking about the USB confs :)

Circling back to the original example here: [allscan-uci-120]

https://github.com/AllStarLink/asl3-menu/pull/79/files#diff-ab6704e58b4ed567524d97afbe64856963bc4b5a2bead47fecd2ccb4c4d04d1c

It seems like that is not intended to be instantiated as an actual config object, and is just a template. I don't see why we can't add `(!)` in that example.

Allan-N:
>Yes, that's fine for the documentation, I was just confused why we were bringing up rpt.conf when talking about the USB confs :)

'cause I ran into issues with `[functions](!)` and `functions = functions` ... so I switched away from adding `(!)`

>It seems like that is not intended to be instantiated as an actual config object, and is just a template. I don't see why we can't add (!) in that example.

Yes, and I'm going back the the asl3-menu PR to check that I can make the menu-driven customizations work everywhere after changing `[functions]` to `[functions-main]` and adding `[functions](functions-main)`.

Stay tuned.  If this PR is closed then I was successful.

InterLinked1:
> > Yes, that's fine for the documentation, I was just confused why we were bringing up rpt.conf when talking about the USB confs :)
> 
> 'cause I ran into issues with `[functions](!)` and `functions = functions` ... so I switched away from adding `(!)`

I'm still confused on what those issues were though.
Could you provide the simpleusb.conf (not rpt.conf) that caused issues and what the issues were?

> > It seems like that is not intended to be instantiated as an actual config object, and is just a template. I don't see why we can't add (!) in that example.
> 
> Yes, and I'm going back the the asl3-menu PR to check that I can make the menu-driven customizations work everywhere after changing `[functions]` to `[functions-main]` and adding `[functions](functions-main)`.
> 
> Stay tuned. If this PR is closed then I was successful.

Sounds good - keep us posted! I'm curious what we'll end up uncovering here.

Allan-N:
>My question was what is the undesirable behavior when doing [not-used](!). Of course without (!), it will be treated as a standard config object. I would argue this config is invalid.

The [1999] instance works just fine.  It's just that the node initially selected to be "active" is one that can't/won't ever become active.

Here's another way to explain the conflict.  Let's focus on the `susb` CLI commands.  What object should they be interacting with?  Right now, the commands only talk to the "active" object.  Yes, you can change/select the object you want to use but, by default, one is chosen for you ... and in this case the object chosen to be "active" is not viable.

Shouldn't the "active" object always be one that is valid?  Should we even allow you to select a object that is not viable?

Allan-N:
>Because it's not a templated, it's treated like a normal object.

Even thought the object was not marked as a template that doesn't mean that it could not be used as a template ... and that's what I \*\*\*was*** doing.

I've updated the menu changes to only use customization categories that are proper template objects ... so my issue has been resolved.

But, if anyone adds (or #include's) a non-templated object that doesn't have a "devstr" then the "active" object may not be one of the viable objects.  This is not a crisis, just an annoyance to those [AMI] applications that use `susb` to query information without first selecting what object/node they wish to interact with (and assuming that the default "active" node will provide reasonable information)

InterLinked1:
> > Because it's not a templated, it's treated like a normal object.
> 
> Even thought the object was not marked as a template that doesn't mean that it could not be used as a template ... and that's what I ***was*** doing.

Yes, we've established that - but if it isn't marked as a template, then it is expected to be a valid configuration object on its own, which in your example it is not.

You keep saying "it doesn't need to be a template", which is true in general but doesn't really apply here from what I can see. You also haven't explained why it *can't* be a template.

If it can be a template, I think it is generally wrong for it not to be. The ability to be used for inheritance without being a template exists for the use case of sections that can be used for both inheritance and are valid config objects.

Since this isn't a valid config object standalone, it should be a template. This is not optional here.

> I've updated the menu changes to only use customization categories that are proper template objects ... so my issue has been resolved.
> 
> But, if anyone adds (or #include's) a non-templated object that doesn't have a "devstr" then the "active" object may not be one of the viable objects. This is not a crisis, just an annoyance to those [AMI] applications that use `susb` to query information without first selecting what object/node they wish to interact with (and assuming that the default "active" node will provide reasonable information)

Based on our discussion, I still would argue such sections are not valid. Those should be marked as templates and if not, that is a bad configuration.

PR #584, opened 2025-03-24
Update extensions.conf to add missing toll-free NPAs

ve7fet:
[check_route] is missing the complete list of toll-free NPAs.

Current and reserved (future) NPAs for toll-free service include: 844, 833, 822, 880, 887, and 889.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -5272,8 +5272,8 @@ static void *rpt(void *this)
 		if (c && !myrpt->macrotimer && starttime && t > starttime) {
 			char cin = c & 0x7f;
 			myrpt->macrotimer = MACROTIME;
-			ast_copy_string(str, str + 1, len - 1);
-			ast_str_truncate(myrpt->macrobuf, -1);
+			ast_copy_string(str, str + 1, len);
+			ast_str_truncate(myrpt->macrobuf, len - 1);
 			if ((cin == 'p') || (cin == 'P'))
 				myrpt->macrotimer = MACROPTIME;
 			rpt_mutex_unlock(&myrpt->lock);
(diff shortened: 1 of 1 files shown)


mkmer:
https://en.wikipedia.org/wiki/Toll-free_telephone_numbers_in_the_North_American_Numbering_Plan

Should we add 881, 882, 883, 884, 885, 886 as well? Or A generic version and remove all 88x:
`exten => _188NNXXXXXX,2,Goto(invalidnum,s,1)`

ve7fet:
Indeed, probably a generic dialplan to block 880-889 would be more efficient.

So, I think that would be:
`exten => _188XNXXXXXX,2,Goto(invalidnum,s,1)`

Since, X matches 0-9 (whereas N matches 2-9, so 880 and 881 would be excluded).

Not sure if there is a more graceful way to match 800-888 with wildcards or pattern matching... I'm no dialplan expert, and it is probably just as easy to leave the existing rules (adding the missing ones), for the sake of clarity.

mkmer:
Certainly makes sense -> dialplans are not my expertise either.  Your proposal looks right.

InterLinked1:
> Not sure if there is a more graceful way to match 800-888 with wildcards or pattern matching... I'm no dialplan expert, and it is probably just as easy to leave the existing rules (adding the missing ones), for the sake of clarity.

Unfortunately not, there is no good way to pattern match toll-free NPAs with the Asterisk pattern matcher aside from enumeration. In FreeSWITCH, a regular expression is allowed so a single expression can capture 8 + 2 repeating digits 3-8 to get them all.

mkmer:
880 -> 889 are all toll free.  889 is reserved for future according to the WIKI, This: 
`exten => _188XNXXXXXX,2,Goto(invalidnum,s,1)`
Should cover them correctly (I think anyway)

PR #588, opened 2025-03-25
app_rpt: Remove iaxkey and associated logic

mkmer:
Turns out there is no code to send the message !IAXKEY, so there is no reason to look for said key to arrive.


InterLinked1 (commented):
Minor nitpick - "iaxkeyand" in the commit message title

Allan-N (approved):

InterLinked1 (approved):

PR #590, opened 2025-03-25
app_rpt: do_link_post snprintf using wrong size

### LINKED ISSUE #589: statpost truncated node string
Statpost is presenting a node string with "T " and no node value.

mkmer:
Updates to snprintf used the sizeof() on a malloc'd string resulting in truncated messages. 
Resolves #589

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3284,27 +3284,20 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
  */
 static inline int do_link_post(struct rpt *myrpt)
 {
-	int nstr, str_size;
-	char lst, *str;
+	int nstr;
+	char lst;
+	struct ast_str *str;
 	time_t now;
 	struct rpt_link *l;
 
 	myrpt->linkposttimer = LINKPOSTTIME;
-	str_size = 0;
-	for (l = myrpt->links.next; l != &myrpt->links; l = l->next) {
-		/* if is not a real link, ignore it */
-		if (l->name[0] == '0') {
-			continue;
-		}
-		str_size += strlen(l->name) + 1;
-	}
-	str_size += 256;
-	str = ast_malloc(str_size);
+
+	str = ast_str_create(RPT_AST_STR_INIT_SIZE);
 	if (!str) {
 		return -1;
 	}
 	nstr = 0;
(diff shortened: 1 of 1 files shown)


Allan-N (approved):

apps/app_rpt.c
```diff
@@ -3284,21 +3284,22 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
...
 	for (l = myrpt->links.next; l != &myrpt->links; l = l->next) {
 		/* if is not a real link, ignore it */
 		if (l->name[0] == '0') {
 			continue;
 		}
-		nstr += strlen(l->name) + 1;
+		str_size += strlen(l->name) + 1;
 	}
-	str = ast_malloc(nstr + 256);
+	str_size += 256;
```
InterLinked1, line 3301: I know this was here before, but what's the significance of 256?

apps/app_rpt.c
```diff
@@ -3318,14 +3319,14 @@ static inline int do_link_post(struct rpt *myrpt)
 			lst = 'C';
 		if (nstr)
 			strcat(str, ",");
-		snprintf(str + strlen(str), sizeof(str) - strlen(str), "%c%s", lst, l->name);
+		snprintf(str + strlen(str), str_size - strlen(str), "%c%s", lst, l->name);
 		nstr = 1;
 	}
-	snprintf(str + strlen(str), sizeof(str) - strlen(str), "&apprptvers=%d.%d.%d", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
+	snprintf(str + strlen(str), str_size - strlen(str), "&apprptvers=%d.%d.%d", VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH);
```
InterLinked1, line 3325: Why do we even have multiple snprintf calls to build the URL? A single one would be much simpler and more efficient. No need to keep manipulating buffer sizes, etc.

apps/app_rpt.c
mkmer: No Idea, just went with it. I probably should go with ast_str here and let it float.

Allan-N (approved):

InterLinked1 (approved):

PR #593, opened 2025-03-26
chan_simpleusb: add 1 null frame when USB buffer is emtpy allowing for lowlatency=1

mkmer:
By adding 1 null frame (typically at the beginning of a transmission), chan_simpleusb can then keep up with the usb device. Generally, the channel update is > 20ms (20.1ms -> 20.5ms) which allowed the usb buffer to empty. Relocation of the process transmit queue logic because we need to write frames even if we don't receive one. This eliminates the need to set snd_usb_audio lowlatency=0

### DIFF
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -2061,52 +2061,23 @@ static struct ast_frame *simpleusb_read(struct ast_channel *c)
 		ast_mutex_unlock(&o->echolock);
 	}
 
-	/* Read audio data from the USB sound device.
-	 * Sound data will arrive at 48000 samples per second
-	 * in stereo format.
-	 */
-	res = read(o->sounddev, o->simpleusb_read_buf + o->readpos, sizeof(o->simpleusb_read_buf) - o->readpos);
-	if (res < 0) {				/* Audio data not ready, return a NULL frame */
-		if (errno != EAGAIN) {
-			o->readerrs = 0;
-			o->hasusb = 0;
-			return &ast_null_frame;
-		}
-		if (o->readerrs++ > READERR_THRESHOLD) {
-			ast_log(LOG_ERROR, "Stuck USB read channel [%s], un-sticking it!\n", o->name);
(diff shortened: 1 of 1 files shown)


channels/chan_simpleusb.c
```diff
@@ -2061,52 +2061,23 @@ static struct ast_frame *simpleusb_read(struct ast_channel *c)
...
 	for (;;) {
 		num_frames = 0;
 		ast_mutex_lock(&o->txqlock);
 		AST_LIST_TRAVERSE(&o->txq, f1, frame_list) num_frames++;
 		ast_mutex_unlock(&o->txqlock);
-		
 		i = used_blocks(o);
-		if (num_frames && ((num_frames > 3) || ((!o->txkeyed) && (!o->txtestkey))) && (i <= o->queuesize)) {
+		if (o->txkeyed) {
+			ast_debug(7, "blocks used %d, Dest Buffer %d", i, o->simpleusb_write_dst);
```
InterLinked1, line 2073: Is this debug message really necessary/useful?

mkmer: If we get back into this situation where we need to see how the buffer is performing.
It give us the buffer level and timestamp.

mkmer: Once we get through reviews on simpleusb, I'll fix up usbradio.

InterLinked1: > If we get back into this situation where we need to see how the buffer is performing.

Has this actually been useful? Or is this purely hypothetical?

My concern stems from the fact it seems it could be fairly noisy, perhaps always creating log messages whenever keyed up? Or is this only in certain circumstances?

InterLinked1 (changes requested):
The commit message title has a typo in it (empty)

channels/chan_simpleusb.c
mkmer: I used it to figure out the problem in the first place...

Allan-N (approved):

PR #595, opened 2025-03-27
app_rpt: incorrect evaluation pointer totx

### LINKED ISSUE #594: audioarchive log file always reports "TXKEY" on links
On UNKEY, the log file reports:

```
20250327083352,RXKEY,MAIN
20250327083352,TXKEY,29873
20250327083352,TXKEY,572920
20250327083353,RXUNKEY,MAIN
20250327083353,TXKEY,29873
20250327083353,TXKEY,572920
```

The second set should be `TXUNKEY, xxxx`

mkmer:
Node log should be looking at the value of totx not pointer value.
Resolves #594

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4159,7 +4159,7 @@ static inline void rxkey_helper(struct rpt *myrpt, struct rpt_link *l)
 }
 
 /*! \retval -1 to exit and terminate the node, 0 to continue */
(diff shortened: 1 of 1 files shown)


Allan-N (approved):
That's right.  In most places we have `int totx` but here we have `int *totx`.   Not essential but changing the latter from `totx` to `totx_p` would help highlight the difference.

Also, GOOD CATCH!

mkmer:
Yes, great call out.  Let's go with `totx_p` so it's more obvious.

PR #596, opened 2025-03-28
app_rpt: If we are unkeyed, send an AST_FRAME_CNG.

mkmer:
AST_FRAME_CNG stops the jitter buffer and codec from interpolating missing frames.
app_rpt intentionally stops sending voice frames to indicate "unkey" to the receiving side.
Specifically when a loud noise (like a squelch burst) occurs in the last frame, the jitter buffer and the codec try to interpolate the missing frames for up to 10 frames.  Sending an AST_FRAME_CNG frame triggers the interpolater not interpolate

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4215,22 +4215,18 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 						ast_indicate(l->chan, AST_CONTROL_RADIO_KEY);
 				} else {
 					ast_indicate(l->chan, AST_CONTROL_RADIO_UNKEY);
+					if (l->lastframe) {
+						memset(&wf, 0, sizeof(wf));
+						wf.frametype = AST_FRAME_CNG;
+						ast_write(l->chan, &wf);
+						l->lastframe = 0;
+					}
 				}
 				if (myrpt->p.archivedir) {
 					donodelog_fmt(myrpt, totx ? "TXKEY,%s" : "TXUNKEY,%s", l->name);
 				}
 			}
 			l->lasttx = *totx;
-
-			if (l->chan && l->lastframe && !l->lasttx) { /* Send a silence frame to clear jitter buffer */
-				ast_write(l->chan, &ast_null_frame);
-				wf.frametype = AST_FRAME_CNG;
-				wf.subclass.integer = 0;
-				wf.delivery.tv_sec = 0;
-				wf.delivery.tv_usec = 0;
-				ast_write(l->chan, &wf);
-				l->lastframe = 0;
-			}
 		}
 
 		rpt_mutex_lock(&myrpt->lock);
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2299,7 +2299,7 @@ static int attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	l->linkmode = 0;
 	l->lastrx1 = 0;
 	l->lastrealrx = 0;
-	l->lastframe = 0;
+	l->last_frame_sent = 0;
 	l->rxlingertimer = RX_LINGER_TIME;
 	l->newkeytimer = NEWKEYTIME;
 	l->link_newkey = RADIO_KEY_NOT_ALLOWED;
@@ -4215,9 +4215,9 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 						ast_indicate(l->chan, AST_CONTROL_RADIO_KEY);
 				} else {
 					ast_indicate(l->chan, AST_CONTROL_RADIO_UNKEY);
-					if (l->lastframe) {
+					if (l->last_frame_sent) {
 						ast_write(l->chan, &wf);
-						l->lastframe = 0;
+						l->last_frame_sent = 0;
 					}
 				}
 				if (myrpt->p.archivedir) {
@@ -4426,7 +4426,7 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 					 * 
 					 */
 					ast_write(l->chan, f);
-					l->lastframe = 1;
+					l->last_frame_sent = 1;
 				}
 			}
 			if (f->frametype == AST_FRAME_CONTROL && f->subclass.integer == AST_CONTROL_HANGUP) {
@@ -6560,7 +6560,7 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 		ast_copy_string(l->name, b1, MAXNODESTR);
 		l->isremote = 0;
 		l->chan = chan;
-		l->lastframe = 0;
+		l->last_frame_sent = 0;
 		l->connected = 1;
 		l->thisconnected = 1;
 		l->hasconnected = 1;
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4163,7 +4163,9 @@ static inline void rxkey_helper(struct rpt *myrpt, struct rpt_link *l)
 static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *who, int *restrict totx, char *restrict myfirst)
 {
 	struct rpt_link *l, *m;
-	struct ast_frame wf = { .frametype = AST_FRAME_CNG };
+	struct ast_frame wf = {
+		.frametype = AST_FRAME_CNG
+	};
 
 	/* @@@@@ LOCK @@@@@ */
 	rpt_mutex_lock(&myrpt->lock);
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4163,7 +4163,7 @@ static inline void rxkey_helper(struct rpt *myrpt, struct rpt_link *l)
 static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *who, int *restrict totx, char *restrict myfirst)
 {
 	struct rpt_link *l, *m;
-	struct ast_frame wf;
+	struct ast_frame wf = { .frametype = AST_FRAME_CNG };
 
 	/* @@@@@ LOCK @@@@@ */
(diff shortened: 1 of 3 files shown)


apps/app_rpt.c
```diff
@@ -4162,6 +4163,7 @@ static inline void rxkey_helper(struct rpt *myrpt, struct rpt_link *l)
 static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *who, int *restrict totx, char *restrict myfirst)
 {
 	struct rpt_link *l, *m;
+	struct ast_frame wf;
```
Allan-N, line 4166: If you change this to :
```
            struct ast_frame wf = {
                .frametype = AST_FRAME_CNG,
            };
```
then you don't need the memset() and wf.frametype= below

apps/app_rpt.c
```diff
@@ -2299,6 +2299,7 @@ static int attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	l->linkmode = 0;
 	l->lastrx1 = 0;
 	l->lastrealrx = 0;
+	l->lastframe = 0;
```
Allan-N, line 2302: I'm not wild about the name of this variable.  I think it's supposed to indicate whether we need to send a we-are-done-for-a-while frame.  But, on as-is, I'm reading this as "this is the last frame".  Can we think of a name that is more telling?  Maybe "add_final_frame"?  or "add_completion_frame"?  or possibly "not_muted"

Allan-N: Thank you for the explanation.  Using `last_frame_sent` works for me.

mkmer:
It looks like they added this prediction code ~10 years ago... 
https://github.com/asterisk/asterisk/commit/077adf48b8410740c1b002be353652d41d160aea

mkmer:
I don't think this will "hurt" anything, but if it doesn't "fix" it, we shouldn't put it in.
It definitely changed the signal that is sent through the IAX2 channel (aka encode signal)
Before the change, we have this crazy signal followed by DC offset.
image
image

After the change, no more DC offset and just a single (non oscillating) wave :
image

Allan-N (approved):

apps/app_rpt.c
```diff
@@ -4162,6 +4163,7 @@ static inline void rxkey_helper(struct rpt *myrpt, struct rpt_link *l)
 static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *who, int *restrict totx, char *restrict myfirst)
 {
 	struct rpt_link *l, *m;
+	struct ast_frame wf = { .frametype = AST_FRAME_CNG };
```
InterLinked1, line 4166: The typical struct declaration in Asterisk here would be:

```
struct ast_frame wf = {
<TAB>.frametype = AST_FRAME_CNG
};
```

apps/app_rpt/app_rpt.h
```diff
@@ -389,27 +389,28 @@ struct rpt;
...
+	char phonemode;
+	char phonevox;		   /* vox the phone */
+	char phonemonitor;	   /* no tx or funs for the phone */
+	char name[MAXNODESTR]; /* identifier (routing) string */
+	char lasttx;
+	char lasttx1;
+	char lastrx;
+	char lastrealrx;
+	char lastrx1;
+	char last_frame_sent; /* We have written a single frame */
```
InterLinked1, line 403: Since this is a boolean, use `unsigned int:1`, not `char`

(I suspect most of these chars also suffer from the same problem, but that's another issue)

InterLinked1 (changes requested):

apps/app_rpt.c
mkmer: It's not going to pass clang-format...

apps/app_rpt.c
```diff
@@ -4163,7 +4163,9 @@ static inline void rxkey_helper(struct rpt *myrpt, struct rpt_link *l)
 static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *who, int *restrict totx, char *restrict myfirst)
 {
 	struct rpt_link *l, *m;
-	struct ast_frame wf = { .frametype = AST_FRAME_CNG };
+	struct ast_frame wf = {
+			.frametype = AST_FRAME_CNG 
+		};
```
InterLinked1, line 4168: These 2 lines are indented 1 too far

apps/app_rpt/app_rpt.h
```diff
@@ -400,7 +400,7 @@ struct rpt_link {
 	char lastrx;
 	char lastrealrx;
 	char lastrx1;
-	char last_frame_sent; /* We have written a single frame */
+	unsigned int last_frame_sent; /* We have written a single frame */
```
InterLinked1, line 403: This is missing the :1 at the end
That ensures only 1 bit is allocated as opposed to sizeof(unsigned int)

InterLinked1 (changes requested):

apps/app_rpt.c
mkmer: Not that I really care, but the format you are proposing here for an assignment (NOT a struct detention) Doesn't appear to match ANY standard and clang-format can't be configured to do it this way.

apps/app_rpt/app_rpt.h
mkmer: Why wouldn't we jsut use a `bool` here?

InterLinked1: bool doesn't exist in C, `unsigned int:1` is the standard way of doing it and what Asterisk does

apps/app_rpt.c
InterLinked1: I'd be surprised if it wasn't a standard... you see this all over the Asterisk codebase.

Allan-N:
#include <stdbool.h>

... and asterisk does use this in codecs/codec_dahdi.c and res/res_timing_pthread.c

apps/app_rpt/app_rpt.h
```diff
@@ -389,27 +389,28 @@ struct rpt;
...
+	char phonemode;
+	char phonevox;		   /* vox the phone */
+	char phonemonitor;	   /* no tx or funs for the phone */
+	char name[MAXNODESTR]; /* identifier (routing) string */
+	char lasttx;
+	char lasttx1;
+	char lastrx;
+	char lastrealrx;
+	char lastrx1;
+	unsigned int last_frame_sent : 1; /* We have written a single frame */
```
InterLinked1, line 403: No space before or after the colon

apps/app_rpt.c
```diff
@@ -4162,7 +4163,11 @@ static inline void rxkey_helper(struct rpt *myrpt, struct rpt_link *l)
 static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *who, int *restrict totx, char *restrict myfirst)
 {
 	struct rpt_link *l, *m;
-
+	// clang-format off
```
InterLinked1, line 4166: Is there any way to avoid this?

I didn't have a problem with adding the clang format when it was in a separate file, but it would be nice to avoid markup in the code itself for it, particularly as these comments are not compliant.

Does the trailing comma workaround mentioned here work at all? https://stackoverflow.com/questions/47409797/how-to-wrap-struct-initializer-in-clang-format

InterLinked1 (changes requested):

apps/app_rpt.c
mkmer: Wow - just a trailing comma and it's happy.  I must have spent 2 hours messing with parameters...

apps/app_rpt/app_rpt.h
mkmer: This too was configured with spaces around `:` - updated the `clang-format` file to no spaces.

InterLinked1 (approved):

mkmer:
I thought there was some performance issues with less than a cpu unit size (30 years Ago anyway).  But maybe that's fixed now?

InterLinked1:
> I thought there was some performance issues with less than a cpu unit size (30 years Ago anyway). But maybe that's fixed now?

Well, yes, technically the memory is in chunks of some block size, so the CPU needs to load one into a register and then apply the bitmask to extract the appropriate bit. While technically more work, I'm not sure if that would add cycles or not. It may be slightly slower due to alignment issues but I'm not sure on that.

The other side is semantics. If you look at an `unsigned int:1;`, you know it's a boolean and it makes following the code easier. That's really personally my main perceived benefit, more than any space savings (and right now there is none anyways due to padding, unless other `unsigned int:1`s are added next to it. )If you look at a `char`, you have zero clue what that means. Is it storing a char? A small integer? A single bit only? And it's certainly not done in Asterisk anymore or really current software at all. So I would avoid `char` unless it's actually the most appropriate type to use, just because it doesn't convey any meaning to someone looking at the code.

mkmer:
I believe that is what I'm recalling from 90's "c" class -> plus as PLC programmer, we use `bool` for "everything".  It has to do with comparison testing vs potential memory savings - YES I can squeeze 8 or 16 bits in one memory location, but then the CPU has a bunch of code to mask out the other bits in EVERY test performed.  I would think type `bool` would be (at least for me) more clear that it's a True/False type of comparison.  
Once I "learn" that `int x:1` means `bool x` , then I suppose it doesn't matter :)
And yes, most of the `char ` defines just above this one are boolean tests .

Either way, I prefer the "correct typing" as it allows the compiler to tell us when we are being stupid ;)

tsawyer:
> I don't think this will "hurt" anything, but if it doesn't "fix" it, we shouldn't put it in. It definitely changed the signal that is sent through the IAX2 channel (aka encode signal) Before the change, we have this crazy signal followed by DC offset.
> After the change, no more DC offset and just a single (non oscillating) wave

My HSR Duplex system has a horrible pop at the end of transmissions. I thought it was the nature of the beast (sorry Marshal) but maybe not. I'm hoping this commit will fix the pop (due to the elimination of the DC offset) as well as the twang.

PR #597, opened 2025-03-29
app_rpt: static for functions only called in telemetry

mkmer:
Bit of housekeeping.
static for functions only called in `rpt_telemetry.c`
move `send_tele_link()` to `rpt_telemetry.c` as it's only used in `rpt_telemetry.c`.

### DIFF
--- a/apps/app_rpt/rpt_telemetry.c
+++ b/apps/app_rpt/rpt_telemetry.c
@@ -64,7 +64,13 @@ void rpt_telem_select(struct rpt *myrpt, int command_source, struct rpt_link *my
 	myrpt->telemmode = TELEM_HANG_TIME;
 }
 
-int handle_meter_tele(struct rpt *myrpt, struct ast_channel *mychannel, char *args)
+/*
+ * Parse a METER request for telemetry thread.
+ * This is passed in a comma separated list of items from the function table entry
+ * There should be 3 or 4 fields in the function table entry: device, channel,
+ * meter face, and  optionally: filter
+ */
+static int handle_meter_tele(struct rpt *myrpt, struct ast_channel *mychannel, char *args)
 {
 	int i, res, files, filter, val;
 	int pin = 0;
@@ -386,7 +392,28 @@ void cancel_pfxtone(struct rpt *myrpt)
 	}
 }
 
-int send_tone_telemetry(struct ast_channel *chan, char *tonestring)
+static void send_tele_link(struct rpt *myrpt, char *cmd)
+{
+	char str[1028];
+	struct ast_frame wf;
+	struct rpt_link *l;
+
+	snprintf(str, sizeof(str) - 1, "T %s %s", myrpt->name, cmd);
+	init_text_frame(&wf, "send_tele_link");
+	wf.datalen = strlen(str) + 1;
+	wf.data.ptr = str;
+	l = myrpt->links.next;
+	/* give it to everyone */
+	while (l != &myrpt->links) {
+		if (l->chan && (l->mode == 1))
+			rpt_qwrite(l, &wf);
+		l = l->next;
+	}
+	rpt_telemetry(myrpt, VARCMD, cmd);
+}
+
+/*! Send telemetry tones */
+static int send_tone_telemetry(struct ast_channel *chan, char *tonestring)
 {
 	char *p, *stringp;
 	char *tonesubset;
@@ -434,7 +461,7 @@ int send_tone_telemetry(struct ast_channel *chan, char *tonestring)
 	return res;
 }
 
-int telem_any(struct rpt *myrpt, struct ast_channel *chan, char *entry)
+static int telem_any(struct rpt *myrpt, struct ast_channel *chan, char *entry)
 {
 	int res;
 	char c;
@@ -501,7 +528,10 @@ static struct telem_defaults tele_defs[] = {
 	{ "pfxtone", "|t(350,440,30000,3072)" }
 };
 
-int telem_lookup(struct rpt *myrpt, struct ast_channel *chan, char *node, char *name)
+/*! \brief Looks up a telemetry name in the config file, and does a telemetry response as configured.
+ * 4 types of telemetry are handled: Morse ID, Morse Message, Tone Sequence, and a File containing a recording.
+ */
+static int telem_lookup(struct rpt *myrpt, struct ast_channel *chan, char *node, char *name)
 {
 
 	int res;
@@ -533,7 +563,11 @@ int telem_lookup(struct rpt *myrpt, struct ast_channel *chan, char *node, char *
 	return res;
 }
 
-void handle_varcmd_tele(struct rpt *myrpt, struct ast_channel *mychannel, char *varcmd)
+/*! \brief Processes various telemetry commands that are in the myrpt structure
+ * Used extensively when links are built or torn down and other events are processed
+ * by the rpt_master threads.
+ */
+static void handle_varcmd_tele(struct rpt *myrpt, struct ast_channel *mychannel, char *varcmd)
 {
 	char *strs[100], *p, buf[100], c;
 	int i, j, k, n, res, vmajor, vminor;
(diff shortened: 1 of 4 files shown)


apps/app_rpt/rpt_telemetry.c
```diff
@@ -501,7 +527,9 @@ static struct telem_defaults tele_defs[] = {
 	{ "pfxtone", "|t(350,440,30000,3072)" }
 };
 
-int telem_lookup(struct rpt *myrpt, struct ast_channel *chan, char *node, char *name)
+/*! \brief This function looks up a telemetry name in the config file, and does a telemetry response as configured.
```
Allan-N, line 530: Can we limit the new comment blocks to < 80 characters?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -533,7 +561,9 @@ int telem_lookup(struct rpt *myrpt, struct ast_channel *chan, char *node, char *
 	return res;
 }
 
-void handle_varcmd_tele(struct rpt *myrpt, struct ast_channel *mychannel, char *varcmd)
+/*! \brief Routine to process various telemetry commands that are in the myrpt structure
```
Allan-N, line 564: Can we limit the new comment blocks to < 80 characters?

apps/app_rpt/rpt_telemetry.c
InterLinked1: 80 columns is unreasonably restrictive, and there's no basis for this in the Asterisk coding guidelines

InterLinked1 (commented):
Commit message has typo for filenames - app_telemetry.c instead of rpt_telemetry.c

apps/app_rpt/rpt_telemetry.c
Allan-N: It was a request ... so that ---I--- don't have to unnecessarily widen my editing windows to view changes without line wrapping.

Allan-N: Also, 80 columns is not unreasonable, especially given that the default terminal window size on most systems is 80x24.

InterLinked1: I'm going to disagree with this. A single person's convenience when reviewing a change should not really be a consideration in the code that gets committed, it should be the experience when looking at the actual codebase. Review experience is a temporary inconvenience anyways. And as long as are talking about personal preferences, I would personally prefer longer rows to make better use of vertical space. Not everyone is going to agree.

Opinion aside, I don't think the code should be catering to any individual's personal preferences, and columns in the ~120-150 col range are quite standard in the Asterisk codebase today. The only limits of that nature of which I'm aware of are a hard 80 column limit and soft 72 column limit for commit message messages specifically. I'll support "soft" limits on truly long lines, like beyond that 120-150 range (I don't have a specific number in mind, but something in there seems reasonable), but 80 columns is just not realistic.

That being said, the words "This function" can be removed from the doxygen, they are redundant. That will bring the column count down :)

Allan-N: [Asterisk : Code formatting](https://docs.asterisk.org/Development/Policies-and-Procedures/Coding-Guidelines/?h=code+formatting#code-formatting)

Roughly, Asterisk code formatting guidelines are generally equivalent to the following:
```
# indent -i4 -ts4 -br -brs -cdw -lp -ce -nbfda -npcs -nprs -npsl -nbbo -saf -sai -saw -cs -l90 foo.c
```
this means in verbose:
```
...
-l90: line length 90 columns
```

InterLinked1: > [Asterisk : Code formatting](https://docs.asterisk.org/Development/Policies-and-Procedures/Coding-Guidelines/?h=code+formatting#code-formatting)
> 
> Roughly, Asterisk code formatting guidelines are generally equivalent to the following:
> 
> ```
> # indent -i4 -ts4 -br -brs -cdw -lp -ce -nbfda -npcs -nprs -npsl -nbbo -saf -sai -saw -cs -l90 foo.c
> ```
> 
> this means in verbose:
> 
> ```
> ...
> -l90: line length 90 columns
> ```

Thanks for pointing that out. The edit history unfortunately was lost in the Atlassian wiki -> markdown docs a couple years back now, but I can say that kind of content has not been edited in a very long time now. I don't know of any current guidance on that, or even if there is any. 90 columns may have been a standard 15 or 20 years ago when that was written, but I would argue it's not accurate anymore, or at least reflective of the current codebase.

I wouldn't say a shorter column count is *wrong*, either. I know of one Asterisk contributor that does limit line lengths to the 80-100 range - perhaps he also uses vim :). I'm just not aware of that being done or enforced for anyone else, the typical line lengths I see are more in the 100-140 range. So I certainly don't think a line length less than 100 should be coerced to be shorter than that. For certain types of comments and lines, it also makes sense to try to keep them on a single line if feasible.

mkmer: I thought we had already discussed this.  I even quoted that guidance back then.  We picked 130 for the "soft" limit in clang-format in one of the other PR's (probably 2 months ago?)
We can change the hard and soft limits if they are getting too long, or to whatever you guys prefer,  but the whole point of clang-format is to get these things "forced check" so it's NOT part of a PR discussion :)  

My original statement around the whole format thing is "WHY are we making one up???"  There are 7 predefined styles and for some reason we have to customize the heck out if???
```
LLVM A style complying with the [LLVM coding standards](https://llvm.org/docs/CodingStandards.html)

Google A style complying with [Google’s C++ style guide](https://google.github.io/styleguide/cppguide.html)

Chromium A style complying with [Chromium’s style guide](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/styleguide/styleguide.md)

Mozilla A style complying with [Mozilla’s style guide](https://firefox-source-docs.mozilla.org/code-quality/coding-style/index.html)

WebKit A style complying with [WebKit’s style guide](https://www.webkit.org/coding/coding-style.html)

Microsoft A style complying with [Microsoft’s style guide](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference)

GNU A style complying with the [GNU coding standards](https://www.gnu.org/prep/standards/standards.html)
```

InterLinked1: 130 for a soft limit seems reasonable to me, yes.

I'm not familiar with the details all those styles, so can't speak to that, but I think there are significant deviations from all of them with the Asterisk style. From prior research I know ones like GNU, LLVM, etc. are WAY off. I want to say Asterisk style closely resembles K&R for the most part, with a major difference being the way that switch cases are formatted.

Allan-N: As I recall, we picked 130 so that our format checking wouldn't complain too much about the existing code that clearly has many lines with really long comments.  My last comment here is that "just because you can doesn't mean that you should".  What catches my eye, from a readability point of view, is seeing nearby code with shorter comments and then our newly added code with much longer comments ... and I say "is this really necessary?".  Am I going to block the commit, NO.  Will I ask that you take a look and consider a change, yes, because --I-- think the code would "look" better..

Allan-N (approved):
As I recall, we picked 130 so that our format checking wouldn't complain too much about the existing code that clearly has many lines with really long comments.  My last comment here is that "just because you can doesn't mean that you should".  What catches my eye, from a readability point of view, is seeing nearby code with shorter comments and then our newly added code with much longer comments ... and I say "is this really necessary?".  Am I going to block the commit, NO.  Will I ask that you take a look and consider a change, yes.

mkmer:
One thing I think is interesting here:  Because this is in the same file, the compiler noticed this was too small for some calls:
https://github.com/AllStarLink/app_rpt/blob/041d0843b81ec0c42b84c3133db3b76f977ee4cc/apps/app_rpt/rpt_telemetry.c#L397

I simply increased the size here, but maybe the "right" answer is to shrink it elsewhere... (In another PR if we do)

PR #598, opened 2025-03-30
chan_usbradio: Add 1 null frame when USB buffer is emtpy.

mkmer:
By adding 1 null frame (typically at the beginning of a transmission), chan_usbradio can then keep up with the usb device. Generally, the channel update is > 20ms (20.1ms -> 20.5ms) which allowed the usb buffer to empty.  This eliminates the need to set snd_usb_audio lowlatency=0

### DIFF
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -1522,6 +1522,7 @@ static int used_blocks(struct chan_usbradio_pvt *o)
 static int soundcard_writeframe(struct chan_usbradio_pvt *o, short *data)
 {
 	int res;
+	short outbuf[FRAME_SIZE * 2 * 2 * 6];
 
 	/* If the sound device is not open, setformat will open the device */
 	if (o->sounddev < 0) {
@@ -1552,7 +1553,11 @@ static int soundcard_writeframe(struct chan_usbradio_pvt *o, short *data)
 		}
 		return 0;
 	}
-
+	if (res == 0) { /* We are not keeping the buffer full, add 1 frame */
+		memset(outbuf, 0, sizeof(outbuf));
+		write(o->sounddev, ((void *) outbuf), FRAME_SIZE * 2 * 2 * 6);
+		ast_debug(7, "A null frame has been added");
+	}
 	res = write(o->sounddev, ((void *) data), FRAME_SIZE * 2 * 2 * 6);
 	if (res < 0) {
 		ast_log(LOG_ERROR, "Channel %s: Sound card write error %s\n", o->name, strerror(errno));
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -1522,7 +1522,7 @@ static int used_blocks(struct chan_usbradio_pvt *o)
 static int soundcard_writeframe(struct chan_usbradio_pvt *o, short *data)
 {
 	int res;
-	short outbuf[FRAME_SIZE * 2 * 2 * 6];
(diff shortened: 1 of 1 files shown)


InterLinked1 (commented):
Commit message says "simpleusb" - should that say usbradio?

channels/chan_usbradio.c
```diff
@@ -1522,6 +1522,7 @@ static int used_blocks(struct chan_usbradio_pvt *o)
 static int soundcard_writeframe(struct chan_usbradio_pvt *o, short *data)
 {
 	int res;
+	short outbuf[FRAME_SIZE * 2 * 2 * 6];
```
Allan-N, line 1525: I believe that the buffer is supposed to be sized to hold "FRAME_SIZE * 2 * 2 * 6 (2 bytes per sample, 2 channels, 6 for upsample to 48K)".  In this case, you have an extra "* 2" as a "short" holds 2 bytes.

... and, for comparison, you can look at chan_simpleusb where you will see `short outbuf[FRAME_SIZE * 2 * 6];`

channels/chan_usbradio.c
```diff
@@ -1552,7 +1553,11 @@ static int soundcard_writeframe(struct chan_usbradio_pvt *o, short *data)
 		}
 		return 0;
 	}
-
+	if (res == 0) { /* We are not keeping the buffer full, add 1 frame */
+		memset(outbuf, 0, sizeof(outbuf));
+		write(o->sounddev, ((void *) outbuf), FRAME_SIZE * 2 * 2 * 6);
```
Allan-N, line 1558: This write() ... and the one just below ... could be updated with `FRAME_SIZE * 2 * 2 * 6` --> `sizeof(outbuf)`

Allan-N (changes requested):

channels/chan_usbradio.c
mkmer: So this is the one just below it wrong too (which I copied)?  Write says "n bytes" for size_t.  
Should I have short FRAME_SIZE *2 *6 (already 2x for short), but the write should be sizeof(outbuf)*2 since it's asking for bytes?

mkmer: - I guess not.  sizeof(outbuf) returns the size in bytes not defined units of short.

Allan-N: If you are not going to use sizeof() then the calculation should match the sizing.

When you had `short outbuf[FRAME_SIZE * 2 * 2 * 6]` you would have needed to write `FRAME_SIZE * 2 * 2 * 6` times `2` ... because a "short" is 2 bytes.

Allan-N: >I guess not. sizeof(outbuf) returns the size in bytes not defined units of short.

Yes, this is sizeof(), not ARRAY_LEN()

Allan-N: Oh, wait, the write() below was writing out "data" and not "outbuf" ... so sizeof(outbuf) wouldn't make sense.  This is another one of those cases where the function args should include both "data" and "datalen" ... so one doesn't need to make any assumptions about the size of the passed in argument.

Allan-N (approved):

InterLinked1 (approved):

channels/chan_usbradio.c
mkmer: Riight - it's the second call that threw me off - I copied it without fully understanding what was going on... Guilty as charged.

PR #604, opened 2025-03-31
app_rpt: allow more "rpt cmd <node> ..." CLI commands

Allan-N:
The app_rpt CLI parser had expections that all commands would have at least 4 arguments. This limitation prevented using commands like "rpt cmd 2000 status 12" (with only 3 args).  We now check the number of arguments based on the command being used.

### DIFF
--- a/apps/app_rpt/rpt_cli.c
+++ b/apps/app_rpt/rpt_cli.c
@@ -986,7 +986,8 @@ static int rpt_do_cmd(int fd, int argc, const char *const *argv)
 	struct rpt *myrpt = NULL;
 	int nrpts = rpt_num_rpts();
 
-	if (argc != 6) {
+	if (argc < 4) {
+		/* we need at least "rpt cmd <node> ..." */
 		return RESULT_SHOWUSAGE;
 	}
 
@@ -1004,11 +1005,17 @@ static int rpt_do_cmd(int fd, int argc, const char *const *argv)
 	}							/* if thisRpt < 0 */
 
 	/* Look up the action */
-	thisAction = function_table_index(argv[3]);
+	thisAction = rpt_function_lookup(argv[3]);
 	if (thisAction < 0) {
 		ast_cli(fd, "Unknown action name %s.\n", argv[3]);
 		return RESULT_FAILURE;
-	}							/* if thisAction < 0 */
+	} /* if thisAction < 0 */
+
+	if (argc < (4 + rpt_function_minargs(thisAction))) {
+		/* for this function we need to have (at least)
+		   "rpt cmd <node> <function-name> [required-function-args]" */
+		return RESULT_SHOWUSAGE;
+	}
 
 	/* at this point, it looks like all the arguments make sense... */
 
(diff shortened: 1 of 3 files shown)


apps/app_rpt/rpt_cli.c
```diff
@@ -986,7 +986,7 @@ static int rpt_do_cmd(int fd, int argc, const char *const *argv)
 	struct rpt *myrpt = NULL;
 	int nrpts = rpt_num_rpts();
 
-	if (argc != 6) {
+	if (argc < 4) {
```
mkmer, line 989: Should this magic number have a `#define NAME = 4` for all of the uses?

apps/app_rpt/rpt_cli.c
```diff
@@ -1403,8 +1415,8 @@ static char *handle_cli_lookup(struct ast_cli_entry *e, int cmd, struct ast_cli_
 	switch (cmd) {
 	case CLI_INIT:
 		e->command = "rpt lookup";
-		//e->usage = rpt_usage;
-		e->usage = NULL;		/*! \todo 20220111 NA rpt_usage doesn't exist! */
+		e->usage = "Usage: rpt lookup <nodename>\n"
```
mkmer, line 1418: Why does this command need a node?  When I run it on our hub, I can choose any of the 4 repeaters and it shows me ALL of the repeaters.
```
Hub01*CLI> rpt lookup 287891
Node: 287890     Data: radio@127.0.0.1:4569/287891,NONE
Node: 287891     Data: radio@127.0.0.1:4569/287891,NONE
Node: 287892     Data: radio@127.0.0.1:4569/287891,NONE
Node: 287893     Data: radio@127.0.0.1:4569/287891,NONE
```

apps/app_rpt/rpt_cli.c
```diff
@@ -1017,8 +1021,15 @@ static int rpt_do_cmd(int fd, int argc, const char *const *argv)
...
 		rpt_vars[thisRpt].cmdAction.state = CMD_STATE_BUSY;
 		rpt_vars[thisRpt].cmdAction.functionNumber = thisAction;
-		snprintf(rpt_vars[thisRpt].cmdAction.param, sizeof(rpt_vars[thisRpt].cmdAction.param), "%s,%s", argv[4], argv[5]);
-		ast_copy_string(rpt_vars[thisRpt].cmdAction.digits, argv[5], sizeof(rpt_vars[thisRpt].cmdAction.digits));
+		rpt_vars[thisRpt].cmdAction.param[0] = 0;
+		if (argc > 4) {
+			ast_copy_string(rpt_vars[thisRpt].cmdAction.param, argv[4], sizeof(rpt_vars[thisRpt].cmdAction.param));
+		}
+		rpt_vars[thisRpt].cmdAction.digits[0] = 0;
+		if (argc > 5) {
```
mkmer, line 1029: Maybe (if 4 becomes a `NAME`,`NAME + 1` here?

mkmer (commented):
I like this "fix" - it's really annoying how it requires dummy inputs.
This will break ALLMON v3 commands.  It will be a bit difficult to format these. Maybe ALLMON can look at the version to create the correct command (I'm not sure)
It generates commands like: this
```rpt cmd 287891 status 12 xxx```

File: apps/app_rpt/rpt_cli.c (hunk 2 above)
Comment (Allan-N) [reply apps/app_rpt/rpt_cli.c:1418]: Well, that looks to be a bug in `rpt_do_lookup()`.

Comment (Allan-N) [reply apps/app_rpt/rpt_cli.c:1418]: and it's not actually returning your 4 repeaters.  Instead, your are getting the node # with the same info.

**Comment by Allan-N** [comment]
>This will break ALLMON v3 commands. It will be a bit difficult to format these. Maybe ALLMON can look at the version to create the correct command (I'm not sure)
It generates commands like: this
rpt cmd 287891 status 12 xxx

These changes do not (appear to) break Allmon3.  The changes are updating the minimum number of arguments you need to provide.   I am not capping the maximum number of args (i.e. you can add as many extras as you want ... with the individual commands deciding what to do with any extras).

File: apps/app_rpt/rpt_cli.c (hunk 1 above)
Comment (Allan-N) [reply apps/app_rpt/rpt_cli.c:989]: I started down the path of adding a `#define` but wasn't happy.  So, I've opted for the middle ground and added some comments about how the command is viewed/handled.

FYI: I remain confused with the usage of `.cmdAction.param` and `.cmdAction.digits`.

File: apps/app_rpt/rpt_cli.c (hunk 3 above)
Comment (Allan-N) [reply apps/app_rpt/rpt_cli.c:1029]: I remain confused about the usage of `.cmdAction.param` and `.cmdAction.digits`.

File: apps/app_rpt/rpt_cli.c (hunk 2 above)
Comment (mkmer) [reply apps/app_rpt/rpt_cli.c:1428]: Something is still odd about this report.  Nodes are like this:
```
[nodes]
287890=radio@127.0.0.1:4569/287890,NONE
287891=radio@127.0.0.1:4569/287891,NONE
287892=radio@127.0.0.1:4569/287892,NONE
287893=radio@127.0.0.1:4569/287893,NONE
```

but the lookup returns this:
```
Node: 287890     Data: radio@127.0.0.1:4569/287893,NONE
Node: 287891     Data: radio@127.0.0.1:4569/287893,NONE
Node: 287892     Data: radio@127.0.0.1:4569/287893,NONE
Node: 287893     Data: radio@127.0.0.1:4569/287893,NONE
```
Note the `/287893` - if I look at node 287890 I get `/287890` for all entries.  
I suspect I don't understand what the command is trying to tell me...

Comment (Allan-N) [reply apps/app_rpt/rpt_cli.c:1428]: The output from `rpt lookup <node-to-lookup>` is unrelated to these changes but, since we're hear, I'll try to explain what I think is happening.  First, it looks like you have a system with 4 nodes.  The lookup command is querying the connection info for the requested node **_from the view of each of your 4 nodes_**.  If you only have a single [nodes] category then the results of the lookup should be the same.  But, consider the following hypothetical scenario :
```
[nodes]
12345 = radio@ 10.0.1.100:4568/1001,NONE

[nodes-2001]
12345 = radio@10.0.1.200:4570/1001,NONE

[2000](node-main)
;nodes=nodes         ;<--- this is assumed
...

[2001](node-main)
nodes=nodes-2001     ;<--- use an alternate [nodes] category
...
```
With this configuration a `rpt lookup 12345` would have output like :
```
Node: 2000     Data: radio@ 10.0.1.100:4568/1001,NONE   <--- from [nodes]
Node: 2001     Data: radio@ 10.0.1.200:4568/1001,NONE   <--- from [nodes-2001]
```
Also, if the queried node is not in [nodes] then we pass back the DNS (or rpt_extnodes) result

mkmer (approved):

InterLinked1 (approved):

PR #605, opened 2025-04-01
app_rpt: Add enum names and use them

mkmer:


apps/app_rpt/app_rpt.h
```diff
@@ -231,48 +231,166 @@ typedef struct {
...
-enum{DAQ_CMD_IN, DAQ_CMD_ADC, DAQ_CMD_OUT, DAQ_CMD_PINSET, DAQ_CMD_MONITOR};
-enum{DAQ_SUB_CUR = 0, DAQ_SUB_MIN, DAQ_SUB_MAX, DAQ_SUB_STMIN, DAQ_SUB_STMAX, DAQ_SUB_STAVG};
-enum{DAQ_PT_INADC = 1, DAQ_PT_INP, DAQ_PT_IN, DAQ_PT_OUT};
-enum{DAQ_TYPE_UCHAMELEON};
+enum rpt_daq_mode {
+	DAQ_PS_IDLE = 0,
+	DAQ_PS_START,
+	DAQ_PS_BUSY,
+	DAQ_PS_IN_MONITOR
+};
```
InterLinked1, line 369: There should be a blank line between all these enums here

apps/app_rpt/rpt_functions.h
```diff
@@ -2,37 +2,40 @@
 /* Define function protos for function table here */
 
 /*! \brief Internet linking function */
-int function_ilink(struct rpt *myrpt, char *param, char *digitbuf, int command_source, struct rpt_link *mylink);
+enum rpt_function_response function_ilink(
```
InterLinked1, line 5: I'd kind of prefer it stay on a single line because this "looks weird" to me... I realize that's a bit subjective though and I'm not saying that's the only "right way". Curious what Allan thinks.

Sometimes longer declarations are broken up into multiple lines if needed in Asterisk, but usually the lengths are a bit longer and there are some arguments on the first line. I've at least never seen it like this before.

mkmer: Something is up with `clang-format` here -> It broke this into two lines locally, then failed on the repo.  A bit frustrating as I'm using the exact same config (./.dev/.clang-format) as the container.  It's not being clear on what it doesn't like either :)  
The line is broken up because it's >130 (by quite a bit) now with the enum in front.

InterLinked1: If we want to break it up, I won't object to that, but I'd prefer it be done in a way consistent with the rest of Asterisk... breaking it up as "late" as possible rather than right at the beginning.

Allan-N: My $0.02:
One per line gives us the opportunity to add per-enum comments (if the enums are not sufficiently self-explanatory).
Also, one per line also makes it easier to expand the list.

mkmer: I think we're talking about the function call ->
`enum big_name func(balh, blah, blah) {`

Or
```
enum big_name func(
   blah, blah, blah)
{
```
Where blah blah is > 120 long

mkmer:
For whatever reason - `clang-format` things everything needs an indent.  I don't think that is right.

apps/app_rpt/rpt_functions.h
Allan-N: Oh, if we're talking about function prototypes, calls, ... then, again, my focus is on readability.  If there's one or two args then it's easy to "see" how many args are needed and the ordering.  As the number of arguments increases then the line length follows and it become harder and harder to see how many args you need, the ordering, and what the type of args to be passed.  Moving each arg to its own line helps some ("me") see what's expected.  Also, one per line leaves room for comments and makes it much easier to manage the list should updates be needed.

... and y'all already know how I feel about long lines :)

InterLinked1: We don't do one argument per line, the standard practice is to add a new line when the length gets long, but not prematurely.

So:
```
enum something func(arg1, arg2, ...arg4,
<TAB>arg5, arg6, etc.);
```

I'm not sure what you mean by comments. Any commentary would go in the doxygen for the function, not mixed in the declaration.

Allan-N: The asterisk coding standard rules!  End of discussion.

apps/app_rpt/rpt_functions.c
```diff
@@ -51,7 +51,8 @@ char *dtmf_tones[] = {
 
 static char remdtmfstr[] = "0123456789*#ABCD";
 
-int function_ilink(struct rpt *myrpt, char *param, char *digits, int command_source, struct rpt_link *mylink)
+enum rpt_function_response function_ilink(
```
InterLinked1, line 54: As many arguments as possible still need to be pulled onto the first line here
Same with other instances

channels/chan_simpleusb.c
```diff
@@ -151,9 +151,25 @@ static char stoppulser;
...
-enum { SD_IGNORE, SD_HID, SD_HID_INVERT, SD_PP, SD_PP_INVERT };	// no,external,externalinvert,software
-enum { PAGER_NONE, PAGER_A, PAGER_B };
+enum simpleusb_carrier_detect {
+	CD_IGNORE,
+	CD_HID,
+	CD_HID_INVERT,
+	CD_PP,
+	CD_PP_INVERT
+};
+enum simpleusb_squelch_detect {
```
InterLinked1, line 161: Blank line between enums here

channels/chan_simpleusb.c
```diff
@@ -151,9 +151,25 @@ static char stoppulser;
...
+	SD_IGNORE,
+	SD_HID,
+	SD_HID_INVERT,
+	SD_PP,
+	SD_PP_INVERT
+}; // no,external,externalinvert,software
```
InterLinked1, line 167: I know this comment was here before. Is this referring to what each of the values means? Should it be documentation?

channels/xpmr/xpmr.h
```diff
@@ -225,17 +225,41 @@
...
-enum {RX_AUDIO_NONE,RX_AUDIO_SPEAKER,RX_AUDIO_FLAT};
-enum {TX_AUDIO_NONE,TX_AUDIO_FLAT,TX_AUDIO_FILTERED,TX_AUDIO_PROC};
-enum {CD_IGNORE,CD_XPMR_NOISE,CD_XPMR_VOX,CD_HID,CD_HID_INVERT};
-enum {SD_IGNORE,SD_HID,SD_HID_INVERT,SD_XPMR};    				 // no,external,externalinvert,software
-enum {RX_KEY_CARRIER,RX_KEY_CARRIER_CODE};
-enum {TX_OUT_OFF,TX_OUT_VOICE,TX_OUT_LSD,TX_OUT_COMPOSITE,TX_OUT_AUX};
-enum {TOC_NONE,TOC_PHASE,TOC_NOTONE};
-#endif
+#define BIN_PROG_3 LP_PIN09
+enum usbradio_rx_audio {
```
InterLinked1, line 229: Blank lines between enums

channels/xpmr/xpmr.h
```diff
@@ -700,9 +724,9 @@ typedef struct	t_pmr_chan
 	i16 gainVoice;
 	i16 gainSubAudible;
 
-	i16 txMixA;				// Off, Ctcss, Voice, Composite
-	i16 txMixB;				// Off, Ctcss, Voice, Composite
-	
+	enum usbradio_tx_mix txMixA; // Off, Ctcss, Voice, Composite
```
InterLinked1, line 727: Same here, this comment looks like it should be documentation somewhere?

channels/xpmr/xpmr.h
mkmer, line 229: I'm not sure I like what's going on with this enumeration.  It was all cross pollinated with usbradio, xpmr, and voter.
Should there be a common header or something where this could live, and we could it for all files?  (I'm not an expert on this stuff).

InterLinked1: Maybe res_usbradio.h if it's only for the USB stuff?
If it's also for voter, I don't think there is anything that exists currently.

Allan-N (approved):

include/asterisk/res_radio.h
```diff
@@ -0,0 +1,68 @@
+/*
```
InterLinked1, line 1: There is no `res_radio` module, so I don't think it's appropriate to add a header file called that... at least I've never seen that done before.

At the same time, it's not very elegant to put it in an `app_rpt` related header and include that from the channel drivers. But we're already doing that to some extent so that might be slightly less unelegant.

Also, you can omit Jim Dixon and Steve from the header as well and just leave yourself as copyright.

mkmer: I was thinking I just "moved" the stuff... I'll take credit :)
As for header file, we don't reference the `app_rpt.h` header from the `chan_*` files (at least that I've seen).  At one point I tried to do that and Danny said we shouldn't.
I suppose we could put it in `res_usbradio.h`, voter would then become tied to it.

Allan-N: I don't think `res_usbradio.h` makes sense.  We're looking for a header that is used across multiple components (apps, channels, res, ...) and `.../include/asterisk/*.h` seems like the right place.  Wouldn't this be similar to `strings.h`?

Also, I would agree that `app_rpt.h` is likely not the right header as it defines bits that only apply to `app_rpt.so`.

InterLinked1: I agree `res_usbradio` doesn't make sense.
It's an odd scenario because there is a lot of interaction between the radio modules, which is not really true anywhere else in Asterisk. There are dependency modules that do common tasks, but that's not quite what's going on here either.
Will the new universal radio channel driver fix most of these issues?

Allan-N: I need to revise my thinking here.  We're looking to have a #include file that is shared across our channel drivers (currently chan_simpleusb, chan_usbradio, and chan_voter).  The content is not focused on "usb" like `res_usbradio.[ch]`.  The content is also separate from `app_rpt[.so]`.  To avoid duplication, we want a shared header.  The name `res_radio.h` removes the "usb" but maybe it's not the best name.

Here's some other ideas:
- rpt_channel.h
- rpt_channel_common.h
- app_rpt_channel.h

Also, I suspect the new (planned) universal radio channel driver will also use this header.

InterLinked1: If the new radio channel driver is truly universal, won't we be able to get rid of all the other ones, and then we won't need the header file anymore? Or am I misunderstanding?

If it's only used by channel drivers, then maybe `channels/rpt_channel.h` would be good.

mkmer: The "universal" driver will only be for USB, it will not replace `chan_voter`, `chan_tlb`, or `chan_echolink`.  
In general, if we have some "common" structures, enums or functions in the channel drivers, what is the "right" way to do that?  Maybe a header file that lives in the channels folder?
There are a few other functions in the channel drivers that are common as well that we could benefit from reusing (finddelim, others?)

InterLinked1: IIRC, finddelim has nothing to do with channel driver stuff. Really that should probably be in `include/utils.h` if it's really that important. I suspect some of those "utility functions" are probably at least partially duplicative of existing code

Allan-N: For clarity, are you suggesting that we take the app_rpt functions that are also being duplicated in our channel drivers and :
- change their names to include an `ast_radio_` prefix (so that the functions will be available for linking)
- make sure that the function prototypes are available in a [new] shared header (e.g. `include/rpt_common.h`)
- change our channel drivers to use the functions that will not be exported by app_rpt.so ... and remove the no longer needed copies.

Note: any changes to consolidate the duplicated functions need not be done with this PR (it's more about the shared enums)

Allan-N: >I think we probably need a broader review/conversation around what all those functions are exactly and what they are doing, before making any changes like making a new resource module.

OK.  Let's move the discussion on common/shared functions to https://github.com/AllStarLink/app_rpt/issues/659

mkmer: > There is no `res_radio` module, so I don't think it's appropriate to add a header file called that... at least I've never seen that done before.
> 


There ARE other header files that do not have a "module" or associated `.c` files.  
One example: `\asterisk_core\include\asterisk\res_aeap_message.h`
Should we just create a header file that lives in `\channels\` , leave this as is, other choices?

PR #606, opened 2025-04-01
app_rpt: Use previously added macro for sscanf format

mkmer:
app_rpt: Use previously added macro for sscanf format


InterLinked1 (commented):
I don't think a body is needed, but "missed one" is a little lacking in context on its own... how about

`app_rpt: Use previously added macro for sscanf format` or something like that. At least somebody looking at it knows the backstory then.

Allan-N (approved):

Issue #607, opened 2025-04-02
ASL3.4.1-1 : Crashing on echolink connection


mkmer:
Local repeater crashing on echo link connection:
```
Connected to Asterisk 22.2.0+asl3-3.4.1-1.deb12 currently running on Moline (pid = 3017884)
    -- Executing [29972@radio-secure:1] Rpt("echolink/el0-0", "29972") in new stack
       > Received OK from Echolink server naeast.echolink.org.
       > Received OK from Echolink server servers.echolink.org.
       > Directory completed downloading(partial,compressed), 704 records.
  == Spawn extension (radio-secure, 29972, 1) exited non-zero on 'Surrogate/echolink/el0-0'
    -- Remote UNIX connection
    -- Remote UNIX connection disconnected
[2025-04-02 11:22:59.557] ERROR[3017929]: channel.c:1038 __ast_queue_frame: FRACK!, Failed assertion bad magic number 0x0 for object 0x7fa238001940 (0)
```

[core-asterisk-2025-04-02T15-22-59Z-info.txt](https://github.com/user-attachments/files/19571178/core-asterisk-2025-04-02T15-22-59Z-info.txt)
[core-asterisk-2025-04-02T15-22-59Z-locks.txt](https://github.com/user-attachments/files/19571175/core-asterisk-2025-04-02T15-22-59Z-locks.txt)
[core-asterisk-2025-04-02T15-22-59Z-thread1.txt](https://github.com/user-attachments/files/19571176/core-asterisk-2025-04-02T15-22-59Z-thread1.txt)
[core-asterisk-2025-04-02T15-22-59Z-brief.txt](https://github.com/user-attachments/files/19571177/core-asterisk-2025-04-02T15-22-59Z-brief.txt)
[core-asterisk-2025-04-02T15-22-59Z-full.txt](https://github.com/user-attachments/files/19571179/core-asterisk-2025-04-02T15-22-59Z-full.txt)

mkmer:
Root cause of echolink crash is here:
https://github.com/AllStarLink/app_rpt/commit/e554ae424e653f28ce468571faf05ab1e953f318

Appears the yank fails on an echolink channel.

InterLinked1:
[...]

PR #608, opened 2025-04-02
chan_echolink: Echolink channel driver does not handle ast_channel_yank()

mkmer:
`chan_echolink` was storing a channel pointer which it then attempted to use
after the `ast_channel_yank()`, causing a crash.  This stored pointer was
not necessary and has be refactored to remove it.

Fixes https://github.com/AllStarLink/app_rpt/issues/607

### DIFF
--- a/channels/chan_echolink.c
+++ b/channels/chan_echolink.c
@@ -356,7 +356,6 @@ struct el_node {
 	float jitter;
 	struct el_instance *instp;
 	struct el_pvt *pvt;
-	struct ast_channel *chan;
 	struct timeval last_packet_time;
 	uint32_t rx_audio_packets;
 	uint32_t tx_audio_packets;
@@ -481,13 +480,11 @@ struct el_rxqel {
  * This is stored in the asterisk channel private technology for reference.
  */
 struct el_pvt {
-	struct ast_channel *owner;
 	struct el_instance *instp;
 	char app[16];
 	char stream[80];
 	char ip[EL_IP_SIZE];
 	char firstsent;
-	char firstheard;
 	char txkey;
 	int rxkey;
 	int keepalive;
@@ -1325,7 +1322,6 @@ static int el_call(struct ast_channel *ast, const char *dest, int timeout)
 		ast_log(LOG_WARNING, "el_call called on %s, neither down nor reserved.\n", ast_channel_name(ast));
 		return -1;
 	}
-
 	/* Make sure we have a destination */
 	if (!*dest) {
 		ast_log(LOG_WARNING, "Call on %s failed - no destination.\n", ast_channel_name(ast));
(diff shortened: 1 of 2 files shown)


InterLinked1:
This isn't really a "fix", and it doesn't really track with the backtrace you posted at least from looking at it initially. The channel driver doesn't handle yank operations so this makes no sense.
I think we need to do some more analysis and figure out what the real issue is here.

mkmer:
Originally, I was generating a "fix" to get our beta back on track for release.  After a bit more analysis into WHY the channel driver couldn't handle a yank, we found it was storing a channel pointer, which isn't required and not the intended architecture (best I can tell) - using the channel handed to us in the callback solves to issue (probably correctly).

InterLinked1:
> Originally, I was generating a "fix" to get our beta back on track for release. After a bit more analysis into WHY the channel driver couldn't handle a yank, we found it was storing a channel pointer, which isn't required and not the intended architecture (best I can tell) - using the channel handed to us in the callback solves to issue (probably correctly).

Okay, looks much better and makes more sense now!

Did you mean to remove those two frame blocks?

Also, please update the commit message accordingly based on your update (I suppose the PR description, since your commits aren't squashed)

channels/chan_echolink.c
```diff
@@ -3365,6 +3370,8 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *pvt, const char
 				ast_mutex_unlock(&el_db_lock);
 				return -1;
 			}
+
+			ast_queue_frame(chan, &answer_fr);
```
InterLinked1, line 3373: Why are we manually queueing an answer frame here, rather than using the answer APIs?

mkmer: Just moved what  we had to here where we actually have a channel.
Which call should I use?

InterLinked1: Yes, `ast_answer` is usually good unless there's a specific reason not to use the default behavior (in which case `ast_raw_answer` might be better).

`ast_raw_answer` would be closer to the behavior of just sending the answer frame since it doesn't have logic to automatically wait for audio.

channels/chan_echolink.c
```diff
@@ -3586,22 +3582,6 @@ static void *el_reader(void *data)
 						found_key = (struct el_node **) tfind(&instp->el_node_test, &el_node_list, compare_ip);
 						if (found_key) {
 							node = *found_key;
-							if (!node->pvt->firstheard) {
-								node->pvt->firstheard = 1;
```
Allan-N, line 3590: Looks like all of the ->firstheard references have been removed

Allan-N (approved):

channels/chan_echolink.c
mkmer: Needed `ast_raw_answer()` - as there is a text message that was not tolerant of the lock created by `ast_answer()`

InterLinked1:
Here is a better commit message, I don't like the wording of it as it is now:

```
chan_echolink: Don't store pointer to channel.

chan_echolink was storing the channel pointer, which is invalid if a masquerade
occurs, such as in app_rpt's rpt_exec. This stored pointer was not necessary and
chan_echolink has been refactored to remove it.
```

InterLinked1 (approved):

PR #610, opened 2025-04-03
chan_tlb: TLB channel driver does not handle ast_channel_yank()

### LINKED ISSUE #609: chan_TLB: stored channel pointer not tolerant of `ast_channel_yank()`
Same problem as issue #607 , except a little harder to address.

mkmer:
chan_tlb: Don't store pointer to channel.
chan_tlb was storing the channel pointer, which is invalid if a masquerade
occurs, such as in app_rpt's rpt_exec. This stored pointer was not necessary and
chan_tlb has been refactored to remove it.
Fixes #609

### DIFF
--- a/channels/chan_tlb.c
+++ b/channels/chan_tlb.c
@@ -96,6 +96,8 @@ tlb.conf file.
 
 #define	MAX_RXKEY_TIME 4
 #define	RTPBUF_SIZE 400			/* actually 320 would be sufficient */
+#define TLB_INIT_STR 40
+
 enum { TLB_GSM, TLB_G726, TLB_ULAW };
 
 /*!
@@ -245,7 +247,7 @@ struct TLB_instance;
 struct TLB_pvt;
 
 /*!
- * \brief Echolink connected node struct.
+ * \brief TLB connected node struct.
  * These are stored internally in a binary tree.
  */
 struct TLB_node {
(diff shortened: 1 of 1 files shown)


channels/chan_tlb.c
```diff
@@ -2260,41 +2252,39 @@ static void *TLB_reader(void *data)
...
+							if (textq) {
+								textq->buff = ast_str_create(TLB_INIT_STR);
+								if (textq->buff) {
+									ast_str_set(&textq->buff, 0, "D 0 %s %u %c", p->instp->astnode, ++(p->dtmfidx), dchar);
+									insque((struct qelem *) textq, (struct qelem *) p->textq.qe_back);
+								} else {
+									ast_free(textq);
+								}
+							}
+							ast_mutex_unlock(&p->lock);
```
Allan-N, line 2266: When/where was the lock acquired?

mkmer: Line 1442 -> we mess around with `p` all over in here.  Maybe it's wrong to lock it for the whole call.  At a minimum I think we need to lock around the `insque` and `remque`  calls (which is not being done)

Allan-N: I should have looked at the changes in an editor (vs. looking at the diff on GitHub).  Thank you for pointing out the matching lock.

Allan-N (approved):

InterLinked1 (approved):

PR #611, opened 2025-04-04
chan_echolink: Address memory leak and code quality updates

mkmer:
Cleanup empty queues and free memroy when hanging up.
mutex lock around `insque()` and `remque()` calls as they are in executed in different threads.
Add mutex lock around access to el_nodlist access.

### DIFF
--- a/channels/chan_echolink.c
+++ b/channels/chan_echolink.c
@@ -484,9 +484,9 @@ struct el_pvt {
 	char app[16];
 	char stream[80];
 	char ip[EL_IP_SIZE];
-	char firstsent;
-	char txkey;
-	int rxkey;
+	unsigned int firstsent:1; /* First packet seen from echolink */
+	unsigned int txkey:1;	  /* Transmit keyed */
+	int rxkey;				  /* Receive keyed timer */
 	int keepalive;
 	int txindex;
 	struct ast_frame fr;
@@ -497,6 +497,7 @@ struct el_pvt {
 	struct ast_trans_pvt *xpath;
 	unsigned int nodenum;
 	char *linkstr;
+	ast_mutex_t lock;
 };
 
 /*!
@@ -742,6 +743,7 @@ static int compare_eldb_callsign(const void *pa, const void *pb)
 
 /*!
  * \brief Find an echolink node from the internal user database by nodenum.
+ * \note Must be called locked.
  * \param nodenum	Pointer to node number to find.
  * \retval NULL		If the IP address was not found.
  * \return If found returns an eldb struct.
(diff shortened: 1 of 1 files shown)


channels/chan_echolink.c
```diff
@@ -484,9 +484,9 @@ struct el_pvt {
 	char app[16];
 	char stream[80];
 	char ip[EL_IP_SIZE];
-	char firstsent;
-	char txkey;
-	int rxkey;
+	unsigned int firstsent:1; /*First packet seen from echolink*/
```
InterLinked1, line 487: comment spacing - space before First and after Echolink
Same below

channels/chan_echolink.c
```diff
@@ -742,6 +743,7 @@ static int compare_eldb_callsign(const void *pa, const void *pb)
 
 /*!
  * \brief Find an echolink node from the internal user database by nodenum.
+ * Must be called locked.
```
InterLinked1, line 746: These should be a `\note`

channels/chan_echolink.c
```diff
@@ -1492,6 +1505,21 @@ static int el_hangup(struct ast_channel *ast)
 		ast_log(LOG_WARNING, "Asked to hangup channel not connected.\n");
 		return 0;
 	}
+
+	ast_mutex_lock(&pvt->lock);
+	while (pvt->rxqast.qe_forw != &pvt->rxqast) {
+		qpast = (struct qelem *) pvt->rxqast.qe_forw;
```
InterLinked1, line 1511: Casts from void are not necessary and should not be done in C (assuming that's what this is)

mkmer: Assuming I understand: This is not casting from void.  `rxqast` is a `struct el_rxqast`, and `rxqel` is a `struct el_rxqel`.
This is a slight refactor of this:
https://github.com/AllStarLink/app_rpt/blob/73f70323c84692f4561680b5a7aeb28bd8451180/channels/chan_echolink.c#L2265-L2267
Allowing to reuse the variable for multiple queues.

channels/chan_echolink.c
```diff
@@ -484,9 +484,9 @@ struct el_pvt {
 	char app[16];
 	char stream[80];
 	char ip[EL_IP_SIZE];
-	char firstsent;
-	char txkey;
-	int rxkey;
+	unsigned int firstsent:1; /* First packet seen from echolink */
+	unsigned int txkey:1;	  /* Transmit keyed timer*/
```
InterLinked1, line 488: space before start

InterLinked1 (changes requested):

InterLinked1 (approved):

channels/chan_echolink.c
```diff
@@ -3363,7 +3388,8 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *pvt, const char
 				ast_mutex_unlock(&el_db_lock);
 				return -1;
 			}
-			ast_raw_answer(chan);
+			fr.src = type;
+			ast_queue_frame(chan, &fr);
```
mkmer, line 3392: While `ast_raw_answer()` appeared to work, we need to use `ast_queue_frame()` because we are in a channel driver here.

Allan-N (approved):

PR #613, opened 2025-04-04
chan_tlb: Address memory leak on hangup and code cleanup

mkmer:
Change all "el" to "tlb" references.
Add mutex around remque() and insque() calls.
Use ast_indicate() for key/unkey.


channels/chan_tlb.c
```diff
@@ -2056,7 +2073,8 @@ static int do_new_call(struct TLB_instance *instp, struct TLB_pvt *p, const char
 				}
 				tlb_set_nativeformats(chan, p->txcodec, p->rxcodec);
 				ast_debug(1, "tlb: tx codec set to %s\n", tlb_codecs[p->txcodec].name);
-				ast_raw_answer(chan);
+				fr.src = type;
+				ast_queue_frame(chan, &fr);
```
mkmer, line 2077: While using `ast_raw_answer` appered to "work" we need to `use ast_queue_frame` as we are in a driver here.

InterLinked1: Thanks for going down that rabbit hole... should we add a comment about that so the next guy knows?

Allan-N (approved):

PR #615, opened 2025-04-04
chan_echolink: restore removed hangup conditions

### LINKED ISSUE #614: Echolink not "haning up" channels after call
Not sure how I missed this in testing, but we are not hanging up the channels after a call is completed.

After moving `softhangup()` out of `find_delete()` there are 2 locations where we need to hangup, but no longer have a channel reference to hangup.

https://github.com/AllStarLink/app_rpt/blob/73f70323c84692f4561680b5a7aeb28bd8451180/channels/chan_echolink.c#L2175-L2178

and here:

https://github.com/AllStarLink/app_rpt/blob/73f70323c84692f4561680b5a7aeb28bd8451180/channels/chan_echolink.c#L3686-L3689

We have a channel here, so probably just need a `softhangup()`

https://github.com/AllStarLink/app_rpt/blob/73f70323c84692f4561680b5a7aeb28bd8451180/channels/chan_echolink.c#L2373-L2375

mkmer:
Relocaiton of `softhangup()` outside of the `find_delete()` function left channels hanging.
The channel's private data is available and sticks with a channel masquerade.  Simply add a flag in the private data allows us to hangup on the next channel read/write.
Fixes #614

### DIFF
--- a/channels/chan_echolink.c
+++ b/channels/chan_echolink.c
@@ -489,6 +489,7 @@ struct el_pvt {
 	int rxkey;
 	int keepalive;
 	int txindex;
+	unsigned int hangup:1; /* indicate the channel should hang up */
 	struct ast_frame fr;
 	struct el_rxqast rxqast;
 	struct el_rxqel rxqel;
@@ -1441,6 +1442,7 @@ static struct el_pvt *el_alloc(const char *data)
 			ast_log(LOG_ERROR, "Cannot get DSP!\n");
 			return NULL;
 		}
+		pvt->hangup = 0;
 		ast_dsp_set_features(pvt->dsp, DSP_FEATURE_DIGIT_DETECT);
 		ast_dsp_set_digitmode(pvt->dsp, DSP_DIGITMODE_DTMF | DSP_DIGITMODE_MUTECONF | DSP_DIGITMODE_RELAXDTMF);
 		pvt->xpath = ast_translator_build_path(ast_format_slin, ast_format_gsm);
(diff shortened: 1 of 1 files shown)


InterLinked1 (commented):
Don't say "channel yank" in the commit message, that is a specific API. In general, this is when a channel is masqueraded, if you're referring to that.

Also, the changes for tlb and echolink look similar, you should combine them into one PR, not have one for each

mkmer:
> Don't say "channel yank" in the commit message, that is a specific API. In general, this is when a channel is masqueraded, if you're referring to that.
> 
> Also, the changes for tlb and echolink look similar, you should combine them into one PR, not have one for each

`ast_channel_yank(chan);` is the problem/function call - none of the docs around this function say anything about masquerading.  
I'd say the function is named wrong if we are supposed to call it masquerading.

InterLinked1:
> > Don't say "channel yank" in the commit message, that is a specific API. In general, this is when a channel is masqueraded, if you're referring to that.
> > Also, the changes for tlb and echolink look similar, you should combine them into one PR, not have one for each
> 
> `ast_channel_yank(chan);` is the problem/function call - none of the docs around this function say anything about masquerading. I'd say the function is named wrong if we are supposed to call it masquerading.

That is what the process of taking a channel and moving its guts into another channel structure is called. Take a look at: https://www.asterisk.org/asterisk-architecture-the-bridging-framework-part-i/

Any masquerade would cause these same issues, it's nothing specific to `ast_channel_yank`.

Allan-N (approved):

PR #616, opened 2025-04-04
chan_tlb.c: restore removed hangup conditions

mkmer:
Relocaiton of `softhangup()` outside of the `find_delete()` function left channels hanging.

### DIFF
--- a/channels/chan_tlb.c
+++ b/channels/chan_tlb.c
@@ -340,6 +340,7 @@ struct TLB_pvt {
 	char txkey;
 	int rxkey;
 	int keepalive;
+	unsigned int hangup:1; /* Hangup channel requested */
 	struct ast_frame fr;
 	int txindex;
 	struct TLB_rxqast rxqast;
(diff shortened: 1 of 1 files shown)


channels/chan_tlb.c
```diff
@@ -340,6 +340,7 @@ struct TLB_pvt {
 	char txkey;
 	int rxkey;
 	int keepalive;
+	int hangup:1; /* Hangup channel requested */
```
InterLinked1, line 343: I'd make this unsigned int:1 for consistency, not just int:1
There is a reason for that, though I don't know what it is off hand

mkmer: Now that I'm looking around, there are other cases of `uint8_t` for `xxx:1` 
Wonder if we should use that...

InterLinked1: I proposed `unsigned int:1` primarily because it's widely used in Asterisk and has well established semantics. I haven't seen this done with `uint8_t`.

Allan-N (approved):

Issue #526, opened 2025-03-02
Crash when rxctcssfreqs / txctcssfreqs list counts do not match


Allan-N:
From [Community #22965](https://community.allstarlink.org/t/asl3-asterisk-failing-to-start-code-dumped-status-11-segv/22965) :

Asterisk will crash if the size of the `rxctcssfreqs` and `txctcssfreqs` do not have the same number of tones.

InterLinked1:
Since this seems to be config-related, however we decide to fix it, this is a perfect candidate for adding a test for this as well.

mkmer:
This is probably not good:  
https://github.com/AllStarLink/app_rpt/blob/b542ccd42c7d106c824ac402345c0dda115c7420/channels/chan_usbradio.c#L4731-L4734
Only checking one size but looking at both....

mkmer:
Does it make sense to have 2 tx ctcss carriers?  Seems a bit odd

InterLinked1:
Looks like that entire function is also written wrong it seems - it's called from a CLI command, but instead of printing to that file descriptor, it uses printf, which is wrong.

PR #618, opened 2025-04-09
xpmr.c: Replace `printf()` with `ast_log()` restoring logging.

mkmer:
Related to this issue :
https://github.com/AllStarLink/app_rpt/issues/526

### DIFF
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -65,6 +65,7 @@
 #endif
 
 #define N_FMT(duf) "%30" #duf /* Maximum sscanf conversion to numeric strings */
+#include "asterisk.h"
 
 #include <stdio.h>
 #include <ctype.h>
@@ -122,14 +123,13 @@ void strace2(t_sdbg *sdbg)
 /*
 	Hardware Trace Signals via the PC Parallel Port
 */
-void	pptp_init (void)
+void pptp_init(void)
 {
-	if (ppdrvdev == 0)
-    	ppdrvdev = open("/dev/ppdrv_device", 0);
-
-    if (ppdrvdev < 0)
-    {
-        ast_log(LOG_ERROR, "open /dev/ppdrv_ppdrvdev returned %i\n",ppdrvdev);
+	if (ppdrvdev == 0) {
+		ppdrvdev = open("/dev/ppdrv_device", 0);
+	}
+	if (ppdrvdev < 0) {
+		ast_log(LOG_ERROR, "open /dev/ppdrv_ppdrvdev returned %i\n", ppdrvdev);
 		exit(0);
 	}
 	ioctl(ppdrvdev, PPDRV_IOC_PINMODE_OUT, DTX_CLK | DTX_DATA | DTX_ENABLE | DTX_TXPWR | DTX_TX | DTX_TP1 | DTX_TP2);
@@ -164,10 +164,10 @@ i16 string_parse(char *src, char **dest, char ***ptrs)
 	char *ptstr[1000];
 	i16 i, slen, numsub;
 
-	TRACEJ(2,("string_parse(%s)\n",src));
+	TRACEJ(2, "string_parse(%s)\n", src);
 
 	slen=strlen(src);
-	TRACEJ(2,(" source len = %i\n",slen));
+	TRACEJ(2, " source len = %i\n", slen);
 
 	pd=*dest;
 	if(pd) ast_free(pd);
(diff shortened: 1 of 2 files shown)


channels/xpmr/xpmr.c
```diff
@@ -129,7 +130,7 @@ void	pptp_init (void)
 
     if (ppdrvdev < 0)
     {
-        ast_log(LOG_ERROR, "open /dev/ppdrv_ppdrvdev returned %i\n",ppdrvdev);
+		ast_log(AST_LOG_ERROR, "open /dev/ppdrv_ppdrvdev returned %i\n", ppdrvdev);
```
InterLinked1, line 133: Could you fix the alignment of the two lines above while you're doing this?
Also, LOG_ERROR is the same as AST_LOG_ERROR, etc.

channels/xpmr/xpmr.c
```diff
@@ -2200,7 +2205,9 @@ t_pmr_chan	*createPmrChannel(t_pmr_chan *tChan, i16 numSamples)
 		pSps->nx=taps_int_lpf_300_1_2;
 		pSps->size_x=4;
 		pSps->x=(void*)(ast_calloc(pSps->nx,pSps->size_x));
-		if(pSps==NULL)printf("Error: calloc(), createPmrChannel()\n");
+		if (pSps == NULL) {
+			ast_log(AST_LOG_ERROR, "Error: calloc(), createPmrChannel()\n");
```
InterLinked1, line 2209: This log message isn't needed (alloc failure) and can just be removed
Same for a few others here
That being said, it looks like the code is not safe when allocation fails as it continues anyways, so the code will need to abort

mkmer: It's an odd test here : we calloc to `pSps->x`, but test `pSps` for NULL.  
That's bad all the way around as we've been setting suff in `pSps` up to here :)

Allan-N (commented):
I'm not so sure about these changes.  While I have no issue with the switch from `printf()` to `ast_log()` I do question moving away from using the `TRACE...()` defines.

Seems like the old code, if enabled, would check the `tracelevel` value to control the printf's ... and that value could be updated using CLI `radio set debug [level]` commands.

Without that extra layer we're stuck with mixing all of the chan_usbradio log messages and the xpmr log messages together.

mkmer:
I had tried reworking the defines to use `ast_log()` and ran into a problem (I can't remember what it was...) 
I tried putting these at a higher debug levels so we could separate them based on level. (Not saying it's right or wrong)

Allan-N:
>I had tried reworking the defines to use ast_log() and ran into a problem

Here's a guess.  You started off with a #define (e.g. `TRACEC(level, a)`) and tried to replace the `printf` with a call to `ast_debug()`. The gotcha is that `ast_debug()` is also a #define ... and that's never fun.

InterLinked1:
> > I had tried reworking the defines to use ast_log() and ran into a problem
> 
> Here's a guess. You started off with a #define (e.g. `TRACEC(level, a)`) and tried to replace the `printf` with a call to `ast_debug()`. The gotcha is that `ast_debug()` is also a #define ... and that's never fun.
> 
> You might look at adding your own macro, one that looks a lot like `ast_debug()`, but is also `tracelevel` aware.

There is an `ast_trace`, in addition to `ast_debug`, if that helps.

I don't think "being mixed together" is really a concern, the log messages will have the filename of the source file in them so one could filter to either the USB module or to xpmr.c

mkmer:
OK - learned a little macro magic that I didn't know before ;)  I think this will do what we need (aka send debug stuff somewhere we can see it)

channels/xpmr/xpmr.h
```diff
@@ -80,30 +80,35 @@
...
-#define TRACEJ(level,a) {if(XPMR_TRACE_LEVEL>=level){printf a;} }
-#define TRACES(level,a) {if(mySps->parentChan->tracelevel >= level){printf a;} }
-#define TRACET(level,a) {if(pChan->tracelevel>=level){printf("%08i %02i",pChan->frameCountRx,pChan->rptnum);printf a;} }
-#define TRACEXR(a) {printf a;}
-#define TRACEM(level,sys,a) {if(pChan->tracelevel>=level || (pChan->tracesys[sys])){printf a;} }
-#else						  
+#if (XPMR_TRACE == 1)
+#define TRACEX(a) \
+	{ \
+		ast_debug a; \
```
Allan-N, line 86: All of the places using TRACEX() are commented out ... but if any had been uncommented you 'd see that this won't work.

mkmer: I think it will as it's the same as "before" where `printf()` just had `printf a`.  It is replacing `a` with the stuff in `()`.
I did un-comment and compile and it does work.

Allan-N: Ah, yes, the commented out TRACEX calls still use the older style with the parens
```
	// TRACEX((1, " -MeasureBlock()=%i\n",mySps->apeak));
```
so, instead of the result being `printf(1, " -MeasureBlock()=%i\n",mySps->apeak)` we will now have `ast_debug(1, " -MeasureBlock()=%i\n",mySps->apeak)`.

This just looked out of place without the __VA_ARGS__ change you made to the other defines

PR #620, opened 2025-04-09
xpmr.c: Prevent out of range reference to `pChan->pTxCode[i]`

mkmer:
At least partially addresses Issue https://github.com/AllStarLink/app_rpt/issues/526

### DIFF
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -331,9 +331,9 @@ i16 code_string_parse(t_pmr_chan *pChan)
 				TRACEF(1, ("pChan->rxctcss[%i]=%s  pChan->rxCtcssMap[%i]=%i RXONLY\n", i, pChan->rxctcss[i], ri, ti));
 			} else {
 				pChan->numrxctcssfreqs = 0;
+				ast_log(LOG_ERROR, "ERROR: Invalid CTCSS configuration. CTCSS has been disabled\n");
 				for (ii = 0; ii < CTCSS_NUM_CODES; ii++) {
 					pChan->rxCtcssMap[ii] = CTCSS_NULL;
-					ast_log(LOG_ERROR, "ERROR: Invalid CTCSS configuration. TX CTCSS has been disabled\n");
 				}
 			}
 		}
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -298,10 +298,15 @@ i16 code_string_parse(t_pmr_chan *pChan)
 				maxctcssindex = ri;
 			}
 
-			sscanf(pChan->pTxCode[i], N_FMT(f), &f);
-			ti = CtcssFreqIndex(f);
-			if (f > maxctcsstxfreq) {
-				maxctcsstxfreq = f;
+			if (i < pChan->numtxcodes) {
+				sscanf(pChan->pTxCode[i], N_FMT(f), &f);
+				ti = CtcssFreqIndex(f);
+				if (f > maxctcsstxfreq) {
+					maxctcsstxfreq = f;
+				}
+			} else {
+				ti = -1;
+				f = 0;
 			}
 
 			if (ri > CTCSS_NULL && ti > CTCSS_NULL) {
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -312,6 +312,8 @@ i16 code_string_parse(t_pmr_chan *pChan)
 				ti = CtcssFreqIndex(f);
 				if (f > maxctcsstxfreq && ti > CTCSS_NULL) {
 					maxctcsstxfreq = f;
+				} else {
+					f = 0;
 				}
 			} else {
 				ti = -1;
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -255,7 +255,16 @@ i16 code_string_parse(t_pmr_chan *pChan)
 	pChan->numtxcodes = string_parse( pChan->pTxCodeSrc, &(pChan->pTxCodeStr), &(pChan->pTxCode));
 
 	if(pChan->numrxcodes!=pChan->numtxcodes)printf("ERROR: numrxcodes != numtxcodes \n");
-	
+
+	if (pChan->numrxcodes >= CTCSS_NUM_CODES) {
+		ast_log(LOG_ERROR, "To many rxctcss codes: %d defined, only %d allowed", pChan->numrxcodes, CTCSS_NUM_CODES);
+		pChan->numrxcodes = CTCSS_NUM_CODES;
+	}
+	if (pChan->numtxcodes >= CTCSS_NUM_CODES) {
+		ast_log(LOG_ERROR, "To many txctcss codes: %d defined, only %d allowed", pChan->numtxcodes, CTCSS_NUM_CODES);
+		pChan->numtxcodes = CTCSS_NUM_CODES;
+	}
+
 	pChan->rxCtcss->enabled=0;
 	pChan->rxCtcss->gain=1*M_Q8;
 	pChan->rxCtcss->limit=8192;
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -301,13 +301,13 @@ i16 code_string_parse(t_pmr_chan *pChan)
 			if (i < pChan->numtxcodes) {
 				sscanf(pChan->pTxCode[i], N_FMT(f), &f);
 				ti = CtcssFreqIndex(f);
-				if (f > maxctcsstxfreq && ti > CTCSS_NULL) {
-					maxctcsstxfreq = f;
-				} else {
+				if (ti == CTCSS_NULL) {
 					f = 0;
+				} else if (f > maxctcsstxfreq) {
+					maxctcsstxfreq = f;
 				}
 			} else {
-				ti = -1;
+				ti = CTCSS_NULL;
 				f = 0;
 			}
 
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -379,7 +379,9 @@ i16 code_string_parse(t_pmr_chan *pChan)
 	{
 		sscanf(p, N_FMT(f), &f);
 		ti = CtcssFreqIndex(f);
-		if (f > maxctcsstxfreq) {
+		if (ti == CTCSS_NULL) {
+			ast_log(LOG_ERROR, "ERROR: Invalid default TX CTCSS code detected and ignored. %s\n", pChan->pTxCodeDefault);
+		} else if (f > maxctcsstxfreq) {
 			maxctcsstxfreq = f;
 		}
 
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -310,7 +310,7 @@ i16 code_string_parse(t_pmr_chan *pChan)
 			if (i < pChan->numtxcodes) {
 				sscanf(pChan->pTxCode[i], N_FMT(f), &f);
 				ti = CtcssFreqIndex(f);
-				if (f > maxctcsstxfreq) {
+				if (f > maxctcsstxfreq && ti > -1) {
 					maxctcsstxfreq = f;
 				}
 			} else {
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -310,7 +310,7 @@ i16 code_string_parse(t_pmr_chan *pChan)
 			if (i < pChan->numtxcodes) {
 				sscanf(pChan->pTxCode[i], N_FMT(f), &f);
 				ti = CtcssFreqIndex(f);
-				if (f > maxctcsstxfreq && ti > -1) {
+				if (f > maxctcsstxfreq && ti > CTCSS_NULL) {
 					maxctcsstxfreq = f;
 				}
 			} else {
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -254,8 +254,9 @@ i16 code_string_parse(t_pmr_chan *pChan)
 	pChan->numrxcodes = string_parse( pChan->pRxCodeSrc, &(pChan->pRxCodeStr), &(pChan->pRxCode));
 	pChan->numtxcodes = string_parse( pChan->pTxCodeSrc, &(pChan->pTxCodeStr), &(pChan->pTxCode));
 
-	if(pChan->numrxcodes!=pChan->numtxcodes)printf("ERROR: numrxcodes != numtxcodes \n");
-
+	if (pChan->numrxcodes != pChan->numtxcodes) {
+		ast_log(LOG_ERROR, "numrxcodes != numtxcodes \n");
+	}
 	pChan->rxCtcss->enabled=0;
 	pChan->rxCtcss->gain=1*M_Q8;
 	pChan->rxCtcss->limit=8192;
@@ -295,7 +296,7 @@ i16 code_string_parse(t_pmr_chan *pChan)
 			sscanf(p, N_FMT(f), &f);
 			ri = CtcssFreqIndex(f);
 			if (ri == CTCSS_NULL) {
-				ast_log(LOG_ERROR, "ERROR: Invalid RX CTCSS code detected and ignored. %i %s\n", i, pChan->pRxCode[i]);
+				ast_log(LOG_ERROR, "Invalid RX CTCSS code detected and ignored. %i %s\n", i, pChan->pRxCode[i]);
 
 			} else if (ri > maxctcssindex) {
 				maxctcssindex = ri;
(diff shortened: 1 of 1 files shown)


channels/xpmr/xpmr.c
```diff
@@ -255,7 +255,16 @@ i16 code_string_parse(t_pmr_chan *pChan)
 	pChan->numtxcodes = string_parse( pChan->pTxCodeSrc, &(pChan->pTxCodeStr), &(pChan->pTxCode));
 
 	if(pChan->numrxcodes!=pChan->numtxcodes)printf("ERROR: numrxcodes != numtxcodes \n");
-	
+
+	if (pChan->numrxcodes >= CTCSS_NUM_CODES) {
+		ast_log(LOG_ERROR, "To many rxctcss codes: %d defined, only %d allowed", pChan->numrxcodes, CTCSS_NUM_CODES);
```
InterLinked1, line 260: Too many

channels/xpmr/xpmr.c
```diff
@@ -298,10 +307,15 @@ i16 code_string_parse(t_pmr_chan *pChan)
 				maxctcssindex = ri;
 			}
 
-			sscanf(pChan->pTxCode[i], N_FMT(f), &f);
-			ti = CtcssFreqIndex(f);
-			if (f > maxctcsstxfreq) {
-				maxctcsstxfreq = f;
+			if (i < pChan->numtxcodes) {
+				sscanf(pChan->pTxCode[i], N_FMT(f), &f);
+				ti = CtcssFreqIndex(f);
```
Allan-N, line 312: Should we check if `(ti > maxctcssindex)` to ensure we have a valid tone?

channels/xpmr/xpmr.c
```diff
@@ -255,7 +255,16 @@ i16 code_string_parse(t_pmr_chan *pChan)
 	pChan->numtxcodes = string_parse( pChan->pTxCodeSrc, &(pChan->pTxCodeStr), &(pChan->pTxCode));
 
 	if(pChan->numrxcodes!=pChan->numtxcodes)printf("ERROR: numrxcodes != numtxcodes \n");
-	
+
+	if (pChan->numrxcodes >= CTCSS_NUM_CODES) {
```
Allan-N, line 259: In this case, and in this file, I don't think we actually have a CTCSS_NUM_CODES limit.  The string_parse() function appears to allocate an array of pointers sized for however many tones are provided/parsed.

mkmer: If it's wrong at ln 259, then it must be wrong here?
https://github.com/AllStarLink/app_rpt/blob/e61817d07d90740acb68ce492365cd4a21262748/channels/xpmr/xpmr.c#L279-L283
These arrays are limited to [38] - so if we find >38 in the list we will have a problem down here (assuming I've interpreted this right).

Allan-N (changes requested):

channels/xpmr/xpmr.c
```diff
@@ -310,7 +310,7 @@ i16 code_string_parse(t_pmr_chan *pChan)
 			if (i < pChan->numtxcodes) {
 				sscanf(pChan->pTxCode[i], N_FMT(f), &f);
 				ti = CtcssFreqIndex(f);
-				if (f > maxctcsstxfreq && ti > -1) {
+				if (f > maxctcsstxfreq && ti > CTCSS_NULL) {
```
Allan-N, line 313: if `ti` is not valid, shouldn't we also be setting `f = 0`?

Allan-N (changes requested):

channels/xpmr/xpmr.c
```diff
@@ -312,6 +312,8 @@ i16 code_string_parse(t_pmr_chan *pChan)
 				ti = CtcssFreqIndex(f);
 				if (f > maxctcsstxfreq && ti > CTCSS_NULL) {
 					maxctcsstxfreq = f;
+				} else {
+					f = 0;
```
Allan-N, line 316: Sorry to be a pain here but you've got 3 variables in play.  First, we have the frequency, "f" ... that should be zero'd if it is not a valid tone.  Then, we have `ti` that reflects the tone index.  Lastly, we have `maxctcsstxfreq` that should be updated IF we have a valid tone.

Allan-N (changes requested):

channels/xpmr/xpmr.c
mkmer: I think it's right -> if `ti == CTCSS_NULL`, we don't update `maxctcsstxfreq` AND set `f = 0`
If `i` is outside of the number of `numtxcodes`, we skip all of it and set `ti = -1` and `f = 0`

Allan-N: You are correct.  My apologies.

mkmer: It probably could be easier to read... But "easier"/refactor isn't jumping out at me right now.

mkmer: Actually, it's still wrong - if f < maxfreq, we don't want f = 0.

mkmer: OK - I'm thinking this is a bit easier to read, and it's doing the right thing now :)

mkmer:
I was able to replicate the crash on my system (mismatched number for tx/rx codes).
This PR does eliminate the crash when this scenario is present.

channels/xpmr/xpmr.c
Allan-N: I think you're close but not quite there.  How about :
```
                        if (i < pChan->numtxcodes) {
                                sscanf(pChan->pTxCode[i], N_FMT(f), &f);
                                ti = CtcssFreqIndex(f);
                                if (ti == CTCSS_NULL) {
                                        if (f != 0.0) {
                                                f = -1.0; /* tone freq not valid */
                                        } 
                                } else if (f > maxctcsstxfreq) {
                                        maxctcsstxfreq = f;
                                }
                        } else {
                                ti = CTCSS_NULL;
                                f = -1.0; /* tone freq not provided */
                        }

Comment (mkmer) [reply channels/xpmr/xpmr.c]: The more I look at this, the more lost I become!
We look up the rx frequency index in the CTCSS Table (from the configured frequency in the config file).
Then we lookup the tx frequncy an set its index in a map `rxCtcssMap[ri]=ti;`.  So if we have 2 entries with the same RX frequency but different tx frequencies, the last one will win.
If we have 1 invalid tx frequency, we flush the whole table (so I suspect we will not txctcss at all, but will still rx ctcss)

Comment (Allan-N) [reply channels/xpmr/xpmr.c]: >Then we lookup the tx frequncy an set its index in a map rxCtcssMap[ri]=ti;. So if we have 2 entries with the same RX frequency but different tx frequencies, the last one will win.

I suspect the first RX mapping will win (search for the first matching RX freq and use that mapping).  ... but you can have multiple RX frequencies that will output the same TX frequency.

>If we have 1 invalid tx frequency, we flush the whole table (so I suspect we will not txctcss at all, but will still rx ctcss)

Yes, that is odd.  Maybe we should keep any valid mappings and just ignore/log the invalid ones?

Comment (mkmer) [reply channels/xpmr/xpmr.c]: Talked with a commercial radio guy today - this solution is called "split tone": 

- Input match to any output (RX/TX pairs - same or different values)
- multiple RX inputs matched a TX (say 3 different inputs with 1 output).
- No tx output (aka `f = 0` on tx)

Not allowed to have multiple output ctcss on a single input ctcss.

Other than having multiple input ctcss with the same frequency in the list, it looks like we have this covered based on his description anyway.   We don't crash or fault in this case, just the last input ctcss to output ctcss will win.

Amen :)

File: channels/xpmr/xpmr.c [hunk 6]
```diff
@@ -316,8 +325,10 @@ i16 code_string_parse(t_pmr_chan *pChan)
 				TRACEF(1,("pChan->rxctcss[%i]=%s  pChan->rxCtcssMap[%i]=%i RXONLY\n",i,pChan->rxctcss[i],ri,ti));
 			} else {
 				pChan->numrxctcssfreqs=0;
-				for(ii=0;ii<CTCSS_NUM_CODES;ii++) pChan->rxCtcssMap[ii]=CTCSS_NULL;
-				TRACEF(1,("WARNING: Invalid Channel code detected and ignored. %i %s %s \n",i,pChan->pRxCode[i],pChan->pTxCode[i]));
+				for (ii = 0; ii < CTCSS_NUM_CODES; ii++) {
+					pChan->rxCtcssMap[ii] = CTCSS_NULL;
+				}
+				ast_log(LOG_ERROR, "ERROR: Invalid Channel code detected and ignored. %i %s %s \n", i, pChan->pRxCode[i], pChan->pTxCode[i]);
```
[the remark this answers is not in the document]
mkmer: I'm really "unhappy" with the complications around this whole section of code... still not seeing a better way at the moment.  
Relocated the log messages, keeping the index in range for "this" error message, and added a couple to clearly indicate when we "turn off" things that are expected to work.

channels/xpmr/xpmr.c
```diff
@@ -298,26 +298,44 @@ i16 code_string_parse(t_pmr_chan *pChan)
...
-			if (f > maxctcsstxfreq) {
-				maxctcsstxfreq = f;
+			if (i < pChan->numtxcodes) { /* more rx codes than tx codes */
+				sscanf(pChan->pTxCode[i], N_FMT(f), &f);
+				ti = CtcssFreqIndex(f);
+				if (ti == CTCSS_NULL) {
+					if (f != 0.0) {
+						f = -1.0; /* tone freq not valid */
+						ast_log(LOG_ERROR,
+							"ERROR: Invalid Channel code detected and ignored. %i %s %s \n",
```
Allan-N, line 308: Let's not confuse people.  Can we have the error say "Invalid TX CTCSS code ..."

channels/xpmr/xpmr.c
```diff
@@ -298,26 +298,44 @@ i16 code_string_parse(t_pmr_chan *pChan)
...
+							pChan->pRxCode[i],
+							pChan->pTxCode[i]);
+					}
+				} else if (f > maxctcsstxfreq) {
+					maxctcsstxfreq = f;
+				}
+			} else {
+				ti = CTCSS_NULL;
+				f = -1.0; /* tone freq not provided */
+				ast_log(LOG_ERROR, "ERROR: Invalid ctcss configuration.  Number of rx codes > number of tx codes\n");
```
Allan-N, line 319: ctcss --> CTCSS

channels/xpmr/xpmr.c
```diff
@@ -298,26 +298,44 @@ i16 code_string_parse(t_pmr_chan *pChan)
...
-				TRACEF(1,("pChan->rxctcss[%i]=%s  pChan->rxCtcssMap[%i]=%i\n",i,pChan->rxctcss[i],ri,ti));
+				TRACEF(1, ("pChan->rxctcss[%i]=%s  pChan->rxCtcssMap[%i]=%i\n", i, pChan->rxctcss[i], ri, ti));
 			} else if (ri > CTCSS_NULL && f == 0) {
 				pChan->b.ctcssRxEnable=1;
-				pChan->rxCtcssMap[ri]=CTCSS_RXONLY;
+				pChan->rxCtcssMap[ri] = CTCSS_RXONLY;
 				pChan->numrxctcssfreqs++;
-				TRACEF(1,("pChan->rxctcss[%i]=%s  pChan->rxCtcssMap[%i]=%i RXONLY\n",i,pChan->rxctcss[i],ri,ti));
+				TRACEF(1, ("pChan->rxctcss[%i]=%s  pChan->rxCtcssMap[%i]=%i RXONLY\n", i, pChan->rxctcss[i], ri, ti));
 			} else {
```
Allan-N, line 333: The first test adds valid RX->TX mappings, the second adds valid RXONLY mappings, and we've already logged bad data.  I'm questioning whether we want to toss what's valid because of 1 (or more) bad entries.  I'm thinking we should just remove the `else {}` case (i.e. no need to disable everything)

channels/xpmr/xpmr.c
```diff
@@ -298,26 +298,44 @@ i16 code_string_parse(t_pmr_chan *pChan)
 				maxctcssindex = ri;
```
Allan-N, line 298: As below, do we want to check if `ri == CTCSS_NULL` and log an error about an invalid RX CTCSS code?

mkmer: Yes, probably should.
Looking a bit further down the code "page" there are more sscanf and potential challenges that follow... 
Maybe the right answer in the "else" is just to `return` and stop processing anything until they fix it (just don't crash).
I'll have to dig in a bit to see if it's "ok" to just `return`

channels/xpmr/xpmr.c
mkmer: IT only disables TX codes... Generally, if the config is screwed up, we should just kill it all and log it.  Why try to run with a 1/2 backed config?

Allan-N (approved):

channels/xpmr/xpmr.c
mkmer: On this - I don't think we need to do anything else as we will  not set one or both of these if we fail to find a valid code:
```
pChan->b.ctcssRxEnable = 1;
pChan->b.ctcssTxEnable = 1;
```

channels/xpmr/xpmr.c
```diff
@@ -294,30 +294,47 @@ i16 code_string_parse(t_pmr_chan *pChan)
...
-				TRACEF(1,("pChan->rxctcss[%i]=%s  pChan->rxCtcssMap[%i]=%i RXONLY\n",i,pChan->rxctcss[i],ri,ti));
+				TRACEF(1, ("pChan->rxctcss[%i]=%s  pChan->rxCtcssMap[%i]=%i RXONLY\n", i, pChan->rxctcss[i], ri, ti));
 			} else {
-				pChan->numrxctcssfreqs=0;
-				for(ii=0;ii<CTCSS_NUM_CODES;ii++) pChan->rxCtcssMap[ii]=CTCSS_NULL;
-				TRACEF(1,("WARNING: Invalid Channel code detected and ignored. %i %s %s \n",i,pChan->pRxCode[i],pChan->pTxCode[i]));
+				pChan->numrxctcssfreqs = 0;
+				for (ii = 0; ii < CTCSS_NUM_CODES; ii++) {
+					pChan->rxCtcssMap[ii] = CTCSS_NULL;
+					ast_log(LOG_ERROR, "ERROR: Invalid CTCSS configuration. TX CTCSS has been disabled\n");
```
Allan-N, line 336: Do we want this log inside the loop?

channels/xpmr/xpmr.c
```diff
@@ -362,7 +379,9 @@ i16 code_string_parse(t_pmr_chan *pChan)
 	{
 		sscanf(p, N_FMT(f), &f);
 		ti = CtcssFreqIndex(f);
-		if (f > maxctcsstxfreq) {
+		if (ti == CTCSS_NULL) {
```
Allan-N, line 382: Can `f` be 0.0?  If so, do we need to factor that into the CtcssFreqIndex() check?

mkmer: I don't think you should configure 0.0 in the default if you don't want ctcss, just don't configure it.
If you're creating a list, then 0.0 would be in the tx list so that rx/tx has pairs (assuming you want a couple of rx with no tx). 
You can also just have a list of rx, no tx at all and this will now work (where before it crashed).

channels/xpmr/xpmr.c
mkmer: That would make sure they see it :)

channels/xpmr/xpmr.c
Allan-N: OK

channels/xpmr/xpmr.c
```diff
@@ -294,30 +294,47 @@ i16 code_string_parse(t_pmr_chan *pChan)
 		{
 			sscanf(p, N_FMT(f), &f);
 			ri = CtcssFreqIndex(f);
-			if (ri > maxctcssindex) {
+			if (ri == CTCSS_NULL) {
+				ast_log(LOG_ERROR, "ERROR: Invalid RX CTCSS code detected and ignored. %i %s\n", i, pChan->pRxCode[i]);
```
InterLinked1, line 298: There's no need to include ERROR in the log message. It's already being sent to the ERROR log level.

InterLinked1 (changes requested):

channels/xpmr/xpmr.c
```diff
@@ -294,30 +295,47 @@ i16 code_string_parse(t_pmr_chan *pChan)
...
+						f = -1.0; /* tone freq not valid */
+						ast_log(LOG_ERROR, "Invalid TX CTCSS code detected and ignored. %i %s\n", i, pChan->pTxCode[i]);
+					}
+				} else if (f > maxctcsstxfreq) {
+					maxctcsstxfreq = f;
+				}
+			} else {
+				ti = CTCSS_NULL;
+				f = -1.0; /* tone freq not provided */
+				ast_log(LOG_ERROR, "Invalid CTCSS configuration. Number of rx codes > number of tx codes\n");
```
InterLinked1, line 319: Can we also log the values?

InterLinked1 (approved):

channels/xpmr/xpmr.c
mkmer: I'm not sure of the value... Which values are you thinking, the count of rx and tx codes?

InterLinked1: > I'm not sure of the value... Which values are you thinking, the count of rx and tx codes?

The values used in the comparison.
Was just a thought, if you don't think it would make sense, feel free to ignore.

Allan-N: The log messages include the problematic `pChan->pRxCode[i]` and `pChan->pTxCode[i]` strings

PR #621, opened 2025-04-10
Update rpt.conf to Remove Uncommon Options

ve7fet:
Remove remote base and DAQ config options that are rarely used (remote base) or currently unsupported (DAQ), and are just adding clutter and confusion to rpt.conf.

ASL manual will document all the available config options and their usage, should users want/need to enable those options.

This cleans up the default template for new users.

### DIFF
--- a/configs/rpt/rpt.conf
+++ b/configs/rpt/rpt.conf
@@ -152,11 +152,6 @@ linkunkeyct = ct8                   ; sent when a transmission received over the
 ;lnkactmacro = *52                  ; Function to execute when link activity timer expires.
 ;lnkacttimerwarn = 30seconds        ; Message to play when the link activity timer has 30 seconds left.
 
-;remote_inact_timeout =             ; Specifies the amount of time without keying from the link. Set to 0 to disable timeout. (15 * 60)
-;remote_timeout =                   ; Session time out for remote base. Set to 0 to disable. (60 * 60)
-;remote_timeout_warning_freq =      ; 30
-;remote_timeout_warning =           ; (3 * 60)
-
 ;nounkeyct = 0                      ; Set to a 1 to eliminate courtesy tones and associated delays.
 
 holdofftelem = 0                    ; Hold off all telemetry when signal is present on receiver or from connected nodes
@@ -181,10 +176,6 @@ parrotmode = 0                      ; 0 = Parrot Off (default = 0)
 
 parrottime = 1000                   ; Set the amount of time in milliseconds
                                     ; to wait before parroting what was received
-
-;rxnotch=1065,40                    ; (Optional) Notch a particular frequency for a specified
-                                    ; b/w. app_rpt must have been compiled with
-                                    ; the notch option
 
 ;startup_macro =                    ; Best use in your node stanza (below) when more than one node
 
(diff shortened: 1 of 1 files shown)


mkmer:
Thinking out loud: would it make sense to group uncommon "stuff" together with a comment about it being uncommon?
OR maybe have a `rpt.conf.all` file that has every option available in it, and a `rpt.conf` with most common?
I suppose that's what the manual is for - once we have it all in there.

Allan-N:
We could also move the uncommon "stuff" into one or more #include files.

InterLinked1:
> Thinking out loud: would it make sense to group uncommon "stuff" together with a comment about it being uncommon? OR maybe have a `rpt.conf.all` file that has every option available in it, and a `rpt.conf` with most common?

The sample file is the one with every option in it.

Allan-N:
... and the plan is to add documentation to the code and then generate the sample files (with everything documented) using https://github.com/InterLinked1/astsamplegen

ve7fet:
> Since there are already comments for these, these should be moved to the sample file (which doesn't exist yet) before being removed

I have commented pretty much every option I have found (so far) in the pending rpt.conf PR that is going into the manual. There are a few "undocumented" ones I still need to track down. I've gone through all the existing rpt.conf examples floating around, and tried to capture all the settings to get them all documented.

My thoughts on this PR is that for the majority of users, there are options in the installed sample rpt.conf that they will never use (remote base stuff) or aren't even currently supported (DAQ... I have an idea on how to support that in the future though). So, for readability purposes, I feel it makes sense to get rid of the clutter, and only include the options that they must have to make a functioning system, or are likely to want to customize.

In addition, some of the stuff, like the [remote-functions], is unreachable. Unless you specifically name that stanza in the node config to pull it in, it is orphaned, and will be ignored.

As I work through building more manual pages, my intent is to include sample code blocks of configs that can be added to the user's rpt.conf to enable those features... but that will be a work in progress, and will take a number of iterations, until everything is in the manual, and all the cross-references are added.

ve7fet:
> We could also move the uncommon "stuff" into one or more #include files.

That's probably a decent way to manage some of the more uncommon features. It would keep some of the clutter out of the main rpt.conf file, and group the related/required settings for the option into one place.

InterLinked1:
> ... and the plan is to add documentation to the code and then generate the sample files (with everything documented) using https://github.com/InterLinked1/astsamplegen

This is still sort of a proof of concept... not sure if the tool is that mature yet.
Either way, the documentation the OP is proposing to remove would need to be documented. Right now, there is no documentation and no sample.

ve7fet:
> > ... and the plan is to add documentation to the code and then generate the sample files (with everything documented) using https://github.com/InterLinked1/astsamplegen
> 
> This is still sort of a proof of concept... not sure if the tool is that mature yet. Either way, the documentation the OP is proposing to remove would need to be documented. Right now, there is no documentation and no sample.

See:
[https://github.com/AllStarLink/ASL3-Manual/pull/109](https://github.com/AllStarLink/ASL3-Manual/pull/109)
[https://github.com/AllStarLink/ASL3-Manual/pull/110](https://github.com/AllStarLink/ASL3-Manual/pull/110)
[https://github.com/AllStarLink/ASL3-Manual/pull/111](https://github.com/AllStarLink/ASL3-Manual/pull/111)

ve7fet:
> I'm talking about XML documentation in the code, not any external documentation that may exist elsewhere. For the purposes of this repo, that doesn't count.

Can you point to this XML documentation you refer to? I must not be looking in the right place.

Regarding sample docs... sure I can write some samples to put in, and add some includes to pull the samples in to rpt.conf.

InterLinked1:
> > I'm talking about XML documentation in the code, not any external documentation that may exist elsewhere. For the purposes of this repo, that doesn't count.
> 
> Can you point to this XML documentation you refer to? I must not be looking in the right place.

That's just it, there isn't any currently for any of the radio modules, it's part of the backlog to add some so that all the options are documented there.
The CLI documentation, auto-generated Asterisk web docs, etc. all stem from the XML documentation.

Here is an example of a module that has documentation: https://github.com/asterisk/asterisk/blob/master/apps/confbridge/conf_config_parser.c#L44

> Regarding sample docs... sure I can write some samples to put in, and add some includes to pull the samples in to rpt.conf.

I think Allan wants to experiment with not working on samples for the moment to see if we can autogenerate them. If somebody wants to do that, that's certainly great, but I think the XML documentation alone will probably be more valuable right now.

Allan-N:
>I think Allan wants to experiment with not working on samples for the moment to see if we can autogenerate them. If somebody wants to do that, that's certainly great, but I think the XML documentation alone will probably be more valuable right now.

Correct.  We can spend our time updating the .conf files, adding .sample files, adding comments into the code, and updating all of the above whenever changes to be needed.  Or, we can focus our efforts.  By adding the XML documentation (into the code) we can use tools to create/update all of the other content.  That also puts the truth in one place.

ve7fet:
> > I think Allan wants to experiment with not working on samples for the moment to see if we can autogenerate them. If somebody wants to do that, that's certainly great, but I think the XML documentation alone will probably be more valuable right now.
> 
> Correct. We can spend our time updating the .conf files, adding .sample files, adding comments into the code, and updating all of the above whenever changes to be needed. Or, we can focus our efforts. By adding the XML documentation (into the code) we can use tools to create/update all of the other content. That also puts the truth in one place.

Ok, I can have a look and see if I can help with some of the XML documentation... once I finish working on the ASL3 manual conversions I'm working on.

In regards to this PR. I've moved the removed settings to sample files, and adjusted rpt.conf to pull in the additional conf files for those features.

InterLinked1:
> > > I think Allan wants to experiment with not working on samples for the moment to see if we can autogenerate them. If somebody wants to do that, that's certainly great, but I think the XML documentation alone will probably be more valuable right now.
> > 
> > 
> > Correct. We can spend our time updating the .conf files, adding .sample files, adding comments into the code, and updating all of the above whenever changes to be needed. Or, we can focus our efforts. By adding the XML documentation (into the code) we can use tools to create/update all of the other content. That also puts the truth in one place.
> 
> Ok, I can have a look and see if I can help with some of the XML documentation... once I finish working on the ASL3 manual conversions I'm working on.
> 
> In regards to this PR. I've moved the removed settings to sample files, and adjusted rpt.conf to pull in the additional conf files for those features.

The sample configs aren't quite right... there should only be one `rpt.conf.sample`, and no "subfiles" for components of that. Samples also usually don't have templates.
If this is just temporary and you plan to remove those / replace with the documentation soon, I'm not as concerned.

Allan-N:
I really don't want to close this PR.  I also don't think it's ready to be merged.

Long term, we should be updating the code with all of the variables, defaults values, descriptions, etc and then use that to generate / update the online documentation, sample files, etc.  But, I don't see us doing that with this PR.

For this PR, I think the question is whether we want to extract out the uncommon (blocks of) options into separate #include/#tryinclude .conf files.

PR #622, opened 2025-04-10
gps.conf, app_gps: fix configuration issues and crashes

Allan-N:
This change addresses two issues:
1. the [general] section should not have been marked as a template
2. adding additional sections resulted in crashes

### DIFF
--- a/apps/app_gps.c
+++ b/apps/app_gps.c
@@ -471,7 +471,8 @@ static int getserialline(int fd, char *str, int max)
  */
 static void *aprs_connection_thread(void *data)
 {
-	char *call, *password, *val, buf[300];
+	char *call, *password, buf[300];
+	const char *val;
 	struct ast_config *cfg = NULL;
 	struct ast_flags zeroflag = { 0 };
 	struct ast_sockaddr addr = { {0,} };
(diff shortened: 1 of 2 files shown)


apps/app_gps.c
```diff
@@ -1721,43 +1714,43 @@ static int load_module(void)
 		ast_log(LOG_NOTICE, "Unable to load config %s\n", config);
 		return AST_MODULE_LOAD_DECLINE;
 	}
-	val = (char *) ast_variable_retrieve(cfg, ctg, "comport");
+	val = (char *) ast_variable_retrieve(cfg, "general", "comport");
```
mkmer, line 1717: Should this be a `#define gps_context "general"` type of thing?

Allan-N: Not a bad idea but this would be the first place in all of app_rpt (or asterisk) to use a `#define`

mkmer: I think we can make `const char *val` and remove the cast.

mkmer (approved):

apps/app_gps.c
Allan-N: Done!

InterLinked1 (approved):
More specifics would in the commit description would have been nice... but I think I was able to follow everything - was the crash the pthread_join on the thread that was previously detached?

apps/app_gps.c
```diff
@@ -615,56 +615,55 @@ static int report_aprs(const char *ctg, const char *lat, const char *lon, const
...
 	if (val) {
 		comment = ast_strdupa(val);
 	} else {
-		comment = ast_strdupa(APRS_DEFAULT_COMMENT);
+		comment = APRS_DEFAULT_COMMENT;
 	}
-	val = (char *) ast_variable_retrieve(cfg, ctg, "power");
+	val = ast_variable_retrieve(cfg, ctg, "power");
 	if (val) {
 		power = (char) strtol(val, NULL, 0);
```
mkmer, line 632: I get it, at some point we need to "stop"... Should we fix this cast too?  Why strtol vs strtoi when we chop it into a char...

Allan-N: Yes, I think it's time to say "stop".  Why are "power", "height", and "gain" char's?  Why are we not doing any error checking on the string to value conversions?  Why?

PR #625, opened 2025-04-13
app_rpt: add "archivedatefmt" variable for archive date string

Allan-N:
Setting the `archivedir` variable enables a simple log and audio recording functionality.  Our documentation claimed that the file names would reflect the date and time down to the 1/100 of a second.  In reality, the names were limited to 1 second.

This change adds a new `archivedatefmt` variable allowing one to specify the date/time formatting.  The default format string remains the same but one can now add additional sub-second digits.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -687,6 +687,20 @@ int function_table_index(const char *s)
 	return -1;
 }
 
+/*! \brief format date string for archive log/file */
+static char *donode_make_datestr(char *buf, size_t bufsize, time_t *timep, const char *datefmt)
+{
+	struct timeval t = ast_tvnow();
+	struct ast_tm tm;
+
+	ast_localtime(&t, &tm, NULL);
+	if (timep) {
+		*timep = t.tv_sec;
+	}
+	ast_strftime(buf, bufsize, datefmt ? datefmt : "%Y%m%d%H%M%S", &tm);
+	return buf;
+}
+
 /*! \brief node logging function */
 void donodelog(struct rpt *myrpt, char *str)
 {
@@ -701,10 +715,9 @@ void donodelog(struct rpt *myrpt, char *str)
 	if (!nodep) {
 		return;
 	}
-	time(&nodep->timestamp);
 	ast_copy_string(nodep->archivedir, myrpt->p.archivedir, sizeof(nodep->archivedir));
-	strftime(datestr, sizeof(datestr) - 1, "%Y%m%d%H%M%S", localtime(&nodep->timestamp));
-	snprintf(nodep->str, sizeof(nodep->str) - 1, "%s %s,%s\n", myrpt->name, datestr, str);
+	donode_make_datestr(datestr, sizeof(datestr), &nodep->timestamp, myrpt->p.archivedatefmt);
+	snprintf(nodep->str, sizeof(nodep->str), "%s %s,%s\n", myrpt->name, datestr, str);
 	ast_mutex_lock(&nodeloglock);
 	insque((struct qelem *) nodep, (struct qelem *) nodelog.prev);
 	ast_mutex_unlock(&nodeloglock);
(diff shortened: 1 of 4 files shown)


configs/rpt/rpt.conf
```diff
@@ -215,14 +215,23 @@ parrottime = 1000                   ; Set the amount of time in milliseconds
...
-; The "archivedir" and "archiveformat" lines can be enabled here (affecting
-; all nodes) or in the per-node stanzas (for recording of individual nodes).
+; The "archivedatefmt" line can be used to specify the filename used for
+; the recordings.  By default, the files will use a "%Y%m%d%H%M%S" date
+; format that maps to YYYYMMDDHHMMSS.  If desired, you can specify any
+; date format supported by the C function strftime(3) API.  For more
+; precision, you can also include the '%[n]q' format to add fractions
+; of a second, with leading zeros.
+;
+; The "archivedir", "archiveformat", and "archivedatestr" lines can be
```
mkmer, line 225: "archivedatestr"?  I don't see where we try to load this in the config code...

apps/app_rpt.c
```diff
@@ -701,10 +715,9 @@ void donodelog(struct rpt *myrpt, char *str)
 	if (!nodep) {
 		return;
 	}
-	time(&nodep->timestamp);
 	ast_copy_string(nodep->archivedir, myrpt->p.archivedir, sizeof(nodep->archivedir));
-	strftime(datestr, sizeof(datestr) - 1, "%Y%m%d%H%M%S", localtime(&nodep->timestamp));
-	snprintf(nodep->str, sizeof(nodep->str) - 1, "%s %s,%s\n", myrpt->name, datestr, str);
+	donode_make_datestr(datestr, sizeof(datestr), &nodep->timestamp, myrpt->p.archivedatefmt);
```
mkmer, line 719: There is only 1 place we use `nodep->timestamp()` -> I think we can eliminate the entire variable as well as the input (`timep`) to this function and get the date around ln 5684 with this function...

configs/rpt/rpt.conf
Allan-N: Typo.  This should have been "archivedatefmt"

apps/app_rpt.c
Allan-N: Calls to `donodelog()` capture the timestamp in the queued message.  When the message is pulled off of the queue the same timestamp is used to get the message written to the correct YYYYYMMDD.txt file.

mkmer: It doesn't look like we extract the value in the queue, rather we use the current value of `nodep->timestamp`.  If we loaded 10 messages into the queue, the last call of `donodelog()` will have updated the pointer and we will put the old stuff in the new log (at a boundary).
So - I still think we can use the function there and eliminate the pointer,  "OR" fix the way we write to the file by extracting the time from the `remque` messages to match.

Allan-N: Restating :

Each call to `donodelog()` allocates a `struct nodelog`, has the log's timestamp set (with the call to `dnoode_make_datestr()`, and is inserted into the `nodelog` queue :
https://github.com/AllStarLink/app_rpt/blob/420baf558bd062fd7615c90c32b1be9737ef8807/apps/app_rpt.c#L714-L723

Then, when we pull the logs off the queue we use the timestamp of the message to form the filename (using just the date) :
https://github.com/AllStarLink/app_rpt/blob/420baf558bd062fd7615c90c32b1be9737ef8807/apps/app_rpt.c#L5674-L5708

This way, a message that was captured at 20250412235959 won't end up being appended to the 20250413.txt file

mkmer: Sure does - I (as usual) missed something obvious :)

mkmer (approved):

apps/app_rpt.c
Allan-N: Took me a few minutes to understand why there was a timestamp in `struct nodelog`, where it made sense to set the value, and where we didn't need to bother.  Until I started digging it was NOT obvious.

InterLinked1 (approved):

Issue #130, opened 2023-03-19
rpt_cli: Add tab completion for node names


InterLinked1:
Users should be able to tab complete the RPT node numbers/names.

PR #630, opened 2025-04-15
app_rpt: add cli completion for node names

mkmer:
Fixes issue rpt_cli: Add tab completion for node names #130

### DIFF
--- a/apps/app_rpt/rpt_cli.c
+++ b/apps/app_rpt/rpt_cli.c
@@ -1064,7 +1064,7 @@ static int rpt_do_setvar(int fd, int argc, const char *const *argv)
 	return 0;
 }
 
-static char *rpt_get_nodes(const char *line, const char *word, int pos, int state, int rpos)
+static char *complete_node_list(const char *line, const char *word, int pos, int rpos)
 {
 	int i = 0;
 	int nrpts = rpt_num_rpts();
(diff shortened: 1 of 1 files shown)


apps/app_rpt/rpt_cli.c
```diff
@@ -1064,6 +1064,24 @@ static int rpt_do_setvar(int fd, int argc, const char *const *argv)
 	return 0;
 }
 
+static char *rpt_get_nodes(const char *line, const char *word, int pos, int state, int rpos)
```
Allan-N, line 1067: Might there be a better function name?  rpt_add_cli_matching_nodes() ?

... and it looks like the "state" argument is not used.

apps/app_rpt/rpt_cli.c
```diff
@@ -1198,7 +1216,7 @@ static char *handle_cli_dump(struct ast_cli_entry *e, int cmd, struct ast_cli_ar
 			"	Dumps struct debug info to log\n";
 		return NULL;
 	case CLI_GENERATE:
-		return NULL;
+		return rpt_get_nodes(a->line, a->word, a->pos, a->n, 2);
```
Allan-N, line 1219: Interesting that rpt_get_nodes() ... or whatever the function name might be changed to ... always returns NULL.  Does this suggest that we should split the function call from the return?
```
		rpt_get_nodes(a->line, a->word, a->pos, a->n, 2);
		return NULL;
```

mkmer: Yes, There appears to be "two" ways to use the Generate function - 
Use `ast_cli_completion_add()` and `return NULL;`  or return a `char *` and the calling function will keep calling until you `return NULL;`  -> looks like the latter was "old" behavior, and former is the "new" way.

apps/app_rpt/rpt_cli.c
mkmer: Yes, it morphed from just returning all nodes to returning the match.
Looking around the other asterisk functions doing this, I think `complete_xxx` is more in keeping with the rest.

Allan-N (approved):

InterLinked1 (approved):

PR #632, opened 2025-04-15
app_rpt: report actual "connection time"

Allan-N:
The "connection time" available via AMI and displayed by web apps like Allmon3 has not been correct. This was due to accumulating the time using incremental updates. We now report the time since connection was initiated.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2282,7 +2282,7 @@ static int attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	}
 	*tele++ = 0;
 	l->elaptime = 0;
-	l->connecttime = 0;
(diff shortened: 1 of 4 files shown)


mkmer (approved):

InterLinked1 (approved):

mkmer:
Does this mean ALL of the timers are incorrectly timing???

Allan-N:
>Does this mean ALL of the timers are incorrectly timing???

Yes, it looks like we lose a bit of time each loop.  I'm not sure if switching to a monotonic timer would help or if we need to have all timers capture the start time and then calculate elapsed or find an alternate way to make the increment/decrement model that's prevalent in the code be more reliable.

For this PR, i fixed just the connection time.  We can choose to look at the rest of the timers at a later date.

PR #633, opened 2025-04-16
app_rpt: Use `rpt_pbx_run` in place of `ast_pbx_start`

mkmer:
app_rpt: Create `rpt_pbx_run()` to eliminate race conditions between local and remote hangups
where the pbx could/would hangup the channel before `rpt_call()` could finish.
Use `ast_pbx_run_args()` with `no_hangup_chan=1` leaving full control of the channel to `rpt_call()`.
This addresses crashes related to short/invalid dialplans.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1169,13 +1169,56 @@ static void rpt_filter(struct rpt *myrpt, volatile short *buf, int len)
 
 #endif
 
- /*
-  *
-  * WARNING:  YOU ARE NOW HEADED INTO ONE GIANT MAZE OF SWITCH STATEMENTS THAT DO MOST OF THE WORK FOR
-  *           APP_RPT.  THE MAJORITY OF THIS IS VERY UNDOCUMENTED CODE AND CAN BE VERY HARD TO READ.
-  *           IT IS ALSO PROBABLY THE MOST ERROR PRONE PART OF THE CODE, ESPECIALLY THE PORTIONS
-  *           RELATED TO THREADED OPERATIONS.
-  */
+struct rpt_autopatch {
+	struct rpt *myrpt;
+	struct ast_channel *mychannel;
+	unsigned int pbx_exited:1;
+};
+
+/*!
+ * \brief Create an autopatch specific pbx run thread with no_hangup_chan = 1 arg.
+ * \param data 	Structure of rpt_autopatch
+ */
+static void *rpt_pbx_autopatch_run(void *data)
+{
+	struct rpt_autopatch *autopatch = data;
+	struct rpt *myrpt = autopatch->myrpt;
+	enum ast_pbx_result res;
+	struct ast_pbx_args pbx_args;
+
+	memset(&pbx_args, 0, sizeof(pbx_args));
+	pbx_args.no_hangup_chan = 1;
+
+	res = ast_pbx_run_args(autopatch->mychannel, &pbx_args);
+	if (res) { /* could not start PBX */
+		rpt_mutex_lock(&myrpt->lock);
+		myrpt->callmode = CALLMODE_FAILED;
+		rpt_mutex_unlock(&myrpt->lock);
+	} else if (myrpt->patchfarenddisconnect || (myrpt->p.duplex < 2)) { /* PBX has finished dialplan */
+		ast_debug(1, "callmode=%i, patchfarenddisconnect=%i, duplex=%i\n", myrpt->callmode, myrpt->patchfarenddisconnect, myrpt->p.duplex);
+		rpt_mutex_lock(&myrpt->lock);
+		myrpt->callmode = CALLMODE_DOWN;
+		rpt_mutex_unlock(&myrpt->lock);
+		myrpt->macropatch = 0;
+		if (!myrpt->patchquiet) {
+			rpt_telemetry(myrpt, TERM, NULL);
+		}
+	} else { /* Send congestion until patch is downed by command */
+		rpt_mutex_lock(&myrpt->lock);
+		myrpt->callmode = CALLMODE_FAILED;
+		rpt_mutex_unlock(&myrpt->lock);
+	}
+	autopatch->pbx_exited = 1;
+	return NULL;
+}
+
+/*
+ *
+ * WARNING:  YOU ARE NOW HEADED INTO ONE GIANT MAZE OF SWITCH STATEMENTS THAT DO MOST OF THE WORK FOR
+ *           APP_RPT.  THE MAJORITY OF THIS IS VERY UNDOCUMENTED CODE AND CAN BE VERY HARD TO READ.
+ *           IT IS ALSO PROBABLY THE MOST ERROR PRONE PART OF THE CODE, ESPECIALLY THE PORTIONS
+ *           RELATED TO THREADED OPERATIONS.
+ */
 
 /*
  *  This is the main entry point from the Asterisk call handler to app_rpt when a new "call" is detected and passed off
@@ -1190,6 +1233,8 @@ void *rpt_call(void *this)
 	int stopped, congstarted, dialtimer, lastcidx, aborted, sentpatchconnect;
 	struct ast_channel *mychannel, *genchannel;
 	struct ast_format_cap *cap;
+	struct rpt_autopatch *patch_thread_data;
+	pthread_t threadid;
 
 	cap = ast_format_cap_alloc(AST_FORMAT_CAP_FLAG_DEFAULT);
 	if (!cap) {
@@ -1256,14 +1301,6 @@ void *rpt_call(void *this)
 
 	/* Reverse engineering of callmode Mar 2023 NA
 	 * XXX These should be converted to enums once we're sure about these, for programmer sanity.
-	 *
-	 * If 1 or 4, then we wait.
-	 * 0 = abort this call
-	 * 1 = no auto patch extension
-	 * 2 = auto patch extension exists
-	 * 3 = ?
-	 * 4 = congestion
-	 *
 	 * We wait up to patchdialtime for digits to be received.
 	 * If there's no auto patch extension, then we'll wait for PATCH_DIALPLAN_TIMEOUT ms and then play an announcement.
 	 */
@@ -1330,7 +1367,6 @@ void *rpt_call(void *this)
 		ast_hangup(mychannel);
 		ast_hangup(genchannel);
 		rpt_mutex_lock(&myrpt->lock);
-		myrpt->callmode = CALLMODE_DOWN;
 		myrpt->macropatch = 0;
 		channel_revert(myrpt);
 		rpt_mutex_unlock(&myrpt->lock);
@@ -1348,7 +1384,6 @@ void *rpt_call(void *this)
 			ast_free(instr);
 		}
 	}
-
 	ast_channel_context_set(mychannel, myrpt->patchcontext);
 	ast_channel_exten_set(mychannel, myrpt->exten);
 
@@ -1356,94 +1391,106 @@ void *rpt_call(void *this)
 		ast_channel_accountcode_set(mychannel, myrpt->p.acctcode);
 	ast_channel_priority_set(mychannel, 1);
 	ast_channel_undefer_dtmf(mychannel);
-	if (ast_pbx_start(mychannel) < 0) {
-		ast_log(LOG_ERROR, "Unable to start PBX!\n");
-		ast_hangup(mychannel);
-		ast_hangup(genchannel);
+	if (rpt_call_bridge_setup(myrpt, mychannel)) {
 		rpt_mutex_lock(&myrpt->lock);
 		myrpt->callmode = CALLMODE_DOWN;
 		rpt_mutex_unlock(&myrpt->lock);
+		ast_hangup(mychannel);
+		ast_hangup(genchannel);
 		pthread_exit(NULL);
 	}
-	usleep(10000);
-	rpt_mutex_lock(&myrpt->lock);
-	myrpt->callmode = CALLMODE_UP;
-
-	if (ast_channel_pbx(mychannel)) {
-		if (rpt_call_bridge_setup(myrpt, mychannel, genchannel)) {
-			myrpt->callmode = CALLMODE_DOWN;
-			rpt_mutex_unlock(&myrpt->lock);
-			pthread_exit(NULL);
-		}
-	} else {
-		/* XXX Can this ever happen (since we exit if ast_pbx_start fails)? */
-		ast_log(LOG_WARNING, "%s has no PBX?\n", ast_channel_name(mychannel));
+	patch_thread_data = ast_calloc(1, sizeof(struct rpt_autopatch));
+	if (!patch_thread_data) {
+		rpt_mutex_lock(&myrpt->lock);
+		myrpt->callmode = CALLMODE_DOWN;
+		rpt_mutex_unlock(&myrpt->lock);
+		ast_hangup(mychannel);
+		ast_hangup(genchannel);
+		pthread_exit(NULL);
+	}
+	patch_thread_data->myrpt = myrpt;
+	patch_thread_data->mychannel = mychannel;
+	res = ast_pthread_create(&threadid, NULL, rpt_pbx_autopatch_run, patch_thread_data);
+	if (res < 0) {
+		ast_log(LOG_ERROR, "Unable to start PBX!\n");
+		rpt_mutex_lock(&myrpt->lock);
+		myrpt->callmode = CALLMODE_DOWN;
+		rpt_mutex_unlock(&myrpt->lock);
+		ast_hangup(mychannel);
+		ast_hangup(genchannel);
+		pthread_exit(NULL);
 	}
 
+	rpt_mutex_lock(&myrpt->lock);
+	myrpt->callmode = CALLMODE_UP;
 	sentpatchconnect = 0;
+	congstarted = 0;
 	while (myrpt->callmode != CALLMODE_DOWN) {
-		if (!ast_channel_pbx(mychannel) && (myrpt->callmode != CALLMODE_FAILED)) {
-			/* If patch is setup for far end disconnect */
-			if (myrpt->patchfarenddisconnect || (myrpt->p.duplex < 2)) {
-				ast_debug(1, "callmode=%i, patchfarenddisconnect=%i, duplex=%i\n",
-					myrpt->callmode, myrpt->patchfarenddisconnect, myrpt->p.duplex);
-				myrpt->callmode = CALLMODE_DOWN;
-				myrpt->macropatch = 0;
-				if (!myrpt->patchquiet) {
-					rpt_mutex_unlock(&myrpt->lock);
-					rpt_telemetry(myrpt, TERM, NULL);
-					rpt_mutex_lock(&myrpt->lock);
-				}
-			} else {			/* Send congestion until patch is downed by command */
-				myrpt->callmode = CALLMODE_FAILED;
-				rpt_mutex_unlock(&myrpt->lock);
-				/* start congestion tone */
-				rpt_play_congestion(genchannel);
-				rpt_mutex_lock(&myrpt->lock);
-			}
+		if (!congstarted && myrpt->callmode == CALLMODE_FAILED) { /* Send congestion until patch is downed by command */
+			rpt_mutex_unlock(&myrpt->lock);
+			/* start congestion tone */
+			rpt_play_congestion(genchannel);
+			rpt_mutex_lock(&myrpt->lock);
+			congstarted = 1;
 		}
-		if (ast_channel_is_bridged(mychannel) && ast_channel_state(mychannel) == AST_STATE_UP)
-			if ((!sentpatchconnect) && myrpt->p.patchconnect && ast_channel_is_bridged(mychannel)
-				&& (ast_channel_state(mychannel) == AST_STATE_UP)) {
+		if (myrpt->callmode != CALLMODE_FAILED) {
+			ast_channel_lock(mychannel);
+			if ((!sentpatchconnect) && myrpt->p.patchconnect && ast_channel_is_bridged(mychannel) &&
+				(ast_channel_state(mychannel) == AST_STATE_UP)) {
+				ast_channel_unlock(mychannel);
 				sentpatchconnect = 1;
 				rpt_mutex_unlock(&myrpt->lock);
-				rpt_telemetry(myrpt, PLAYBACK, (char*) myrpt->p.patchconnect);
+				rpt_telemetry(myrpt, PLAYBACK, (char *) myrpt->p.patchconnect);
 				rpt_mutex_lock(&myrpt->lock);
+			} else {
+				ast_channel_unlock(mychannel);
 			}
-		if (myrpt->mydtmf) {
-			struct ast_frame wf = { AST_FRAME_DTMF, };
+			if (myrpt->mydtmf) {
+				struct ast_frame wf = {
+					AST_FRAME_DTMF,
+				};
 
-			wf.subclass.integer = myrpt->mydtmf;
-			if (ast_channel_is_bridged(mychannel) && ast_channel_state(mychannel) == AST_STATE_UP) {
-				rpt_mutex_unlock(&myrpt->lock);
-				ast_queue_frame(mychannel, &wf);
-				ast_senddigit(genchannel, myrpt->mydtmf, 0);
-				rpt_mutex_lock(&myrpt->lock);
+				wf.subclass.integer = myrpt->mydtmf;
+				ast_channel_lock(mychannel);
+				if (ast_channel_is_bridged(mychannel) && ast_channel_state(mychannel) == AST_STATE_UP) {
+					ast_channel_unlock(mychannel);
+					rpt_mutex_unlock(&myrpt->lock);
+					ast_queue_frame(mychannel, &wf);
+					ast_senddigit(genchannel, myrpt->mydtmf, 0);
+					rpt_mutex_lock(&myrpt->lock);
+				} else {
+					ast_channel_unlock(mychannel);
+				}
 			}
-			myrpt->mydtmf = 0;
 		}
+		myrpt->mydtmf = 0;
 		rpt_mutex_unlock(&myrpt->lock);
 		usleep(MSWAIT * 1000);
 		rpt_mutex_lock(&myrpt->lock);
 	}
-	ast_debug(1, "exit channel loop\n");
+	ast_debug(1, "exit channel loop mode %d\n", myrpt->callmode);
 	rpt_mutex_unlock(&myrpt->lock);
 	rpt_stop_tone(genchannel);
-	if (ast_channel_pbx(mychannel))
+
+	if (!patch_thread_data->pbx_exited) {
 		ast_softhangup(mychannel, AST_SOFTHANGUP_DEV);
+		pthread_join(threadid, NULL);
+	} else {
+		ast_hangup(mychannel);
+	}
 	ast_hangup(genchannel);
+
 	rpt_mutex_lock(&myrpt->lock);
 	myrpt->callmode = CALLMODE_DOWN;
 	myrpt->macropatch = 0;
 	channel_revert(myrpt);
 	rpt_mutex_unlock(&myrpt->lock);
 
-	/* first put the channel on the conference in announce mode */
-	if (myrpt->p.duplex >= 2 && myrpt->p.duplex <= 4) {
+	/* Put pchannel back on the conference in announce mode */
+	if (myrpt->p.duplex == 4 || myrpt->p.duplex == 3) {
 		rpt_conf_add_announcer_monitor(myrpt->pchannel, myrpt);
-	} else {
-		rpt_conf_add_speaker(myrpt->pchannel, myrpt);
 	}
+	ast_free(patch_thread_data);
 	pthread_exit(NULL);
 }
 
@@ -2006,11 +2053,6 @@ static void handle_link_phone_dtmf(struct rpt *myrpt, struct rpt_link *mylink, c
 				rpt_telemetry(myrpt, COMPLETE, NULL);
 				return;
 			}
-#if 0
-			if ((myrpt->rem_dtmfidx < 0) && ((myrpt->callmode == CALLMODE_DIALING) || (myrpt->callmode == CALLMODE_UP))) {
-				myrpt->mydtmf = c;
-			}
-#endif
 		}
 	}
 	if (myrpt->cmdnode[0] && strcmp(myrpt->cmdnode, "aprstt")) {
--- a/apps/app_rpt/rpt_bridging.c
+++ b/apps/app_rpt/rpt_bridging.c
@@ -468,27 +468,25 @@ int __rpt_conf_add(struct ast_channel *chan, struct rpt *myrpt, enum rpt_conf_ty
 	return 0;
 }
 
-int rpt_call_bridge_setup(struct rpt *myrpt, struct ast_channel *mychannel, struct ast_channel *genchannel)
+int rpt_call_bridge_setup(struct rpt *myrpt, struct ast_channel *mychannel)
 {
 	int res;
 
-	/* first put the channel on the conference in announce mode */
-	if (myrpt->p.duplex == 2) {
-		res = rpt_conf_add_announcer_monitor(myrpt->pchannel, myrpt);
-	} else {
-		res = rpt_conf_add_speaker(myrpt->pchannel, myrpt);
-	}
-	if (res) {
-		ast_hangup(mychannel);
-		ast_hangup(genchannel);
-		return -1;
+	/* Put pchannel back on the conference in speaker mode */
+	if (myrpt->p.duplex == 4 || myrpt->p.duplex == 3) {
+		if (rpt_conf_add_speaker(myrpt->pchannel, myrpt)) {
+			return -1;
+		}
 	}
 
 	/* get its conference number */
 	res = dahdi_conf_get_channo(mychannel);
 	if (res < 0) {
 		ast_log(LOG_WARNING, "Unable to get autopatch channel number\n");
-		ast_hangup(mychannel);
+		/* Put pchannel back on the conference in announce mode */
+		if (myrpt->p.duplex == 4 || myrpt->p.duplex == 3) {
+			rpt_conf_add_announcer_monitor(myrpt->pchannel, myrpt);
+		}
 		return -1;
 	}
 
@@ -504,7 +502,10 @@ int rpt_call_bridge_setup(struct rpt *myrpt, struct ast_channel *mychannel, stru
 	 * on a channel number being used as a conference.
 	 */
 	if (dahdi_conf_add(myrpt->voxchannel, res, DAHDI_CONF_MONITOR)) {
-		ast_hangup(mychannel);
+		/* Put pchannel back on the conference in announce mode */
+		if (myrpt->p.duplex == 4 || myrpt->p.duplex == 3) {
+			rpt_conf_add_announcer_monitor(myrpt->pchannel, myrpt);
+		}
 		return -1;
 	}
 	return 0;
--- a/apps/app_rpt/rpt_bridging.h
+++ b/apps/app_rpt/rpt_bridging.h
@@ -96,7 +96,7 @@ int rpt_equate_tx_conf(struct rpt *myrpt);
 #define rpt_tx_conf_add_announcer(chan, myrpt) rpt_conf_add(chan, myrpt, RPT_TXCONF, RPT_CONF_CONFANN)
 
 /*! \note Used in app_rpt.c */
-int rpt_call_bridge_setup(struct rpt *myrpt, struct ast_channel *mychannel, struct ast_channel *genchannel);
+int rpt_call_bridge_setup(struct rpt *myrpt, struct ast_channel *mychannel);
 
 /*! \note Used in app_rpt.c */
 int rpt_mon_setup(struct rpt *myrpt);


apps/app_rpt/rpt_bridging.c
```diff
@@ -488,7 +488,7 @@ int rpt_call_bridge_setup(struct rpt *myrpt, struct ast_channel *mychannel, stru
 	res = dahdi_conf_get_channo(mychannel);
 	if (res < 0) {
 		ast_log(LOG_WARNING, "Unable to get autopatch channel number\n");
-		ast_hangup(mychannel);
+		ast_softhangup(mychannel, AST_SOFTHANGUP_DEV); /* The PBX has control of this channel */
```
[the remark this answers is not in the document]
mkmer: That is how the original code was: https://github.com/AllStarLink/app_rpt/commit/c48ec20f8bc7103cecc66a3fed19084fac0a9075
I'm thinking we should do the "same thing" for all return -1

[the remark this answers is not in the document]
mkmer: Actually - refactoring the "hangups" outside of this function will reduce all of the duplicates - if we `return -1` hangup "stuff"...

apps/app_rpt/rpt_bridging.c
```diff
@@ -504,7 +504,7 @@ int rpt_call_bridge_setup(struct rpt *myrpt, struct ast_channel *mychannel, stru
 	 * on a channel number being used as a conference.
 	 */
 	if (dahdi_conf_add(myrpt->voxchannel, res, DAHDI_CONF_MONITOR)) {
-		ast_hangup(mychannel);
+		ast_softhangup(mychannel, AST_SOFTHANGUP_DEV); /* The PBX has control of this channel */
```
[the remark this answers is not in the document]
mkmer: Ha - yes, I went with this option - much cleaner.

apps/app_rpt.c
```diff
@@ -1416,8 +1416,11 @@ void *rpt_call(void *this)
 	ast_debug(1, "exit channel loop\n");
 	rpt_mutex_unlock(&myrpt->lock);
 	rpt_stop_tone(genchannel);
-	if (ast_channel_pbx(mychannel))
+	if (ast_channel_pbx(mychannel)) {
```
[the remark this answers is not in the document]
mkmer: Looking at the crash data - it is the `softhangup` that's killing us here:
It's interesting that ln 1359 `ast_channel_pbx(mychannel)` is false so we log the message, but then at ln 1423 we get a true test and `softhangup()` crashes.

```
[2025-04-16 08:24:49.071] WARNING[485014]: app_rpt.c:1367 rpt_call:  has no PBX?
[2025-04-16 08:24:49.071] DEBUG[485014]: app_rpt.c:1375 rpt_call: callmode=3, patchfarenddisconnect=1, duplex=2
[2025-04-16 08:24:49.091] DEBUG[485014]: app_rpt.c:1416 rpt_call: exit channel loop
[2025-04-16 08:24:49.091] ERROR[485014]: channel.c:2495 ast_softhangup: FRACK!, Failed assertion bad magic number 0x500a0d7 for object 0x7f9680005120 (0)
[2025-04-16 08:24:49.092] ERROR[485014]:   Got 6 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1d3) [0x564a666336d3]
# 1: /usr/sbin/asterisk(ast_softhangup+0x35) [0x564a66675095]
# 2: /usr/lib/x86_64-linux-gnu/asterisk/modules/app_rpt.so(+0x6b3f) [0x7f96a5206b3f]
# 3: /usr/sbin/asterisk(+0x1d3aad) [0x564a66797aad]
# 4: /lib/x86_64-linux-gnu/libc.so.6(+0x891f5) [0x7f96a88a81f5]
# 5: /lib/x86_64-linux-gnu/libc.so.6(+0x10989c) [0x7f96a892889c]

[2025-04-16 08:24:49.092] ERROR[485014]: channel.c:1038 __ast_queue_frame: FRACK!, Failed assertion bad magic number 0x500a0d7 for object 0x7f9680005120 (0)
[2025-04-16 08:24:49.092] ERROR[485014]:   Got 8 backtrace records
# 0: /usr/sbin/asterisk(__ao2_lock+0x1d3) [0x564a666336d3]
# 1: /usr/sbin/asterisk(+0xa75f4) [0x564a6666b5f4]
# 2: /usr/sbin/asterisk(ast_softhangup_nolock+0x3d) [0x564a66674dfd]
# 3: /usr/sbin/asterisk(ast_softhangup+0x3f) [0x564a6667509f]
# 4: /usr/lib/x86_64-linux-gnu/asterisk/modules/app_rpt.so(+0x6b3f) [0x7f96a5206b3f]
# 5: /usr/sbin/asterisk(+0x1d3aad) [0x564a66797aad]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x891f5) [0x7f96a88a81f5]
# 7: /lib/x86_64-linux-gnu/libc.so.6(+0x10989c) [0x7f96a892889c]

```

Allan-N (approved):

mkmer:
After a bit more research, There is a race between the PBX closing down the channel and the `rpt_call()` code since PBX is in it's own thread.  Since we already "know" that it's dead, just hangup the other genchannel and exit early.

Also, noticed that when things fail, we were not returning the `pchannel` to the correct conference mode... This "could" result in no repeated audio after said failure.

apps/app_rpt/rpt_bridging.c
```diff
@@ -468,28 +468,31 @@ int __rpt_conf_add(struct ast_channel *chan, struct rpt *myrpt, enum rpt_conf_ty
...
 	}
 
 	/* get its conference number */
 	res = dahdi_conf_get_channo(mychannel);
 	if (res < 0) {
 		ast_log(LOG_WARNING, "Unable to get autopatch channel number\n");
-		ast_hangup(mychannel);
-		return -1;
+		/* Restore the conference mode */
+		if (myrpt->p.duplex == 2 || myrpt->p.duplex == 4) { /* p.duplex == 4 -> move pchannel back to repeated audio */
```
Allan-N, line 490: This block of code looks to be doing exactly the same thing as above (l.476...480)

apps/app_rpt/rpt_bridging.c
```diff
@@ -504,7 +507,12 @@ int rpt_call_bridge_setup(struct rpt *myrpt, struct ast_channel *mychannel, stru
 	 * on a channel number being used as a conference.
 	 */
 	if (dahdi_conf_add(myrpt->voxchannel, res, DAHDI_CONF_MONITOR)) {
-		ast_hangup(mychannel);
+		/* Restore the conference mode */
+		if (myrpt->p.duplex == 2 || myrpt->p.duplex == 4) { /* p.duplex == 4 -> move pchannel back to repeated audio */
```
Allan-N, line 511: This block of code looks to be doing exactly the same thing as above (l.476...480)

apps/app_rpt.c
```diff
@@ -1416,8 +1423,9 @@ void *rpt_call(void *this)
 	ast_debug(1, "exit channel loop\n");
 	rpt_mutex_unlock(&myrpt->lock);
 	rpt_stop_tone(genchannel);
-	if (ast_channel_pbx(mychannel))
+	if (mychannel && ast_channel_pbx(mychannel)) {
```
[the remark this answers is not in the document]
mkmer: I'm fairly sure this is the problem - the PBX is in the process of hanging up the channel (above we get a False, then here we get a True.  This could only happen if mychannel moved to say NULL and the pointed to some random spot).  All `ast_channel_pbx` does is return `mychannel->pbx`.  I think I've addressed it above by exiting the routine before we get here.

apps/app_rpt/rpt_bridging.c
mkmer: It is, 
1st -> `return -1 ` is a failure to move the conferences around (aka do nothing)... 
the other `return -1`s are AFTER we moved the channel around and we really need to put it back or there will be no tx audio.
Now that we are exiting on a `return -1`, we don't make it to the end of the routine where this was supposed to happen (not to mention there were other paths that did not handle it and should).

Looking at it holistically, I thing we only need to check ==4 to move it in/out of the conference, but I'm not 100% sure.

apps/app_rpt.c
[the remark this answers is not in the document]
Allan-N: My question is about the mychannel "variable".  How would the variable, the pointer to the channel, change from a non-NULL value to a NULL value?  If what's it's pointing to is no longer valid then we have a bigger issue ... like after we do X we can no longer use the value for anything.

apps/app_rpt/rpt_bridging.c
Allan-N: I'm [still] confused.
- Let's assume that duplex==2.
- On l.477, we call `res = rpt_conf_add_announcer_monitor(myrpt->pchannel,  myrpt);`
- If res != 0 (success) we continue and on l.486 we call `res = dahdi_conf_get_channo(mychannel);`
- if res < 0 (failure) we will on l.491 (again) call `rpt_conf_add_announcer_monitor(myrpt->pchannel, myrpt);`

How does this last call (l.491) that looks like the same as the earlier one (l.477) "Restore the conference mode" ?

apps/app_rpt.c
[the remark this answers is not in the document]
mkmer: I see -> yes it's not the value of mychannel, is what we are pointing at that is freed...
I'm back to "how do we deal with the other thread".

- PBX has the ability to hangup and act on the channel
- App_rpt needs to be able to tell PBX to hangup (softhangup), but there is a race.

I have reduced the opportunity for the race by exiting at ln 1376 (The first time we "know" the pbx is dead), but it's still possible for the PBX to hangup

mkmer:
This is the original core dump driving this PR.
[core-asterisk-2025-04-15T18-01-50Z-info.txt](https://github.com/user-attachments/files/19794936/core-asterisk-2025-04-15T18-01-50Z-info.txt)
[core-asterisk-2025-04-15T18-01-50Z-locks.txt](https://github.com/user-attachments/files/19794937/core-asterisk-2025-04-15T18-01-50Z-locks.txt)
[core-asterisk-2025-04-15T18-01-50Z-thread1.txt](https://github.com/user-attachments/files/19794938/core-asterisk-2025-04-15T18-01-50Z-thread1.txt)
[core-asterisk-2025-04-15T18-01-50Z-brief.txt](https://github.com/user-attachments/files/19794940/core-asterisk-2025-04-15T18-01-50Z-brief.txt)
[core-asterisk-2025-04-15T18-01-50Z-full.txt](https://github.com/user-attachments/files/19794941/core-asterisk-2025-04-15T18-01-50Z-full.txt)

apps/app_rpt.c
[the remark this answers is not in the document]
Allan-N: Random thoughts:

After calling ast_channel_pbx() we see a call to usleep(10000).  Why?  To me, this suggests a possible issue with timing.  Adding to this is the "Can this ever happen" comment if ast_channel_pbx() returns FALSE.  If the call to ast_pbx_start() succeeds then is the expectation that ast_channel_pbx() will always return TRUE?  If so, then why do we keep checking?

The other question I'll ask is whether we can move any of the channel/conference/bridge setup to before the call to ast_pbx_start() ... and if doing so might address any of the multi-threading issues.

apps/app_rpt/rpt_bridging.c
```diff
@@ -468,27 +468,25 @@ int __rpt_conf_add(struct ast_channel *chan, struct rpt *myrpt, enum rpt_conf_ty
...
-	if (myrpt->p.duplex == 2) {
-		res = rpt_conf_add_announcer_monitor(myrpt->pchannel, myrpt);
-	} else {
-		res = rpt_conf_add_speaker(myrpt->pchannel, myrpt);
-	}
-	if (res) {
-		ast_hangup(mychannel);
-		ast_hangup(genchannel);
-		return -1;
+	/* Put pchannel back on the conference in speaker mode */
```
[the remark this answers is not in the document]
mkmer: It's already setup that way when we initialize the system....

apps/app_rpt.c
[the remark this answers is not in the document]
mkmer: I was asking myself the same question regarding the delay.  I have refactored the code setting up the bridge BEFORE starting the PBX and removing the no longer necessary logic.
This appears to "work" and address the crash scenarios around fast hangups, etc.
I would like to see others give it a whirl...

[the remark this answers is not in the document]
mkmer: So - turns out, if we don't wait a bit, the PBX thread has not set `ast_channel_pbx()` to "something non NULL".
There MUST be a better way to know what the PBX thread is doing, waiting some random time is "not cool".
My first thought is to wait for `ast_channel_pbx()` to got TRUE, then let things continue, but we would need a timer in case it never goes TRUE. 
Still hunting for some other way to get the status of the PBX and watch for the thread status.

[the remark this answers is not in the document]
Allan-N: <just-asking-questions>

Q? do we know that `mychannel` is (or will soon be) associated with a PBX?  If so, then do we really need to check that it is?  or can we just call ast_softhangup()?

or do we not know whether the channel is (or will soon be) associated with a PBX and we need to use a different hangup?

or does calling ast_softhangup() crash when it's in transition?

</end-of-questions>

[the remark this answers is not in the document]
InterLinked1: > So - turns out, if we don't wait a bit, the PBX thread has not set `ast_channel_pbx()` to "something non NULL". There MUST be a better way to know what the PBX thread is doing, waiting some random time is "not cool". My first thought is to wait for `ast_channel_pbx()` to got TRUE, then let things continue, but we would need a timer in case it never goes TRUE. Still hunting for some other way to get the status of the PBX and watch for the thread status.

This is because `ast_pbx_start` doesn't set the PBX directly, it launches another detached thread to run the PBX and that thread is the one that sets it.

I think it should be safe to assume that if `ast_pbx_start` returns success, then it is going to end up with a PBX, but I haven't really considered other implications or looked at this change yet in detail.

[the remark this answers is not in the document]
mkmer: Yes we can assume there is a PBX, but it's possible the call thread NEVER sees the value become populated.  Part of the issue I've been chasing here around having a "sort" dialplan - usually it's a BAD dialplan, but either way it ends quickly.
The original code had crash issues.

Right now I'm thinking we need to subscribe to stasis messages on the channel, then watch for a hangup over there, rather than looking at ->pbx in the structure.
At the moment, I'm struggling with the messages returned by the subscription.

apps/app_rpt.c
```diff
@@ -1156,13 +1157,41 @@ static void rpt_filter(struct rpt *myrpt, volatile short *buf, int len)
...
+	struct rpt *myrpt = userdata;
+	if (ast_channel_hangup_request_type() == stasis_message_type(msg)) {
+		ast_debug(1, "Hangup requested");
+		if (myrpt->patchfarenddisconnect || (myrpt->p.duplex < 2)) {
+			ast_debug(
+				1, "callmode=%i, patchfarenddisconnect=%i, duplex=%i\n", myrpt->callmode, myrpt->patchfarenddisconnect, myrpt->p.duplex);
+			myrpt->callmode = CALLMODE_HANGUP;
+			myrpt->macropatch = 0;
+			if (!myrpt->patchquiet) {
+				rpt_mutex_unlock(&myrpt->lock);
```
Allan-N, line 1178: Seems a bit odd that `myrpt` would already be locked when we get the callback

apps/app_rpt.c
```diff
@@ -1343,7 +1372,17 @@ void *rpt_call(void *this)
...
 	ast_channel_priority_set(mychannel, 1);
 	ast_channel_undefer_dtmf(mychannel);
-	if (ast_pbx_start(mychannel) < 0) {
+	subscription = stasis_subscribe(ast_channel_topic(mychannel), channel_callback, myrpt);
+
+	if (rpt_call_bridge_setup(myrpt, mychannel)) {
+		myrpt->callmode = CALLMODE_DOWN;
+		ast_hangup(mychannel);
+		ast_hangup(genchannel);
+		rpt_mutex_unlock(&myrpt->lock);
```
Allan-N, line 1381: Where did we get locked?

Allan-N (approved):

apps/app_rpt.c
```diff
@@ -1156,13 +1157,43 @@ static void rpt_filter(struct rpt *myrpt, volatile short *buf, int len)
...
-  *           RELATED TO THREADED OPERATIONS.
-  */
+/*!
+ * \brief Subscription callback for all channel messages.
+ * \param data Data pointer given when creating the subscription.
+ * \param sub This subscription.
+ * \param message The message itself.
+ */
+
+static void channel_callback(void *userdata, struct stasis_subscription *sub, struct stasis_message *msg)
```
InterLinked1, line 1167: Adding a stasis subscription seems like massive overkill. It seems like there would have to be a more elegant way to do this.

Since the root issue seems to be that we didn't know whether to use ast_hangup or ast_softhangup, why not do the same thing we did in the previous related issue for issue #15 ? If we masquerade the channel, the PBX will never hang it up and then we can always safely call `ast_hangup` when we're done with it. That would eliminate a lot of this complexity.

mkmer: The PBX also needs to hangup the channel  (or our task needs to know)  if the far end hangsup.  Is there a way to handle that?

InterLinked1: If the channel is masqueraded, the PBX won't hang up the channel anymore. That's what the commit e554ae424e653f28ce468571faf05ab1e953f318 did for the `Rpt()` channel.

mkmer: Probably my lack of understanding, but how will our task know the PBX has finished?  PBX is running a dialplan, and the only way I see for it to feedback what it's doing is stasis.
Additionally, there is a race at the beginning where the PBX ends before we can see the shared variable change (from NULL to not NJLL then back to NULL), there's is a race at the end where our app wants to end the call AND the PBX has a hangup (at the "same" time). 
Looking at how the PBX works, it starts and finishes hanging up before the pointer is set to NULL (channel->pbx), so there is a real chance our app tries a `softhangup` in the middle of all of that.

InterLinked1: If the channel is masqueraded before starting the PBX, the PBX wouldn't hang up so you could do a normal hangup instead of a softhangup. Would that help in this situation?

mkmer: How will we deal with the far end hanging up (the person we called hung up on use)?
Right now that's is being delt with via stasis, before it was delt with by watching the `channel->pbx` going null...

InterLinked1: Ah, I see what you're saying... you want the channel to hang up immediately in that case (which makes sense).

In that case, I suppose it probably makes sense to let the PBX do the hangup.
If `ast_pbx_start()` returns success, why do we need to monitor the channel for hangup anyways? The PBX thread should ensure it gets hung up properly, and we can throw away our reference to it. It doesn't seem like we do anything really with `mychannel` after the PBX is started on it. Would that work?

mkmer: >If ast_pbx_start() returns success, why do we need to monitor the channel for hangup anyways? 

We also need to be able to hangup the channel from the state machine controlling the call (aka the near side).
In the "general case" If someone presses `#` on the radio side, we hangup the call.
If some hangs up on the far end, there are 2 paths: 
1- just hangup and return to "normal"
2- play fast busy (congestion) on the `genchannel` until someone on the radio side presses `#`.

The `#` is configurable, but for this discussion I was sticking with the default.

In any of the cases, we still need to hangup the `genchannel` and put the `pchannel` back where it was before the call if we are in duplex mode 4. (Aka there is stuff to do when the call is hung up in any of the cases).

InterLinked1: > > If ast_pbx_start() returns success, why do we need to monitor the channel for hangup anyways?
> 
> We also need to be able to hangup the channel from the state machine controlling the call (aka the near side). In the "general case" If someone presses `#` on the radio side, we hangup the call.

In that case, would it suffice to queue a softhangup for the channel at that point, if it still exists? Rather than storing the channel, you could store the channel name since it'll be unique, and then look it up when needed and hang it up if it's still around.

> If some hangs up on the far end, there are 2 paths: 1- just hangup and return to "normal" 2- play fast busy (congestion) on the `genchannel` until someone on the radio side presses `#`.

I noticed that part as well. What happens after the congestion? Can the channel do anything else after that point?
If not, maybe it would make sense to queue congestion on the channel. That would indicate congestion and the far end would generate its own reorder tone locally. That way, we don't need to keep using the channel for that.

> The `#` is configurable, but for this discussion I was sticking with the default.
> 
> In any of the cases, we still need to hangup the `genchannel` and put the `pchannel` back where it was before the call if we are in duplex mode 4. (Aka there is stuff to do when the call is hung up in any of the cases with some differences between hangup reasons).

But those are 2 other channels, not `mychannel`, right?

mkmer: Hmmm.... Should we just watch for the channel to "disappear" to indicate a far end hangup?  Rather than watching for channel->pbx to go NULL? That maybe cleaner for sure.  Then if we locally hangup, we can softhangup, OR if the channel disappears (goes NULL?) we break out of the loop and do the "other stuff".

On the "congestion" - we simply play the sound on genchannel until someone presses `#` on the radio side - I believe it's to help the people on the radio side "know" the patch hungup.  
Anyway, it plays "forever" until `#` is pressed.  Seems like it should have a timer and give after some amount of time... but that's for a different PR :)

InterLinked1: > Hmmm.... Should we just watch for the channel to "disappear" to indicate a far end hangup? Rather than watching for channel->pbx to go NULL? That maybe cleaner for sure. Then if we locally hangup, we can softhangup, OR if the channel disappears (goes NULL?) we break out of the loop and do the "other stuff".

No, we shouldn't actively be checking that either. But do we need to wait necessarily at that point? I haven't looked at the big picture but if not, just throwing it over to the PBX and forgetting about it might work.

If we can do that, we only need to check once in the case of a local hangup.

> On the "congestion" - we simply play the sound on genchannel until someone presses # on the radio side - I believe it's to help the people on the radio side "know" the patch hungup.
Anyway, it plays "forever" until # is pressed. Seems like it should have a timer and give after some amount of time... but that's for a different PR :)

It may help with this one actually as that could potentially be eliminated and just replaced with indicating congestion rather than playing in band reorder.

mkmer: OK - I've worked up something that is a bit less intrusive, appears to handle all of the scenarios mentioned earlier, and doesn't need stasis.  
Essentially - we just look at the channel rather than the `channel->pbx` -> There is NO race and the PBX can take it's time or go really fast... no matter.

apps/app_rpt.c
```diff
@@ -1196,6 +1197,7 @@ void *rpt_call(void *this)
 	ast_debug(1, "Requested channel %s\n", ast_channel_name(mychannel));
 	rpt_disable_cdr(mychannel);
 	ast_answer(mychannel);
+	mychannel_name = ast_channel_name(mychannel);
```
InterLinked1, line 1200: You would want to make a copy of this, if the channel goes away this memory will no longer be valid
Additionally the channel should be locked when doing that

apps/app_rpt.c
```diff
@@ -1335,75 +1336,74 @@ void *rpt_call(void *this)
...
 		if (myrpt->mydtmf) {
 			struct ast_frame wf = { AST_FRAME_DTMF, };
 
 			wf.subclass.integer = myrpt->mydtmf;
-			if (ast_channel_is_bridged(mychannel) && ast_channel_state(mychannel) == AST_STATE_UP) {
+			if (ast_channel_is_bridged(ast_channel_get_by_name(mychannel_name)) &&
+				ast_channel_state(ast_channel_get_by_name(mychannel_name)) == AST_STATE_UP) {
 				rpt_mutex_unlock(&myrpt->lock);
-				ast_queue_frame(mychannel, &wf);
+				ast_queue_frame(ast_channel_get_by_name(mychannel_name), &wf);
```
InterLinked1, line 1406: Any time you use ast_channel_get_by_name, the channel needs to be unreffed afterwards. You should only call it once in each case as each call returns a new reference
Also, it can return NULL so I don't think you can pass directly to ast_queue_Frame

apps/app_rpt.c
```diff
@@ -1335,75 +1336,74 @@ void *rpt_call(void *this)
...
-		}
-	} else {
-		/* XXX Can this ever happen (since we exit if ast_pbx_start fails)? */
-		ast_log(LOG_WARNING, "%s has no PBX?\n", ast_channel_name(mychannel));
-	}
-
 	sentpatchconnect = 0;
 	while (myrpt->callmode != CALLMODE_DOWN) {
-		if (!ast_channel_pbx(mychannel) && (myrpt->callmode != CALLMODE_FAILED)) {
+		if (!ast_channel_get_by_name(mychannel_name) && (myrpt->callmode != CALLMODE_FAILED)) {
```
InterLinked1, line 1369: This is still effectively polling channel existence in a loop, which I don't like.
Why do we need to monitor mychannel again? I don't think I understood that.

mkmer: WE need to know when the far other end hangs up.  We also need to loop here to see if the radio user wants to hang up...

InterLinked1: Why is that, exactly? Is that just to set the call mode variable to DOWN and all that stuff?
In that case, running a callback when the channel hangs up might make sense. I think stasis is overkill for that though, a channel datastore would accomplish as much - possibly also with a framehook maybe but that wouldn't be strictly necessary even - though maybe helpful for relaying the DTMF.

apps/app_rpt.c
mkmer: Do we need to lock the channel when it's still in our "possession"  - we have not shared it with any threads yet?

I notice the comments on `ast_channel_is_bridged()` say we are supposed to lock the channel as well - is that true?

apps/app_rpt.c
mkmer: I feel like we are "stuck" here - if we use `mychannel` - PBX could hang up on it and it's pointing to nowhere, If we use get_by_name, I suppose I could check for NULL then use it and unref it (that would keep it around until we finish trying to queue the frame).
The docs on the function are not forth coming on freeing a ref when using it.

apps/app_rpt.c
mkmer: How do we handle hanging up a channel (say in a conference), where the PBX has created link and we want to hang up from the conference side?  That's essentially what we are doing here.

apps/app_rpt.c
InterLinked1: > Do we need to lock the channel when it's still in our "possession" - we have not shared it with any threads yet?

I'm not entirely sure... other parts of Asterisk still know about the channel, though at the same time it can't disappear as long as that reference to it exists. But I suppose a masquerade could happen which would cause the name to change. So I think so, and it wouldn't hurt anyways.

> I notice the comments on ast_channel_is_bridged() say we are supposed to lock the channel as well - is that true?

If it isn't, then the documentation would be wrong so I would assume so. A good guide would be to look at other usages of the API and see if callers lock.

apps/app_rpt.c
InterLinked1: You call `ast_channel_unref` when done with it. Some functions exist to explicit modify the refcount, but get by name is one of the few that implicitly increments the count. There shouldn't be any need to manually adjust ref count.

A sign that you have a ref leak is the channel is supposedly hung up but still lingers in `core show channels`.

I think the channel driver does this as well or something in the core for the initial reference, and if you are given a channel pointer (reference), then you leave that alone (e.g. dialplan modules).

apps/app_rpt.c
InterLinked1: If the far end disconnects, I think the channel should hang up on its own if the PBX was started on it.
If we want to hang it up, we could search by name and hang it up if it exists.

It's the other cases I think are a little less clear what the best option is, such as knowing exactly when the channel hangs up. That's why leveraging an existing mechanism to run code when the channel disappears probably makes the most sense for that scenario.

There isn't a way to generically do that, there is a patch another dev I know wrote but it isn't merged. But creating a datastore would give you a hook to do that. Many framehook function modules do this.

apps/app_rpt.c
```diff
@@ -1335,101 +1336,112 @@ void *rpt_call(void *this)
...
 		usleep(MSWAIT * 1000);
 		rpt_mutex_lock(&myrpt->lock);
 	}
-	ast_debug(1, "exit channel loop\n");
+	ast_debug(1, "exit channel loop mode %d\n", myrpt->callmode);
 	rpt_mutex_unlock(&myrpt->lock);
 	rpt_stop_tone(genchannel);
-	if (ast_channel_pbx(mychannel))
-		ast_softhangup(mychannel, AST_SOFTHANGUP_DEV);
+	channel_ref = ast_channel_get_by_name(mychannel_name);
```
mkmer, line 1429: I think this is a "best" practice for deciding to `softhangup()` or not.
Move forward with this part?

apps/app_rpt.c
```diff
@@ -1335,101 +1336,112 @@ void *rpt_call(void *this)
...
+		if (!channel_ref && (myrpt->callmode != CALLMODE_FAILED)) {
 			/* If patch is setup for far end disconnect */
 			if (myrpt->patchfarenddisconnect || (myrpt->p.duplex < 2)) {
-				ast_debug(1, "callmode=%i, patchfarenddisconnect=%i, duplex=%i\n",
-					myrpt->callmode, myrpt->patchfarenddisconnect, myrpt->p.duplex);
+				ast_debug(1,
+					"callmode=%i, patchfarenddisconnect=%i, duplex=%i\n",
+					myrpt->callmode,
+					myrpt->patchfarenddisconnect,
+					myrpt->p.duplex);
```
mkmer, line 1377: This is still not what @InterLinked1 wants to see.
We need a callback similar to the `statis` version, just not stasis?
I haven't been able to locate a method for creating a callback for a channel. If you have one handy it would be much appreciated.

apps/app_rpt.c
```diff
@@ -1348,101 +1349,112 @@ void *rpt_call(void *this)
...
-	} else {
-		/* XXX Can this ever happen (since we exit if ast_pbx_start fails)? */
-		ast_log(LOG_WARNING, "%s has no PBX?\n", ast_channel_name(mychannel));
-	}
-
 	sentpatchconnect = 0;
 	while (myrpt->callmode != CALLMODE_DOWN) {
-		if (!ast_channel_pbx(mychannel) && (myrpt->callmode != CALLMODE_FAILED)) {
+		channel_ref = ast_channel_get_by_name(mychannel_name);
+		if (!channel_ref && (myrpt->callmode != CALLMODE_FAILED)) {
```
mkmer, line 1383: This is still not what @InterLinked1 wants to see.
We need a callback similar to the statis version, just not stasis?
I haven't been able to locate a method for creating a callback for a channel. If you have one handy it would be much appreciated.

apps/app_rpt.c
```diff
@@ -1348,101 +1349,112 @@ void *rpt_call(void *this)
...
-	ast_debug(1, "exit channel loop\n");
+	ast_debug(1, "exit channel loop mode %d\n", myrpt->callmode);
 	rpt_mutex_unlock(&myrpt->lock);
 	rpt_stop_tone(genchannel);
-	if (ast_channel_pbx(mychannel))
-		ast_softhangup(mychannel, AST_SOFTHANGUP_DEV);
+	channel_ref = ast_channel_get_by_name(mychannel_name);
+	if (channel_ref) {
+		ast_softhangup(channel_ref, AST_SOFTHANGUP_DEV);
+		ast_channel_unref(channel_ref);
```
mkmer, line 1445: I think this is a "best" practice for deciding to softhangup() or not.
Move forward with this part?

apps/app_rpt.c
mkmer: So MAYBE the answer to this problem is to use ast_pbx_run() in our own thread rather than using ast_pbx_start(), then we can watch for the thread to die in the loop when the pbx is finished? No need for a callback...

InterLinked1: `ast_pbx_run` is synchronous though - wouldn't that prevent you from doing the other stuff in the loop that you're doing now?

apps/app_rpt.c
InterLinked1: This is the idea, but just call it `chan` or something, the object is the channel like normal, but it's implicitly increased the refcount, which is the unref is needed. And after you do that, the pointer may no longer be valid because the channel could go away.

apps/app_rpt.c
mkmer: I'm saying we run `ast_pbx_run` in a thread of our own making, that way when it exits we can do the cleanup.
I think it would cover the far end hangup problem.  Then at the softhangup, we would join the thread and wait for it to die, then finish cleanup.

InterLinked1: That might work... you could also set the `no_hangup_chan` flag if that would help... for example: https://github.com/asterisk/asterisk/blob/8924258639ffa919b4b78908f161756814e2bd06/main/message.c#L790

apps/app_rpt.c
```diff
@@ -1168,14 +1173,41 @@ static void rpt_filter(struct rpt *myrpt, volatile short *buf, int len)
...
+		myrpt->callmode = CALLMODE_DOWN;
+		rpt_mutex_unlock(&myrpt->lock);
+		myrpt->macropatch = 0;
+		if (!myrpt->patchquiet) {
+			rpt_telemetry(myrpt, TERM, NULL);
+		}
+	} else { /* Send congestion until patch is downed by command */
+		myrpt->callmode = CALLMODE_FAILED;
+	}
+	pthread_exit(NULL);
```
InterLinked1, line 1202: Just do return NULL, we try to avoid `pthread_exit` in new code

apps/app_rpt.c
```diff
@@ -1348,102 +1383,102 @@ void *rpt_call(void *this)
...
+		myrpt->callmode = CALLMODE_DOWN;
+		rpt_mutex_unlock(&myrpt->lock);
 		ast_hangup(mychannel);
 		ast_hangup(genchannel);
+		pthread_exit(NULL);
+	}
+	patch_thread_data = ast_malloc(sizeof(struct rpt_autopatch));
+	patch_thread_data->myrpt = myrpt;
+	patch_thread_data->mychannel = mychannel;
+	res = ast_pthread_create_detached(&threadid, NULL, rpt_pbx_run, (void *) patch_thread_data);
```
InterLinked1, line 1404: I thought you said we want to join the thread to see if it's completed. If so, why are we created it detached?
Also, no need to cast to void

apps/app_rpt.c
mkmer: OK, but the `pbx_thread` started by `ast_pbx_start` uses `pthread_exit(NULL)`, then `return NULL` (which it never gets to i suppose).

apps/app_rpt.c
mkmer: I do want to attach when we ask it to hangup, then we know it finished.  Guilty of using the wrong one :)

apps/app_rpt.c
```diff
@@ -1169,13 +1174,41 @@ static void rpt_filter(struct rpt *myrpt, volatile short *buf, int len)
...
+{
+	struct rpt_autopatch *autopatch = data;
+	struct rpt *myrpt = autopatch->myrpt;
+	enum ast_pbx_result res;
+	res = ast_pbx_run(autopatch->mychannel);
+	if (res) { /* could not start PBX */
+		rpt_mutex_lock(&myrpt->lock);
+		myrpt->callmode = CALLMODE_FAILED;
+		rpt_mutex_unlock(&myrpt->lock);
+	} else if (myrpt->patchfarenddisconnect || (myrpt->p.duplex < 2)) { /* PBX has finished dialplan (including hangup)*/
```
InterLinked1, line 1191: space before closing /

apps/app_rpt.c
```diff
@@ -1348,102 +1384,102 @@ void *rpt_call(void *this)
...
-			pthread_exit(NULL);
-		}
-	} else {
-		/* XXX Can this ever happen (since we exit if ast_pbx_start fails)? */
-		ast_log(LOG_WARNING, "%s has no PBX?\n", ast_channel_name(mychannel));
-	}
-
 	sentpatchconnect = 0;
+	congstarted = 0;
+	ast_channel_ref(mychannel);
```
InterLinked1, line 1420: Why are we manually referring the channel here?

mkmer: I left a comment - if we don't add a reference, the channel locks get a FRACK error when the PBX hangsup.
Adding the ref keeps the channel around in zombie state so that the loop (which is in the process of exiting) can finish.

apps/app_rpt.c
```diff
@@ -1348,102 +1384,102 @@ void *rpt_call(void *this)
...
-		ast_log(LOG_ERROR, "Unable to start PBX!\n");
+	if (rpt_call_bridge_setup(myrpt, mychannel)) {
+		rpt_mutex_lock(&myrpt->lock);
+		myrpt->callmode = CALLMODE_DOWN;
+		rpt_mutex_unlock(&myrpt->lock);
 		ast_hangup(mychannel);
 		ast_hangup(genchannel);
+		pthread_exit(NULL);
+	}
+	patch_thread_data = ast_malloc(sizeof(struct rpt_autopatch));
```
InterLinked1, line 1402: This can fail

apps/app_rpt.c
InterLinked1: I don't think this is the correct way to do this. You should set the `no_hangup_chan` flag instead: https://github.com/search?q=repo%3Aasterisk%2Fasterisk%20no_hangup_chan&type=code

InterLinked1 (changes requested):

apps/app_rpt.c
mkmer: As I said in the comment, IF we set no hangup, the PBX does not hangup the PJSIP channel it created for the call.  We have no reference to that channel and can't cleanup it up (that I could find anyway).

mkmer: The "alternate" is to get a channel ref by name, which would add the ref, then I would free unref at the end.  Really no difference in the "behavior"

InterLinked1: Okay, if you've validated that and have a comment in the code so other people can see then I don't have an objection - it's just atypical which is why I was wondering.

InterLinked1 (changes requested):

apps/app_rpt.c
```diff
@@ -1348,102 +1386,119 @@ void *rpt_call(void *this)
...
 	}
-	ast_debug(1, "exit channel loop\n");
+	ast_channel_unref(mychannel);
+	ast_debug(1, "exit channel loop mode %d\n", myrpt->callmode);
 	rpt_mutex_unlock(&myrpt->lock);
 	rpt_stop_tone(genchannel);
-	if (ast_channel_pbx(mychannel))
-		ast_softhangup(mychannel, AST_SOFTHANGUP_DEV);
+
+	chan = ast_channel_get_by_name(mychannel_name);
```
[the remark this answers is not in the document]
mkmer: When we hangup outside of the PBX, we need to softhangup.  We don't really know why we exited the loop, only that its now in `CALLMODE_DOWN` - this is the way to see if the PBX still has it.
I suppose we could create a different exit - so CALLMODE_DOWN = we need to hang up, and say CALLMODE_NOPBX (or whatever) to know not to soft hangup.  Then we would just test for CALLMODE_DOWN here rather than check for a channel...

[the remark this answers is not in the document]
InterLinked1: That's a scenario I'm not sure about - may just need to test and see if it works.

Really, the cleanest thing to do would be set no_hangup_chan and then we could always safely hard hangup afterwards, but you said this didn't hang up the other channel as well for some reason which I don't fully understand why that would happen. Is that the case even when you hard hang up this channel? It seems odd that it would linger... do you have any logs you can share for that case?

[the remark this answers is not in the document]
mkmer: OK - I tested `no_hanup_chan=1` again, and it is working as you expected.  I believe the part "I" missed was the `hangup()` when PBX exited.  It looks like I only had the `softhangup()` when it had not yet exited.

Now - if we call out, and the far end hangsup, we `hangup()`.  If we command a hangup from the repeater, we `softhangup()` and wait for the thread to finish.

apps/app_rpt.c
```diff
@@ -1432,7 +1437,6 @@ void *rpt_call(void *this)
 	 * the channel disappearing.  We can not use `no_hangup_chan=1` PBX arg because
```
InterLinked1, line 1437: I think this comment here is old now?

apps/app_rpt.c
```diff
@@ -1169,13 +1175,49 @@ static void rpt_filter(struct rpt *myrpt, volatile short *buf, int len)
...
+		}
+	} else { /* Send congestion until patch is downed by command */
+		rpt_mutex_lock(&myrpt->lock);
+		myrpt->callmode = CALLMODE_FAILED;
+		rpt_mutex_unlock(&myrpt->lock);
+	}
+	autopatch->pbx_exited = 1;
+	return NULL;
+}
+/*
```
InterLinked1, line 1214: Leave a blank line before this comment

InterLinked1 (commented):
With the latest changes, I think this is looking good now! Just a couple last nitpicks

apps/app_rpt.c
```diff
@@ -1169,13 +1175,50 @@ static void rpt_filter(struct rpt *myrpt, volatile short *buf, int len)
...
-  * WARNING:  YOU ARE NOW HEADED INTO ONE GIANT MAZE OF SWITCH STATEMENTS THAT DO MOST OF THE WORK FOR
-  *           APP_RPT.  THE MAJORITY OF THIS IS VERY UNDOCUMENTED CODE AND CAN BE VERY HARD TO READ.
-  *           IT IS ALSO PROBABLY THE MOST ERROR PRONE PART OF THE CODE, ESPECIALLY THE PORTIONS
-  *           RELATED TO THREADED OPERATIONS.
-  */
+/*!
+ * \brief Create an autopatch specific pbx run thread.
+ * \param data 	Structure of rpt_autopatch
+ */
+static void *rpt_pbx_run(void *data)
```
InterLinked1, line 1182: rpt_pbx_run makes it sound like this is the rpt version of `ast_pbx_run` for everything, but this is autopatch specific so maybe name as such? autopatch_pbx_run?

InterLinked1 (commented):
The commit message feels a bit misleading to me now - rpt_pbx_run didn't exist before and this doesn't mention the hangup issues. I think more on point to what we are actually trying to do would be `app_rpt: Avoid `ast_hangup` on channels currently owned by a PBX thread` or something more along those lines

apps/app_rpt.c
```diff
@@ -1169,13 +1175,50 @@ static void rpt_filter(struct rpt *myrpt, volatile short *buf, int len)
...
+ */
+static void *rpt_pbx_autopatch_run(void *data)
+{
+	struct rpt_autopatch *autopatch = data;
+	struct rpt *myrpt = autopatch->myrpt;
+	enum ast_pbx_result res;
+	struct ast_pbx_args pbx_args;
+
+	memset(&pbx_args, 0, sizeof(pbx_args));
+	pbx_args.no_hangup_chan = 1,
```
InterLinked1, line 1190: should end in a semicolon... I guess technically this is valid C though as the build didn't fail

apps/app_rpt.c
```diff
@@ -406,6 +406,12 @@
 	</application>
  ***/
 
+struct rpt_autopatch {
```
InterLinked1, line 409: Since this struct is only used once, instead of putting it at the top of the file, maybe put it just before rpt_autopatch_pbx_run, that way it's closer to its usage

apps/app_rpt.c
mkmer: That's odd... but true

InterLinked1 (approved):

PR #635, opened 2025-04-18
Update/correct AMI "Response:" and "ActionID:" order

Allan-N:
The ordering of the AMI "Response:" and "ActionID:" lines in a reply to an AMI Action is important.  The "Response" needs to be first.


Allan-N:
This change addresses an issue I was having when issuing AMI `RptStatus` / `XStat` requests [cami](https://github.com/InterLinked1/cami)

InterLinked1 (approved):

mkmer (approved):

Issue #626, opened 2025-04-14
Connection time not reported correctly


Allan-N:
As reported on Community : [Allmon3 Conn time column, times are way off](https://community.allstarlink.org/t/allmon3-conn-time-column-times-are-way-off/23140)

PR #636, opened 2025-04-18
app_rpt.c: Account for the residual microseconds when calculating `elap` time

mkmer:


apps/app_rpt.c
```diff
@@ -5236,6 +5236,8 @@ static void *rpt(void *this)
 		/* calculate loop time */
 		looptimenow = ast_tvnow();
 		elap = ast_tvdiff_ms(looptimenow, looptimestart);
+		looptimenow.tv_usec -= (looptimenow.tv_usec - looptimestart.tv_usec) - elap * 1000; /* Put the residual time back on now,
```
Allan-N, line 5239: I'm curious as to why you are updating `looptimenow` vs. just adding the residual into `looptimestart` (below)

Allan-N: Also, the same change should be made @ https://github.com/AllStarLink/app_rpt/blob/0372d0baf0eca29b430baeefe95f0a9c5f6bb7d4/apps/app_rpt.c#L7093-L7095

Allan-N (changes requested):

apps/app_rpt.c
mkmer: 6 to one, 1/2 dozen to other ;)  Actually didn't think we really needed another variable, BUT a variable called "residual" would probably eliminate the need for a comment.

Another "way" to solve this would be a function that puts the residual back on the pointer we passed in.  Since we do it in two places, it may be a better solution, albeit a bit strange changing "now".

Allan-N (approved):

mkmer:
Related to issue #626

PR #640, opened 2025-04-19
app_rpt: Remove duplicated code, narrow variable scope.

InterLinked1:
* Remove duplicated code by leveraging existing function.
* Reduce the scope of who and a few other variables where possible.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2243,11 +2243,37 @@ static int handle_remote_phone_dtmf(struct rpt *myrpt, char c, char *restrict ke
 	return 0;
 }
 
+/*!
+ * \brief Create dial string from parsed channel name
+ * \param[in] s Input string
+ * \param[out] s1
+ * \param[out] buf
+ * \param len
+ * \return s2
+ */
+static char *parse_node_format(char *s, char **restrict s1, char *buf, size_t len)
+{
+	char *s2;
+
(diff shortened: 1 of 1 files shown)


apps/app_rpt.c
```diff
@@ -2243,11 +2243,29 @@ static int handle_remote_phone_dtmf(struct rpt *myrpt, char c, char *restrict ke
 	return 0;
 }
 
+static char *parse_node_format(char *s, char **restrict s1, char *sx, size_t size)
```
Allan-N, line 2246: Since you are moving this function, and because it's not obvious what the function does, can we add some doc about the arguments, what the expected strings passed in would look like and what the passed in strings and buffers will contain on exit?  Askng because it looks like `s`, `s1`, and `sx` can all end up being updated.

Also, it might help if `s`, `s1`, and `sx` had clearer variables names :)

InterLinked1: To be honest, I'm not entirely sure what the function does. I think it parses possibly a channel name and creates a dial string using that, but I'm not 100% sure. Open to differing ideas on that.

Allan-N: I did a bit of poking around and, to me, it looks like the string relates back to the dial plan and the "data" provided as the Rpt() arg.

Allan-N (approved):

mkmer (approved):

PR #644, opened 2025-04-22
app_rpt: Address negative elap times.

mkmer:
It's possible to have 0 ms values, so we should not update the start time unless we actually have 1ms or more. 
Once we have 1ms, we need to keep the residual usec in the count using `ast_tvsub()` rather than "normal" math.
Use of `ast_tvnow()` can result in negative time measurement with system time changes,  `rpt_tvnow()` uses monotonic time to eliminate negative measurements.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3212,7 +3212,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 
 		/* start tracking connect time */
 		if (l->connecttime.tv_sec == 0) {
-			l->connecttime = ast_tvnow();
+			l->connecttime = rpt_tvnow();
 		}
 
 		/* ignore non-timing channels */
@@ -3756,12 +3756,12 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 	} else if (f->frametype == AST_FRAME_DTMF_BEGIN) {
 		mute_frame_helper(myrpt);
 		dtmfed = 1;
-		myrpt->lastdtmftime = ast_tvnow();
+		myrpt->lastdtmftime = rpt_tvnow();
 	} else if (f->frametype == AST_FRAME_DTMF) {
 		int x;
 		char c = (char) f->subclass.integer;	/* get DTMF char */
 		ast_frfree(f);
-		x = ast_tvdiff_ms(ast_tvnow(), myrpt->lastdtmftime);
+		x = ast_tvdiff_ms(rpt_tvnow(), myrpt->lastdtmftime);
 		if ((myrpt->p.litzcmd) && (x >= myrpt->p.litztime) && strchr(myrpt->p.litzchar, c)) {
 			ast_debug(1, "Doing litz command %s on node %s\n", myrpt->p.litzcmd, myrpt->name);
 			macro_append(myrpt, myrpt->p.litzcmd);
@@ -4231,7 +4231,7 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 		}
 		rpt_mutex_unlock(&myrpt->lock);
 
-		now = ast_tvnow();
+		now = rpt_tvnow();
 		if ((who == l->chan) || (!l->lastlinktv.tv_sec) || (ast_tvdiff_ms(now, l->lastlinktv) >= 19)) {
 			char mycalltx;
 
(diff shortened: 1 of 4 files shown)


apps/app_rpt.c
```diff
@@ -5280,8 +5280,14 @@ static void *rpt(void *this)
 		/* calculate loop time */
 		looptimenow = ast_tvnow();
 		elap = ast_tvdiff_ms(looptimenow, looptimestart);
-		looptimenow.tv_usec -= (looptimenow.tv_usec - looptimestart.tv_usec) - elap * 1000; /* Put the residual time back on now,
-																							   eliminating accumulated error */
+		if (elap < 0) {
```
Allan-N, line 5283: How can elap ever be < 0?   Minimally, the second call to ast_tvnow() would return the same time as the previous call.

apps/app_rpt.c
```diff
@@ -5280,8 +5280,14 @@ static void *rpt(void *this)
...
+			elap = 0;
+		}
+		if (elap > 0) {
+			looptimenow.tv_usec -= (looptimenow.tv_usec - looptimestart.tv_usec) - elap * 1000; /* Put the residual time back on now,
```
Allan-N, line 5287: I think this is where your calculation is off.  Yes, it may be possible for tv_usec to be less than zero but in that case you need to borrow 1 from tv_sec and adjust.

Allan-N (changes requested):

apps/app_rpt.c
mkmer: As I was out riding I realized what was actually going on here -> any sub millisecond call would just ignore the value and reset the start.  I added a fix the worked as long as everything was > 1 ms.  Simply not updating the start unless we have at least 1ms will fix the missing time.

Allan-N (approved):

mkmer:
So - the math was "really bad" - the usec move into sec, so we can't just subtract from the usec part to gain the time back.
For now, only updating the start time IF we have a whole ms is "closer", but still loosing some of the truncated time.

Allan-N (changes requested):
Yes, math is sometimes hard.

With this change, as it is, we are back to losing the residual.  But, we have all of the #'s (end time, start time, elapsed in ms), we know how to convert between milliseconds and microseconds, and we have functions to add/subtract timeval's (`ast_tvadd()`, `ast_tvsub()`).  What are we missing?

One thing that might help would be to switch from `ast_tvdiff_ms()` to `ast_tvdiff_us()`

mkmer:
> Yes, math is sometimes hard.
> 
> With this change, as it is, we are back to losing the residual. But, we have all of the #'s (end time, start time, elapsed in ms), we know how to convert between milliseconds and microseconds, and we have functions to add/subtract timeval's (`ast_tvadd()`, `ast_tvsub()`). What are we missing?
> 
> One thing that might help would be to switch from `ast_tvdiff_ms()` to `ast_tvdiff_us()`

Yes, we were back to loosing "some" residual time, but not as much.
The math was actually "right", just not using the right "number system".
I found the `ast_tvsub()` and `ast_tvadd()` functions that do what we need for the formula to work correctly. 
The latest commit appears to do the math right (no more negative time), and accounts for residual.

The only "other" way to solve this would be to change the entire timing system to use a fixed "start time" and simply use the difference.  I think that would be a "major" change that isn't really necessary.

apps/app_rpt.c
```diff
@@ -3212,7 +3212,7 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
 
 		/* start tracking connect time */
 		if (l->connecttime.tv_sec == 0) {
-			l->connecttime = ast_tvnow();
+			l->connecttime = rpt_tvnow();
```
Allan-N, line 3215: If you are changing `l->connecttime` to use `rpt_tvnow()` then you also need to update the other places where we compare against this time with `ast_tvdiff_ms()` (2x `rpt_cli.c`, 1x `rpt_manager.c`)

apps/app_rpt.c
```diff
@@ -5278,12 +5278,18 @@ static void *rpt(void *this)
...
 		}
 		/* calculate loop time */
-		looptimenow = ast_tvnow();
+		looptimenow = rpt_tvnow();
 		elap = ast_tvdiff_ms(looptimenow, looptimestart);
-		looptimenow.tv_usec -= (looptimenow.tv_usec - looptimestart.tv_usec) - elap * 1000; /* Put the residual time back on now,
-																							   eliminating accumulated error */
-		looptimestart = looptimenow;
-
+		if (elap < 0) { /* The system time has changed? */
```
Allan-N, line 5283: If we are using the MONOTONIC time then this can never happen (the time is no longer affected by time updates)

Allan-N (changes requested):

apps/app_rpt.c
mkmer: Well - that explains the wacked out time on my dev node.  I didn't intend to to this one, but may as well.

Allan-N (approved):

PR #651, opened 2025-04-24
Correct a few places where pointers were used after the memory free'd

Allan-N:
The finddelim() function takes a string and fills a buffer of pointers back to the string.  As such, we should not be referencing those pointers after the passed in string has been free'd.

Note: for consistency, I also re-arranged a few paired calls to ast_free().


mkmer (approved):
Looks good.

InterLinked1 (approved):

PR #656, opened 2025-04-24
rpt_install.sh: Clean up and simplify script and README.

InterLinked1:
* Use simpler cp commands rather than enumerating all the files we want to copy in the script.
* Remove outdated or old notes in the README that are no longer useful or needed, to make the README more useful to current audiences.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # app_rpt
-Refactoring and upgrade of AllStarLink's app_rpt, etc.
-
-ASL Grant Info: https://www.ampr.org/grants-old/grant-allstarlink-radio-over-ip-roip-app-enhancements-infrastructure-improvement-phase-1/
+**Refactoring and upgrade of AllStarLink's app_rpt, etc.**
 
 # Debugging and Submitting Bugs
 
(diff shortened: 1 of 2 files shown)


Allan-N (commented):
I thought I saw you make changes to `phreaknet.sh` that would copy in the `app_rpt` source files?  Do we still need `rpt_install.sh` ?

InterLinked1:
> I thought I saw you make changes to `phreaknet.sh` that would copy in the `app_rpt` source files? Do we still need `rpt_install.sh` ?

That's true, but I think they are scoped differently. That would technically work fine right now, but with @mkmer's proposed changes of optimizing the CI, it may not make sense to use the PhreakScript option and we would still need to do it ourselves.

mkmer:
Thanks for reminding me about the CI container.  The first pass is just to split up the CI function and create a prebuilt asterisk container that we can then merge and build app_rpt into.  I was NOT planning on changing the way it is done.
If we are going with the build script, we will need to figure out the tests (for which we now have 1).

InterLinked1:
> Thanks for reminding me about the CI container. The first pass is just to split up the CI function and create a prebuilt asterisk container that we can then merge and build app_rpt into. I was NOT planning on changing the way it is done. If we are going with the build script, we will need to figure out the tests (for which we now have 1).

There is a standalone PhreakScript command that can be used to install the test suite, if needed.

PR #657, opened 2025-04-26
Use starter container for ci

mkmer:
Speeds up AI runs to ~2 minutes from original ~10 minutes by only compiling and testing app_rpt and reusing the asterisk build.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,18 +14,9 @@ on:
 jobs:
   container:
     runs-on: ubuntu-latest
-    container: debian:12
+    container: 
+      image: ghcr.io/allstarlink/asl3-ci:latest
     steps:
-      - name: Build DAHDI and Asterisk
-        run: |
-          cd /usr/src
-          apt-get update -y
-          apt-get install -y wget procps
-          sysctl net.ipv6.conf.all.disable_ipv6
-          wget https://docs.phreaknet.org/script/phreaknet.sh
-          chmod +x phreaknet.sh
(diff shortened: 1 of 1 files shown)


mkmer:
For "starters", just go with what we already have for CI, just break it into 2 parts.  
Any changes to base asterisk and/or phreaknet will require a rebuild of the base container.

InterLinked1:
> For "starters", just go with what we already have for CI, just break it into 2 parts. Any changes to base asterisk and/or phreaknet will require a rebuild of the base container.

What is the process for base container rebuilds, if it needs to be done manually? How often is this done automatically, daily?

mkmer:
Right now, we create a release and tag it, container rebuilds.

InterLinked1:
> Right now, we create a release and tag it, container rebuilds.

Maybe a dumb question, but I can't find it... where can I see what that container is actually doing?

mkmer:
> > Right now, we create a release and tag it, container rebuilds.
> 
> Maybe a dumb question, but I can't find it... where can I see what that container is actually doing?

https://github.com/AllStarLink/ASL3-CI_container

I am working on an automation to rebuild daily.  In the end I hope to watch a file or repo and rebuild on a release (which for now would be phreaknet)

InterLinked1 (approved):
Worst case, if we can manually trigger a rebuild if needed on that CI (which I think is the default), that should be good enough. The daily rebuild would be even better. But otherwise I think it looks good enough for now.

Allan-N (approved):

PR #658, opened 2025-04-26
app_rpt: validate length of str in handle_link_data()

mkmer:


apps/app_rpt.c
```diff
@@ -1727,7 +1727,11 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 	}
 	if (*str == 'L') {
 		rpt_mutex_lock(&myrpt->lock);
-		ast_str_set(&mylink->linklist, 0, "%s", str + 2); /* Dropping the "L " of the message */
+		if (strlen(str) > 2) {
+			ast_str_set(&mylink->linklist, 0, "%s", str + 2); /* Dropping the "L " of the message */
+		} else {
+			ast_str_set(&mylink->linklist, 0, " ");
```
InterLinked1, line 1733: This can all be simplified to:

`ast_str_set(&mylink->linklist, 0, "%s", strlen(str) > 2 ? str + 2 : " ");`

InterLinked1 (changes requested):

apps/app_rpt.c
Allan-N: Or, like the 'M' case below, do we check if `strlen(str) < 2` and just return?

mkmer: I went with return, but MAYBE we want to clear the link list when we receive a bad `L` type message... maybe not?

InterLinked1 (approved):
I don't think the blank line that was added is needed, but otherwise looks good

Allan-N (approved):

PR #660, opened 2025-04-26
chan_echolink: fix crash associated with incoming connection

Allan-N:
An incoming EchoLink connection to a node already associated with another node with [many] linked nodes was triggering a crash.  The prime suspect for the crash was the el_text() function that was iterating over the linked nodes and generating text to be passed to connected EchoLink nodes.  This code was hard to read, had some fixed-sized buffers, and did a lot of memory copying.  The code has been reworked and appears to have resolved the crashes.

A handful of other changes, mostly for readability, are included.

### DIFF
--- a/channels/chan_echolink.c
+++ b/channels/chan_echolink.c
@@ -488,7 +488,7 @@ struct el_pvt {
 	struct ast_module_user *u;
 	struct ast_trans_pvt *xpath;
 	unsigned int nodenum;
-	char *linkstr;
+	struct ast_str *linkstr;
 	ast_mutex_t lock;
 };
 
@@ -1388,11 +1388,11 @@ static int el_call(struct ast_channel *ast, const char *dest, int timeout)
  */
 static void el_destroy(struct el_pvt *pvt)
 {
-	struct qelem *qpast;
+	struct el_rxqast *qpast;
 
 	ast_mutex_lock(&pvt->lock);
 	while (pvt->rxqast.qe_forw != &pvt->rxqast) {
-		qpast = (struct qelem *) pvt->rxqast.qe_forw;
+		qpast = pvt->rxqast.qe_forw;
 		remque(qpast);
 		ast_free(qpast);
 	}
@@ -1666,9 +1666,9 @@ static int el_text(struct ast_channel *ast, const char *text)
 	struct el_pvt *pvt = ast_channel_tech_pvt(ast);
 	const char *delim = " ";
 	char *cmd, *arg1, *arg4;
-	char *ptr,*saveptr, *cp, *pkt;
+	char *ptr, *saveptr, *cp;
 	char buf[5120], str[200], *strs[MAXLINKSTRS];
-	int i, j, k, x, pkt_len, pkt_actual_len;
+	int i, j, k, x;
 
 	/* see if we are receiving a link text message */
 	if (pvt->instp && (text[0] == 'L')) {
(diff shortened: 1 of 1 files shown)


channels/chan_echolink.c
```diff
@@ -1692,59 +1692,55 @@ static int el_text(struct ast_channel *ast, const char *text)
...
+			k = 0; /* start of current line in buffer */
 			for (x = 0; x < i; x++) {
+				char mode = *strs[x];
+				char *node = strs[x] + 1;
+
 				/* Process AllStar node numbers - skip over those that begin with '3' which are echolink */
-				if ((*(strs[x] + 1) != '3')) {
-					if (strlen(pkt + k) >= 32) {
-						strncat(pkt, "\r    ", pkt_len);
+				if ((*node != '3')) {
```
InterLinked1, line 1708: Extra parentheses not needed

channels/chan_echolink.c
```diff
@@ -1692,59 +1692,55 @@ static int el_text(struct ast_channel *ast, const char *text)
...
+			k = ast_str_strlen(pkt); /* start of current line in buffer */
 			for (x = 0; x < i; x++) {
+				char mode = *strs[x];
+				char *node = strs[x] + 1;
+
 				/* Process echolink node numbers - they start with 3 */
-				if (*(strs[x] + 1) == '3') {
-					if (strlen(pkt + k) >= 32) {
-						strncat(pkt, "\r    ", pkt_len);
+				if (*node == '3') {
```
InterLinked1, line 1729: This '3' seems like a "magic character" to me, should it be a define?

channels/chan_echolink.c
```diff
@@ -1928,10 +1925,10 @@ static void lookup_node_callsign(const void *nodep, const VISIT which, void *clo
  */
 static void send_info(const void *nodep, const VISIT which, const int depth)
 {
+	struct el_node *node = *(struct el_node **) nodep;
```
InterLinked1, line 1928: No casts from void

channels/chan_echolink.c
```diff
@@ -1969,20 +1966,21 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
  */
 static void send_heartbeat(const void *nodep, const VISIT which, const int depth)
 {
+	struct el_node *node = *(struct el_node **) nodep;
```
InterLinked1, line 1969: No casts from void

InterLinked1 (commented):
Does this resolve that issue you mentioned? Should it be linked to close it?

channels/chan_echolink.c
Allan-N: Fixed

channels/chan_echolink.c
Allan-N: To reference an EchoLink node you take it's 6-digit [EL] node number (with any needed leading zeros) and prefix that with "3".  So, to connect to EL node "9999" you would link with node "3009999".

channels/chan_echolink.c
Allan-N: The function prototype for send_info() needs to match what's expected by `twalk()`.  But, for these calls, what you get is a (struct el_node **).  It's much cleaner and readable to perform the cast once and use properly type'd variables.

But, if I'm missing something here then please explain.

InterLinked1: Yup, you're right - I missed the leading * there. In that case it's fine since it saves a line

channels/chan_echolink.c
InterLinked1: From a code perspective, I'd still say that's a magic number to me :)

InterLinked1 (approved):

channels/chan_echolink.c
mkmer: There are a BUNCH of these magic "characters" around.  At some point we need to go through the string compare and character matches and add #defines for what they "really" mean.
If one peaks at the call code, you will find more of these... Maybe it's worth creating an issue so we don't "forget"

mkmer (approved):

channels/chan_echolink.c
Allan-N: Might also think about adding helper functions like `is_node_echolink()`.

InterLinked1: I agree, but those could likely be macros like `IS_ECHOLINK_CHANNEL()` etc.
We could use those whenever we currently manually parse channel names, it would also make the code a lot easier to read in those places.

okcrum:
I just got this while running this patch. No errors or warnings from asterisk.

[core-asterisk-2025-04-27T22-32-22Z-thread1.txt](https://github.com/user-attachments/files/19932129/core-asterisk-2025-04-27T22-32-22Z-thread1.txt)
[core-asterisk-2025-04-27T22-32-22Z-locks.txt](https://github.com/user-attachments/files/19932130/core-asterisk-2025-04-27T22-32-22Z-locks.txt)
[core-asterisk-2025-04-27T22-32-22Z-info.txt](https://github.com/user-attachments/files/19932131/core-asterisk-2025-04-27T22-32-22Z-info.txt)
[core-asterisk-2025-04-27T22-32-22Z-full.txt](https://github.com/user-attachments/files/19932132/core-asterisk-2025-04-27T22-32-22Z-full.txt)
[core-asterisk-2025-04-27T22-32-22Z-brief.txt](https://github.com/user-attachments/files/19932133/core-asterisk-2025-04-27T22-32-22Z-brief.txt)

Allan-N:
@okcrum : Can you confirm that this crash occurred when running `asterisk` with `valgrind`?

@mkmer : Can you take a look at the last crash?  To me, it's looking like we have a voice frame, that we are calling `ast_frame_adjust_volume_float(f, fac);`, but the frame data and/or frame sample count must be messed up ... possibly by the change in how memory is initialized.

okcrum:
This crash occurred running standalone. No valgrind used.

> @okcrum : Can you confirm that this crash occurred when running `asterisk` with `valgrind`?
>

mkmer:
```wf = {frametype = AST_FRAME_CNG, subclass = ``` 
I think this is our "bad boy" -> a bit of digging to see "how" that frame breaks this.

mkmer:
I think something is wrong here (but may be confused):
https://github.com/AllStarLink/app_rpt/blob/fe5161f06f69742a9aca18c0ac47aab4ceda961d/channels/chan_echolink.c#L2331-L2337
our buffer is only 97 bytes long, but the frame samples = 160.  I suppose we are "ok" with the buf + AST_FRIENDLY_OFFSET = 161 total bytes, but WHY would we be multiplying a factor across the whole `fr.data`?  

Alternate idea:
Maybe this has something to do with the "first received" flags.  I'll dig into that next

mkmer:
So - YES this is a problem. WE WERE using frame datalength / 2 for the scale, now that we switched to ast_frame_adjust, it uses samples.  Something is wrong in the frame size.
https://github.com/AllStarLink/app_rpt/pull/571/files#diff-e9c04babdfa10caa4961bbcc4d26b4d2b254adcd13d5c6d3a40dfa342f455653L4184-L4201

I suspect the `samples` is "wrong", but `datalen/2` doesn't seem right either (only 15 of the 33 bytes?)

KB4MDD:
Keep in mind that the GSM frame should be translated to PCM as it arrives to app_rpt.

Allan-N:
>Does this resolve that issue you mentioned? Should it be linked to close it?

This PR resolves ONE of the issues we have found. But, since it's not resolving ALL of the issues I'm unlinking and we can/will close the parent issue when we have fixed all that are currently known.

mkmer:
Thanks Danny, we are receiving these in the "general" case (bit of debug)
`process_link_channels: 2 Frame length 320, Samples 160, Type 2, class slin`

So - somehow we must be getting one where the samples are less than the frame length generating the out of bounds.

PR #661, opened 2025-04-27
chan_tlb: restore answer logic matching echolink updates

mkmer:


InterLinked1 (approved):
This _really_ should have been in the same PR as #650 - please submit related changes in the same PR in the future.

Allan-N (approved):

mkmer:
Maybe it should have been, but I missed the TLB = ECHOLINK when we were fixing echolink.  Really too late change once it's merged.

InterLinked1:
> Maybe it should have been, but I missed the TLB = ECHOLINK when we were fixing echolink. Really too late change once it's merged.

All good, I just mean that it helps organize better if in advance you know the same change is going to be made to two different channel drivers, you can do them both in the same change.

k6ian:
Something broke in this update related to chan_tlb. It will not connect up anymore.

mkmer:
> Something broke in this update related to chan_tlb. It will not connect up anymore.

Please open an issue related to the connection and include log files -> this has been a challenge testing as we don't have a TLB to test with :(  We may need to do a bit of testing with you if that works for you.

PR #663, opened 2025-04-29
chan_echolink, chan_tlb: Fix RTCP packet creation

Allan-N:
The RTCP packet creation functions had a few math problems that resulted in sending a few additonal bytes of data at the end of the packet with some of those bytes being from uninitialized memory. This change corrects these issues and should also resolve a few valgrind errors that were reported.

### DIFF
--- a/channels/chan_echolink.c
+++ b/channels/chan_echolink.c
@@ -234,13 +234,7 @@ do not use 127.0.0.1
 		+ 16  		/* ASL & EL nodes #'s */  \
 		+ 80  		/* room for "a" message */ \
 		+ (50 * 8) 	/* room for linked ASL and EL nodes */
-/* 
- * If you want to compile/link this code
- * on "BIG-ENDIAN" platforms, then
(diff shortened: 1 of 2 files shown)


mkmer (approved):
This looks like some high level confusion to me. I see why it had a well hidden mistake...
My review will not be much help here.
I truly hate the Big/Little endian stuff - makes my head hurt.  At least they didn't toss in nibbles with endian flipped around too (Siemens does that).

Allan-N:
>This looks like some high level confusion to me. I see why it had a well hidden mistake...

What caught my eye was checking `(l & 4)` to see if any padding as needed ... and then my head started to spin.

>I truly hate the Big/Little endian stuff - makes my head hurt.

The strange thing is that the code had endian oriented structs but they were not (or barely) used.  To me, using the individual fields greatly improves the readability of the code and minimizes the "we have this value that we need to shift here and merge that in with this other value shifted there and and and".

PR #665, opened 2025-04-29
app_rpt.c: Check frame before adjusting audio

Allan-N:
When monitoring multiple nodes our call to ast_frame_adjust_volume_float() was crashing due to a frame with unexpected data.  Specifically:

```
$xx = { frametype = AST_FRAME_VOICE,
        subclass = {integer = 0, {format = 0x5556853f28, topology = 0x5556853f28}, frame_ending = 0},
        datalen = 0,
        samples = 160,
        mallocd = 1,
        mallocd_hdr_len = 222,
        offset = 64,
        src = 0x0,
        data = {ptr = 0x0, uint32 = 0, pad = "\000\000\000\000\000\000\000"},
        delivery = {tv_sec = 1745930822, tv_usec = 19338},
        frame_list = {next = 0x0},
        flags = 0,
        ts = 0,
        len = 0,
        seqno = 0,
        stream_num = 0
      }
```

For now, we're going to skip the audio adjustment for these frames.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4336,7 +4336,18 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 				if ((myrpt->p.linkmongain != 1.0) && (l->mode != 1) && (l->wouldtx))
 					fac *= myrpt->p.linkmongain;
 				if (fac != 1.0) {
-					ast_frame_adjust_volume_float(f, fac);
+					if (f->data.ptr && (f->samples == f->datalen / 2)) {
+						ast_frame_adjust_volume_float(f, fac);
+					} else {
+						ast_debug(3,
+							"Skip volume adjust on %s, fac = %f, data = %p, datalen = %d, samples = %d, src = %s\n",
+							ast_channel_name(l->chan),
+							fac,
+							f->data.ptr,
+							f->datalen,
+							f->samples,
+							f->src ? f->src : "(nil)");
+					}
 				}
 
 				l->rxlingertimer = RX_LINGER_TIME;
@@ -4497,7 +4508,18 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 					fac = myrpt->p.ttxgain;
 
 				if (fac != 1.0) {
-					ast_frame_adjust_volume_float(f, fac);
+					if (f->data.ptr && (f->samples == f->datalen / 2)) {
+						ast_frame_adjust_volume_float(f, fac);
+					} else {
+						ast_debug(3,
+							"Skip volume adjust on %s, fac = %f, data = %p, datalen = %d, samples = %d, src = %s\n",
+							ast_channel_name(l->chan),
+							fac,
+							f->data.ptr,
+							f->datalen,
+							f->samples,
+							f->src ? f->src : "(nil)");
+					}
 				}
 				/* foop */
 				if (l->chan && (l->lastrx || (!altlink(myrpt, l))) && ((l->link_newkey != RADIO_KEY_NOT_ALLOWED) || l->lasttx || strcasecmp(ast_channel_tech(l->chan)->type, "IAX2"))) {


InterLinked1:
This doesn't really look like a final solution, just a debugging/testing change. Maybe we should run it on a few systems and see what we can learn from that so a proper fix can be written.

Allan-N:
>This doesn't really look like a final solution, just a debugging/testing change.

No, this is not the final solution.  But, it's a way to keep from crashing and [effectively] get us back to where we were before switching over to `ast_frame_adjust_volume_float()`.  What we need to do is find out who handed us the frame with no data but a non-zero sample count.

Also, the problem isn't hard to reproduce.  Here's a debug log I captured with this change:
```
[2025-04-29 11:04:05.865] DEBUG[87587]: app_rpt.c:4258 process_link_channels: Skip volume adjust on IAX2/12.17.28.195:4569-10126, fac = 0.079433, data = (nil), datalen = 0, samples = 160
```

InterLinked1:
> > This doesn't really look like a final solution, just a debugging/testing change.
> 
> No, this is not the final solution. But, it's a way to keep from crashing and [effectively] get us back to where we were before switching over to `ast_frame_adjust_volume_float()`. What we need to do is find out who handed us the frame with no data but a non-zero sample count.

If you are trying to track where it came from, there should be an assertion on the unexpected path, e.g. `ast_assert()`

> Also, the problem isn't hard to reproduce. Here's a debug log I captured with this change:

Then we should add a test for this as well.

apps/app_rpt.c
```diff
@@ -4497,7 +4507,17 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
...
+						ast_frame_adjust_volume_float(f, fac);
+					} else {
+						ast_debug(3,
+							"Skip volume adjust on %s, fac = %f, data = %p, datalen = %d, samples = %d\n",
+							ast_channel_name(l->chan),
+							fac,
+							f->data.ptr,
+							f->datalen,
+							f->samples);
+					}
```
mkmer, line 4520: Maybe helpful to add `f->src` - I know the example was NULL, but "maybe" something else will show up

InterLinked1:
> This change is restoring the previous behavior where the `app_rpt` version of `volume_adjustment` used `datalen / 2` to work on the frame. The "bad" frame is coming out of an IAX2 channel directly from the asterisk core... not sure why it's creating such a thing.

Interesting, I suppose that's possible, we might also be doing something wrong here or not fully understand why we're getting that. I don't know enough about this myself to suggest anything. We should definitely dig deeper into it.

mkmer:
> > This change is restoring the previous behavior where the `app_rpt` version of `volume_adjustment` used `datalen / 2` to work on the frame. The "bad" frame is coming out of an IAX2 channel directly from the asterisk core... not sure why it's creating such a thing.
> 
> Interesting, I suppose that's possible, we might also be doing something wrong here or not fully understand why we're getting that. I don't know enough about this myself to suggest anything. We should definitely dig deeper into it.

It's odd to me that the frame is partially populated - it's `mallocd`, has a `frametype` set, but has no data pointer and no `datalen`, and no `src`
```
$xx = { frametype = AST_FRAME_VOICE,
        subclass = {integer = 0, {format = 0x5556853f28, topology = 0x5556853f28}, frame_ending = 0},
        datalen = 0,
        samples = 160,
        mallocd = 1,
        mallocd_hdr_len = 222,
        offset = 64,
        src = 0x0,
        data = {ptr = 0x0, uint32 = 0, pad = "\000\000\000\000\000\000\000"},
        delivery = {tv_sec = 1745930822, tv_usec = 19338},
        frame_list = {next = 0x0},
        flags = 0,
        ts = 0,
        len = 0,
        seqno = 0,
        stream_num = 0
```

InterLinked1:
> > > This change is restoring the previous behavior where the `app_rpt` version of `volume_adjustment` used `datalen / 2` to work on the frame. The "bad" frame is coming out of an IAX2 channel directly from the asterisk core... not sure why it's creating such a thing.
> > 
> > 
> > Interesting, I suppose that's possible, we might also be doing something wrong here or not fully understand why we're getting that. I don't know enough about this myself to suggest anything. We should definitely dig deeper into it.
> 
> It's odd to me that the frame is partially populated - it's `mallocd`, has a `frametype` set, and an `offset`, but has no data pointer, no `datalen`, and no `src`
> 
> ```
> $xx = { frametype = AST_FRAME_VOICE,
>         subclass = {integer = 0, {format = 0x5556853f28, topology = 0x5556853f28}, frame_ending = 0},
>         datalen = 0,
>         samples = 160,
>         mallocd = 1,
>         mallocd_hdr_len = 222,
>         offset = 64,
>         src = 0x0,
>         data = {ptr = 0x0, uint32 = 0, pad = "\000\000\000\000\000\000\000"},
>         delivery = {tv_sec = 1745930822, tv_usec = 19338},
>         frame_list = {next = 0x0},
>         flags = 0,
>         ts = 0,
>         len = 0,
>         seqno = 0,
>         stream_num = 0
> ```
> 
> All we do is read a frame from the channel... `f = ast_read(l->chan);`

Where are you seeing this came from an IAX2 channel? The src here seems to be NULL

mkmer:
We know it was on an IAX2 link channel - that's all there was

apps/app_rpt.c
mkmer: Maybe also make them unique in some way (or does debug give us the line number?)

InterLinked1: Yes, it does. No need to manually add line numbers.

InterLinked1:
> We know it was on an IAX2 link channel - that's all there was. Connect to 2 remote links in listen only mode.

But this is the channel we are reading the frame on, not necessarily where the frame was created, right?

> The crash so far is always on the `who = l->chan` side of the link. This represents the IAX driver channel.

Since this seems easy to reproduce apparently and we need to analyze further, would one of you be able to help put together a test case for this? I can help with "testifying" it but don't know exactly what the radio module configuration would be

apps/app_rpt.c
Allan-N: >Maybe helpful to add f->src - I know the example was NULL, but "maybe" something else will show up

Added.

Allan-N:
I created https://github.com/AllStarLink/app_rpt/issues/666 for researching the source of the frames and any remediation.

Allan-N:
>We know it was on an IAX2 link channel - that's all there was. Connect to 2 remote links in listen only mode.

... and both channels we are monitoring need to be keyed/sending audio

>Since this seems easy to reproduce apparently and we need to analyze further, would one of you be able to help put together a test case for this? I can help with "testifying" it but don't know exactly what the radio module configuration would be

Certainly, but let's do that in another PR.  This one is just to get us to a release where asterisk does not crash

InterLinked1:
> > We know it was on an IAX2 link channel - that's all there was. Connect to 2 remote links in listen only mode.
> 
> ... and both channels we are monitoring need to be keyed/sending audio
> 
> > Since this seems easy to reproduce apparently and we need to analyze further, would one of you be able to help put together a test case for this? I can help with "testifying" it but don't know exactly what the radio module configuration would be
> 
> Certainly, but let's do that in another PR. This one is just to get us to a release where asterisk does not crash

I guess that's fine, as long as we work on solidifying a test and solution afterwards, since most of this change will end up getting reverted in short order.

I'll also point out, this sort of change would not be accepted by Asterisk, as it's not a solution by itself. Their philosophy tends to be more along the lines of letting things continue to crash absent a proper solution to it that people are satisfied with. I realize that isn't the experience we want ASL users to have though.

Allan-N:
>Even if you don't assert, might be good to log a WARNING instead of a debug message, until we have a permanent solution. Unless this would happen constantly, that would be too much noise.

If this was a one-off frame then a WARNING message would be OK.  Unfortunately, when I had the debug logs enabled ... and when the problem appeared ... I saw bunches of reports.  We don't need/want to flood the log calling attention to something the end-user has no control over.

>Secondly, a \todo comment in the code might be nice, although hopefully not needed if we can address it quickly.

I hope we'll find the source sooner rather than later.

okcrum:
> > We know it was on an IAX2 link channel - that's all there was. Connect to 2 remote links in listen only mode.
> 
> But this is the channel we are reading the frame on, not necessarily where the frame was created, right?
> 
> > The crash so far is always on the `who = l->chan` side of the link. This represents the IAX driver channel.
> 
> Since this seems easy to reproduce apparently and we need to analyze further, would one of you be able to help put together a test case for this? I can help with "testifying" it but don't know exactly what the radio module configuration would be

You can test with a radioless hub.

InterLinked1:
> > > We know it was on an IAX2 link channel - that's all there was. Connect to 2 remote links in listen only mode.
> > 
> > 
> > But this is the channel we are reading the frame on, not necessarily where the frame was created, right?
> > > The crash so far is always on the `who = l->chan` side of the link. This represents the IAX driver channel.
> > 
> > 
> > Since this seems easy to reproduce apparently and we need to analyze further, would one of you be able to help put together a test case for this? I can help with "testifying" it but don't know exactly what the radio module configuration would be
> 
> You can test with a radioless node.
> 
> * configure asl as a hub: `rxchannel = dahdi/pseudo` and `duplex=2` in rpt.conf
> * I use `hangtime = 0` and `telemdefault = 0`; YMMV
> * connect to two or more nodes in Monitor mode
> 
> EDIT: Let me know if you need a "start from scratch" complete test case.

If you're able to provide a "start from scratch" minimal config for it, yes, that would be great!
(I'm not an app_rpt user myself just yet, so you'll have to pardon my ignorance there a little bit)

Allan-N:
>I'm not an app_rpt user myself just yet

Ah, but you will be VERY soon :)

PR #667, opened 2025-04-29
app_rpt: Enumerate link->mode

mkmer:


mkmer:
Apparently the file is named wrong too:
`rpt/tranceive`

InterLinked1 (approved):

InterLinked1:
> Apparently the file is named wrong too: `rpt/tranceive`

Yikes, I guess we need to fix that too, but that's outside the scope of this change

mkmer:
> > Apparently the file is named wrong too: `rpt/tranceive`
> 
> Yikes, I guess we need to fix that too, but that's outside the scope of this change

I thought it "didn't look right", but it was that way everywhere so I went with it ;)
Question everything...

PR #668, opened 2025-04-29
app_rpt: set 1 bit width in rpt_link structure for boolean variables

mkmer:


Allan-N (approved):

apps/app_rpt/app_rpt.h
```diff
@@ -460,9 +460,9 @@ struct rpt_lstat {
 	char peer[MAXPEERSTR];
 	char name[MAXNODESTR];
 	char mode;
-	char outbound;
+	unsigned int outbound:1;
```
InterLinked1, line 463: Unless there's an important reason these variables should be here, it would be ideal if we can clump all the 1-bit variables together, the compiler can pack the struct to take less space.
In C, structs aren't rearranged by the compiler to optimize space, so we need to manually ensure the `unsigned int:1`s are all together.

mkmer: I thought I had - these are the only 2 in this structure.
Are you saying move them up in the struct?

InterLinked1: No, my bad - GitHub cut out the lines inbetween and I assumed it was all in one big struct.

InterLinked1 (approved):

PR #671, opened 2025-05-01
app_rpt: add "archiveaudio" variable to control audio capture

Allan-N:
Adding a new (to ASL3) "archiveaudio" variable that can be used to control whether the both the simple activity log and audio recordings are included in the archive directory.  The variable defaults to "yes" (capture audio) but if set to "no" then only the log will be created (the audio recordings will not be saved).

This added functionality might be used by those who want to keep track of node activity without the added expense (and storage) associated with saving the audio content.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2993,29 +2993,36 @@ static inline int rpt_any_hangups(struct rpt *myrpt)
 
 static inline void log_keyed(struct rpt *myrpt)
 {
-	char mydate[100], myfname[512];
-	const char *myformat;
-
 	if (myrpt->monstream) {
 		ast_closestream(myrpt->monstream);
 	}
 	myrpt->monstream = 0;
 	if (myrpt->p.archivedir) {
-		long blocksleft;
+		int do_archive = 0;
 
-		donode_make_datestr(mydate, sizeof(mydate), NULL, myrpt->p.archivedatefmt);
-		snprintf(myfname, sizeof(myfname), "%s/%s/%s", myrpt->p.archivedir, myrpt->name, mydate);
-		myformat = myrpt->p.archiveformat ? myrpt->p.archiveformat : "wav49";
-		myrpt->monstream =
-			ast_writefile(myfname, myformat, "app_rpt Air Archive", O_CREAT | O_APPEND, 0, 0644);
-		if (myrpt->p.monminblocks) {
-			blocksleft = diskavail(myrpt);
-			if (blocksleft >= myrpt->p.monminblocks) {
-				donodelog(myrpt, "TXKEY,MAIN");
+		if (myrpt->p.archiveaudio) {
+			do_archive = 1;
+			if (myrpt->p.monminblocks) {
+				long blocksleft;
+
+				blocksleft = diskavail(myrpt);
+				if (blocksleft < myrpt->p.monminblocks) {
+					do_archive = 0;
+				}
 			}
-		} else {
-			donodelog(myrpt, "TXKEY,MAIN");
 		}
+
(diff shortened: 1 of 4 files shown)


apps/app_rpt/app_rpt.h
```diff
@@ -669,6 +669,7 @@ struct rpt {
 		struct rpt_xlat inxlat;
 		struct rpt_xlat outxlat;
 		const char *archivedir;
+		unsigned int archiveaudio:1;
```
InterLinked1, line 672: Can we put this with the other 1-bit fields? Otherwise it won't save any space

InterLinked1 (changes requested):
Also, could you prefix `app_rpt: ` onto your commit message?

apps/app_rpt/app_rpt.h
Allan-N: Of course, there are 1-bit fields above AND below.

InterLinked1: Those should probably be "clumped" at some point (separate PR), but we should still avoid adding new clumps of bits.

Allan-N: Moved up.

apps/app_rpt.c
```diff
@@ -2993,29 +2993,36 @@ static inline int rpt_any_hangups(struct rpt *myrpt)
...
-			if (blocksleft >= myrpt->p.monminblocks) {
-				donodelog(myrpt, "TXKEY,MAIN");
+		int do_archive = 0;
+
+		if (myrpt->p.archiveaudio) {
+			do_archive = 1;
+			if (myrpt->p.monminblocks) {
+				long blocksleft;
+
+				blocksleft = diskavail(myrpt);
```
mkmer, line 3008: I looks like we will now continue to `donodelog()` even if we are running out of disk space (just like it was before), but is that really the right thing to do?
Seems like we should not write anything if we have a configure value of disk space use allotted.

mkmer (approved):

apps/app_rpt.c
Allan-N: >I looks like we will now continue to donodelog() even if we are running out of disk space (just like it was before)

>but is that really the right thing to do?
Seems like we should not write anything if we have a configure value of disk space use allotted.

Allan-N: >I looks like we will now continue to donodelog() even if we are running out of disk space (just like it was before)

Correct.

>but is that really the right thing to do?
>Seems like we should not write anything if we have a configure value of disk space use allotted.

I don't know.  On the one hand, if you enable archiving you need to plan on having to deal with the additional storage requirements.  We now (correctly) check the available space before saving the audio capture ... and that's where most of the storage will be consumed.  We could also check the available space before writing the much smaller informational messages.  A related question is how do we indicate when we are not logging/capturing due to insufficient space.  Do we need to log something when we stop archiving?  and when we resume?  and this could be ping-pong'ing as the available space hovers near the threshold.  With the text message logs I think we'd need something to explain any gaps.

mkmer: Seems like either way, we should log something when the disk is too full (1 time) based on the configuration value.  Then it gets tricky - when do we start logging again, etc.
Probably all for another PR? Or is this the "time" to do something about it....

Allan-N: Yup.  I would suggest that changes to monitor the available storage for the status logs, and any logging associated with stopping/restarting are a bit out of scope for this change (adding a control to enable/disable the audio samples). I'd say we look at expanding the storage availability checks (and logging) for another PR.

Issue #670, opened 2025-04-30
Callsign data not being displayed in Echolink user app while connected to Echolink chan_echolink based nodes


k2cb:
Recently transitioned from running Echolink on a dedicated PC, to now using chan_echolink running on an ASL node, which is then connected to a repeater controller. 

When connecting to the Echolink Windows based node, users can see the callsigns of all connected stations.  Now, when connecting to the chan_echlolink based node, users now only see the Echolink node numbers, just like ASL nodes, and not the actual callsigns of the other connected stations.

This is an issue for large repeater systems where during nets etc,  we can have as many as 20 or more stations connected to the Echolink node.  Net control operators usually log into the Echolink node so they can see who is connected real-time.

Since switching over, we have had numerous inquires as to why the users no longer see the callsigns.

Can this functionality be added to chan_echolink?

Eric
K2CB

Allan-N:
This sounds like you are asking for each EL node# to be looked up in the EL database before being added to a list that is shared from the server.

Q? do you have access to a system that is still running EL on a dedicated PC?  If so, could you attach a screen capture showing how an EL node is displayed when connecting to that [PC] system vs. a screen capture showing how the same EL node is displayed to an [ASL] system?

k2cb:
![Image](https://github.com/user-attachments/assets/3539f8a7-562f-43c9-8b82-6ca8c6adfb7c)
![Image](https://github.com/user-attachments/assets/7d2caebc-ceb6-428b-8897-7629fa46ca1e)

k2cb:
Hi Allan,

The first image W2NJR-R is while connected to the AllStar node running chan_echolink, and the second image W2NJR-L is while connected to the Echolink PC.


Thank you.

Eric
K2CB

Allan-N:
>The first image W2NJR-R is while connected to the AllStar node running chan_echolink, and the second image W2NJR-L is while connected to the Echolink PC.

Thank you for the images.  I had an idea (hope) that you were referring to the textual information reporting the linked nodes.  As it turns out, I was just reworked some of that code to fix a few problems.  I will look to see if I can easily (and efficiently) map the EL node # to the EL callsign.

k2cb:
That would be awesome.  
Thank you.

Eric
K2CB

Allan-N:
Note: just to set expectations, this is not at/near the top of my TODO list ... but it's also not at the bottom

Allan-N:
@k2cb

[...]

PR #675, opened 2025-05-05
chan_echolink: Add callsign lookup for echolink nodes

mkmer:
Addressing some of issue #670

### DIFF
--- a/channels/chan_echolink.c
+++ b/channels/chan_echolink.c
@@ -1707,8 +1707,9 @@ static int el_text(struct ast_channel *ast, const char *text)
 						ast_str_append(&pkt, 0, "\r    ");
 						k = ast_str_strlen(pkt) - 4;
 					}
-					if (!j++) {
+					if (!j) {
 						ast_str_append(&pkt, 0, "AllStar:");
+						j = 1;
 					}
 					ast_str_append(&pkt, 0, " %s%s", node, mode == 'T' ? "" : "(M)");
 				}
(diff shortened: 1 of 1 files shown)


channels/chan_echolink.c
```diff
@@ -1729,10 +1731,19 @@ static int el_text(struct ast_channel *ast, const char *text)
 						ast_str_append(&pkt, 0, "\r    ");
 						k = ast_str_strlen(pkt) - 4;
 					}
-					if (!j++) {
+					if (!j) {
 						ast_str_append(&pkt, 0, "Echolink:");
+						j = 1;
+					}
+					node_lookup.nodenum = atoi(node);
```
Allan-N, line 1738: L1738..1741 : I think you want to use `lookup_node_by_nodenum`

Allan-N (changes requested):

channels/chan_echolink.c
mkmer: That will cause us to look up the node beyond our connected nodes list.  Do we need to go that far here? Since these are already connected nodes, it "should" be in the list, if not.... that would be a "huh?" moment (I think)

Allan-N: I was thinking that if we have a node to lookup then we use the our lookup function.  The logic ends up being the same for connected nodes.  And, if for whatever reason, we have a node number that we're not directly connected to then we can pull from the DB (that we already have).

Also, using lookup_node_by_nodenum localizes the number of places we need to manage the locks.

Allan-N (approved):

PR #678, opened 2025-05-07
app_rpt: add telemetry information to archive logs

Allan-N:
The "archivedir" variable is used to enable the simple node activity log (key, unkey, link changes) and, optionally, audio recordings.  Here, we are extending the log content to include telemetry information.

### DIFF
--- a/apps/app_rpt/rpt_telemetry.c
+++ b/apps/app_rpt/rpt_telemetry.c
@@ -395,14 +395,19 @@ void cancel_pfxtone(struct rpt *myrpt)
 
 static void send_tele_link(struct rpt *myrpt, char *cmd)
 {
-	char str[1028];
+	int len;
+	char *str;
 	struct ast_frame wf;
 	struct rpt_link *l;
 
-	snprintf(str, sizeof(str) - 1, "T %s %s", myrpt->name, cmd);
+	len = ast_asprintf(&str, "T %s %s", myrpt->name, cmd);
+	if (len < 0) {
+		return;
+	}
+
 	init_text_frame(&wf, "send_tele_link");
-	wf.datalen = strlen(str) + 1;
 	wf.data.ptr = str;
+	wf.datalen = len + 1;
 	l = myrpt->links.next;
 	/* give it to everyone */
 	while (l != &myrpt->links) {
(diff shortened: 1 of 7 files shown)


mkmer (approved):
LGTM.  These "could" just be debug messages too... I suppose there is value in adding them to the log file for "audio recordings".

Allan-N:
If you'd like, I can look into replacing the `donodelog` calls with a new `telelog` function that would write to the audio archive AND issue a debug message.

mkmer:
Mostly thinking out loud.  IF people don't like these messages in the log OR if it's not enabled, it would be nice to have it in the debug log if it were enabled...

Allan-N:
>>If you'd like, I can look into replacing the donodelog calls with a new telelog function that would write to the audio archive AND issue a debug message.

>Mostly thinking out loud. IF people don't like these messages in the log OR if it's not enabled, it would be nice to have it in the debug log if it were enabled...

First off, I made a few changes to the PR (squashed a warning, created a function for some duplicated code, cleanup).   As far as audio archive logs vs logging, I think we should revisit this later.  My reasoning is that the audio logging is embedded in the code where we would actually "play" telemetry.  This does not include code where we take actions but don't "play" anything.  I think we also need to talk about debug log levels (what do you get by default, what do you have to enable for added logs, etc).

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1125,74 +1182,97 @@ void *rpt_tele_thread(void *this)
...
+		if (res < 0) { /* Then default message */
 			res = ast_streamfile(mychannel, "rpt/callterminated", ast_channel_language(mychannel));
 		}
 		break;
 	case COMPLETE:
 		/* wait a little bit */
-		if (!wait_interval(myrpt, DLY_TELEM, mychannel))
-			res = telem_lookup(myrpt, mychannel, myrpt->name, "functcomplete");
+		if (wait_interval(myrpt, DLY_TELEM, mychannel) == -1) {
+			break;
```
mkmer, line 1232: Should we be setting `res = -1` on interval failures (all locations)?
Seems to be a mixed bag on setting it...
Maybe `res = wait_interval()`?

Allan-N: When adding the `nodelog_fmt()` calls I found the `wait_interval()` usage was not consistent.  That's why you see chunks of rework with a goal of consistency and readability.  With the PR, I tried to ensure that I was not changing the value/usage of the `res` variable as going down that path looks to be a twisty maze.

mkmer: I was "thinking" the lack of `-1` at the breaks "could" be a source of keyups with no audio.  
Looking at it again, I don't think this would change the thread creation either way, but it's possible I'm not looking at it right.

Allan-N: >I was "thinking" the lack of -1 at the breaks "could" be a source of keyups with no audio.

The changes I made should not have changed the keyup behavior.  If there was a broken case before then it should not have changed.  We would still need more insights (maybe the newly added audio logs) into what activity is involved in the errant behavior.

PR #679, opened 2025-05-08
app_rpt: Add default value for "dtmfkeys" variable

Allan-N:


mkmer (approved):
Well, that was easy :)

PR #680, opened 2025-05-08
app_rpt: valgrind reports uninitialized memory usage in mdc_decoder_new

Allan-N:


mkmer (approved):
So it "was" initialized but the "extras" were not? 
LGTM

Allan-N:
>So it "was" initialized but the "extras" were not?

Yes, initialized ... but not fully initialized (->bits, ->synclow, ->synchigh)

PR #682, opened 2025-05-13
chan_tlb: Use ast_queue_frame in channel drivers - not ast_sendtext()

mkmer:


mkmer:
> ast_debug(1, "tlb: Channel %s got DTMF %c\n", ast_channel_name((*found_key)->chan), dchar);

We no longer have `->chan` stored in the structure.  If we wanted to get at this, I think we now need to store the channel name and lookup the pointer via the name.

We could put the debug around l1524 (in `TLB_xwrite`) where we DO have the channel pointer -> debug is a bit further downstream here where other failures could prevent the log message....

That all being said, we "could" re-architect the whole thing to convert `->chan` to `->chan_name`  and just lookup up the channel pointer -> this would roll back TLB to it's "original" design with the addition of the lookup.

Allan-N:
>We no longer have ->chan stored in the structure. If we wanted to get at this, I think we now need to store the channel name and lookup the pointer via the name.

As far as storing the channel name, that sounds like a simple/easy change.

As for re-architecting, I would only switch back if the old way was "better".

mkmer:
> > We no longer have ->chan stored in the structure. If we wanted to get at this, I think we now need to store the channel name and lookup the pointer via the name.
> 
> As far as storing the channel name, that sounds like a simple/easy change.
> 
> As for re-architecting, I would only switch back if the old way was "better".

Something broke with the change.  We "could" return to the original code and get the channel via name and "probably" fix whatever is broken, as it would be "yank" proof and essentially be the original code.

I suppose we should wait for more details on what is broken (It may just be this fix)

PR #686, opened 2025-05-15
app_rpt: rpt_vars[]->ready never set true for remote.  Issue created by PR#532

### LINKED ISSUE #681: ASL3 remote base not working
Trying to make the built in RB work. But it doesn't. Included in node rpt.conf I added:

remote = ft100
ioport = /dev/ttyACM0
iospeed = 4800

I know these configs are correct as I used them in ASL2 remote and also the port and speed work fine on the same PC with hamlib. 

If I change the nodes node line to delete the "Y" at the end of the line and then comment out the remote specific lines notes above this node 2428 can be connected to from another. But when remote base is enabled you cannot connect to the remote node. 

It was suggested to me to open a ticket because one person on the developer list is already running a remote. Maybe she or he can give me pointers

GeorgeC W2DB

PS just to make more certain I;m doing things correctly I swapped drives to a ASL2 installation. Remote worked fine.

mkmer:
Resolves #681

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -5666,13 +5666,14 @@ static void *rpt_master(void *ignore)
 		return NULL;
 	}
 
-	/* start em all */
+	/* start a rpt() thread for each repeater that is not a remote */
 	for (i = 0; i < nrpts; i++) {
 		load_rpt_vars(i, 1); /* Load initial config */
 
-		/* if is a remote, dont start one for it */
+		/* if is a remote, dont start a rpt() thread for it */
 		if (rpt_vars[i].remote) {
-			if (retrieve_memory(&rpt_vars[i], "init")) {	/* Try to retrieve initial memory channel */
+			rpt_vars[i].ready = 1;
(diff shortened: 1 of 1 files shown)


apps/app_rpt.c
```diff
@@ -5672,7 +5672,8 @@ static void *rpt_master(void *ignore)
 
 		/* if is a remote, dont start one for it */
```
Allan-N, line 5673: This comment, especially here, is confusing.  Does the comment really belong "in" (or at least inline with) the `if () { }` block?  and would the `else { }` inline comment be better "in" the `else` block?

Also, what are we not starting?  and how does not starting whatever sync up with marking it as "ready"?

Allan-N (approved):

apps/app_rpt.c
mkmer: "start em all..." that's what were not starting :)  

If we are remote, the repeater threads are NOT started:
https://github.com/AllStarLink/app_rpt/blob/34b33105962433d7403b88d6a3f13700e6efd03c/apps/app_rpt.c#L5707

What should we put in the comments to help clear this up?

PR #688, opened 2025-05-21
app_rpt: Allow tailmessage from dialplan

mkmer:
Adding a context defined by telemetry to the dialplan and an extension `TAIL` allows users to play multiple files and do actions in the dialplan when the `TAILMSG` telemetry case is activated.  
If the context or extension does not exist, previous tailmessage behavior will occur.

Sample Dialplan:
```
[globals]

TAILFILEINDEX=0

[telemetry]
exten => TAIL,1,Set(TAILFILES=custom/tailmsg,/tmp/tailmsg2) ; add , separated tail files to play
	same => n,Set(index=$[${GLOBAL(TAILFILEINDEX)} % ${FIELDQTY(TAILFILES,\,)} + 1])
	same => n,Set(file=${CUT(TAILFILES,\,,${index})})
	same => n,Playback(${file})
	same => n,Set(GLOBAL(TAILFILEINDEX)=$[${TAILFILEINDEX} + 1]) 
	same => n,Hangup() 

```
Example of multiple messages:
```
[globals]

TAILFILEINDEX=0

[telemetry]
; Say message file
exten => TAIL,1,Set(TAILFILES=/tmp/tailmsg,/tmp/tailmsg2) ; add , separated tail files to play
	same => n,Set(index=$[${GLOBAL(TAILFILEINDEX)} % ${FIELDQTY(TAILFILES,\,)} + 1])
	same => n,Set(file=${CUT(TAILFILES,\,,${index})})
	same => n,Playback(${file})
 	;after playing the tail message, play a weather announcement
  	same => n,Playback(/tmp/current-weather)
  	;after playing the weather announcement, play a traffic announcement
  	same => n,Playback(/tmp/current-traffic)
	same => n,Set(GLOBAL(TAILFILEINDEX)=$[${TAILFILEINDEX} + 1]) 
  	same => n,Hangup()
```
This is a bite size version of PR #557 that I think we can build on.

### DIFF
--- a/configs/rpt/extensions.conf
+++ b/configs/rpt/extensions.conf
@@ -7,7 +7,7 @@ writeprotect = yes ; from overwriting the config file. Leave them here.
 HOMENPA = 999 ; change this to your Area Code
 NODE = 1999   ; change this to your node number
 
-TAILFILEINDEX=0
+TAILFILEINDEX=0 ; used for telemetry, do not remove/change
 
 
 [default]
@@ -194,14 +194,18 @@ exten => s,1,Ringing
 
 [telemetry]
 ;  
-; This stanza and extensions are used by "app_rpt" to play a telemetry  
-; file (e.g. tailmessages).  If this extension does not exist, 
-; app_rpt will play the telemetry based on tailmessagelist configuration.
+  
+; This stanza and extensions are used by "app_rpt" to play a telemetry 
+; file (e.g. tailmessages).  
+;  
+; The following TAIL extension provides support for playing messages after  
+; keying.  If this extension does not exist, app_rpt will play the telemetry  
+; based on the rpt.conf "tailmessagelist" variable.  
 ; 
-; The "standard" tail message follows:
-; Modify the global variable TAILFILES to contain the list of tail files separated with ",".
-
-;exten => TAIL,1,Set(TAILFILES=/tmp/tailmsg,/tmp/tailmsg2) ; add , separated tail files to play
+; An equivalent to the rpt.conf "tailmessage" configuration follows:  
+; Modify the TAILFILES value to contain the list of tail files separated with ",".  
+;  
+;exten => TAIL,1,Set(TAILFILES=custom/yourtailmsg,/tmp/yourothertailmsg) ; add , separated tail files to play
 ;	same => n,Set(index=$[${GLOBAL(TAILFILEINDEX)} % ${FIELDQTY(TAILFILES,\,)} + 1])
 ;	same => n,Set(file=${CUT(TAILFILES,\,,${index})})
 ;	same => n,Playback(${file})
@@ -209,14 +213,14 @@ exten => s,1,Ringing
 ;	same => n,Hangup() 
 
 ; Customization to the tail message is possible by adding to this extension as follows:
-;exten => TAIL,1,Set(TAILFILES=/tmp/tailmsg,/tmp/tailmsg2) ; add , separated tail files to play
+;exten => TAIL,1,Set(TAILFILES=custom/yourtailmsg,/tmp/yourothertailmsg) ; add , separated tail files to play
 ;	same => n,Set(index=$[${GLOBAL(TAILFILEINDEX)} % ${FIELDQTY(TAILFILES,\,)} + 1])
 ;	same => n,Set(file=${CUT(TAILFILES,\,,${index})})
 ;	same => n,Playback(${file})
 ; 	;after playing the tail message, play a weather announcement
-;  	same => n,Playback(weather_file)
+;  	same => n,Playback(/tmp/current-weather)
 ;  	;after playing the weather announcement, play a traffic announcement
-;  	same => n,Playback(traffic_file)
+;  	same => n,Playback(/tmp/current-traffic)
 ;	same => n,Set(GLOBAL(TAILFILEINDEX)=$[${TAILFILEINDEX} + 1]) 
 ;  	same => n,Hangup()
 


apps/app_rpt/rpt_telemetry.c
```diff
@@ -1192,11 +1208,16 @@ void *rpt_tele_thread(void *this)
 		break;
 	case TAILMSG:
 		/* wait a little bit longer */
-		if (wait_interval(myrpt, DLY_TELEM, mychannel) == -1) {
-			break;
+		if (!wait_interval(myrpt, DLY_TELEM, mychannel)) {
+			donodelog_fmt(myrpt, "TELEMETRY,%s,TAILMSG,%s", myrpt->name, myrpt->p.tailmessages[myrpt->tailmessagen]);
+			if (ast_exists_extension(mychannel, myrpt->p.telemetry, TELEM_TAIL_FILE_EXTN, 1, NULL)) {
```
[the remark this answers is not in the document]
mkmer: I don't know.  I suppose you would want to match (or have independent versions) for each repeater (right?) That would / could force a different dial plan context for each repeater (or not).  If you don't "need" different stanzas for each, you would use the same stanza for all (right?)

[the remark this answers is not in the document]
Allan-N: I will add that my concern is that someone will opt for a different telemetry stanza and not know that they should also duplicate the extensions.conf stanza.

configs/rpt/extensions.conf
```diff
@@ -172,4 +172,19 @@ exten => s,1,Ringing
...
+[telemetry]
+;  
+; this stanza and extensions are used by "app_rpt" to play a telemetry  
+; file (e.g. tailmessages).  
+;  
+; note: you should not need to make any changes to this stanza  
+; Customization to the tail message is possible by adding to this extension:
+; ex: exten => TAIL,1,playback(${RPT_PLAYBACK_FILE})
+;     same => n,playback("weather_file.wav")
+;     same => n, playback("traffic_file.wav")
```
Allan-N, line 184: How about some consistency with "n,playback" vs. "n, playback" ?

Also, would your example end up playing whatever file was passed in ${RPT_PLAYBACK_FILE} --AND-- then both "weather_file.wav" and "traffic_file.wav" ?  If so, I would think a better example would be one that compares/matches the variable and plays something else instead.

configs/rpt/extensions.conf
```diff
@@ -172,4 +172,19 @@ exten => s,1,Ringing
...
+; this stanza and extensions are used by "app_rpt" to play a telemetry  
+; file (e.g. tailmessages).  
+;  
+; note: you should not need to make any changes to this stanza  
+; Customization to the tail message is possible by adding to this extension:
+; ex: exten => TAIL,1,playback(${RPT_PLAYBACK_FILE})
+;     same => n, playback("weather_file.wav")
+;     same => n, playback("traffic_file.wav")
+;     same => n, Hangup()
+
```
Allan-N, line 186: Can we add a comment here?
```
; used to play a message from the tailmessagelist=
exten => TAIL,1,playback(${RPT_PLAYBACK_FILE})
same  => n, Hangup()
```

configs/rpt/extensions.conf
Allan-N: How about :
```
; exten => TAIL,1,playback(${RPT_PLAYBACK_FILE})
;         ; after playing the tail message, add the current weather annoucement
;         same => n,playback("weather_file.wav")
;         ; after playing the tail message, add the current traffic annoucement
;         same => n,playback("traffic_file.wav")

File: configs/rpt/extensions.conf [hunk 4]
```diff
@@ -172,4 +172,24 @@ exten => s,1,Ringing
 	same => n,Wait(1)
 	same => n,Hangup
 
+[telemetry]
+;  
+; This stanza and extensions are used by "app_rpt" to play a telemetry  
+; file (e.g. tailmessages).  If this extension does not exist, 
+; app_rpt will not play the telemetry based on internal configuration.
```
Allan-N, line 179: "will not play"?  or "will play" ?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -1041,15 +1058,23 @@ static void handle_varcmd_tele(struct rpt *myrpt, struct ast_channel *mychannel,
 	ast_debug(5, "Ending telemetry, active_telem = %p, mytele = %p\n", myrpt->active_telem, mytele); \
 	myrpt->active_telem = NULL;
 
+/*! \brief Set a TELEM_TAIL_FILE variable to the selected tail file name
+ * and execute the dial plan with extentsion TELEM_TAIL_FILE_EXTN
+ */
+inline static void rpt_do_tail_exten(struct ast_channel *chan, const char *filename, const char *context)
```
InterLinked1, line 1064: I don't really see why the TELEM_TAIL_FILE / RPT_PLAYBACK_FILE variable is needed at all. If we're executing dialplan, it should be in the dialplan. There's no need for app_rpt to know about it at all. Or is there?

configs/rpt/extensions.conf
```diff
@@ -172,4 +172,24 @@ exten => s,1,Ringing
...
+; 
+; Customization to the tail message is possible by adding to this extension:
+;     exten => TAIL,1,playback(${RPT_PLAYBACK_FILE})
```
InterLinked1, line 182: Capitalize Playback, since that's its canonical form

configs/rpt/extensions.conf
```diff
@@ -172,4 +172,24 @@ exten => s,1,Ringing
...
+;     ;after playing the tail message, play a weather announcement
+;     same => n, playback("weather_file.wav")
+;	 ;after playing the weather announcement, play a traffic announcement
+;     same => n, playback("traffic_file.wav")
```
InterLinked1, line 186: No quotes are needed and no file extension should be provided

configs/rpt/extensions.conf
mkmer: will

apps/app_rpt/rpt_telemetry.c
mkmer: The variable "TELEM_PLAYBACK_FILE" is part of a rotating tail message system that "if" used needs to share the file name to the dial plan.  IF the user chooses to go with a single file name, they can simply use that file name in the dial plan.
In order to implement similar functionality, we would need a list of tail files in the dial plan and a method to rotate between them.... (Maybe that is possible?)

apps/app_rpt/rpt_telemetry.c
```diff
@@ -35,8 +35,24 @@
...
 
+/*!
+ * \brief Execute dialplan on conference conference
+ */
+static int rpt_do_dialplan(struct ast_channel *dpchannel, char *exten, const char *context)
+{
+	rpt_disable_cdr(dpchannel);
+	ast_channel_context_set(dpchannel, context);
+	ast_channel_exten_set(dpchannel, exten);
+	ast_debug(5, "Playback dialplan extension %s\n", exten);
```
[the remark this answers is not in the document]
mkmer: It sure does.  I had these for debuging my initial "construction" of the code -> left just because....

apps/app_rpt/rpt_telemetry.c
mkmer: Woops, looks like we just rotate it every time it is called -> and it's called once every `tailmessagetime` has expired and the repeater has keyed up.

mkmer:
Your questions made me realize there may be an issue with "squashing" the tail message using `ast_pbx_run`.  Once we hand channel over to a PBX, I don't think we can stop it (or can we just softhangup() to get it to stop?)
I'm thinking it needs to fire off a `pbx` thread then fall down to the `ast_streamwait()` and wait for PBX to finish...

https://github.com/AllStarLink/app_rpt/blob/0aab6177e7d97486dd75ba81189c9112755752fe/apps/app_rpt/rpt_telemetry.c#L2684-L2701

It's not clear to me at the moment how we currently kill the stream mid message.  It appears that when it IS killed, we DON'T increment the counter, but switch to the `tailsquashedtime`

https://github.com/AllStarLink/app_rpt/blob/0aab6177e7d97486dd75ba81189c9112755752fe/apps/app_rpt/rpt_telemetry.c#L2694-L2701

apps/app_rpt/rpt_telemetry.c
InterLinked1: It certainly could be done in dialplan, though I understand about compatibility.
Keeping a long-term view though, if this were ever to be merged into the Asterisk tree, I think stuff like this where it could be done fully in dialplan would raise eyebrows, but we don't need to address that now

InterLinked1: I'm writing this off the cuff so it's somewhat sloppy/unoptimized, and you may want to test this - but something like this:

Basically, using a global variable and incrementing that each time you play, taking care to wrap around back to 0 when needed.

```
[globals]
fileindex=0
numfiles=10
files=file1&file2&custom/file3 etc.

[some-context]
...
same => n,Set(filenum=${INC(fileindex)})
same => n,Set(index=$[${filenum} % ${numfiles}])
same => n,Set(file=${CUT(files,&,${index})})
same => n,Playback(${file})
```

mkmer:
Found where and how we kill tailmessages:
https://github.com/AllStarLink/app_rpt/blob/0aab6177e7d97486dd75ba81189c9112755752fe/apps/app_rpt.c#L5160-L5170
So - we do softhangup and the PBX should stop playing the dialplan.  I'll do a bit of testing to confirm.

apps/app_rpt/rpt_telemetry.c
mkmer: So the only rub here is that we don't know if we were squashed, so we will increment the counter for every call.
Maybe if we do the indexing after Playback, it will not increment on a softhangup... I will investigate that possibility

mkmer:
Have this ready to go:
Dailplan now rotates the messages, sticks at the last tailmessage, and uses the squashtimer.  It now can behave 100% like the current hard coded version, yet allow for customization by the user.

configs/rpt/extensions.conf
```diff
@@ -7,6 +7,11 @@ writeprotect = yes ; from overwriting the config file. Leave them here.
 HOMENPA = 999 ; change this to your Area Code
 NODE = 1999   ; change this to your node number
 
+TAILNUMFILES=2 ; set number of & separated files
```
Allan-N, line 10: Can we move TAILNUMFILES after TAILFILES (define the content, then count the elements)

Or, better yet, can we find a way to eliminate having to hard-code this value in the .conf file?  Could we use `FIELDQTY()`?

and can we leave out TAILFILEINDEX with the assumption that the initial value would be 0?

configs/rpt/extensions.conf
```diff
@@ -189,4 +194,27 @@ exten => s,1,Ringing
 	same => n,Wait(1)
 	same => n,Hangup
 
+[telemetry]
+;  
+; This stanza and extensions are used by "app_rpt" to play a telemetry  
+; file (e.g. tailmessages).  If this extension does not exist, 
+; app_rpt will play the telemetry based on internal configuration.
```
Allan-N, line 201: I'm confused.  If one doesn't make any changes to this extensions.conf file then are we playing the `tailmessagelist` audio files?  How are those file names passed to the dial plan?  Is this via `$(RPT_PLAYBACK_FILE)` ?  If so, why do I see `file=${CUT(files,&,${index})}`?

Also, if the TAIL dial plan is not available I think we can say that the `tailmessagelist` files will simply be played.

configs/rpt/extensions.conf
```diff
@@ -189,4 +194,27 @@ exten => s,1,Ringing
...
+; 
+; Customization to the tail message is possible by adding to this extension:
+;     exten => TAIL,1,Playback(${RPT_PLAYBACK_FILE})
+;     ;after playing the tail message, play a weather announcement
+;     same => n, Playback(weather_file)
+;     ;after playing the weather announcement, play a traffic announcement
+;     same => n, Playback(traffic_file)
+;     same => n, Hangup()
+; The "standard" tail message follows:
+; Modify the global variable TAILNUMFILES to contain the list of tail files separated with "&".
```
Allan-N, line 211: Still confused about $(RPT_PLAYBACK_FILE) vs. TAILFILES.

configs/rpt/extensions.conf
```diff
@@ -189,4 +194,27 @@ exten => s,1,Ringing
...
+; Modify TAILNUMFILES the number of tail files in the list.
```
Allan-N, line 212: See above re: TAILNUMFILES and TAILFILEINDEX.

and if I'm using the `tailmessagelist` in `rpt.conf` then do I also need to be setting TAILNUMFILES here?

configs/rpt/rpt.conf
```diff
@@ -134,7 +134,7 @@ idrecording = |iNOTSET              ; id recording or morse string
 idtime = 540000                     ; id interval time (in ms) (optional) Default 5 minutes (300000 ms)
 politeid = 30000                    ; time in milliseconds before ID timer expires to try and ID in the tail. (optional, default 30000)
 
-;tailmessagelist=/usr/share/asterisk/sounds/custom/yourtailmessage
+;tailmessagelist=/usr/share/asterisk/sounds/custom/yourtailmessage ; This list is ignored if the dialplan (extensions.conf) contains extension TAIL
```
Allan-N, line 137: OK, now you have me really confused.  Your updated `extensions.conf` includes a TAIL extension.  Are we going with A or B?  Can the TAIL extension use TAILFILES (from the dial play); else RPT_PLAYBACK_FILE passed from `rpt.conf`?

and what do you think about having the extensions.conf variable be TAIL_MESSAGE_FILES?

configs/rpt/extensions.conf
mkmer: I was thinking - maybe we should totally comment out this section.  If a user wants to use the extension, these are examples of how to do it.  If they leave it "stock" (without a TAIL extension), we continue to use the list found in `tailmessagelist` as configured in `rpt.conf`.

configs/rpt/rpt.conf
mkmer: Here I wanted to let a user "know" that IF they have the TAIL extension, this list is ignored (otherwise it's a head scratcher why it's not playing this list).

I'm leaning toward commenting out the TAIL extension, leaving it as a "option" to modify, not default.

Allan-N: Hmmm.... If a specify a `tailmessagelist` in rpt.conf AND I also define a list in extensions.conf could we play ALL of the files?

For each call,
- if not null, play RPT_PLAYBACK_FILE
- if TAILFILES, play TAILFILES[TAILFILEINDEX]
- increment/reset TAILFILEINDEX

Yes, if both RPT_PLAYBACK_FILE and TAILFILES are non-null then we'd play 2 files one after the other ... but, to me, that's OK

configs/rpt/extensions.conf
InterLinked1: Not sure I follow - what's wrong with channel variables here?
It could however be simplified to this:

`same => n,Set(MYVAL=${IF(${VARIABLE_EXISTS(VAR1)}?${VAR1}:${VAR2})})`

configs/rpt/rpt.conf
Allan-N: Without using any channel variables it would seem like the logic ends up being :
- if `tailmessagetime` is set then we :
  - play a tail message
    - if TAIL exists in extensions.conf then the dial plan determines what is played
    - if TAIL does not exist then the rpt.conf `tailmessagelist` is used
  - wait `tailmessagetime` before playing the next message  

DId I get this right?  If so, I think this is what should be stated in the rpt.conf and extensions.conf comments.

mkmer: > Without using any channel variables it would seem like the logic ends up being :
> 
> * if `tailmessagetime` is set then we :
AND unkeyed
>   
>   * play a tail message
>     
>     * if TAIL exists in extensions.conf then the dial plan determines what is played
>     * if TAIL does not exist then the rpt.conf `tailmessagelist` is used
>   * wait `tailmessagetime` before playing the next message
> 
> DId I get this right? If so, I think this is what should be stated in the rpt.conf and extensions.conf comments.

configs/rpt/extensions.conf
Allan-N: Or, leave the extension defined in the dial plan.  Then, one would only need to uncomment / change the single TAILFILES var.

configs/rpt/extensions.conf
```diff
@@ -7,6 +7,9 @@ writeprotect = yes ; from overwriting the config file. Leave them here.
 HOMENPA = 999 ; change this to your Area Code
 NODE = 1999   ; change this to your node number
 
+TAILFILEINDEX=0
```
Allan-N, line 10: Maybe add a comment like "used for telemetry, do not remove/change"

configs/rpt/extensions.conf
```diff
@@ -189,4 +192,33 @@ exten => s,1,Ringing
 	same => n,Wait(1)
 	same => n,Hangup
 
+[telemetry]
+;  
+; This stanza and extensions are used by "app_rpt" to play a telemetry
```
Allan-N, line 197: I think this comment block needs to be split.  The first sentence talks about the [telemetry] block itself ... and is good. The second sentence has a tailmessage focus ... and I'm thinking that it needs to be on it's own or combined with the sentence below.  Maybe something like :
```
[telemetry]
;
; This stanza and extensions are used by "app_rpt" to play a telemetry 
; file (e.g. tailmessages).
;
; The following TAIL extension provides support for playing messages after
; keying.  If this extension does not exist, app_rpt will play the telemetry
; based on the rpt.conf "tailmessagelist" variable.
; 
; An equivalent to the rpt.conf "tailmessage" configuration follows:
; Modify the TAILFILES value to contain the list of tail files separated with ",".
;
;exten => TAIL,1,Set(TAILFILES=/tmp/tailmsg,/tmp/tailmsg2) ; add , separated tail files to play
...
```
and then you can add comments for other [telemetry] extensions below

configs/rpt/extensions.conf
```diff
@@ -189,4 +192,33 @@ exten => s,1,Ringing
...
+[telemetry]
+;  
+; This stanza and extensions are used by "app_rpt" to play a telemetry  
+; file (e.g. tailmessages).  If this extension does not exist, 
+; app_rpt will play the telemetry based on tailmessagelist configuration.
+; 
+; The "standard" tail message follows:
+; Modify the global variable TAILFILES to contain the list of tail files separated with ",".
+
+;exten => TAIL,1,Set(TAILFILES=/tmp/tailmsg,/tmp/tailmsg2) ; add , separated tail files to play
```
Allan-N, line 204: How about using `TAILFILES=custom/yourtailmessage,/tmp/temptailmsg` to pick up the same file path used in the rpt.conf example and highlight that files from /tmp should really be thought of as temporary?

configs/rpt/extensions.conf
```diff
@@ -189,4 +192,33 @@ exten => s,1,Ringing
...
+;	same => n,Set(index=$[${GLOBAL(TAILFILEINDEX)} % ${FIELDQTY(TAILFILES,\,)} + 1])
+;	same => n,Set(file=${CUT(TAILFILES,\,,${index})})
+;	same => n,Playback(${file})
+;	same => n,Set(GLOBAL(TAILFILEINDEX)=$[${TAILFILEINDEX} + 1]) 
+;	same => n,Hangup() 
+
+; Customization to the tail message is possible by adding to this extension as follows:
+;exten => TAIL,1,Set(TAILFILES=/tmp/tailmsg,/tmp/tailmsg2) ; add , separated tail files to play
```
Allan-N, line 212: Same comment re: your choice of audio files

configs/rpt/extensions.conf
```diff
@@ -189,4 +192,33 @@ exten => s,1,Ringing
...
+;	same => n,Set(index=$[${GLOBAL(TAILFILEINDEX)} % ${FIELDQTY(TAILFILES,\,)} + 1])
+;	same => n,Set(file=${CUT(TAILFILES,\,,${index})})
+;	same => n,Playback(${file})
+; 	;after playing the tail message, play a weather announcement
+;  	same => n,Playback(weather_file)
```
Allan-N, line 217: Seems like the "weather" and "traffic" files would be something more temporary.  I'm thinking "/tmp/current-weather' and "/tmp/current-traffic".

Also, what happens in the dial plan if a file is specified that does not exist (like /tmp files at startup)?

Allan-N: >If the file doesn't exist, it skips and plays the next file (or hangs up). Logs a warning too.

I was wondering about the error handling.  A warning message is fine, just wanted to know that the world would not fall apart because of a missing file / typo / ...

Allan-N (approved):

configs/rpt/extensions.conf
mkmer: Looks like this:
```
[2025-08-01 09:47:53.413] WARNING[91561][C-00000193]: file.c:821 openstream_internal: File /tmp/tailmsg2 does not exist in any format
[2025-08-01 09:47:53.413] WARNING[91561][C-00000193]: file.c:1330 ast_streamfile: Unable to open /tmp/tailmsg2 (format (ulaw)): No such file or directory
[2025-08-01 09:47:53.413] WARNING[91561][C-00000193]: app_playback.c:520 playback_exec: Playback failed on DAHDI/pseudo-464359821 for /tmp/tailmsg2
```

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2681,23 +2702,26 @@ void *rpt_tele_thread(void *this)
...
+			}
 		}
+		ast_stopstream(mychannel);
+		ast_hangup(mychannel);
 	}
-	ast_stopstream(mychannel);
 	rpt_mutex_lock(&myrpt->lock);
 	if (mytele->mode == TAILMSG) {
-		if (!res) {
+		if (!res && !mytele->killed) { /* The tail message was not squashed (Keyed up while playing tail message)*/
```
InterLinked1, line 2719: should be a space before `*/`

configs/rpt/extensions.conf
```diff
@@ -7,6 +7,9 @@ writeprotect = yes ; from overwriting the config file. Leave them here.
 HOMENPA = 999 ; change this to your Area Code
 NODE = 1999   ; change this to your node number
 
+TAILFILEINDEX=0 ; used for telemetry, do not remove/change
```
InterLinked1, line 10: A slight improvement here - instead of defining it here, since users don't need to change this, it could be set in the telemetry context if the variable doesn't already exist. That way, the user doesn't need to know about it and can't mess it up

configs/rpt/extensions.conf
```diff
@@ -189,4 +192,37 @@ exten => s,1,Ringing
...
+; Customization to the tail message is possible by adding to this extension as follows:
+;exten => TAIL,1,Set(TAILFILES=custom/yourtailmsg,/tmp/yourothertailmsg) ; add , separated tail files to play
+;	same => n,Set(index=$[${GLOBAL(TAILFILEINDEX)} % ${FIELDQTY(TAILFILES,\,)} + 1])
+;	same => n,Set(file=${CUT(TAILFILES,\,,${index})})
+;	same => n,Playback(${file})
+; 	;after playing the tail message, play a weather announcement
+;  	same => n,Playback(/tmp/current-weather)
+;  	;after playing the weather announcement, play a traffic announcement
+;  	same => n,Playback(/tmp/current-traffic)
+;	same => n,Set(GLOBAL(TAILFILEINDEX)=$[${TAILFILEINDEX} + 1])
```
InterLinked1, line 224: The way this is written, if a hangup occurs prior to this line, the variable won't get incremented.
Is that fine?

Issue #687, opened 2025-05-18
rpt_config.c: nodes.allstarlink.org should be configurable in rpt.conf


jxmx:
This would allow for testing changes to node lookups and for people running private ASL networks to create their own DNS lookups rather than file-based lookups.

PR #694, opened 2025-06-02
Add config variable for DNS node lookup

jxmx:
Addesses #687

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -423,7 +423,7 @@ static int nrpts = 0;
 
 /* general settings */
 enum rpt_dns_method rpt_node_lookup_method = DEFAULT_NODE_LOOKUP_METHOD;
-char *rpt_dns_node_domain = DEFAULT_DNS_NODE_DOMAIN;
+const char *rpt_dns_node_domain = DEFAULT_DNS_NODE_DOMAIN;
 int rpt_max_dns_node_length = 6;
 
 int max_chan_stat[] = { 22000, 1000, 22000, 100, 22000, 2000, 22000 };
@@ -5505,6 +5505,7 @@ static int load_config(int reload)
 	int i, n = 0;
 	struct ast_config *cfg;
 	char *val, *this = NULL;
+	const char *cval;
 
 	cfg = ast_config_load("rpt.conf", config_flags);
 	if (!cfg) {
@@ -5540,12 +5541,12 @@ static int load_config(int reload)
 		}
 	}
 
-	val = (char *) ast_variable_retrieve(cfg, "general", "dns_node_domain");
-	if (val) {
+	cval = ast_variable_retrieve(cfg, "general", "dns_node_domain");
+	if (cval) {
 		if (rpt_is_valid_dns_name(val)) {
 			rpt_dns_node_domain = val;
 		} else {
-			ast_log(LOG_ERROR, "Configuration error: dns_node_domain value %s is not a valid format", val);
+			ast_log(LOG_ERROR, "Configuration error: dns_node_domain value %s is not a valid format", cval);
 			rpt_dns_node_domain = DEFAULT_DNS_NODE_DOMAIN;
 		}
 	} else {
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -424,6 +424,7 @@ static int nrpts = 0;
 
 /* general settings */
 enum rpt_dns_method rpt_node_lookup_method = DEFAULT_NODE_LOOKUP_METHOD;
+char *rpt_dns_node_domain = DEFAULT_DNS_NODE_DOMAIN;
 int rpt_max_dns_node_length = 6;
 
 int max_chan_stat[] = { 22000, 1000, 22000, 100, 22000, 2000, 22000 };
@@ -5549,8 +5550,6 @@ static int load_config(int reload)
 			ast_log(LOG_ERROR, "Configuration error: revert to default dns_node_domain value %s", DEFAULT_DNS_NODE_DOMAIN);
 			rpt_dns_node_domain = DEFAULT_DNS_NODE_DOMAIN;
 		}
-	} else {
-		rpt_dns_node_domain = DEFAULT_DNS_NODE_DOMAIN;
 	}
 	ast_log(LOG_NOTICE, "Domain used for DNS node lookup is: %s", rpt_dns_node_domain);
 	
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -300,7 +300,6 @@
 #include <fnmatch.h>
 #include <curl/curl.h>
 #include <termios.h>
-#include <stdbool.h>
 
 #include "asterisk/utils.h"
 #include "asterisk/lock.h"
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -300,6 +300,7 @@
 #include <fnmatch.h>
 #include <curl/curl.h>
 #include <termios.h>
+#include <stdbool.h>
 
 #include "asterisk/utils.h"
 #include "asterisk/lock.h"
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -5550,7 +5550,6 @@ static int load_config(int reload)
 		}
 	}
 	ast_log(LOG_NOTICE, "Domain used for DNS node lookup is: %s", rpt_dns_node_domain);
-	
 	val = (char *) ast_variable_retrieve(cfg, "general", "max_dns_node_length");
 	if (val) {
 		i = atoi(val);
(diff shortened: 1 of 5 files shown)


apps/app_rpt.c
```diff
@@ -5538,6 +5539,21 @@ static int load_config(int reload)
 			rpt_node_lookup_method = DEFAULT_NODE_LOOKUP_METHOD;
 		}
 	}
+
+	val = (char *) ast_variable_retrieve(cfg, "general", "dns_node_domain");
```
mkmer, line 5544: I think we can make `val` a `const char *` and NOT cast `(char *)` here.
Probably make `rpt_dns_node_domain` a constant as well?

apps/app_rpt/rpt_config.c
```diff
@@ -1354,3 +1359,36 @@ void rpt_update_boolean(struct rpt *myrpt, char *varname, int newval)
 		rpt_event_process(myrpt);
 	}
 }
+
+bool rpt_is_valid_dns_name(const char *dns_name) {
```
mkmer, line 1363: Should add documentation to this "new" function.

apps/app_rpt/rpt_config.c
```diff
@@ -1354,3 +1359,36 @@ void rpt_update_boolean(struct rpt *myrpt, char *varname, int newval)
...
+	if (!dns_name || strlen(dns_name) > MAX_DOMAIN_LENGTH) {
```
mkmer, line 1364: Why are we limiting the string length?  I don't believe there is a limit to DNS name sizes (in the DNS spec)

Allan-N: From `<arpa/nameser.h>` :
```
#define NS_MAXCDNAME    255     /*%< maximum compressed domain name */
#define NS_MAXLABEL     63      /*%< maximum length of domain label */
```

apps/app_rpt/rpt_config.c
```diff
@@ -1354,3 +1359,36 @@ void rpt_update_boolean(struct rpt *myrpt, char *varname, int newval)
...
+			label_start = true;
+		} else {
+			if (!isalnum(*ptr) && *ptr != '-') {
+				return false; // Only alphanumeric and hyphens allowed
+			}
+			if (label_start && *ptr == '-') {
+				return false; // Labels can't start with a hyphen
+			}
+			label_length++;
+			if (label_length > MAX_LABEL_LENGTH) {
```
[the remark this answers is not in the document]
jxmx: DNS NAME label elements are 64 octets including the . So here it's 63.

apps/app_rpt/rpt_config.c
```diff
@@ -1354,3 +1359,36 @@ void rpt_update_boolean(struct rpt *myrpt, char *varname, int newval)
...
+	if (!dns_name || strlen(dns_name) > MAX_DOMAIN_LENGTH) {
+		return false;
+	}
+
+	int label_length = 0;
+	bool label_start = true;
+
+	for (const char *ptr = dns_name; *ptr; ++ptr) {
+		if (*ptr == '.') {
+			if (label_start || label_length == 0) {
```
mkmer, line 1372: Is it possible for one or the other here?  I think `if (label_start) { ` is sufficient to cover all of the cases (and make it a bit easier to understand).
Essentially, if we have a `.` and `label_start == TRUE`, there was nothing between the last `.` and this time around.

apps/app_rpt.c
```diff
@@ -5538,6 +5539,19 @@ static int load_config(int reload)
...
+		if (rpt_is_valid_dns_name(val)) {
+			rpt_dns_node_domain = val;
+		} else {
+			ast_log(LOG_ERROR, "Configuration error: dns_node_domain value %s is not a valid format", val);
+			ast_log(LOG_ERROR, "Configuration error: revert to default dns_node_domain value %s", DEFAULT_DNS_NODE_DOMAIN);
+			rpt_dns_node_domain = DEFAULT_DNS_NODE_DOMAIN;
+		}
+	}
+	ast_log(LOG_NOTICE, "Domain used for DNS node lookup is: %s", rpt_dns_node_domain);
+
```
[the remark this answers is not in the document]
jxmx: no

jxmx:
I get some of this, but why are we changing formats if the surrounding code already matches? For example, it's flagging on stuff like a tab vs. space between `#define` when all the other code is uses a tab. Same with blank lines between functions for readability.

apps/app_rpt.c
jxmx: `char *val` seems to be used in a lot of places. Not sure this is the patch to fix that universally?

apps/app_rpt.c
```diff
@@ -5538,6 +5539,21 @@ static int load_config(int reload)
 			rpt_node_lookup_method = DEFAULT_NODE_LOOKUP_METHOD;
 		}
 	}
+
+	val = (char *) ast_variable_retrieve(cfg, "general", "dns_node_domain");
+	if (val) {
+		if (rpt_is_valid_dns_name(val)) {
+			rpt_dns_node_domain = val;
+		} else {
+			ast_log(LOG_ERROR, "Configuration error: dns_node_domain value %s is not a valid format", val);
```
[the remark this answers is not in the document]
jxmx: removed it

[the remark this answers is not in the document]
mkmer: I think it's "ok" to be obvious - just put "all of the words" in one log entry... or go with the 1 liner you have now.

apps/app_rpt.c
mkmer: I "believe" there are many cases where it is "wrong" -> we have been moving to "right".
Of course I could be "wrong" too :)

apps/app_rpt/rpt_config.c
```diff
@@ -1354,3 +1361,44 @@ void rpt_update_boolean(struct rpt *myrpt, char *varname, int newval)
 		rpt_event_process(myrpt);
 	}
 }
+
+int rpt_is_valid_dns_name(const char *dns_name)
+{
+	int label_length, label_start;
+
+	if (!dns_name || strlen(dns_name) > MAX_DNS_NODE_DOMAIN_LEN) {
```
mkmer, line 1369: Why are we limiting the domain string length?  I don't believe there is a max in the DNS spec...

apps/app_rpt/rpt_config.c
```diff
@@ -31,9 +34,13 @@
 #define TLB_QUERY_NODE_EXISTS 1
 #define TLB_QUERY_GET_CALLSIGN 2
 
+/*! \brief DNS max overall and per-label sizes (RFC1035) */
+#define MAX_DNS_NODE_DOMAIN_LEN 253
```
mkmer, line 38: Spec says 255....

mkmer:
> I get some of this, but why are we changing formats if the surrounding code already matches? For example, it's flagging on stuff like a tab vs. space between `#define` when all the other code is uses a tab. Same with blank lines between functions for readability.

It's really not a big deal if you run the format tool locally - fixes everything for you and move on.  Since it's a MASSIVE job to fix the entire code base, we are just picking away at it as we make changes.  Eventually it will all comply (which it will need to be if we ever hope to get it merged back into the Asterisk codebase).

apps/app_rpt.c
jxmx: I'm not saying that it's wrong to do, I'm just questioning procedurally if this pull request is the right place to fix it because it'll touch more than the code i added.

mkmer: I was "thinking" we are getting value from the config, it is/should be a `const` -> by making `const char *val` we remove the need to cast.
I tried making the change and the compiler is "happy" with it, as long as we change all of the pointers to `const`. 

Generally, we should only be "fixing" stuff we are touching (and trying to keep PR's small and standalone), so IF the change requires a major overhaul, we have been leaving it as is.  

Sometimes we just need to shoot the engineer and ship it :)

We "could" talk about it for days... or just leave it as is :)  So much of a philosophical discussion...

Allan-N: We'd start off with `const char *val;`.  Then, we can remove the `(char *)` cast from the `val = ast_variable_retrieve()`.  Then, we would need to change the variables that are assigned the `val` value (e.g. `rpt_node_lookup_method`).  And, since `val` is used assigned/used in many places it's gonna get out of hand quickly.  Would this be a good thing to do?  Yes, but maybe not for this PR.

mkmer: I pushed "my thought" on this.  No need to go of the deep end fixing the whole thing, just what we touched.
If it's not a good idea, I'll undo it :)

apps/app_rpt/rpt_config.c
mkmer: Clarity is always nice... The spec leaves out details for sure.

apps/app_rpt.c
jxmx: This works for me.

Allan-N (approved):
Corrected the result when presented with a domain name including the FQDN  trailing-dot.

To me, this is all good!

mkmer (approved):
LGTM.

PR #696, opened 2025-06-08
app_rpt: fix archive logging crash due to incorrect format string(s)

Allan-N:


Allan-N:
>Strange that this made it.. I thought I had run it locally.

I also had my local system setup with the archive logging.  What I didn't do was exercise all of the commands/functions.  Was doing some testing today and got a crash, looked at the backtrace, and the oops jumped right out.

PR #698, opened 2025-06-10
Resolve parrot mode conflicts between app_rpt, rpt.conf

Allan-N:
The "rpt.conf" file had comments about the "parrotmode" variable when, in fact, we were using the "parrot" variable.  This change clears up the comments in the configuration file and its usage in the code.

### DIFF
--- a/apps/app_rpt/rpt_functions.c
+++ b/apps/app_rpt/rpt_functions.c
@@ -1311,20 +1311,20 @@ enum rpt_function_response function_cop(struct rpt *myrpt, char *param, char *di
 		rpt_telemetry(myrpt, ARB_ALPHA, (void *) "ATDIS");
 		return DC_COMPLETE;
 
-	case 21:					/* Parrot Mode Enable */
+	case 21: /* Parrot Mode Enable */
 		birdbath(myrpt);
-		if (myrpt->p.parrotmode < 2) {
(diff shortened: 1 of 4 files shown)


apps/app_rpt/rpt_functions.c
```diff
@@ -1311,20 +1311,20 @@ enum rpt_function_response function_cop(struct rpt *myrpt, char *param, char *di
 		rpt_telemetry(myrpt, ARB_ALPHA, (void *) "ATDIS");
 		return DC_COMPLETE;
 
-	case 21:					/* Parrot Mode Enable */
+	case 21: /* Parrot Mode Enable */
 		birdbath(myrpt);
-		if (myrpt->p.parrotmode < 2) {
+		if (myrpt->p.parrotmode < PARROT_MODE_ON_ALWAYS) {
```
mkmer, line 1316: When we use ENUM I thought `<` and `>` are not "valid" -> should use `!= PARROT_MODE_ON_ALWAYS`  or `== PARROT_MODE_OFF or == PARROT_MODE_ON_COMMAND`?

Allan-N: The enum values are sequential so this is OK.  But, your suggestion is good.  Branch updated.

PR #702, opened 2025-06-21
iax.conf: Typo fix registartions -> registrations

Elwell:
Trivial typo fix - patched because I couldn't find the file suggested :-)


Allan-N (approved):
Thank you for the fix!

InterLinked1 (approved):

PR #704, opened 2025-06-27
chan_echolink: resolve issues found during code review

Allan-N:
- don't lock/unlock mutex that's already being held


mkmer (approved):
Probably didn't "hurt" anything, but not "right".
LGTM

Issue #600, opened 2025-03-31
archivedir recording (mostly) silence to WAV file when duplex=3


kenalker:
**Describe the bug**
I am using the “archivedir” function to record repeater audio.  I have a single repeater with an RTCM attached to the ASL3 computer.  Audio is recorded to a WAV file (date_and_time_stamp.WAV as filename) each time someone keys up until they unkey (one file per key-up), as expected. 
 The files contain audio when duplex=2 in rpt.conf.  However, when duplex=3 in rpt.conf the files contain no received audio, but they do contain the courtesy tone and ID in CW (if present) at the end of the recording.  WAV files are of the appropriate size to contain audio and `ffprobe -v error -show_format 20250331003700.WAV` reports a file of 8-second duration.

**To Reproduce**
1) Add archivedir=/var/spool/asterisk/monitor to your node stanza in rpt.conf.
2) Key up your repeater and speak.
3) Play back the WAV file created in the /var/spool/asterisk/monitor/##### directory (where ##### is your actual node number)

**Expected behavior**
I expect to have a WAV (*.wav) file recorded that  has spoken audio from the person who keyed up the repeater and spoke.

**Software versions (listed in asl-menu, option 4)**
```
OS            : Debian GNU/Linux 12 (bookworm)
OS Kernel     : 6.1.0-32-amd64
Asterisk      : 22.2.0+asl3-3.3.0-6.deb12
ASL [app_rpt] : 3.3.0
Installed ASL packages :
Package                         Version
==============================  ==============================
allmon3                         1.4.2-1.deb12
asl3                            3.7.1-1.deb
asl3-asterisk                   2:22.2.0+asl3-3.3.0-6.deb12
asl3-asterisk-config            2:22.2.0+asl3-3.3.0-6.deb12
asl3-asterisk-modules           2:22.2.0+asl3-3.3.0-6.deb12
asl3-menu                       1.12-1.deb12
dahdi                           1:3.1.0-2
dahdi-dkms                      1:3.4.0-6+asl
dahdi-linux                     1:3.4.0-6+asl
```

**Have you run a software update and rebooted?**
Yes.

**What is the platform - e.g. Raspberry Pi 4, Raspberry Pi 5, Virtual Machine, Desktop, etc.**
Desktop - Intel(R) Core(TM) i5-2300 CPU @ 2.80GHz

mkmer:
This is caused here:
[...]

PR #708, opened 2025-06-30
app_rpt: audio archive, capture local audio in duplex mode 3

mkmer:
Re-addressing original PR#603 issue: recording silence in audio archive while in duplex mode 3 and "transmitting".


Allan-N:
Again, trying to address https://github.com/AllStarLink/app_rpt/issues/600

apps/app_rpt.c
```diff
@@ -3788,7 +3788,7 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 		f1 = rpt_frame_helper(myrpt, f, ismuted);
 		if (f1) {
 			ast_write(myrpt->localoverride ? myrpt->txpchannel : myrpt->pchannel, f1);
-			if ((myrpt->p.duplex < 2) && myrpt->monstream && (!myrpt->txkeyed) && myrpt->keyed) {
+			if (((myrpt->p.duplex < 2 && myrpt->monstream && !myrpt->txkeyed) || myrpt->p.duplex == 3) && myrpt->keyed) {
```
Allan-N, line 3791: This test needs to be reworked.  As-is, it looks like we only need `myrpt->p.duplex == 3` && `myrpt->keyed` to call ast_writestream() even if we don't have `myrpt->monstream` :(

Allan-N (dismissed):

apps/app_rpt.c
Allan-N: Just commit a change to address the issue that I found.

mkmer: Sure thing - that's what I get for hurrying before vacation, a mistake :)

PR #603, opened 2025-03-31
app_rpt: Update duplex mode 3 to pass audio to audioarchive.

mkmer:
Mode 3 is intended to be a full duplex with local audio muted.  This WAS accomplished via conference settings but that prevented the local audio from reaching the monstream channel. Converting the conference settings to match modes 2 and 4 (full duplex) and simply muting the frames headed to the txchannel allows for monstream to catch the audio and prevent the audio from feeding back on the local speaker. Resolves issue archivedir recording (mostly) silence to WAV file when duplex=3 

Resolves #600

PR #710, opened 2025-07-01
Revert "app_rpt: Update duplex mode 3 to pass audio to audioarchive (part 2)

Allan-N:
While "fixing" audio archive, we "broke" duplex mode 3 original "full duplex" behavior.  The first step was to back out [PR #603](https://github.com/AllStarLink/app_rpt/pull/603).  That change was made with [PR #707](https://github.com/AllStarLink/app_rpt/pull/707) but the change was incomplete.  This change makes it right.


mkmer (approved):
That's a fine how do ya do.... Wonder why this wasn't part of 603 :(
Thanks for hunting it down.

I see now: 601 fixed this at some point after 603 (or something like that).

PR #722, opened 2025-07-19
chan_echolink, chan_tlb: correct strtok_r usage

Allan-N:
Valgrind identified errors in our usage of the strtok_r() API.


channels/chan_echolink.c
```diff
@@ -2092,7 +2092,7 @@ static int find_delete(const struct el_node *key)
 static void process_cmd(char *buf, int buf_len, const char *fromip, struct el_instance *instp)
 {
 	char *cmd, *arg1;
-	char delim = ' ';
+	char *delim = " ";
```
mkmer, line 2095: `const char *` ???

channels/chan_tlb.c
```diff
@@ -1147,7 +1147,7 @@ static int TLB_text(struct ast_channel *ast, const char *text)
 {
 	char buf[200];
 	char *arg4 = NULL, *ptr, *saveptr;
-	char delim = ' ', *cmd;
+	char *delim = " ", *cmd;
```
mkmer, line 1150: `const char *`?

channels/chan_echolink.c
Allan-N: Fixed

channels/chan_tlb.c
Allan-N: Fixed

mkmer (approved):

PR #723, opened 2025-07-19
app_rpt: fix [echolink] choppy audio

Allan-N:
First off, a HUGE thank you to Chuck/AI7SY for his help in identifying the PR that resulted in the choppy audio.

The changes that were made for [#538 ](https://github.com/AllStarLink/app_rpt/pull/538) had a few places where we updated the `struct rpt` frame queue (lastf1, lastf2) but should have updated the `struct rpt_link` frame queue.  This change restores the expected behavior.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3554,34 +3554,55 @@ static inline void outstream_write(struct rpt *myrpt, struct ast_frame *f)
 	}
 }
 
-/*! \brief Zero data in myrpt->lastf1 and lastf2 registers (muting audio)
- * \param myrpt The rpt structure to mute
+/*!
+ * \internal
+ * \brief Free a frame if it exists
  */
-static inline void mute_frame_helper(struct rpt *myrpt)
+static inline void free_frame(struct ast_frame **f)
 {
-	RPT_MUTE_FRAME(myrpt->lastf1);
-	RPT_MUTE_FRAME(myrpt->lastf2);
+	if (!*f) {
+		return;
+	}
+	ast_frfree(*f);
+	*f = NULL;
 }
 
-/*! \brief Shifts frames: myrpt->lastf2 -> return value, myrpt->lastf1 -> myrpt->lastf2, f -> myrpt->lastf1.
- * \param myrpt - the rpt structure
+/*! \brief Zero data in frame_queue->lastf1 and lastf2 registers (muting audio)
+ * \param frame_queue The rpt_frame_queue structure to mute
+ */
+static inline void rpt_frame_queue_mute(struct rpt_frame_queue *frame_queue)
+{
+	RPT_MUTE_FRAME(frame_queue->lastf1);
+	RPT_MUTE_FRAME(frame_queue->lastf2);
+}
+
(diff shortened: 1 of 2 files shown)


apps/app_rpt.c
```diff
@@ -5946,20 +5967,22 @@ static inline int exec_chan_read(struct rpt *myrpt, struct ast_channel *chan, ch
...
-			if (phone_mode && f1) {
-				ast_write(myrpt->txchannel, f1);
-			} else if (!phone_mode) {
-				ast_write(myrpt->txchannel, f);
+		if (f1) {
+			if (!myrpt->remstopgen) {
+				if (phone_mode) {
+					ast_write(myrpt->txchannel, f1);
+				} else {
+					ast_write(myrpt->txchannel, f);
```
mkmer, line 5975: I think we want to write `f` even if `f1` is null don't we?

Allan-N: I don't think so.  We start off with a frame to add to the queue and we want to write the delayed frame,  When we are starting off we don't have any already delayed frames in the pipeline so there's nothing to write.  Else, we'd write the frame we are adding to the queue and then 2 frames later we'd write it again.

Unless the thought here is that this is a special case where we don't delay the frames.

Also, the changes I made here were to revert the 3.3.0 behavior (that also matches the ASL2 code)

mkmer: `f` is the original frame with NO delay.  If we ARE in phone mode, we send the delayed `f1` frame, if we are NOT in phone mode we send `f` (the original frame with no delay)

We "do" modify `f` if it's `ismuted` is true, so we always need to execute `rpt_frame_helper` here, BUT we don't want to chop off the first 2 frames if `!phone_mode` (as we will be not be using `f1` and the first 2 buffers are empty?!?)

apps/app_rpt.c
```diff
@@ -4432,7 +4439,8 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 					ast_write(l->pchan, f);
 				}
 			} else if (f->frametype == AST_FRAME_DTMF_BEGIN) {
-				mute_frame_helper(myrpt);
+				rpt_frame_queue_mute(&l->frame_queue);
+				rpt_frame_queue_mute(&myrpt->frame_queue);
```
mkmer, line 4443: I don't think we need to mess around with the repeater frames here. 
In this location we are processing a link frame (that was DTMF) and muting it (before the link frame is copied into the conferences).  Muting the repeater queue "should" not be necessary and may cause other side effects?

apps/app_rpt.c
```diff
@@ -4443,7 +4451,8 @@ static inline int process_link_channels(struct rpt *myrpt, struct ast_channel *w
 					ast_free(tstr);
 				}
 			} else if (f->frametype == AST_FRAME_DTMF) {
-				mute_frame_helper(myrpt);
+				rpt_frame_queue_mute(&l->frame_queue);
+				rpt_frame_queue_mute(&myrpt->frame_queue);
```
mkmer, line 4455: Same comment as 4442

apps/app_rpt.c
Allan-N: Yup.  Should only be muting one set.

PR #724, opened 2025-07-21
chan_echolink: add few missing locks

mkmer:
Reviewing the choppy audio, I noticed these "missing" locks.


mkmer:
> I'm thinking that this entire function needs to be called locked (like the `el_db_find_XXX` functions). As is, you lock to find the key (a pointer to a node), release the lock, and then use the returned pointer.

I think you are correct -> pesky pointers...

PR #730, opened 2025-07-23
*-tune-menu: (S)wap command should not exit when no alternate devices

Allan-N:


utils/radio-tune-menu.c
```diff
@@ -431,12 +431,12 @@ static int astgetresp(char *cmd)
...
 
 	/* get device list from Asterisk */
-	if (astgetline(COMMAND_PREFIX "tune menu-support 1", buf, sizeof(buf) - 1)) {
+	if (astgetline(COMMAND_PREFIX "tune menu-support 1", buf, sizeof(buf) - 1) < 0) {
 		exit(255);
 	}
 	n = explode_string(buf, strs, ARRAY_LEN(strs), ',', 0);
-	if (n < 1) {
-		fprintf(stderr, "Error parsing USB device information\n");
+	if ((n < 1) || (!*strs[0])) {
```
mkmer, line 438: Why isn't 0 good enough? (aka <1)
https://github.com/AllStarLink/app_rpt/blob/0aab6177e7d97486dd75ba81189c9112755752fe/utils/simpleusb-tune-menu.c#L125-L128
If `strs[0] == 0` we `return 0`

Allan-N: In radio-tune-menu.c, I copied the block from https://github.com/AllStarLink/app_rpt/blob/0aab6177e7d97486dd75ba81189c9112755752fe/utils/radio-tune-menu.c#L496-L499.

and, yes, checking the result count from explode_string() for < 1 would be sufficient in both cases.  I shall fix BOTH.

PR #732, opened 2025-07-24
app_rpt: rpt_call.c address `cap` memory leak

mkmer:
Potential memory leak by not freeing `cap` when failures occur.

### DIFF
--- a/apps/app_rpt/rpt_call.c
+++ b/apps/app_rpt/rpt_call.c
@@ -82,11 +82,14 @@ void rpt_forward(struct ast_channel *chan, char *dialstr, char *nodefrom)
 
 	dest = ast_request("IAX2", cap, NULL, NULL, dialstr, NULL);
 	if (!dest) {
-		if (ast_safe_sleep(chan, 150) == -1)
+		if (ast_safe_sleep(chan, 150) == -1) {
+			ao2_ref(cap, -1);
 			return;
+		}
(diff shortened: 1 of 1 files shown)


apps/app_rpt/rpt_call.c
```diff
@@ -82,11 +82,14 @@ void rpt_forward(struct ast_channel *chan, char *dialstr, char *nodefrom)
 
 	dest = ast_request("IAX2", cap, NULL, NULL, dialstr, NULL);
 	if (!dest) {
-		if (ast_safe_sleep(chan, 150) == -1)
+		if (ast_safe_sleep(chan, 150) == -1) {
+			ao2_ref(cap, -1);
 			return;
+		}
 		dest = ast_request("IAX2", cap, NULL, NULL, dialstr, NULL);
```
InterLinked1, line 89: This looks good, but it seems like we are just waiting 150ms, then retrying the same exact operation again. Maybe we should just remove this code? Seems like unnecessary complexity. If it didn't work the first time, I don't really see the point in retrying.
Then we'd only need the one ao2_ref

Allan-N: I too question why the first request would fail but waiting 150ms and trying again would succeed.  If we don't need the retry then the one ao2_ref call would be sufficient.  But, if the call(s) to ast_request can fail and we return early then we do want to ensure that the cap reference is released.

Allan-N (approved):

PR #735, opened 2025-07-29
app_rpt: Add TOD to dialplan

mkmer:
Allow for Time Of Day to be generated from the dial plan. 
Based on https://github.com/AllStarLink/app_rpt/pull/688 (aka need to merge 688 first)

### DIFF
--- a/apps/app_rpt/rpt_telemetry.c
+++ b/apps/app_rpt/rpt_telemetry.c
@@ -42,6 +42,43 @@
 extern struct rpt rpt_vars[MAXRPTS];
 
 /*!
+ * \brief Say the time
+ * \param mychannel The channel to speak on
+ * \param t The time to say
+ * \param timezone The timezone to use for the time
+ */
+static void rpt_say_time(struct ast_channel *mychannel, time_t t, const char *timezone)
+{
+	time_t t1;
+	struct ast_tm localtm;
+	char *p;
+	int res;
+
+	rpt_localtime(&t, &localtm, timezone);
+	t1 = rpt_mktime(&localtm, NULL);
+	/* Say the phase of the day is before the time */
+	if ((localtm.tm_hour >= 0) && (localtm.tm_hour < 12)) {
+		p = "rpt/goodmorning";
+	} else if ((localtm.tm_hour >= 12) && (localtm.tm_hour < 18)) {
+		p = "rpt/goodafternoon";
+	} else {
+		p = "rpt/goodevening";
+	}
+	if (sayfile(mychannel, p) == -1) {
+		return;
+	}
+	/* Say the time is ... */
+	if (sayfile(mychannel, "rpt/thetimeis") == -1) {
+		return;
+	}
+	/* Say the time */
+	res = ast_say_time(mychannel, t1, "", ast_channel_language(mychannel));
+	if (!res) {
+		res = ast_waitstream(mychannel, "");
+	}
+	ast_stopstream(mychannel);
+}
+/*!
  * \brief Execute dialplan on conference conference
  */
 static enum ast_pbx_result rpt_do_dialplan(struct ast_channel *dpchannel, char *exten, const char *context)
@@ -619,13 +656,10 @@ static int telem_send_ct(struct rpt *myrpt, struct ast_channel *chan, const char
  */
 static void handle_varcmd_tele(struct rpt *myrpt, struct ast_channel *mychannel, char *varcmd)
 {
-	char *strs[100], *p, buf[100], c;
+	char *strs[100], buf[100], c;
 	int i, j, k, n, res, vmajor, vminor;
 	float f;
-	time_t t;
 	unsigned int t1;
-	struct ast_tm localtm;
-
 	n = finddelim(varcmd, strs, ARRAY_LEN(strs));
 	if (n < 1) {
 		return;
@@ -753,34 +787,19 @@ static void handle_varcmd_tele(struct rpt *myrpt, struct ast_channel *mychannel,
 		if (sscanf(strs[1], N_FMT(u), &t1) != 1) {
 			return;
 		}
-		t = t1;
 		if (wait_interval(myrpt, DLY_TELEM, mychannel) == -1) {
 			return;
 		}
 		donodelog_fmt(myrpt, "TELEMETRY,%s,STATS_TIME", myrpt->name);
-		rpt_localtime(&t, &localtm, myrpt->p.timezone);
-		t1 = rpt_mktime(&localtm, NULL);
-		/* Say the phase of the day is before the time */
-		if ((localtm.tm_hour >= 0) && (localtm.tm_hour < 12)) {
-			p = "rpt/goodmorning";
-		} else if ((localtm.tm_hour >= 12) && (localtm.tm_hour < 18)) {
-			p = "rpt/goodafternoon";
-		} else {
-			p = "rpt/goodevening";
-		}
-		if (sayfile(mychannel, p) == -1) {
+		if (ast_exists_extension(mychannel, myrpt->p.telemetry, TELEM_TIME_EXTN, 1, NULL)) {
+			rpt_do_dialplan(mychannel, TELEM_TIME_EXTN, myrpt->p.telemetry);
+			return;
+		} else if (ast_exists_extension(mychannel, TELEMETRY, TELEM_TIME_EXTN, 1, NULL)) {
+			rpt_do_dialplan(mychannel, TELEM_TIME_EXTN, TELEMETRY);
 			return;
 		}
-		/* Say the time is ... */
-		if (sayfile(mychannel, "rpt/thetimeis") == -1) {
-			return;
-		}
-		/* Say the time */
-		res = ast_say_time(mychannel, t1, "", ast_channel_language(mychannel));
-		if (!res) {
-			res = ast_waitstream(mychannel, "");
-		}
-		ast_stopstream(mychannel);
+
+		rpt_say_time(mychannel, t1, myrpt->p.timezone);
 		return;
 	}
 	if (!strcasecmp(strs[0], "STATS_VERSION")) {
@@ -1072,8 +1091,7 @@ void *rpt_tele_thread(void *this)
 	struct ast_channel *mychannel = NULL;
 	int id_malloc = 0, m;
 	char *p, *ident, *nodename;
-	time_t t, t1, t_mono, was_mono;
-	struct ast_tm localtm;
+	time_t t, t_mono, was_mono;
 	char **strs;
 	int i, j, k, ns, rbimode;
 	char mhz[MAXREMSTR], decimals[MAXREMSTR], mystr[200];
@@ -2441,37 +2459,12 @@ treataslocal:
 			pbx = 1;
 			break;
 		} else if (ast_exists_extension(mychannel, TELEMETRY, TELEM_TIME_EXTN, 1, NULL)) {
-			rpt_do_dialplan(mychannel, TELEM_TIME_EXTN, myrpt->p.telemetry);
+			rpt_do_dialplan(mychannel, TELEM_TIME_EXTN, TELEMETRY);
 			pbx = 1;
 			break;
 		}
 
-		t = time(NULL);
-		rpt_localtime(&t, &localtm, myrpt->p.timezone);
-		t1 = rpt_mktime(&localtm, NULL);
-		/* Say the phase of the day is before the time */
-		if ((localtm.tm_hour >= 0) && (localtm.tm_hour < 12)) {
-			p = "rpt/goodmorning";
-		} else if ((localtm.tm_hour >= 12) && (localtm.tm_hour < 18)) {
-			p = "rpt/goodafternoon";
-		} else {
-			p = "rpt/goodevening";
-		}
-		if (sayfile(mychannel, p) == -1) {
-			imdone = 1;
-			break;
-		}
-		/* Say the time is ... */
-		if (sayfile(mychannel, "rpt/thetimeis") == -1) {
-			imdone = 1;
-			break;
-		}
-		/* Say the time */
-		res = ast_say_time(mychannel, t1, "", ast_channel_language(mychannel));
-		if (!res) {
-			res = ast_waitstream(mychannel, "");
-		}
-		ast_stopstream(mychannel);
+		rpt_say_time(mychannel, time(NULL), myrpt->p.timezone);
 		imdone = 1;
 		break;
 	case STATS_VERSION:
--- a/apps/app_rpt/rpt_telemetry.c
+++ b/apps/app_rpt/rpt_telemetry.c
@@ -822,7 +822,6 @@ static void handle_varcmd_tele(struct rpt *myrpt, struct ast_channel *mychannel,
 		if (ast_exists_extension(mychannel, myrpt->p.telemetry, TELEM_TIME_EXTN, 1, NULL)) {
 			struct ast_datastore *datastore = ast_datastore_alloc(&telemetry_datastore, NULL);
 			if (!datastore) {
-				ast_log(LOG_ERROR, "Failed to allocate datastore for telemetry\n");
 				return;
 			}
 			time_data = ast_malloc(sizeof(time_t));
@@ -837,7 +836,6 @@ static void handle_varcmd_tele(struct rpt *myrpt, struct ast_channel *mychannel,
 		} else if (ast_exists_extension(mychannel, TELEMETRY, TELEM_TIME_EXTN, 1, NULL)) {
 			struct ast_datastore *datastore = ast_datastore_alloc(&telemetry_datastore, NULL);
 			if (!datastore) {
-				ast_log(LOG_ERROR, "Failed to allocate datastore for telemetry\n");
 				return;
 			}
 			time_data = ast_malloc(sizeof(time_t));
@@ -2510,7 +2508,6 @@ treataslocal:
 		if (ast_exists_extension(mychannel, myrpt->p.telemetry, TELEM_TIME_EXTN, 1, NULL)) {
 			struct ast_datastore *datastore = ast_datastore_alloc(&telemetry_datastore, NULL);
 			if (!datastore) {
-				ast_log(LOG_ERROR, "Failed to allocate datastore for telemetry\n");
 				imdone = 1;
 				break;
 			}
--- a/apps/app_rpt/rpt_telemetry.c
+++ b/apps/app_rpt/rpt_telemetry.c
@@ -36,7 +36,6 @@
 #include "rpt_rig.h"
 
 #define TELEM_TAIL_FILE_EXTN "TAIL"
-#define TELEM_TAIL_FILE "RPT_PLAYBACK_FILE"
 
 extern struct rpt rpt_vars[MAXRPTS];
 
@@ -1055,14 +1054,6 @@ static void handle_varcmd_tele(struct rpt *myrpt, struct ast_channel *mychannel,
 	ast_debug(5, "Ending telemetry, active_telem = %p, mytele = %p\n", myrpt->active_telem, mytele); \
 	myrpt->active_telem = NULL;
 
-/*! \brief Set a TELEM_TAIL_FILE variable to the selected tail file name
- * and execute the dial plan with extentsion TELEM_TAIL_FILE_EXTN
- */
-inline static enum ast_pbx_result rpt_do_tail_exten(struct ast_channel *chan, const char *filename, const char *context)
-{
-	pbx_builtin_setvar_helper(chan, TELEM_TAIL_FILE, filename);
-	return rpt_do_dialplan(chan, TELEM_TAIL_FILE_EXTN, context);
-}
 /*
  * Threaded telemetry handling routines - goes hand in hand with handle_varcmd_tele (see above)
  * This routine does a lot of processing of what you "hear" when app_rpt is running.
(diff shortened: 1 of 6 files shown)


configs/rpt/extensions.conf
```diff
@@ -217,4 +217,15 @@ same  => n,Playback(${file})
 same  => n,Set(GLOBAL(TAILFILEINDEX)=$[${TAILFILEINDEX} + 1])
 same  => n, Hangup()
 
+; Say the time of day local
+exten => TOD,1,GotoifTime(0:00-11:59,*,*,*?morning:other)
+same  => n(morning),Playback(rpt/goodmorning)
```
InterLinked1, line 222: I believe the standard prompt library already contains prompts for good morning, good afternoon, etc. There shouldn't be any need to use rpt specific ones

Additionally, this context would probably be cleaner with a few ExecIfTimes than a bunch of branching with GotoIfTime

configs/rpt/extensions.conf
```diff
@@ -217,4 +217,15 @@ same  => n,Playback(${file})
...
+same  => n, Goto(time)
+same  => n(other), GotoifTime(12:00-17:59,*,*,*?afternoon:evening)
+same  => n(afternoon), Playback(rpt/goodafternoon)
+same  => n, Goto(time)
```
InterLinked1, line 226: There shouldn't be any space after `same => n,`

configs/rpt/extensions.conf
InterLinked1, line 222: Just a suggestion, but convention is typically to indent `same => n,` by one tab stop. Makes it easier to read

configs/rpt/extensions.conf
```diff
@@ -217,4 +217,15 @@ same  => n,Playback(${file})
...
+same  => n(evening), Playback(rpt/goodevening)
+same  => n(time), Playback(rpt/thetimeis)
+same  => n, SayUnixTime(,EST,IMP)
```
InterLinked1, line 229: Why is EST hardcoded here? Why not just use the system time zone? I think in most cases, that should be correct

configs/rpt/extensions.conf
mkmer: Those audio files are not found in the installed shared folder...

InterLinked1: Hmm, you're right - I thought they were standard prompts, but they might only be in the Pat Fleet library (which is what I use).
Well, the nice thing about dialplan is now users can easily change this depending on what prompts they have, so it doesn't matter that much.

apps/app_rpt/rpt_telemetry.c
```diff
@@ -2429,6 +2430,17 @@ void *rpt_tele_thread(void *this)
 			break;
 		}
 		donodelog_fmt(myrpt, "TELEMETRY,%s,%s", myrpt->name, mytele->mode == STATS_TIME ? "STATS_TIME" : "STATS_TIME_LOCAL");
+
```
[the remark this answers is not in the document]
mkmer: Actaully, I think this chunk of code is saying the time passed in from the remote node via `&t1` 
We then convert it on ln 746 / 747 -> in the end `t1` is the remote nodes time - and that's what we "say" locally.

We definitely would need a channel variable to pass that value out to the dial plan -> I'm thinking people are not customizing this path (or using it) too often. 

If we want to add a dial plan for this, it would be different (or we would pass the time out for both versions to keep TOD generic).  Maybe a PR later to add this?

mkmer:
I'll probably add this (or something like it) as a "way" to eliminate the "o'clock" (in the docs).
```
exten => TOD,1,NoOp(Announce time without 'o'clock')
 same => n,Set(UNIXTIME=${EPOCH})
 same => n,Set(HOUR=${STRFTIME(${UNIXTIME},,%-I)})
 same => n,Set(MINUTE=${STRFTIME(${UNIXTIME},,%-M)})
 same => n,Set(AMPM=${STRFTIME(${UNIXTIME},,%p)})

 ; Check if minute is 0
 same => n,GotoIf($["${MINUTE}" = "0"]?houronly)

 ; Announce full time (e.g., "two fifteen PM")
 same => n,SayUnixTime(${UNIXTIME},,IMp)
 same => n,Hangup()

 ; Hour only case (e.g., "two PM" instead of "two o’clock PM")
 same => n(houronly),SayDigits(${HOUR})
 same => n,SayAlpha(${AMPM})
 same => n,Hangup()
```

apps/app_rpt/rpt_telemetry.c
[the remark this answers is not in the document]
InterLinked1: Using a dialplan function.

Here is an example of this sort of thing: https://github.com/asterisk/asterisk/commit/2198126231ba0bec5df16295805f00a8b21277b4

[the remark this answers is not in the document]
Allan-N: How would that function be used in the dialplan?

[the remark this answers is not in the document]
InterLinked1: The same way a variable is - e.g. `Set(foo=${RECORDING_INFO(duration)})`

Functions didn't exist in Asterisk originally which is why `app_rpt` has none. These days, they are the standard way of doing this sort of thing, and variables aren't used in new code.

It's also more powerful, you could have for example `${RPT(timeofday)}`, `${RPT(foobar)}`, all within one function, etc.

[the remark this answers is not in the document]
InterLinked1: The channel is used for memory, so if `app_rpt` added something to the datastore, the function can retrieve it.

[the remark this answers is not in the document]
Allan-N: Adding a custom function looks to me like a reasonable option (and we've already added one to app_gps.c)

mkmer:
> > I'll probably add this (or something like it) as a "way" to eliminate the "o'clock" (in the docs).
> 
> I don't think this kind of branching is needed. You should be able to just use SayUnixTime, optionally providing a custom format string for the time to skip the o'clock. Specifically, by skipping the 'M' in that case.
> 
> e.g. this whole context could just be:
> 
> ```
> exten => TOD,1,SayUnixTime(,,${IF(${STRFTIME(${UNIXTIME},,%-M)})?IMp:Ip)})
>    same => n,Hangup()
> ```
This appears to work (For some reason it doesn't like testing `${STRFTIME(${UNIXTIME},,%-M)` directly
`SayUnixTime(,,${IF($[${STRFTIME(${UNIXTIME},,%-M) = "0"}]?IMp:Ip)})`

"Final Answer" would look like this:
```
exten => TOD,1,ExecIfTime(0:00-11:59,*,*,*?Playback(rpt/goodmorning))
 same => n,ExecIfTime(12:00-17:59,*,*,*?Playback(rpt/goodafternoon))
 same => n,ExecIfTime(18:00-23:59,*,*,*?Playback(rpt/goodevening))
 same => n,Playback(rpt/thetimeis)
 same => n,SayUnixTime(,,${IF($[${STRFTIME(${UNIXTIME},,%-M) = "0"}]?IMp:Ip)})
 same => n,Hangup()
```

InterLinked1:
> > > I'll probably add this (or something like it) as a "way" to eliminate the "o'clock" (in the docs).
> > 
> > 
> > I don't think this kind of branching is needed. You should be able to just use SayUnixTime, optionally providing a custom format string for the time to skip the o'clock. Specifically, by skipping the 'M' in that case.
> > e.g. this whole context could just be:
> > ```
> > exten => TOD,1,SayUnixTime(,,${IF(${STRFTIME(${UNIXTIME},,%-M)})?IMp:Ip)})
> >    same => n,Hangup()
> > ```
> 
> This appears to work (For some reason it doesn't like testing `${STRFTIME(${UNIXTIME},,%-M)` directly `SayUnixTime(,,${IF($[${STRFTIME(${UNIXTIME},,%-M) = "0"}]?IMp:Ip)})`

That seems odd, since quotes have no significance in dialplan, they are just literal text. If the STRFTIME expression doesn't return a quoted string, I'm not sure how they would be equal. Does it still work if you put quotes around the function too?

> "Final Answer" would look like this:
> 
> ```
> exten => TOD,1,ExecIfTime(0:00-11:59,*,*,*?Playback(rpt/goodmorning))
>  same => n,ExecIfTime(12:00-17:59,*,*,*?Playback(rpt/goodafternoon))
>  same => n,ExecIfTime(18:00-23:59,*,*,*?Playback(rpt/goodevening))

If you wanted to streamline this further, you could even put all the ExecIfTimes in a single priority, by nesting them in the false branches instead of just using the true ones. But maybe it's not worth it here.

(I generally like to optimize the number of lines in my dialplan, but sometimes that comes at the cost of readability)

>  same => n,Playback(rpt/thetimeis)
>  same => n,SayUnixTime(,,${IF($[${STRFTIME(${UNIXTIME},,%-M) = "0"}]?IMp:Ip)})
>  same => n,Hangup()
> ```

Overall, this looks good me - assuming the STRFTIME comparison works

apps/app_rpt/rpt_telemetry.c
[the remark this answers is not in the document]
mkmer: > Adding a custom function looks to me like a reasonable option (and we've already added one to app_gps.c)

I don't see and `func_xxx.c` programs in our repo... Probably confused about what you mean here.

configs/rpt/extensions.conf
```diff
@@ -227,7 +227,8 @@ exten => TOD,1,ExecIfTime(0:00-11:59,*,*,*?Playback(rpt/goodmorning))
 	same => n,ExecIfTime(12:00-17:59,*,*,*?Playback(rpt/goodafternoon))
 	same => n,ExecIfTime(18:00-23:59,*,*,*?Playback(rpt/goodevening))
 	same => n,Playback(rpt/thetimeis)
-	same => n,SayUnixTime(,,IMP)
+	same => n,SayUnixTime(,,IMp) ; Standard "say time"
+	;same => n,SayUnixTime(,,${IF($[${STRFTIME(${UNIXTIME},,%-M) = 0}]?IMp:Ip)}) ; Alternate "stay time" without o'clock
```
Allan-N, line 231: "stay time" --> "say time"

apps/app_rpt/rpt_telemetry.c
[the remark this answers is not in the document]
Allan-N: >I don't see and func_xxx.c programs in our repo... Probably confused about what you mean here.

Correct.  While @KB4MDD did not add a standalone .c file, `app_gps.c` does have includes the code to provide a "function" with the registrations @ https://github.com/AllStarLink/app_rpt/blob/a46026ab7061ab7f9eefe6d8280a342bcfde075b/apps/app_gps.c#L1907-L1909

apps/app_rpt/rpt_telemetry.c
```diff
@@ -37,6 +37,7 @@
 #include "rpt_rig.h"
 
 #define TELEM_TAIL_FILE_EXTN "TAIL"
+#define TELEM_TIME_EXTN "TOD"
```
Allan-N, line 40: Instead of "TOD", might we use a string that's a bit more descriptive ... maybe "TimeOfDay" ... or just "TIME"?

mkmer: TIME :)
I had this finished and local git claimed it was synced... but it was not.
Tis here now

apps/app_rpt/rpt_telemetry.c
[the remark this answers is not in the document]
InterLinked1: Under what circumstances is the time "passed in" and how does that work?
Are you saying it might announce a time that is not the current time?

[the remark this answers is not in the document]
mkmer: > Under what circumstances is the time "passed in" and how does that work? Are you saying it might announce a time that is not the current time?

Yes, if a user triggers the "remote time" -> the local node responds with the time that was passed in here:
https://github.com/AllStarLink/app_rpt/blob/0aab6177e7d97486dd75ba81189c9112755752fe/apps/app_rpt/rpt_telemetry.c#L734-L770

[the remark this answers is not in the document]
mkmer: I suppose we could "always" populate the time register -> either with the local time, or the remote time.  Then just call the `TIME` extension and run the new function to get the time we want to say.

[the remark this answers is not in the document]
Allan-N: >I think we will need a second extension for playing "remote" time. There's no other way to "know" who is "calling". If we "pull" the values with a dialplan function, we would still need to "know" that we need to pull it (I'm suggesting via the extension that was called")
It may be possible to clean up in internal code to use one function.
It's possible I'm missing something you had in mind?

Both chunks of code say "good xxx, the time is ...".  The difference is that one uses the local time, the other is passed as an arg.  I'm suggesting that we take whatever time we want spoken and pass that to a single extension using a function variable.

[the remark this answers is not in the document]
InterLinked1: > I suppose we could "always" populate the time register -> either with the local time, or the remote time

Or do so for the remote time, and not for the local - in which case you know to announce the local by the absence of this info (i.e. function returns empty value)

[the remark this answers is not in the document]
mkmer: Internally, I have made a common "rpt_say_time" function.  
Now we need to store time value, create a function to pull the time value, and update the dialplan to use the extracted time. 
Pointers on storing and extracting?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -37,9 +37,75 @@
...
+
+static const struct ast_datastore_info telemetry_datastore = { .type = TELEM_DATASTORE, .destroy = rpt_telem_callback_destroy };
+
+int rpt_telem_read_datastore(struct ast_channel *chan, const char *cmd, char *data, char *buf, size_t len)
+{
+	struct ast_datastore *datastore = NULL;
+	time_t *value;
+	datastore = ast_channel_datastore_find(chan, &telemetry_datastore, NULL);
+	if (!datastore || !datastore->data) {
+		snprintf(buf, len, "X"); // Default or error value
```
Allan-N, line 63: "X" ?  Or should we return -1 ?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -37,9 +37,98 @@
...
+		return 0;
+	}
+
+	value = (time_t *) datastore->data;
+	snprintf(buf, len, "%ld", (long) *value); // Convert time_t to string for dialplan
+	return 0;
+}
+
```
Allan-N, line 71: Just a thought.  Might we add another helper:
```
static const char *rpt_telem_extension(struct ast_channel *chan, const char *primary, const char *alternate) {
    if (ast_exists_extension(mychannel, primary, TELEM_TIME_EXTN, 1, NULL)) {
        return primary;
    }
    if (ast_exists_extension(mychannel, alternate, TELEM_TIME_EXTN, 1, NULL)) {
        return alternate;
    }
    return NULL;
}
```
and then callers would use :
```
    const char *ext = rpt_telem_extension(mychannel, myrpt->p.telemetry, TELEMETRY);
    if (ext) {
        struct ast_datastore *datastore = ast_datastore_alloc(&telemetry_datastore, NULL);
        ...
```

apps/app_rpt/rpt_telemetry.c
```diff
@@ -37,9 +37,98 @@
...
+ * This function is used to free the data stored in the telemetry datastore.
+ */
+static void rpt_telem_callback_destroy(void *data)
+{
+	ast_free(data);
+}
+
+static const struct ast_datastore_info telemetry_datastore = { .type = TELEM_DATASTORE, .destroy = rpt_telem_callback_destroy };
+
+int rpt_telem_read_datastore(struct ast_channel *chan, const char *cmd, char *data, char *buf, size_t len)
```
InterLinked1, line 57: I don't really like the idea of splitting parts of the function between files like this - i.e. rpt_read_telem_datastore_function is in app_rpt.c, and then rpt_telem_read_datastore has to be made public in order to be accessed from there.

It would be cleaner maybe to replace the direct function register/unregister in app_rpt with a telemetry_init/cleanup call to this file, where all the function stuff is declared. Technically one more public declaration, but if we add more functions here, we don't have to modify the header files. What do you think?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -37,9 +37,98 @@
...
+{
+	struct ast_datastore *datastore = NULL;
+	time_t *value;
+	datastore = ast_channel_datastore_find(chan, &telemetry_datastore, NULL);
+	if (!datastore || !datastore->data) {
+		snprintf(buf, len, "X"); // Default or error value
```
InterLinked1, line 63: It doesn't look like this was already there, so no // comments

apps/app_rpt/rpt_telemetry.c
```diff
@@ -37,9 +37,98 @@
...
+
+static int rpt_telem_datastore(struct ast_channel *chan, time_t t)
+{
+	struct ast_datastore *datastore = ast_datastore_alloc(&telemetry_datastore, NULL);
+	time_t *time_data = NULL;
+
+	if (!datastore) {
+		return -1;
+	}
+	time_data = ast_malloc(sizeof(time_t));
```
InterLinked1, line 86: Do you think it's possible in the future, we might ever want to store anything else?
If so, I would add a struct with a time_t member rather than using time_t directly. That allows for future expansion/flexibility

apps/app_rpt/rpt_telemetry.c
```diff
@@ -37,9 +37,98 @@
...
+ * \param data The data to free
+ *
+ * This function is used to free the data stored in the telemetry datastore.
+ */
+static void rpt_telem_callback_destroy(void *data)
+{
+	ast_free(data);
+}
+
+static const struct ast_datastore_info telemetry_datastore = { .type = TELEM_DATASTORE, .destroy = rpt_telem_callback_destroy };
```
InterLinked1, line 55: I think the members need to be on separate lines per the coding style

apps/app_rpt/rpt_telemetry.c
mkmer: The example says to `return 0` with a string that "means something".  If I returned a "0" the time would report 7AM -> if I return "X", it reports the current time.
I suspect any non numeric value in the `UNIXSAYTIME()` function will just say the current time.

apps/app_rpt/rpt_telemetry.c
mkmer: I thought so too, but the lint tool want's to put it back in one line (probably due to the short length).  I can override, or is this fine for a short one?

apps/app_rpt/rpt_telemetry.c
mkmer: Makes sense to me.  I was not really happy with it either, just couldn't come up with a "clean" way when I was writing the code.

apps/app_rpt/rpt_telemetry.c
```diff
@@ -37,9 +37,134 @@
...
+ */
+static void rpt_telem_callback_destroy(void *data)
+{
+	ast_free(data);
+}
+
+static const struct ast_datastore_info telemetry_datastore = { 
+	.type = TELEM_DATASTORE, 
+	.destroy = rpt_telem_callback_destroy 
+};
```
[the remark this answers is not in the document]
mkmer: Need to workout clang rules to fix this.  Let's just push it in this way.

apps/app_rpt/rpt_telemetry.h
```diff
@@ -13,3 +14,24 @@ void *rpt_tele_thread(void *this);
...
+/*!
+ * \brief Read time_t from telemetry datastore
+ * \param chan The channel to read from
+ * \param cmd The command to execute
+ * \param data Pointer to the data buffer
+ * \param buf Buffer to store the result
+ * \param len Length of the buffer
+ * \return 0 on success, -1 on failure
+ */
+int rpt_telem_read_datastore(struct ast_channel *chan, const char *cmd, char *data, char *buf, size_t len);
```
Allan-N, line 27: This function is now fully contained in rpt_telemetry.c ... so it need not be here in the .h file and can be marked "static"

configs/rpt/extensions.conf
```diff
@@ -224,5 +222,13 @@ exten => s,1,Ringing
 ;	same => n,Set(GLOBAL(TAILFILEINDEX)=$[${TAILFILEINDEX} + 1]) 
 ;  	same => n,Hangup()
 
+; Say the time of day local
+exten => TIME,1,ExecIfTime(0:00-11:59,*,*,*?Playback(rpt/goodmorning))
```
Allan-N, line 226: What time are we comparing against here ?  Are we looking at RPT_TELEM_TIME() ?

configs/rpt/extensions.conf
```diff
@@ -224,5 +222,13 @@ exten => s,1,Ringing
 ;	same => n,Set(GLOBAL(TAILFILEINDEX)=$[${TAILFILEINDEX} + 1]) 
 ;  	same => n,Hangup()
 
+; Say the time of day local
```
Allan-N, line 225: As I had commented in the ASL3 Manual doc PR :

"Say the time of day" (without "local", "global", ...)

configs/rpt/extensions.conf
mkmer: I think we can just do this:
```
same => n,Set(HOUR=${STRFTIME(${RPT_TELEM_TIME()},,%-I)})
```
Then check the HOUR for the ranges....

Allan-N: Like :)

Q? do you set a local variable with one call to ${RPT_TELEM_TIME()} and then work with that local variable as needed ... to avoid having to make multiple round trips into app_rpt?

mkmer: We could (there was an earlier example of getting the epoch time into a var, then using `strftime` to split it into the H, M, and S values. I think we would do that and get RPT_TELEM_TIME() (which is epoch time)

The customization of this is for dialplan gurus... I'm thinking these are "cookbook" dialplans that people can start with.  Since folks can (and will) change , it's not like our hard coded stuff.

mkmer: It would say "good morning (Allan), the time is 2PM" -> as it's morning for you (10am), and it's 2PM on the remote repeater :)

Like I was saying -> I think it's odd if it said "good afternoon" when it's 10am your time.
It can be made to convert the reported time and say the preamble based on the conversion, but "I" don't think it should...

I'm not sure how many people actually use this path either.

Allan-N: > I'm not sure how many people actually use this path either.

Agreed.  And, fortunately, with this being the dialplan they could customize to their liking ... "Good afternoon (Mike), the remote time is nineteen hundred U T C"

PR #741, opened 2025-08-03
app_rpt: fix connect time reported by "rpt lstats" command

Allan-N:


Allan-N:
>Just a bit more than I first noticed. 

The "fix" was to ensure we initialized `s` before using the content.  The rest of the change was cleanup and optimization.

PR #744, opened 2025-08-05
app_rpt: Add timeout reset delay via `toresettime` parameter.

mkmer:
Picket fencing on a stuck key causes timeout to reset too quickly.  Need a bit of delay on unkey before resetting the Timeout Timer and rekeying the machine.
This can be disabled to match previous behavior.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2892,6 +2892,7 @@ static inline void dump_rpt(struct rpt *myrpt, const int lasttx, const int laste
 	ast_debug(2, "myrpt->tonotify = %d\n", myrpt->tonotify);
 	ast_debug(2, "myrpt->retxtimer = %d\n", myrpt->retxtimer);
 	ast_debug(2, "myrpt->totimer = %d\n", myrpt->totimer);
+	ast_debug(2, "myrpt->toresettimer = %d\n", myrpt->toresettimer);
 	ast_debug(2, "myrpt->tailtimer = %d\n", myrpt->tailtimer);
 	ast_debug(2, "myrpt->tailevent = %d\n", myrpt->tailevent);
 	ast_debug(2, "myrpt->linkactivitytimer = %d\n", myrpt->linkactivitytimer);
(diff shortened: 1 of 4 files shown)


apps/app_rpt.c
```diff
@@ -5006,15 +5007,18 @@ static void *rpt(void *this)
...
 		/* add in parrot stuff */
 		totx = totx || (myrpt->parrotstate > 1);
-		/* Reset time out timer variables if there is no activity */
-		if (!totx) {
+
+		if (!myrpt->totimer && totx) { /* we are timed out, set toresettimer delaying reset of totimer. */
+			myrpt->toresettimer = myrpt->p.toresettime;
+		}
+		/* If we have a p.toresettime, delay totimer reset until unkeyed for p.toresettime */
+		if ((!totx && (myrpt->totimer || !myrpt->p.toresettime)) || (!myrpt->toresettimer && myrpt->p.toresettime)) {
```
Allan-N, line 5015: This "if" statement seems way too complex.  If it needs to be as-is can we at least add additional comments to explain all of the if's/and's/or's?

apps/app_rpt/app_rpt.h
```diff
@@ -466,6 +466,7 @@ struct rpt_chan_stat {
 #define	HANGTIME 5000
 #define SLEEPTIME 900		/* default # of seconds for of no activity before entering sleep mode */
 #define	TOTIME 180000
+#define TORESETTIME 1000 /* default timeout reset time to 1 second */
```
Allan-N, line 469: The value is in milliseconds (ms)?

configs/rpt/rpt.conf
```diff
@@ -127,8 +127,8 @@ accountcode = RADIO                 ; account code (optional)
 
 hangtime = 2000                     ; squelch tail hang time (in ms) (optional, default 5 seconds, 5000 ms)
 althangtime = 4000                  ; longer squelch tail
-totime = 180000                     ; transmit time-out time (in ms) (optional, default 3 minutes 180000 ms)
-
+totime = 180000                     ; transmit time-out time (in ms) (optional, default 3 minutes 180000 ms, maximum 9999999 ms)
+toresettime = 1000                  ; transmit time-out reset time (in ms) (optional, default to 1 seconds 10000 ms, maximum 10 seconds)
```
Allan-N, line 131: How about `default to 1000 ms (1s), maximum 10000 ms (10s)`?

apps/app_rpt.c
Allan-N: Thank you for adding the words.

FWIW, AI suggested something like (and I'm not saying this is any better then the added words) :
```
bool need_action;

if (!myrpt->p.toresettime) {
    /* reset-timer feature not in use, just respect plain TOT */
    need_action = !totx;
} else {
    /* reset-timer feature in use, act if reset interval expired or normal TOT hit */
    need_action = !myrpt->toresettimer || (!totx && myrpt->totimer);
}

if (need_action) {
    /* original body */
}

PR #749, opened 2025-08-07
app_rpt.c: Remove totx pointer

mkmer:
totx pointer is NOT used across functions and is not required to pass into `process_link_channels`

The while loop initializes totx, all evaluation of totx occurs before the call to `process_link_channels`.  The "final" output from `process_link_channels` is reset to 0.

Additionally, each loop of the links inside of `process_link_channels()` is assigning values to totx (thus tossing out any previous values).


InterLinked1 (commented):
Looks good, but could you wrap the commit message at ~72-80 characters? The title could also be shorter/moved into the description

PR #752, opened 2025-08-09
remove chan_beagle

okcrum:
Makefile.diff, now to remove some things


InterLinked1 (approved):

okcrum:
Logic fix needed from looking at clint

Issue #209, opened 2023-08-21
Erroneous jb_warning_output


tsawyer:
We talked on Slack about turning off the IAX2 jitter buffer to prevent these messages. I found that, due to the extensive use of the internet to connect nodes, that's not practical. 

Is this a bug in IAX2 or are we setting the JB incorrectly?
```
[2023-08-21 10:42:47.058] WARNING[517779][C-00000039]: chan_iax2.c:1237 jb_warning_output: Resyncing the jb. last_delay 279, this delay 1697, threshold 1074, new offset -1684
[2023-08-21 10:42:54.528] WARNING[517778][C-00000039]: chan_iax2.c:1237 jb_warning_output: Resyncing the jb. last_delay 0, this delay -1699, threshold 1074, new offset 15
```

Typical AllStar codec and JB settings.
```
disallow = all   ; The permitted codecs for outgoing connections
;Audio Quality   Bandwidth
;allow = ulaw    ; best    87 kbps
;allow = adpcm   ; good    55 kbps
;allow = gsm     ; medicore    36 kbps
allow = ulaw
allow = adpcm
allow = gsm


jitterbuffer = yes
forcejitterbuffer = yes
dropcount = 2
maxjitterbuffer = 4000
maxjitterinterps = 10
resyncthreshold = 1000
maxexcessbuffer = 80
minexcessbuffer = 10
jittershrinkrate = 1
tos = 0x1E
autokill = yes
delayreject = yes
; iaxthreadcount = 30
; iaxmaxthreadcount = 150

```

PR #775, opened 2025-08-24
app_rpt: Tune jb settings

okcrum:
The settings we use as defaults for the jitterbuffer are not very useful. They appear to be an attempt to mitigate issue #209. After a number of measurements, I propose the indicated default values for `maxjitterbuffer` and `resyncthreshold`:

maxjitterbuffer = 1000
resyncthreshold = 500           ; set to -1 to disable

These settings are adequate for the vast majority of nodes.

Review of the jitter buffer changes and warnings
The current default jitterbuffer is adaptive, so we are setting its maximum size to 1000 ms. I have seen it reach 450-500 ms when jitter is running 300-350 ms and packet loss is > 10% with no artifacts heard. Jitterbuffer warnings indicate a frame timestamp issue: 3 consecutive audio frames or 1 control frame that are more than `(2 * jitter) + resyncthreshold` out of time from the current value. This is a large value at even 500 ms, and the jb warnings tend to happen in pairs, where the timestamps jump out of spec and then right back in to spec. I see these rarely, 2 in the last month.

Summary
The proposed values are good upper limits on real world observed connections compared to what we currently ship. A comment is added that one can set `resyncthreshold` to -1 if getting excessive warnings that can't be otherwise fixed.


mkmer (approved):
Seems like a good update.   I think 4 seconds of jitter is essentially unusable link anyway :)

InterLinked1 (approved):

Issue #774, opened 2025-08-22
Remove "RANGER" Special Handling for Nodes starting with 9


jxmx:
Since the RANGER concept is long gone, don't treat nodes starting with 9 specially.

PR #776, opened 2025-08-27
app_rpt: Remove "RANGER" Special Handling for Nodes starting with 9

okcrum:
RANGER is a deprecated project/protocol

Resolves: #774


mkmer (approved):
Nice to have a bit of history to know what can go!
You should update the docs to match (if you haven't) over in: https://allstarlink.github.io/config/rpt_conf/?h=ranger#link-commands

apps/app_rpt/rpt_config.c
```diff
@@ -855,21 +855,21 @@ void load_rpt_vars(int n, int init)
...
 	if (!val) {
-		rpt_vars[n].p.althangtime = (ISRANGER(rpt_vars[n].name) ? 1 : HANGTIME);
+		rpt_vars[n].p.althangtime = HANGTIME;
 	}
 	if (rpt_vars[n].p.althangtime < 1) {
 		rpt_vars[n].p.althangtime = 1;
 	}
 
-	RPT_CONFIG_VAR_INT_DEFAULT(totime, "totime", (ISRANGER(rpt_vars[n].name) ? 9999999 : TOTIME));
+	RPT_CONFIG_VAR_INT_DEFAULT(totime, "totime", TOTIME);
```
mkmer, line 872: This one was a mistake on my part.  We were supposed to be limiting the range of the value not checking for RANGER.  The current main branch has the correct code here.

Allan-N (changes requested):
I still see some references to RANGER in `rpt.conf`

Allan-N (approved):

PR #784, opened 2025-09-10
app_rpt.c: Handle `duplex=0` with `linktolink=yes`

### LINKED ISSUE #773: 'rpt playback' behaves like 'rpt localplay' when duplex=0
Whenever a node's configuration is set to `duplex = 0`, the command `rpt playback` behaves like `rpt localplay` in that the audio is not sent out to other connected nodes. When using `duplex=0` and `linktolink=yes` to interface an ASL node with a repeater controller (i.e. it is not the controller itself), it's convenient to allow local audio to be played back to the whole network.

mkmer:
Allowing cli rpt playback to send audio out the links as expected.

Fixes #773 

This one appears to have been around since ASL2.  Further testing to confirm it doesn't break anything else would be appreciated as I'm not sure I completely understand what this configuration is doing.

### DIFF
--- a/apps/app_rpt/rpt_link.c
+++ b/apps/app_rpt/rpt_link.c
@@ -182,7 +182,7 @@ int altlink1(struct rpt *myrpt, struct rpt_link *mylink)
 			tlist = tlist->next;
 		}
 	}
-	if ((!myrpt->p.duplex) || (!nonlocals)) {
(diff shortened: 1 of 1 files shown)


Allan-N (approved):

Allan-N:
> `localplay` sends nothing. Is this also desired behavior?

I would expect `localplay` to always play "locally".  How much telemetry should be suppressed with `duplex = 0`?

PR #787, opened 2025-09-23
app_rpt.c: Allow Tone macro on each keyup by adding "R" to macro string

mkmer:
Current design only executes tonemacro the first time a CTCSS tone is received.  The added "R" to the macro string allows for execution of the `tonemacro` for each keyup where the tone is present.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3884,11 +3884,11 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 					if (!repeat) { /* Small optimization, only copy the string if we care what it was */
 						ast_copy_string(myrpt->lasttone, f->data.ptr, sizeof(myrpt->lasttone));
 					} else { /* otherwise do the lower cost clearing */
-						myrpt->lasttone[0] = \0;
+						myrpt->lasttone[0] = '\0';
 					}
 				}
 			} else { /* There is no tone for this keyup */
-				myrpt->lasttone[0] = 0;
+				myrpt->lasttone[0] = '\0';
 				send_link_pl(myrpt, "0");
 			}
 		}


mkmer:
I think it's a great idea - this way it's not "all or nothing" - we can have the "best of both worlds" :)
This should now behave as you were thinking (not too big of a change)

apps/app_rpt.c
```diff
@@ -3865,22 +3865,29 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
...
+					ast_copy_string(myrpt->lasttone, f->data.ptr, sizeof(myrpt->lasttone));
 					rpt_mutex_unlock(&myrpt->lock);
-				} else if (strcmp((char *) f->data.ptr, myrpt->lasttone)) {
-					val = (char *) ast_variable_retrieve(myrpt->cfg, myrpt->p.tonemacro, (char *) f->data.ptr);
+				} else {
+					val = ast_variable_retrieve(myrpt->cfg, myrpt->p.tonemacro, f->data.ptr);
 					if (val) {
-						ast_debug(1, "Tone %s doing %s on node %s\n", (char *) f->data.ptr, val, myrpt->name);
-						macro_append(myrpt, val);
+						repeat = strchr(val, TONEMACRO_REPEAT);
```
Allan-N, line 3879: TONEMACRO_REPEAT is a character, not a string.

To test for a leading 'R', you'd want to check if `(val[0] == TONEMACRO_REPEAT)`

apps/app_rpt.c
```diff
@@ -3865,22 +3865,29 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
...
+						if (repeat) { /* remove repeat char from tone string */
+							ast_copy_string(repeat, repeat + 1, strlen(repeat));
+						}
+						/* If this is a new tone or the tone string contains the repeat command, execute the macro */
+						if (strcmp(f->data.ptr, myrpt->lasttone) || repeat) {
+							ast_debug(1, "Tone %s doing %s on node %s\n", (char *) f->data.ptr, val, myrpt->name);
+							macro_append(myrpt, val);
```
Allan-N, line 3886: I would remove lines 3880-3882 and change this line to :
```
							macro_append(myrpt, repeat ? val + 1 : val);
```

apps/app_rpt.c
```diff
@@ -3865,22 +3865,29 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
...
+						}
 					}
 					rpt_mutex_lock(&myrpt->lock);
```
Allan-N, line 3889: Do you want to put these lines (3889-3891) inside an "if (!epeat) { ... }" block ... so that you only save the tone when not repeating?

apps/app_rpt.c
mkmer: I was looking for R anywhere in the string, but I suppose since we are now expecting it leading, we can look for it there.

apps/app_rpt.c
```diff
@@ -3865,22 +3865,26 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
...
-				} else if (strcmp((char *) f->data.ptr, myrpt->lasttone)) {
-					val = (char *) ast_variable_retrieve(myrpt->cfg, myrpt->p.tonemacro, (char *) f->data.ptr);
+				} else {
+					val = ast_variable_retrieve(myrpt->cfg, myrpt->p.tonemacro, f->data.ptr);
 					if (val) {
-						ast_debug(1, "Tone %s doing %s on node %s\n", (char *) f->data.ptr, val, myrpt->name);
-						macro_append(myrpt, val);
+						repeat = (val[0] == TONEMACRO_REPEAT);
+						/* If this is a new tone or the tone string contains the repeat command, execute the macro */
+						if (strcmp(f->data.ptr, myrpt->lasttone) || repeat) {
```
Allan-N, line 3881: Minor: if you swap the conditions. checking "repeat"' first, then we can avoid the strcmp().  I.e.
```
						if (repeat || strcmp(f->data.ptr, myrpt->lasttone)) {
```

apps/app_rpt.c
Allan-N: Yes, you are correct.  I was thinking that if `repeat` was true (non-zero) then we would fall to the `else { myrpt->lasttone[0] = 0;` which is much quicker/efficient than copying the string that need not be saved 'cause it won't be compared against.

apps/app_rpt.c
mkmer: I don't disagree. It made sense when I was writing it, but it doesn't make sense now :)

Allan-N (approved):

Issue #739, opened 2025-08-02
The lstats command is not keeping the node connect time column with the node number as nodes connect


wd5m:
The connect time for one node moves to a new node when it connects.  I suspect a recent [change](https://github.com/AllStarLink/app_rpt/pull/627) to lstats sorting may have caused this, but I have not looked at that change details. See lstats examples below. 

First list lstats with a node (55436) that has been connected for a long time.

```
asl3*CLI> rpt lstats 516226
NODE      PEER                RECONNECTS  DIRECTION  CONNECT TIME        CONNECT STATE
----      ----                ----------  ---------  ------------        -------------
55436     44.31.122.26        3           OUT        47:26:06:564        ESTABLISHED
```

Now list lstats after a new node (516224) connects.
```
asl3*CLI> rpt lstats 516226
NODE      PEER                RECONNECTS  DIRECTION  CONNECT TIME        CONNECT STATE
----      ----                ----------  ---------  ------------        -------------
55436     44.31.122.26        3           OUT        00:00:04:990        ESTABLISHED
516224    216.147.121.150     0           IN         47:26:32:863        ESTABLISHED
```

I'm on the latest ASL3 release.
root@asl3:~# asl-show-version
********** AllStarLink [ASL] Version Info **********

OS            : Debian GNU/Linux 12 (bookworm)
OS Kernel     : 6.1.0-37-amd64

Asterisk      : 22.4.1+asl3-3.5.4-1.deb12
ASL [app_rpt] : 3.5.4

Installed ASL packages :

  Package                         Version
  ==============================  ==============================
  allmon3                         1.5.1-1.deb12
  asl3                            3.8-1.deb
  asl3-asterisk                   2:22.4.1+asl3-3.5.4-1.deb12
  asl3-asterisk-config            2:22.4.1+asl3-3.5.4-1.deb12
  asl3-asterisk-modules           2:22.4.1+asl3-3.5.4-1.deb12
  asl3-menu                       1.14-1.deb12
  asl3-update-nodelist            1.5.1-1.deb12
  dahdi                           1:3.1.0-2
  dahdi-dkms                      1:3.4.0-6+asl
  dahdi-linux                     1:3.4.0-6+asl

Allan-N:
I "think" I have this figured out.  Need to do some testing ... tomorrow :)

PR #795, opened 2025-10-01
Workflow

mkmer:
Permissions adjustments

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -6,6 +6,7 @@ on:
 
 permissions:
   contents: read
+  metadata: read
   pull-requests: write
 
 jobs:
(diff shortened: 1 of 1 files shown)


InterLinked1:
> > I don't understand what either the problem or the solution is here - the commit message has no useful information, and #739 is an issue, not a PR.
> 
> Aht yes, the issue and PR are same last two digits swapped (Yikes). The problem: When the PR created by a "non" authorized user is executed the comments can not be posted. I found a similar problem where the solution was adding write permissions "fixes" the 403 error:
> 
> ```
> ERROR:CPP Linter:response returned 403 from POST https://api.github.com/repos/AllStarLink/app_rpt/issues/793/comments with message: {"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/issues/comments#create-an-issue-comment","status":"403"}
> ```
> 
> Without the comments, the user has no clue what it clang-format doesn't like.

Thanks for the explanation, now I understand. Would be nice to include the `When the PR created by a "non" authorized user is executed the comments can not be posted. I found a similar problem where the solution was adding write permissions "fixes" the 403 error` part in the commit message.

Issue #763, opened 2025-08-14
channel.c:1105 __ast_queue_frame: Exceptionally long queue length


ve7fet:
So, my ASL3 went off the rails tonight.

This is ASL3 (versions below), running on a Pi 4. It has three nodes, two are `chan_voter` (435451/435452) and one is `chan_simpleusb` (1999).

Received an email from a member that the repeater had been stuck on transmit. "for a very long time". I checked Allmon3, and it appeared to not be able to connect to the manager, as nothing was being shown for the nodes. 

Logged in via SSH, Asterisk was running. Connected to the console, and started seeing warnings from `channel.c`:

```
Connected to Asterisk 22.4.1+asl3-3.5.4-1.deb12 currently running on ve7rpt (pid = 2001)
[2025-08-13 22:18:32.866] WARNING[2033]: channel.c:1105 __ast_queue_frame: Exceptionally long queue length (0 voice / 398 total) queuing to IAX2/44.32.40.212:4569-4215
[2025-08-13 22:18:42.866] WARNING[2040]: channel.c:1105 __ast_queue_frame: Exceptionally long queue length (0 voice / 399 total) queuing to IAX2/44.32.40.212:4569-4215
[2025-08-13 22:18:52.866] WARNING[2039]: channel.c:1105 __ast_queue_frame: Exceptionally long queue length (0 voice / 400 total) queuing to IAX2/44.32.40.212:4569-4215
```

Did a `module reload` from the CLI, and got dumped back to my shell prompt. Reconnected to the Asterisk CLI to catch some of the messages:

```
Connected to Asterisk 22.4.1+asl3-3.5.4-1.deb12 currently running on ve7rpt (pid = 2001)
[2025-08-13 22:18:32.866] WARNING[2033]: channel.c:1105 __ast_queue_frame: Exceptionally long queue length (0 voice / 398 total) queuing to IAX2/44.32.40.212:4569-4215
[2025-08-13 22:18:42.866] WARNING[2040]: channel.c:1105 __ast_queue_frame: Exceptionally long queue length (0 voice / 399 total) queuing to IAX2/44.32.40.212:4569-4215
[2025-08-13 22:18:52.866] WARNING[2039]: channel.c:1105 __ast_queue_frame: Exceptionally long queue length (0 voice / 400 total) queuing to IAX2/44.32.40.212:4569-4215
```

Repeater stopped transmitting, and normal operations appeared to have resumed.

[...]

PR #803, opened 2025-10-16
app_rpt: prevent null pointer when calling `softhangup()`, observed in issue #763

mkmer:
The first core dump while trying to reload the repeater found in issue #763 : Attempted to `softhangup()` with a null channel pointer.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4796,16 +4796,8 @@ static void *rpt(void *this)
 		}
 
 		if (myrpt->reload) {
-			struct rpt_tele *telem;
-
-			rpt_mutex_lock(&myrpt->lock);
-			telem = myrpt->tele.next;
-			while (telem != &myrpt->tele) {
-				ast_softhangup(telem->chan, AST_SOFTHANGUP_DEV);
-				telem = telem->next;
-			}
+			flush_telem(myrpt);
 			myrpt->reload = 0;
-			rpt_mutex_unlock(&myrpt->lock);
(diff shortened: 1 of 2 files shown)


InterLinked1 (approved):

apps/app_rpt.c
```diff
@@ -4796,16 +4796,8 @@ static void *rpt(void *this)
...
 		if (myrpt->reload) {
-			struct rpt_tele *telem;
-
-			rpt_mutex_lock(&myrpt->lock);
-			telem = myrpt->tele.next;
-			while (telem != &myrpt->tele) {
-				ast_softhangup(telem->chan, AST_SOFTHANGUP_DEV);
-				telem = telem->next;
-			}
+			flush_telem(myrpt);
```
Allan-N: The `flush_telem()` function ensures that `telem->mode != SETREMOTE`.  Is it OK that we are also adding that test here?

mkmer: I believe it is fine, as the SETREMOTE option (as well as any of the other "non wait" options) is moving on without waiting.
https://github.com/AllStarLink/app_rpt/blob/ecd4eedbe9b03896c6ba3f8fca2a98e3dced310c/apps/app_rpt/rpt_telemetry.c#L1304-L1306
This telem "should" hangup by itself before we get past https://github.com/AllStarLink/app_rpt/blob/ecd4eedbe9b03896c6ba3f8fca2a98e3dced310c/apps/app_rpt.c#L4810

PR #804, opened 2025-10-16
app_rpt: on update_parrot() and update_timers() failure, add missing call to rpt_mutex_unlock()

mkmer:
It appears we are locked at these locations and should unlock before leaving the while loop.  Of course this only causes trouble when things "fail" and we try to shut down the repeater thread.


apps/app_rpt.c
InterLinked1: I see we unlock here, but where do we lock again after the unlocks above? Seems like we have double unlocks here (even prior to this patch)

Allan-N: The `break;` statements get us out of the big `while (ms >= 0) {` loop that starts @ l.4786

Unless I'm missing something, we don't need to lock again.

Allan-N (approved):

PR #806, opened 2025-10-16
app_rpt: add missing unlock "rainy day" path, remove double lock in rpt_manager_do_xstat()

mkmer:
Double lock in `rpt_manager_do_xstat()` is likely related to issue #763

### DIFF
--- a/apps/app_rpt/rpt_functions.c
+++ b/apps/app_rpt/rpt_functions.c
@@ -672,8 +672,8 @@ enum rpt_function_response function_remote(struct rpt *myrpt, char *param, char 
 		myrpt->loginlevel[0] = 0;
 		cp = ast_strdup(param);
 		cp1 = strchr(cp, ',');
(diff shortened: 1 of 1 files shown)


apps/app_rpt/rpt_functions.c
```diff
@@ -685,7 +685,10 @@ enum rpt_function_response function_remote(struct rpt *myrpt, char *param, char
 			donodelog_fmt(myrpt, "LOGIN,%s,%s", myrpt->loginuser, myrpt->loginlevel);
 			ast_debug(1, "loginuser %s level %s\n", myrpt->loginuser, myrpt->loginlevel);
 			rpt_telemetry(myrpt, REMLOGIN, NULL);
+		} else {
```
Allan-N: Rather than adding the `else` statement would it make more sense to move the lock from l.675 into the `if`?

PR #810, opened 2025-10-20
app_rpt: Use ao2 container for link management

mkmer:
One of the Asterisk requirements is "don't roll your own linked list".  This is the first step at replacing app_rpt linked lists with ao2 container, utilizing Asterisk linked lists.

If we move forward with this change, other lists will be updated in additional PR's.

## Summary by CodeRabbit

* **Refactor**
  * Improved link management system architecture for enhanced reliability and resource handling.
  * Strengthened lifecycle management of link operations and connections.

* **Bug Fixes**
  * Better error handling and cleanup during link operations and disconnections.

### DIFF
--- a/apps/app_rpt/rpt_manager.c
+++ b/apps/app_rpt/rpt_manager.c
@@ -645,7 +645,7 @@ static int rpt_manager_do_stats(struct mansession *s, const struct message *m, s
 			RPT_LIST_TRAVERSE(links_copy, l, l_it) {
 				j++;
 				ast_str_append(&str, 0, "%s", l->name);
-				if (j < numoflinks - 1) {
+				if (j < numoflinks) {
 					ast_str_append(&str, 0, ",");
 				}
 			}


mkmer:
> Just curious, what was the design consideration steering you to ao2 lists as opposed to the AST_LIST / AST_RWLIST APIs?

Reading the docs, it sounded like ao2 was the way to go.  (It may not be).  As I researched what needed to be done, I found most of the new/current modules are using it. 

We REALLY don't need a double linked list for this, so there is that (granted AST_LIST is not double linked either).

mkmer:
Sorry about the major refactor mid PR here.  I found several mistakes on references after some run time.  I think it makes more sense to use `for()` in place of `while()` to minimize the number of `ast_ref()` calls, thus reducing possibility of mistakes.

mkmer:
Following up on this statement:
> We REALLY don't need a double linked list for this, so there is that (granted AST_LIST is not double linked either).

After a bit more review, we really should use a double linked list for the links.  We often remove a node from the middle of the list, so it's much more efficient to remove a double linked list item (or an AO2 object).

apps/app_rpt/rpt_link.h
```diff
@@ -4,6 +4,40 @@
...
+* ...
+* RPT_AO2_LIST_TRAVERSE(&container, current, iterator) {
+	(do something with current here)
+* }
+* ao2_iterator_destroy(&iterator);
+* \endcode
+
+*/
+
+#define RPT_AO2_LIST_TRAVERSE(container, var, iterator) \
```
Allan-N: I like the TRAVERSE macro.  Just not wild about then name.  What do you think about changing the name to `RPT_LIST_TRAVERSE()` ?  and possibly elevating the macro to `app_rpt.h`?

and, if we want to have "ao2" as an implementation detail, maybe think about :
```
typedef struct ao2_container rpt_list;
typedef ao2_iterator_t rpt_list_iterator;
#define rpt_link_ref(list) ao2_ref(list)
#define rpt_list_iterator_destroy(iterator) ao2_iterator_destroy(iterator)
...
```

apps/app_rpt/rpt_link.h
```diff
@@ -42,19 +76,19 @@ void send_link_dtmf(struct rpt *myrpt, char c);
 
 void send_link_keyquery(struct rpt *myrpt);
 
-/*!
+/*! 1Code has comments. Press enter to view.
```
Allan-N: Huh?

apps/app_rpt/rpt_link.c
Allan-N: In addition, we spend cycles to compute `numoflinks` with the return value not being used and really has nothing to do with the Rssi.  I'd suggest we drop the count and change the function to a void return.

apps/app_rpt/rpt_link.h
```diff
@@ -4,6 +4,40 @@
...
+* static ao2_container container;
+* ao2_iterator_t iterator;
+* ...
+* struct list_entry {
+*    ...
+* }
+* ...
+* struct list_entry *current;
+* ...
+* RPT_AO2_LIST_TRAVERSE(&container, current, iterator) {
```
Allan-N: FYI: your example here has the `{` brace in the same line as RPT_AO2_LIST_TRAVERSE but there's many places in this PR that have moved the brace to the next line.

apps/app_rpt/rpt_cli.c
```diff
@@ -117,26 +117,26 @@ static int rpt_do_stats(int fd, int argc, const char *const *argv)
 			/* Traverse the list of connected nodes */
 			reverse_patch_state = "DOWN";
 			numoflinks = 0;
-			l = myrpt->links.next;
-			while (l && (l != &myrpt->links)) {
+			RPT_AO2_LIST_TRAVERSE(myrpt->links, l, l_it)
```
Allan-N: If we make a copy of the `myrpt->links` container then we don't need to worry about the contents changing and that means we don't to be constrained by the size of `listoflinks[]`.  In fact, we can drop `listoflinks[]` entirely and walk our container copy to get the node names, have no need for `numoflinks`, and likely get rid of MAX_STAT_LINKS :)

apps/app_rpt/rpt_cli.c
```diff
@@ -570,15 +573,16 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
...
-			l = myrpt->links.next;
-			while (l && (l != &myrpt->links)) {
-				if (l->name[0] == '0') {	// Skip '0' nodes 
-					l = l->next;
+			RPT_AO2_LIST_TRAVERSE(myrpt->links, l, l_it)
+			{
+				if (l->name[0] == '0') { // Skip '0' nodes
 					continue;
 				}
 				if (!(s = ast_calloc(1, sizeof(struct rpt_lstat)))) {
```
Allan-N: Rather than allocating a `struct rpt_lstat` and filling it with some of the data from the link can we just create add this link to another container ... and use that container for reporting?  This way, we'd have myrpt->links would have one reference and our reporting container would have a second reference.

apps/app_rpt/rpt_manager.c
```diff
@@ -515,26 +517,26 @@ static int rpt_manager_do_stats(struct mansession *s, const struct message *m, c
...
 					ast_log(LOG_WARNING, "Maximum number of links exceeds %d in rpt_do_stats()!", MAX_STAT_LINKS);
+					ao2_ref(l, -1);
 					break;
 				}
 				if (l->name[0] == '0') {	/* Skip '0' nodes */
 					reverse_patch_state = "UP";
-					l = l->next;
 					continue;
 				}
 				listoflinks[numoflinks] = ast_strdup(l->name);
```
Allan-N: Do we use a second container to hold the links/nodes to report?

apps/app_rpt/rpt_link.c
```diff
@@ -271,11 +254,11 @@ int FindBestRssi(struct rpt *myrpt)
 	if (myrpt->rxchankeyed) {
 		maxrssi = myrpt->rxrssi;
 	}
-
-	l = myrpt->links.next;
-	while (l && (l != &myrpt->links)) {
+	RPT_AO2_LIST_TRAVERSE(myrpt->links, l, l_it)
+	{
 		if (numoflinks >= MAX_STAT_LINKS) {
```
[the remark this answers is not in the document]
mkmer: Interesting observation about this function - in either case we could remove this counter AND there is no need to return numlinks either (from what I can tell)

apps/app_rpt/rpt_functions.c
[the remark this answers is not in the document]
mkmer: Yup

apps/app_rpt/rpt_link.c
```diff
@@ -714,13 +684,13 @@ int connect_link(struct rpt *myrpt, char *node, enum link_mode mode, int perma)
 	}
 	ast_copy_string(myrpt->lastlinknode, node, sizeof(myrpt->lastlinknode));
 	/* establish call */
-	l = ast_calloc(1, sizeof(struct rpt_link));
+	l = ao2_alloc(sizeof(struct rpt_link), rpt_link_destroy);
```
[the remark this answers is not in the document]
Allan-N: The complement of many malloc/calloc... APIs tend to be free (or dealloc).  Then, we have create/delete/destroy.  I'm just trying to reconcile what an rpt_link is and how to use it.  Do we provide a full abstraction layer with many helpers, a partial abstraction, or the bare minimum.

Not digging my heals in here, just sharing what's in my head.

Oh, whether the object has/uses an associated reference is an implementation detail.

apps/app_rpt/rpt_link.c
```diff
@@ -651,30 +593,30 @@ int connect_link(struct rpt *myrpt, char *node, enum link_mode mode, int perma)
...
 		if (!strcmp(l->name, node)) {
-			break;
+			break; /* found , leave referenced */
 		}
-		l = l->next;
 	}
+	ao2_iterator_destroy(&l_it);
 	/* if found */
-	if (l != &myrpt->links) {
+	if (l) {
```
Allan-N: We have a problem here.  If the link was "found", inside the loop with the `break;`, the you have a link with a reference.  But, if no match was found then `l` will be the last link in the list but with NO reference.  Probably need to introduce a `match` variable and not use `l`.

Allan-N (changes requested):

apps/app_rpt/rpt_link.c
mkmer: Actually, this one should probably be an `ao2_callback()`.

apps/app_rpt.c
```diff
@@ -4673,10 +4703,21 @@ static void *rpt(void *this)
 		rpt_hangup_rx_tx(myrpt);
 		pthread_exit(NULL);
 	}
+	myrpt->links = ao2_container_alloc_list(0, /* AO2 object flags. 0 means to use the default behavior */
```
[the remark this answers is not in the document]
mkmer: Thinking out loud here: Not sure it's worth keeping the list sorter 100% of the time... we only do the rbtree/sorted list on the command line (which is almost never?!?)  Is it?

apps/app_rpt/rpt_cli.c
```diff
@@ -317,101 +317,70 @@ static int rpt_do_stats(int fd, int argc, const char *const *argv)
...
-				connecttime = ast_tvdiff_ms(now, s->connecttime);
+				if (l->name[0] == '0') { /* Skip '0' nodes */
+					continue;
+				}
+				if (l->chan) {
+					pbx_substitute_variables_helper(l->chan, "${IAXPEER(CURRENTCHANNEL)}", peer, MAXPEERSTR - 1);
+				} else {
+					strcpy(peer, "(none)");
+				}
+				now = rpt_tvnow();
```
Allan-N: Can we move this line back out of the RPT_LIST_TRAVERSE loop?

apps/app_rpt/rpt_cli.c
```diff
@@ -568,38 +543,22 @@ static int rpt_do_xnode(int fd, int argc, const char *const *argv)
...
+
+				if (l->name[0] == '0') { // Skip '0' nodes
+					continue;
+				}
+				if (l->chan) {
+					pbx_substitute_variables_helper(l->chan, "${IAXPEER(CURRENTCHANNEL)}", peer, MAXPEERSTR - 1);
+				} else {
+					strcpy(peer, "(none)");
+				}
+				now = rpt_tvnow();
```
Allan-N: Can we move this line back out of the RPT_LIST_TRAVERSE loop?

apps/app_rpt/rpt_manager.c
```diff
@@ -213,38 +209,32 @@ static int rpt_manager_do_xstat(struct mansession *ses, const struct message *m,
...
-				s->connecttime = l->connecttime;
-				s->thisconnected = l->thisconnected;
-				memcpy(s->chan_stat, l->chan_stat, NRPTSTAT * sizeof(struct rpt_chan_stat));
-				insque((struct qelem *) s, (struct qelem *) s_head.next);
-				memset(l->chan_stat, 0, NRPTSTAT * sizeof(struct rpt_chan_stat));
-				l = l->next;
+			links_copy = ao2_container_alloc_list(0, /* AO2 object flags. 0 means to use the default behavior */
+				0,									 /* AO2 container flags. */
+				NULL,								 /* Sorting function. NULL means the list will not be sorted */
+				NULL);								 /* Comparison function */
```
Allan-N: Just a thought.  What would be the impact of providing a comparison function here like `rpt_link_find_by_name()`?

Would this simplify the places where we are searching for a link by name.

`ao2_callback(myrpt->links, 0, rpt_link_find_by_name, ...)`
--> `ao2_find(myrpt-links, NULL, 0)`

apps/app_rpt/rpt_functions.c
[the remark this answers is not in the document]
mkmer: I don't think so, If we have an `l`, we "do stuff" and return `DC_COMPLETE`.  At the `break;` (if we got there) we couldn't find an `l` so it's NULL.
It may be better to follow the "rule" of little code in the if, big chunk out of it.

apps/app_rpt.c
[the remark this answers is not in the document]
Allan-N: >I believe we wanted a copy so that we don't lock up the repeater task while doing "stuff" (usually some sort of report if not all of them are reports). I believe every place we use a dup are places where we originally made copies of the list to process unlocked further after unlocking.

With the ao2_container's, do we even need to hold the (myrpt?) lock while traversing?  OK, maybe on the call to `ao2_ref(myrpt->links, +1)`.  But, after that, we know we've got a ref to the container.  Is that not enough to iterate?

apps/app_rpt/rpt_link.h
```diff
@@ -84,7 +84,6 @@ void rpt_update_links(struct rpt *myrpt);
  */
 int connect_link(struct rpt *myrpt, char *node, enum link_mode mode, int perma);
 
-/*! \brief Free link and associated internal memory.
- * \param link Link structure to free
+/*! \brief destroy ao2 object
```
Allan-N: I'd kinda like to see this function moved up in the file, just below `rpt_link_remove()`.

It might also help to add :
```
* \param obj rpt_link object called by the ao2_alloc() destructor function
```

apps/app_rpt/rpt_manager.c
```diff
@@ -213,38 +209,34 @@ static int rpt_manager_do_xstat(struct mansession *ses, const struct message *m,
...
-				s->mode = l->mode;
-				s->outbound = l->outbound;
-				s->reconnects = l->reconnects;
-				s->connecttime = l->connecttime;
-				s->thisconnected = l->thisconnected;
-				memcpy(s->chan_stat, l->chan_stat, NRPTSTAT * sizeof(struct rpt_chan_stat));
-				insque((struct qelem *) s, (struct qelem *) s_head.next);
-				memset(l->chan_stat, 0, NRPTSTAT * sizeof(struct rpt_chan_stat));
-				l = l->next;
+			links_copy = ao2_container_alloc_list(0, /* AO2 object flags. 0 means to use the default behavior */
```
Allan-N: Can we use ao2_container_clone() here ?

apps/app_rpt/rpt_manager.c
```diff
@@ -514,26 +502,27 @@ static int rpt_manager_do_stats(struct mansession *s, const struct message *m, c
...
-			numoflinks = 0;
-			l = myrpt->links.next;
-			while (l && (l != &myrpt->links)) {
-				if (numoflinks >= MAX_STAT_LINKS) {
-					ast_log(LOG_WARNING, "Maximum number of links exceeds %d in rpt_do_stats()!", MAX_STAT_LINKS);
-					break;
-				}
-				if (l->name[0] == '0') {	/* Skip '0' nodes */
+
+			links_copy = ao2_container_alloc_list(0, /* AO2 object flags. 0 means to use the default behavior */
```
Allan-N: Can we use ao2_container_clone() here ?

apps/app_rpt.c
```diff
@@ -1746,11 +1745,13 @@ static int distribute_to_all_links(struct rpt *myrpt, struct rpt_link *mylink, c
 			}
 			if (dest) {
 				/* if it is, send it and we're done */
+				ao2_ref(l, -1);
```
InterLinked1: Do we need to unref outside this branch as well?

mkmer: `RPT_LIST_TRAVERSE` handles the unref.  1748 is a "premature" exit so we need to unref

apps/app_rpt.c
```diff
@@ -4659,6 +4680,17 @@ static inline void voxtostate_to_voxtotimer(struct rpt *myrpt)
 	}
 }
 
+static int sendtext_cb(void *obj, void *arg, int flags)
+{
+	struct rpt_link *link = obj;
+	char *str = arg;
```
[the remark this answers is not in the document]
mkmer: Don't see why not.

InterLinked1 (approved):

PR #811, opened 2025-10-20
app_rpt: memory leak in attempt_reconnect() failure paths

mkmer:
While we have bigger problems if the `cap` fails to allocate, cleaning up the now unlinked link pointer should be included.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2331,6 +2331,7 @@ static int attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	char sx[320];
 	struct ast_frame *f1;
 	struct ast_format_cap *cap;
+	int res = 0;
 
 	if (node_lookup(myrpt, l->name, tmp, sizeof(tmp) - 1, 1)) {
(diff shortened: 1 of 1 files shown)


apps/app_rpt.c
```diff
@@ -2356,30 +2356,38 @@ static int attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
...
+	l->lastrealrx = 0;
+	l->last_frame_sent = 0;
+	l->rxlingertimer = RX_LINGER_TIME;
+	l->newkeytimer = NEWKEYTIME;
+	l->link_newkey = RADIO_KEY_NOT_ALLOWED;
 
 	cap = ast_format_cap_alloc(AST_FORMAT_CAP_FLAG_DEFAULT);
 	if (!cap) {
 		ast_log(LOG_ERROR, "Failed to alloc cap\n");
+		rpt_mutex_lock(&myrpt->lock);
```
Allan-N: Just gonna toss out an idea here.  Could we try to allocate everything needed for the new link before unlinking the old one?  If we've got what we need then we unlink and associated the new channel.  If not, the current link simply remains.

PR #812, opened 2025-10-20
app_rpt: use `ast_tv()` for setting timeval variables

mkmer:


Allan-N (commented):
Honestly, I'm not wild about this change ... at least in its current form.  Yes, it's an inline function but we're allocating a structure, returning a pointer to the structure, and then performing a structure copy.  Isn't two assignments much more efficient?

Seems like we'd be better off with a macro.  Maybe something like :
```
#define TV_INIT(ptr, sec, usec) \
    do {                        \
        (ptr)->tv_sec = (sec);  \
        (ptr)->tv_usec = (usec);\
    } while (0)

...
    struct timeval tv;
    TV_INIT(&tv, 0, 0);
```

or, for constants :
```
#define TV_INIT_CONST(sec, usec) ((struct timeval){ .tv_sec = (sec), .tv_usec = (usec) })
...
     struct timeval tv = TV_INIT_CONST(1, 500000);
```

and, yes, the asterisk headers/functions should really be offering something like the above.

mkmer:
I put this up because I noticed it in the base asterisk code in "many" places.  If it's not the right thing, than we can just close it out.

Allan-N (approved):

PR #825, opened 2025-11-03
app_rpt: Restore ability to disable telemode

mkmer:
When set to 0, timer logic would reset to 1 enabling the mode.  This PR restores original timer logic.


mkmer:
> > if we set it to 0 (to disable it)
> 
> Look at `update_timer()`. We only decrement if the current value is greater than the min/hold value. If the value was "set" to zero then it stays at zero.

I missed 1/2 of the first if statement -> No need for this change.

PR #829, opened 2025-11-14
rpt.conf: Default lookup method change to dns from both

jxmx:
This is to change the default method of node connection lookup to DNS exclusively instead of DNS+File.


InterLinked1 (commented):
The commit message doesn't explain why this change is being made. I'm assuming checking both causes some issue we are trying to solve?

PR #831, opened 2025-11-17
app_rpt: Address timeout timer not resetting when all activity stops

### LINKED ISSUE #830: Node timeout fails to reset and causes loss of incoming (receive) audio
Asterisk: 22.5.2+asl3-3.6.2-1.deb12
ASL [app_rpt]: 3.6.2

After upgrading to the latest release, I have noticed a strange behavior with a few of my nodes.

The first thing I noticed, is a while after restarting Asterisk, if a user connects using a web transceiver app (DVSwitch, for example), the app will not hear rx audio from the node. Yet once the PTT is pressed, the app user will start to hear rx audio. Almost like it works in reverse, but the app still appears to transmit, as the rf stations state they hear the app users, but the app users do not hear the rf stations replying.

At the same time, I noticed that Allmon3 shows the node status as "Transmit - Playback from remote node xxxxxx" instead of the usual "Transmit - Network Source" when this issue occurs.

A simple restart of Asterisk returns the node to normal behavior. The WT apps begin working properly, and Allmon3 returns to displaying the standard "Transmit - Network Source". I have had to reset the affected nodes two or three times in the past week or so since the upgrade.

After further observation -

I happened to be connected to the node via the DV Switch app when the problem occurred again this morning. Everything was working fine, then all of a sudden I heard a nodeXXXXXX timeout message played by the node I was connected to. After that, I lost all inbound audio from the node. Same for other stations connected to the node. As in all past occurrences, pressing the PTT button on the app resulted in us hearing the inbound audio. Releasing the PTT, the inbound audio stops.

At the same time, when the this occurs, we are seeing the abnormal "Playback" message display on Allmon3.

I do not believe that the station talking at the time when this occurred truly timed out the node, but I could be wrong. We did hear the timeout message just as the problem began.

In addition, another user (ZS1RBT James) is reporting the following:

"Following upgrade to the above version a few days ago, connections do not reset correctly following a timeout of a transmission received from a remote node. I hear ‘Node xxx time out’ on the radio, then nothing further until I restart Asterisk (or reboot).

In my previous version, things would reset when the remote node unkeyed and everything would revert back to normal.

Now, following a timeout, I can see incoming transmissions in Supermon, but the bridge to the output side does not activate (simpleusb). And Asterisk CLI (-rvvv) does not report any activity on my ASL node."

I can confirm, after further testing, that it in fact appears to be related to the timeout timer function.

Eric
K2CB

mkmer:
PR #770  prevents timeout timer from resetting without keying up locally.
Fixes #830 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout/unkey reset behavior for remote sessions by adding a separate remote unkey interval, preventing unintended immediate resets and preserving correct notification/telemetry ordering.
  * Adjusted TIMEOUT notifications so telemetry and alerts occur in the correct sequence relative to key/unkey transitions and remote conditions.

* **Refactor**
  * Reorganized timer and reset flow to clearly separate local vs remote unkey handling, kerchunk gating, and state transitions for more predictable behavior.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3384,7 +3384,7 @@ static inline int update_timers(struct rpt *myrpt, const int elap, const int tot
 
 	/* Keep track of time keyed */
 	if (myrpt->keyed) {
-		if (myrpt->keyed + elap < 0) {
+		if (myrpt->keyed_time_ms + elap < 0) {
 			myrpt->keyed_time_ms = INT_MAX;
 		} else {
 			myrpt->keyed_time_ms += elap;
@@ -4966,11 +4966,14 @@ static void *rpt(void *this)
 		}
 		/* add in parrot stuff */
 		totx = totx || (myrpt->parrotstate > 1);
-
-		/* If not in time out condition and not wanting to transmit */
 		if (!totx && myrpt->totimer) {
-			/* Note: This is called every time through the loop when not wanting to transmit and not in the timed out condition
-			 * This is the execution path taken when a user unkeys when not timed out to intentionally reset the time out timer
+			/*
+			 * This is the execution path taken when a user unkeys (!totx) and not
+			 * yet timed out (myrpt->totimer > 0).  Here, we intentionally reset the
+			 * time out timer.
+			 *
+			 * Note: This is called every time through the loop when not wanting to
+			 * transmit and not in the timed out condition
 			 */
 			myrpt->totimer = myrpt->p.totime;
 			myrpt->tounkeyed = 0;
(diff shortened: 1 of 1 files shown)


apps/app_rpt.c
```diff
@@ -5059,27 +5025,19 @@ static void *rpt(void *this)
 					 * and the TOT timer will get reset with no delay when the user unkeys and rekeys in a time out condition
```
Allan-N: since we have two timers (local/!remote, remote) I think it would be good to callout here that we're talking about a local unkey.

... or is it whenever whomever is keying lets go?

mkmer: I reworked this a bit, maybe it's cleared up a bit?  I didn't go through the existing comments, so there really was a bunch of old comments not representing the new behavior.

Issue #334, opened 2024-06-19
Log Voter Timeout Disconnects


tsawyer:
Log Voter Client `   -- Voter client OatMtn disconnect (timeout)` message with a NOTICE status.

Mason10198:
Suggest also logging out-of-bounds packets (packet took longer than `buflen` to get to host), similar to out-of-bounds packet logging in `VOTER client console.

Question - is there any circumstance (packet OOB, packet OOO, corrupt, etc) that causes a `chan_voter` instance to drop a client (remove client from VOTER display) other than a keepalive timeout, or does the loss of a client from the host perspective ALWAYS signify a keepalive timeout?

tsawyer:
> When someone forgets that they created an account several years ago and tries to create an account with their callsign, the allstarlink portal does not let them know there is already an account with this callsign, the callsign field just turns red and does not give an explanation.  They end up opening a helpdesk ticket to tell us that our website is broken.

To my knowledge only a timeout (five missed pings) cause a RTCM (voter or missed client) to disconnect. Disconnects are indicated by a RTCM display dropout and a lost link CW message if a transmitter is connected.

Having said that, I've long had my suspicions that dropouts occur when there is no apparent network problem.  So maybe you're on to something here.

Mason10198:
I have noticed at times one of my TX/RX clients will flash on/off rapidly (a few times per second) in the VOTER display, Allmon, and Supermon. When this "flashing" occurs, there are gaps in the TX audio for that site.

The VOTER clients console shows no messages, and neither does the host Asterisk console whenever this "flashing" occurs.

The client in question is two 5GHz Ubiquiti hops away from the host, so it is difficult to determine the cause of these "drops". TX and RX buffers are set well above the actual latency of the network.

More logging in `chan_voter` would greatly help troubleshooting this.

InterLinked1:
> More logging in `chan_voter` would greatly help troubleshooting this.

Have you enabled debug logging for the module on your system? (e.g. `core set debug 10 chan_voter`)

Mason10198:
> Have you enabled debug logging for the module on your system? (e.g. `core set debug 10 chan_voter`)

Yes, I have enabled debug logging, but it generates a significant amount of output (a few hundred lines per second), which is quite difficult to sift through:
```
[...]

Issue #122, opened 2023-03-08
Chan_voter exits and false timeouts


tsawyer:
This is from SCRN-ASL3 which has a lot of RTCMs (voters) attached. The owner of this says the Voter Client disconnect messages (see attached image) are false. I have had my suspicions about false positives on my system as well.  In any case it would be worth going over chan_voter to see if something is amiss.

Also, possibly related is SCRN-ASL3 is exiting (no core files) after running for 3 or 4 days. 
 
<img width="976" alt="Screenshot 2023-03-08 at 9 05 39 AM" src="https://user-images.githubusercontent.com/1526362/223783832-b3f425a4-356f-4eef-8082-1b03a8734c67.png">

ve7fet:
PR https://github.com/AllStarLink/app_rpt/pull/844 should help with some of the diagnostics with timeouts. A number of debug messages have been reorganized to better assist in troubleshooting. The "noisy" messages have been moved to higher debug levels (6 or 7), so that other messages don't get lost in the "noise".

This is quite an old issue, and it looks like the timeout messages were set as verbose messages only. The current code has them as LOG_NOTICE messages, which will log when they occurred, as well as identify which part of the code they came from (these timeout messages appear in two blocks of code).

In regards to timeouts, there are a couple of things to consider:

1. Clients designated as a **master** (of which there can only be ONE defined on a server/host) when using normal voting have a timeout of 100ms. That is why they need to be on the same LAN as the host, and co-located (on the same LAN on the other end of a microwave link is NOT going to be good enough). This is defined by `#define	MASTER_TIMEOUT_MS 100`.
2. All other clients in all other cases should be tested against `#define	CLIENT_TIMEOUT_MS 3000`, which means that if they haven't been heard from in 3 seconds, they are going to time out. From what I can see (and I may be wrong), there is no such thing as, "this client missed X packets...".

It is conceivable that a client could be timing out "5 times in less than 2 minutes", if there is a network issue going on. Not enough detail was provided on the architecture/layout to help diagnose.

The referenced PR will automatically close this issue, when it is approved. If this problem continues to persist after those changes, please file a new issue with more details to start a new investgation.

PR #844, opened 2025-11-30
chan_voter: Additional debug message changes for easier troubleshooting

ve7fet:
-Clean up grammar in debug messages for clarity
-Add context to "out of bounds" debug message for mix clients -Add and log a more descriptive error when bufflen too small for mix mode clients -First pass at adding client name to additional debug strings (note it can't seem to resolve client names reliably in normal (voting) mode, but resolves fine in mix mode) -Revise TX On/TX Off from debug level 2 to debug level 1 for consistency with like messages -Spelling corrections in comments
-Change debug around line 3906 to level 7 (it is WAY too chatty) -Add debug level 3 message when a client appears to have stopped receiving (RX_TIMEOUT_MS (200ms) detected) -Change disconnect (timeout) messages from verbose 3 to log notice per Issue 334 and Issue 122 which will also help diag which function is asserting the notice (there are two different parts of the code). Note that a client designated as a "master" will timeout in 100ms, versus a regular client which has 3s before it times out. Hence why a master MUST be on the same LAN as the host -Change debug lines at around 4207 to level 7, they are very chatty, and of limited use

Resolves: #334 #122

## Summary by CodeRabbit

* **Chores**
  * Standardized capitalization and terminology in log and status messages for improved clarity and consistency throughout the application.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -3884,7 +3884,8 @@ static void *voter_reader(void *data)
 					.subclass.integer = AST_CONTROL_RADIO_UNKEY,
 					.src = __PRETTY_FUNCTION__,
 				};
-				ast_debug(3, "VOTER client %s was receiving but now has stopped (RX_TIMEOUT_MS)!\n", ((client) ? client->name : "UNKNOWN"));
+				ast_debug(3, "VOTER client %s was receiving but now has stopped (RX_TIMEOUT_MS)!\n",
+					(client && client->name) ? client->name : "UNKNOWN");
 				ast_queue_frame(p->owner, &wf);
 				p->rxkey = 0;
 				p->lastwon = NULL;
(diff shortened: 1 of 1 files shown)


channels/chan_voter.c
```diff
@@ -3903,8 +3905,8 @@ static void *voter_reader(void *data)
 			continue;
 		}
 		vph = (VOTER_PACKET_HEADER *) buf;
-		ast_debug(6, "Got rx packet, len %d payload %d challenge %s digest %08x\n", (int) recvlen, ntohs(vph->payload_type),
-			vph->challenge, ntohl(vph->digest));
+		ast_debug(7, "Got RX packet, len %d payload %d challenge %s digest %08x from client %s\n", (int) recvlen,
```
mkmer: I guess I should have looked at this a bit closer:
We don't know the client at this point in the code so we can't get valid client info yet.

PR #867, opened 2025-12-14
chan_voter: Detect and handle buflen too small with mix mode clients

### LINKED ISSUE #847: chan_voter: Add check for minimum bufflen when using mix mode clients
When using mix mode (general purpose) clients, the minimum size of `buflen` in `voter.conf` is `160`. If `buflen` is smaller than this, clients will fail to connect, and the user may not know why without digging into debug messages (this will be fixed with a new error message in a coming change).

The `buflen` can be smaller for "normal" voting clients, without issue (ie `buflen = 120` works fine on good networks).

Need to look at the feasibility of checking on the fly when a client connects as a mix client to force the `buflen` to a minimum of `160` to prevent "out of bounds" messages and failures to connect. Also note this MAY be affected by per-stanza definitions of `buflen` that need to be reviewed (in addition to the global variable).

ve7fet:
Re-submitting an earlier fix that strictly deals with this issue.

This fix properly detects when a client tries connecting in mix mode, and the `buflen` in `voter.conf` is too small (previously known as an "out of bounds" error). Now, we will log an error telling the user to fix `buflen` in `voter.conf`, and block the mix mode clients from connecting (they can't send audio to the channel when they are misconfigured anyways).

Resolves #847

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection validation by enforcing minimum buffer requirements, preventing invalid connections and connection failures.
  * Enhanced error reporting with clearer, user-friendly messages when connection issues are detected.
  * Implemented smarter logging to reduce repeated error messages while maintaining diagnostic visibility for troubleshooting.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -4709,9 +4709,9 @@ static void *voter_reader(void *data)
 								/* The CLIENT has to send us flags to tell us it is configured for mix mode (GPS PPS = NONE)
 								 * so this is where we check the flags from the client, and update client->mix accordingly.
 								 * Mix mode requires a buflen >= 160 in voter.conf, which is equivalent to client->buflen = 1280
-								 * (buflen * 8, also FRAME_SIZE * 8). This keeps the starting drain index > 0 when we 
-								 * configure it. 
-								 * 
+								 * (buflen * 8, also FRAME_SIZE * 8). This keeps the starting drain index > 0 when we
+								 * configure it.
+								 *
 								 * If a client connects as mix mode, we need to enforce the minimum buflen, otherwie the
 								 * client will connect, but cannot send us audio because the buffer isn't big enough.
 								 *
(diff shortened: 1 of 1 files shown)


channels/chan_voter.c
```diff
@@ -4705,8 +4704,33 @@ static void *voter_reader(void *data)
...
-								client->mix = 1;
+								/* The CLIENT has to send us flags to tell us it is configured for mix mode (GPS PPS = NONE)
+								 * so this is where we check the flags from the client, and update client->mix accordingly.
+								 * Mix mode requires a buflen >= 160 in voter.conf, which is equivalent to client->buflen = 1280
+								 * (buflen * 8). If a client connects as mix mode, we need to enforce the minimum buflen,
+								 * otherwie the client will connect, but cannot send us audio because the buffer isn't big enough.
+								 *
+								 * Check the buflen, throw an error if it is too small, and block the client from connecting.
+								 */
+								if (client->buflen < 1280) {
```
mkmer: Should we make this "magic number" a #define, and "maybe" relate it to the 160ms? (either do the math to create 1280 in the define, the other way aorund? )

InterLinked1 (commented):
Looks good to me, if you address Mike's comments

mkmer (approved):

PR #877, opened 2025-12-19
app_rpt: Refactor rpt_manager using ast_str

mkmer:
Fixes a potential buffer overrun in `rpt_manager_do_stats()`

## Summary by CodeRabbit

* **Refactor**
  * Improved internal string assembly and memory management for manager status reporting, consolidating output paths and reducing manual buffer handling.

* **Chores**
  * Cleaned up command dispatch internals and unified status output formatting.

Note: No user-visible behavior changes in this release.

### DIFF
--- a/apps/app_rpt/rpt_manager.c
+++ b/apps/app_rpt/rpt_manager.c
@@ -760,10 +760,11 @@ static int manager_rpt_status(struct mansession *s, const struct message *m)
 
 	case MGRCMD_RPTSTAT:
 		/* Return Nodes: and a comma separated list of nodes */
-		ast_str_set(&str, 0, "Nodes: ");
 		if (!nrpts) {
+			rpt_manager_success(s, m);
 			astman_append(s, "<NONE>\r\n");
 		} else {
+			ast_str_set(&str, 0, "Nodes: ");
 			for (i = 0; i < nrpts; i++) {
 				ast_str_append(&str, 0, "%s", rpt_vars[i].name);
 				if (i < nrpts - 1) {


apps/app_rpt/rpt_manager.c
```diff
@@ -665,17 +664,17 @@ static int rpt_manager_do_stats(struct mansession *s, const struct message *m, c
 			astman_append(s, "TxTimeSinceSystemInitialization: %02d:%02d:%02d:%02d\r\n", hours, minutes, seconds,
 						  (int) totaltxtime);
 
-			sprintf(str, "NodesCurrentlyConnectedToUs: ");
+			ast_str_append(&str, 0, "NodesCurrentlyConnectedToUs: ");
 			if (!numoflinks) {
-				strcat(str, "<NONE>");
+				ast_str_set(&str, 0, "<NONE>");
```
Allan-N: Shouldn't this be ast_str_append() ?

apps/app_rpt/rpt_manager.c
```diff
@@ -728,11 +733,9 @@ static int manager_rpt_status(struct mansession *s, const struct message *m)
 	int nrpts = rpt_num_rpts();
 
 	time(&now);
-
-	len = 1024;					/* Allocate a working buffer */
-	if (!(str = ast_malloc(len)))
-		return -1;
-
+	if (!str) {
```
Allan-N: Can we move this block above the `time(&now);` statement (why call the function if you are just going to return) ?

or, better yet, move the `time()` call down to just before the one place in the function where the value is used.

Allan-N (changes requested):

apps/app_rpt/rpt_manager.c
mkmer: You and code rabbit :)  No, if there are no links we just overwrite with \<NONE>. (that' what strcat() was doing)

Allan-N: > No, if there are no links we just overwrite with <NONE>

I'm confused.  With no links should are we suppose to have `NodesCurrentlyConnectedToUs: <NONE>` ?
or just `<NONE>` ?

apps/app_rpt/rpt_manager.c
```diff
@@ -756,34 +760,20 @@ static int manager_rpt_status(struct mansession *s, const struct message *m)
...
-				if ((res = snprintf(str + strlen(str), len, "%s,", rpt_vars[i].name)) < 0) {
-					ast_free(str);
-					return 0;
-				}
-			} else {
-				if ((res = snprintf(str + strlen(str), len, "%s", rpt_vars[i].name)) < 0) {
-					ast_free(str);
-					return 0;
+		if (!nrpts) {
+			astman_append(s, "<NONE>\r\n");
```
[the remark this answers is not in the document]
mkmer: Nodes : \<NONE> ?
It's the "easy stuff" that gets me every time...

Issue #875, opened 2025-12-19
node_lookup_bydns is a blocking (synchronous) function


mkmer:
Key to the smooth operation of `app_rpt`, all calls inside of the main channel servicing loop https://github.com/AllStarLink/app_rpt/blob/3a0ac6d71bf364a28d0137a7b850153f80011986/apps/app_rpt.c#L4854
should be <20ms.  Some onetime calls "maybe" less than 80ms could be covered by the channel buffers.
Connecting to a remote node requires a DNS lookup (using the latest configuration defaults):
Reconnect:
https://github.com/AllStarLink/app_rpt/blob/3a0ac6d71bf364a28d0137a7b850153f80011986/apps/app_rpt.c#L2407

Conenct_link():
https://github.com/AllStarLink/app_rpt/blob/3a0ac6d71bf364a28d0137a7b850153f80011986/apps/app_rpt/rpt_link.c#L587

ILINK command 4:
https://github.com/AllStarLink/app_rpt/blob/3a0ac6d71bf364a28d0137a7b850153f80011986/apps/app_rpt/rpt_functions.c#L190

ALL of these call -> `node_lookup` :
https://github.com/AllStarLink/app_rpt/blob/3a0ac6d71bf364a28d0137a7b850153f80011986/apps/app_rpt/rpt_config.c#L523

Calls `node_lookup_bydns`:
https://github.com/AllStarLink/app_rpt/blob/3a0ac6d71bf364a28d0137a7b850153f80011986/apps/app_rpt/rpt_config.c#L368

which calls blocking `ast_dns_resolve()`
https://github.com/AllStarLink/app_rpt/blob/3a0ac6d71bf364a28d0137a7b850153f80011986/apps/app_rpt/rpt_config.c#L430

With a result of:
```
[2025-12-18 20:04:25.690] DEBUG[1343220]: app_rpt/rpt_config.c:403 node_lookup_bydns: Resolving DNS SRV records for: _iax._udp.2997.nodes.allstarlink.org
[2025-12-18 20:04:27.610] WARNING[1343224]: channel.c:1116 __ast_queue_frame: Exceptionally long voice queue length (97 voice / 97 total) queuing to Local/LocalTX@default-00000001;1
[2025-12-18 20:04:27.610] WARNING[1343227]: channel.c:1116 __ast_queue_frame: Exceptionally long voice queue length (97 voice / 97 total) queuing to Local/MonChan@default-00000002;1
[...]

PR #879, opened 2025-12-19
app_rpt: Refactor `connect_link()` and `attempt_reconnect()` as a thread

mkmer:
Using a thread allows `rpt_app` main thread to continue servicing channels while reconnect procedure makes block calls to DNS lookup.
Fixes https://github.com/AllStarLink/app_rpt/issues/875

## Summary by CodeRabbit

* **Refactoring**
  * Improved application responsiveness by converting link connection and reconnection operations to asynchronous threading, allowing the system to remain responsive during these operations.

### DIFF
--- a/apps/app_rpt/rpt_functions.c
+++ b/apps/app_rpt/rpt_functions.c
@@ -354,7 +354,7 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
 				ast_free(connect_data);
 				rpt_mutex_lock(&myrpt->lock);
 				myrpt->connect_thread_count--;
-				assert(myrpt->connect_thread_count >= 0);
+				ast_assert(myrpt->connect_thread_count >= 0);
 				rpt_mutex_unlock(&myrpt->lock);
 			}
 		}


apps/app_rpt/rpt_link.h
```diff
@@ -74,17 +74,23 @@ void __kickshort(struct rpt *myrpt);
...
+void *connect_link(void *data);
 
 /*! \brief Free link and associated internal memory.
  * \param link Link structure to free
  */
-void rpt_link_free(struct rpt_link *link);
\ No newline at end of file
+void rpt_link_free(struct rpt_link *link);
+
+struct rpt_connect_data {
```
InterLinked1: Documentation?
If it was simple, that would be one thing, but `perma` is certainly not self-documenting.

InterLinked1 (changes requested):

apps/app_rpt/rpt_link.h
mkmer: Well... I just copied what was there and it wasn't much" - no parameters WERE documented...
```
*! 
 * \brief Connect a link 
 * \retval -2 Attempt to connect to self 
 * \retval -1 No such node
 * \retval 0 Success
 * \retval 1 No match yet
 * \retval 2 Already connected to this node
 */
 ```
 I'll see if I can figure out what it actually does :)

File: apps/app_rpt.c [hunk 2]
```diff
@@ -2451,14 +2468,22 @@ static int attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
...
 	}
 	rpt_mutex_lock(&myrpt->lock);
 	/* put back in queue */
 	rpt_link_add(myrpt, l);
 	rpt_mutex_unlock(&myrpt->lock);
 	ast_log(LOG_NOTICE, "Reconnect Attempt to %s in progress\n", l->name);
-	return res;
+cleanup:
+	ast_free(reconnect_data);
+	myrpt->connect_thread_count--;
```
InterLinked1: Do these operations need to be surrounded by the mutex?

apps/app_rpt.c
```diff
@@ -3309,8 +3336,19 @@ static inline void periodic_process_links(struct rpt *myrpt, const int elap)
...
+				reconnect_data = ast_calloc(1, sizeof(struct rpt_reconnect_data));
+				if (!reconnect_data) {
+					rpt_mutex_lock(&myrpt->lock);
 					l->retrytimer = RETRY_TIMER_MS;
+					continue;
+				}
+				reconnect_data->myrpt = myrpt;
+				reconnect_data->l = l;
+				myrpt->connect_thread_count++;
+				if (ast_pthread_create_detached(&reconnect_data->threadid, NULL, attempt_reconnect, (void *) reconnect_data) < 0) {
```
InterLinked1: There's no need to cast to void

apps/app_rpt.c
```diff
@@ -5495,9 +5534,14 @@ static void *rpt(void *this)
 
 	myrpt->ready = 0;
 	usleep(100000);
-	/* wait for telem to be done */
-	while (myrpt->tele.next != &myrpt->tele)
+	while (myrpt->tele.next != &myrpt->tele) {
```
InterLinked1: I realize we were already doing this sort of, but I'm not keen on the idea of just using `usleep` and polling at some arbitrary interval to check whether we're done, and I don't think it would fly in the Asterisk repo.

A more "Asterisk-y" and efficient way might be to use an alertpipe to notify this thread whenever telemetry is done, e.g. before attempt_reconnect exits, after it's decremented connect_thread_count, it writes to the alertpipe if all threads have finished, and then this thread here simply has to poll the alertpipe and the sleeps can be eliminated.

apps/app_rpt.c
mkmer: reconnect_data is only used in the thread, so I think that's safe "for sure".
`myrpt->connect_thread_count--;` is only "looked at" when we reload or exit app_rpt -> making sure all of the threads are "done" before freeing the memory.  I don't believe this needs protection (but definitely could be wrong)

apps/app_rpt.c
mkmer: This is in the "exit" of our app_rpt thread.  The only thing changed is the comment...

 Are you saying add alert pipes to all telemetry, then create a list of all active telemetry and "wait" on all of them to alert finishing?   Or referencing the new thread counter.  Maybe the "other" way to deal with the threads would be to keep a full list of threads and join the active ones until they complete.  We would still need to go through the list, pull out the "next" thread and wait on it. Some other may finish first, which would be fine I suppose.
 On the dialing and redialing -> probably the same mechanism would work if we keep a thread pointer around...
 
 I don't think we should change the telemetry part in this PR (maybe a new one?)

InterLinked1: Yeah, we can definitely optimize in another PR, maybe just focusing on these arbitrary usleeps. Just wanted to comment while I was thinking about it.

I suppose serially going through a list of threads and joining them would work too, since we have to wait for all of them anyways. Either that approach or the alertpipe, it moves the task of figuring out when we're done to the kernel so we don't need to do that in user space.

apps/app_rpt.c
Allan-N: > Aren't ++ and -- atomic and not interruptible?

Current CPUs don't make atomic operations easy.  Have a look at :
https://github.com/asterisk/asterisk/blob/c67d3d279619fcef980596038d6e857392486a6a/include/asterisk/lock.h#L652-L778

apps/app_rpt.c
Allan-N: Another thought / possibility for a followup PR would be to look at ast_cond_signal / ast_cond_wait

apps/app_rpt/rpt_link.h
```diff
@@ -74,17 +74,24 @@ void __kickshort(struct rpt *myrpt);
...
-void rpt_link_free(struct rpt_link *link);
\ No newline at end of file
+void rpt_link_free(struct rpt_link *link);
+
+/*! \brief Structure used to share data with connect_data thread */
+struct rpt_connect_data {
+	struct rpt *myrpt;
+	char *digitbuf; /* Node number in string format */
+	enum link_mode mode;
+	unsigned int perma:1; /* permanent  link */
```
Allan-N: one space between "permanent" and "link"

apps/app_rpt/rpt_link.h
```diff
@@ -74,17 +74,24 @@ void __kickshort(struct rpt *myrpt);
...
+void *connect_link(void *data);
 
 /*! \brief Free link and associated internal memory.
  * \param link Link structure to free
  */
-void rpt_link_free(struct rpt_link *link);
\ No newline at end of file
+void rpt_link_free(struct rpt_link *link);
+
+/*! \brief Structure used to share data with connect_data thread */
```
Allan-N: Can we move the struct definition to before rpt_link_connect (connect_link) ?

apps/app_rpt.c
```diff
@@ -419,6 +419,13 @@ int max_chan_stat[] = { 22000, 1000, 22000, 100, 22000, 2000, 22000 };
 
 int nullfd = -1;
 
+/*! \brief Structure used to share data with attempt_reconnect thread */
+struct rpt_reconnect_data {
+	struct rpt *myrpt;
+	struct rpt_link *l;
+	pthread_t threadid;
```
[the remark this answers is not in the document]
Allan-N: Hmmm... Are you just using this variable to store the detached threadid?  If so, do you even need to save the value?

apps/app_rpt.c
mkmer: Issue:
https://github.com/AllStarLink/app_rpt/issues/886

apps/app_rpt.c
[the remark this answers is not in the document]
mkmer: I need somewhere to put it when creating a thread.  We give it a pointer to store things in, seems like the structure was the best place to have it live... We don't look at it in our code after firing it off

apps/app_rpt/rpt_link.c
```diff
@@ -797,5 +810,15 @@ int connect_link(struct rpt *myrpt, char *node, enum link_mode mode, int perma)
 	rpt_link_add(myrpt, l);
 	__kickshort(myrpt);
 	rpt_mutex_unlock(&myrpt->lock);
-	return 0;
+cleanup:
+	ast_free(connect_data->digitbuf);
+	ast_free(connect_data);
+	rpt_mutex_lock(&myrpt->lock);
+	myrpt->connect_thread_count--;
+	if (myrpt->connect_thread_count < 0) {
```
Allan-N: I missed this one earlier ... and kinda surprised that the compiler didn't flag this as an issue but how would `connect_thread_count` ever be less than zero?  especially when the value was declared as `unsigned int connect_thread_count;`

Maybe update the variable declaration?

InterLinked1: If it should never happen, should we assert, rather than merely logging a warning?

mkmer: It will only "hurt" if we get to shutdown and it's <0 -> it would mean we are not sure if any threads are running since it's in an invalid. 
"Maybe" we leave this as is, it will serve the purpose for now, but totally refactor it along with the telemetry thread logic to shutdown cleanly?

Issue #858, opened 2025-12-06
chan_voter: duplex=0 causes voter clients to lock on transmit


ve7fet:
There is an undocumented `duplex` option in `voter.conf`, on a per-instance basis.

By default, it is always set to `1`. It can be overridden, and set to `duplex = 0`.

Not sure what it's purpose really is, I can't follow the code that well as to what exactly it is doing when set to `0`. It doesn't appear to make an RTCM/VOTER into a simplex device?

Regardless, if `duplex = 0`, and you have a single voting client, configured as `master` and `transmit`, keying that client's receiver causes the transmitter to lock on indefinitely until Asterisk is shut down (or possibly until Asterisk times out).

Opening this for awareness for future investigation.

In the meantime, it is not advised to set `duplex = 0` for normal use.

Note, this is not to be confused with "in-cabinet repeat", which is known as `duplex=3` in `rpt.conf`. That is entirely different, and is implemented in the VOTER/RTCM by setting the `hostdeemp` flag on the client configuration line.

ve7fet:
Further investigation...

The `duplex` setting is only ever used in the `voter_reader` function, in two places:

https://github.com/AllStarLink/app_rpt/blob/253936a1aebc177ddb5c559c0fce74f1d00fc1d3/channels/chan_voter.c#L5222-L5240

and

https://github.com/AllStarLink/app_rpt/blob/253936a1aebc177ddb5c559c0fce74f1d00fc1d3/channels/chan_voter.c#L5294-L5312

And in both of those cases, the client impacted is the "current master":

https://github.com/AllStarLink/app_rpt/blob/253936a1aebc177ddb5c559c0fce74f1d00fc1d3/channels/chan_voter.c#L4929

The code only runs if `duplex = 0`, and our transmitter is keyed, and the first thing it does is force the receiver off (`p->rxkey = 0;`)... but the client is still going to be sending audio frames, if it is receiving a signal, and other clients could be sending us received audio frames too. This looks like it says, "stop receiving, just send audio frames to the transmitter". So why would we want to do that, and why only to the master client?

Inherently, the VOTER/RTCM clients are full duplex devices. They will always send receive audio to the host, if there is audio to send (if you don't want to send receive audio, don't hook up a receiver). The transmitter is controlled by a directive in `voter.conf` (`transmit`)... which if set, will cause the client(s) to always send audio out the transmitter, if there is audio to send.

[...]

Issue #720, opened 2025-07-16
chan_voter.c: Expose DMWDIAG as a voter.conf option


ve7fet:
In `voter.c` there is a compile time option to transmit digital milliwatt instead of regular audio for debugging/testing:
 
https://github.com/AllStarLink/app_rpt/blob/2c721d13d13fcc39b10123eac9427d72d5957564/channels/chan_voter.c#L250

Since the majority of users do not compile this project from source, this option is virtually stranded/unusable.

I propose making this a `voter.conf` option to expose it to end users more easily.

PR #884, opened 2025-12-22
chan_voter: Remove un-documented duplex option, add "voter tune" CLI

ve7fet:
The duxplex option in voter.conf has never been documented, does
not work properly, and does not seem to have any practical use.
Removed this option, and associated code.

A compile-time option existed to send a generated 1kHz tone out
the transmitters, replacing all repeated audio. This update
extends that option to userspace with the introduction of the
"voter dmwdiag" Asterisk CLI command.

Resolves: https://github.com/AllStarLink/app_rpt/issues/858 https://github.com/AllStarLink/app_rpt/issues/720

UserNote: New CLI command, "voter dmwdiag <instance-id> [y/n]"

## Release Notes

* **New Features**
  * Added CLI command to query and enable/disable a test tone feature for voter instances at runtime, with per-instance control (previously required recompilation to adjust)

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -1999,7 +1999,7 @@ static char *handle_cli_tune(struct ast_cli_entry *e, int cmd, struct ast_cli_ar
 	case CLI_INIT:
 		e->command = "voter tune";
 		e->usage = "Usage: voter tune instance_id [y/n]\n"
-				   "       Specifies/Queries tune mode (1kHz test tone) for VOTER instance\n";
+				   "       Specifies/Queries tune mode (send 1kHz test tone to TX) for VOTER instance\n";
 		return NULL;
 	case CLI_GENERATE:
 		return voter_complete_node_list(a->line, a->word, a->pos, 2);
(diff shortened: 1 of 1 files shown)


ve7fet:
Well, it is easily changed. I think `txtone` will confuse someone to think it is for a CTCSS tone. Perhaps `tx1ktest`?

InterLinked1:
> To me, the `voter dmwdiag` command name is not intuitively obvious and I'd like to see us adopt a better name. Sadly, we already have `voter test` and `voter tone`. What do you think about :
> 
> * voter txtone
> * voter tx1k
> * voter 1k

For a milliwatt test tone, I think "mw" is more recognizable than "tx1k", and "txtone" isn't very specific, so in that sense, I get where "dmw" is coming from. Digital milliwatt diagnostic, I guess?

Allan-N:
>So, do we have a consensus on what to do here? Are we changing the name (to what?), or going to leave it as is?

I made my comments without putting in any stakes in the ground.  

>I've thought about it a bit more, and I am not sure we'd extend the functionality. I can't see a use for sending "silence", you can just do that effectively enough by keying the channel, and a 60% deviated tone may be tricky to implement... and shouldn't be a replacement for proper alignment anyways.

As a developer/engineer, I understand the desire to be very explicit (digital, milli-watt, 1kHz tone, 3/5kHz deviation, etc).  As someone just starting out, I don't know that it makes a bit of difference whether it's digital or analog ... it's just a test tone.

As for what tones one might want to enable/inject, there have got to be many.  Clearly, the full deviation tone is an important one.  I was thinking that "silence" might be good to help track down possible noise floor issues.  The other thing I think about is what test equipment folks have available for tuning.  Not everyone has a $$$ service monitor.  Maybe there's other tones and/or levels that lower the bar to getting ones radio properly calibrated?

Where does that leave me?  I still like `voter tune` :
```
voter tune instance_id [ ( "full" | "normal" | "quiet" ) ] [ ( "on" | "off" ) ]
```
with the optional `[ ( "full" | "normal" | "quiet" ) ]` being something we can add later.

But, I'm not going to grumble if you say "I really want to stick with `voter dmwdiag`

Allan-N (approved):

channels/chan_voter.c
```diff
@@ -332,7 +332,7 @@ Use "core show help voter <command>"" to display usage.
 
 #include "../apps/app_rpt/pocsag.c"
 
-/* This array is used by the voter dmwdiag CLI command to send a 1kHz tone at
+/* This array is used by the voter tune CLI command to send a 1kHz tone at
```
InterLinked1: Could we add some of these details to the CLI command description as well so users can see it? That way it's a bit clearer to people what it does if they don't have existing knowledge of other "tune" commands

InterLinked1: The full documentation should still be part of the module, not in some external manual.

PR #887, opened 2025-12-29
app_rpt: Add references to `l` when unlinking to allow finishing up, then free

mkmer:
Turns out removing an element from a list automatically dereferences.  

## Summary by CodeRabbit

**Bug Fixes**
- Enhanced stability during link management operations to prevent premature resource deallocation and improve overall system reliability during connection handling.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2440,9 +2440,8 @@ static void *attempt_reconnect(void *data)
 	ast_format_cap_append(cap, ast_format_slin, 0);
 
 	rpt_mutex_lock(&myrpt->lock);
-	/* remove from queue */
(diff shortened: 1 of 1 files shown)


apps/app_rpt.c
```diff
@@ -2441,6 +2441,7 @@ static void *attempt_reconnect(void *data)
 
 	rpt_mutex_lock(&myrpt->lock);
 	/* remove from queue */
+	ao2_ref(l, +1);					  /* We don't want the link to free after removing from the list */
```
Allan-N: What do you think about changing `rpt_link_remove()` to be :
```
void rpt_link_remove(struct ao2_container *links, struct rpt_link *l, int keep);
```
with the `keep` argument specifying whether the reference to `l` should be retained?

PR #888, opened 2025-12-31
app_rpt: Fix retry mechanism - prevent multiple connection threads

mkmer:
Resetting the link retry timer BEFORE a successful connect can (and does) result in multiple connection threads when connection takes longer than the retry time.  We only expect 1 at a time, thus abandoning connections.
Keeping track of the connection in progress both allows us to prevent firing off multiple threads.

## Summary by CodeRabbit

* **Refactor**
  * Reworked reconnect logic to track connection state per link instead of globally, preventing concurrent reconnect attempts and improving stability.
  * Teardown now waits for any active per-link reconnects to finish before releasing link resources.

* **Bug Fixes**
  * Improved error/synchronization handling around reconnect threads to avoid races and resource-release issues.

### DIFF
--- a/apps/app_rpt/rpt_functions.c
+++ b/apps/app_rpt/rpt_functions.c
@@ -342,7 +342,6 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
 			if (ast_pthread_create_detached(&connect_threadid, NULL, rpt_link_connect, (void *) connect_data) < 0) {
 				ast_free(connect_data->digitbuf);
 				ast_free(connect_data);
-				mylink->connect_in_progress = 0;
 			}
 		}
 		rpt_telem_select(myrpt, command_source, mylink);


apps/app_rpt/rpt_functions.c
```diff
@@ -346,16 +339,10 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
 			connect_data->perma = perma;
 			connect_data->command_source = command_source;
 			connect_data->mylink = mylink;
-			rpt_mutex_lock(&myrpt->lock);
-			myrpt->connect_thread_count++;
-			rpt_mutex_unlock(&myrpt->lock);
-			if (ast_pthread_create_detached(&rpt_connect_threadid, NULL, rpt_link_connect, (void *) connect_data) < 0) {
+			if (ast_pthread_create_detached(&connect_threadid, NULL, rpt_link_connect, (void *) connect_data) < 0) {
```
Allan-N: Did you want to set `mylink->connect_in_progress = 1;` before creating the thread?  Else, did you really want to be clearing it, below?

apps/app_rpt/rpt_functions.c
```diff
@@ -171,16 +171,9 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
 		connect_data->perma = perma;
 		connect_data->command_source = command_source;
 		connect_data->mylink = mylink;
-		rpt_mutex_lock(&myrpt->lock);
-		myrpt->connect_thread_count++;
-		rpt_mutex_unlock(&myrpt->lock);
-		if (ast_pthread_create_detached(&rpt_connect_threadid, NULL, rpt_link_connect, (void *) connect_data) < 0) {
+		if (ast_pthread_create_detached(&connect_threadid, NULL, rpt_link_connect, (void *) connect_data) < 0) {
```
Allan-N: Did you want to set `mylink->connect_in_progress = 1;` before creating the thread?  and clear the flag if the thread could not be created?

apps/app_rpt/rpt_functions.c
mkmer: I originally thought that mylink was the link, we pass mylink in for telemetry "stuff".  I missed removing one of the `=0`. The link we want to set `->connect_in_progress` is the link we create inside the thread.  First part of the thread checks if we can create it, if not we do failure "stuff". Further in, if we can create a new call, we alloc an `l` then set the flag.

PR #891, opened 2026-01-01
chan_simpleusb.c : Fix variable declaration in switch case statement

guosuisheng:
The query page status case ('?') in the paging application had a  variable declaration (struct ast_frame wf) directly after the case  label without enclosing braces. This violates C syntax rules as case  labels can only be followed by statements, not declarations.

## Summary by CodeRabbit

* **Refactor**
  * Improved mutex management and code scoping in the query page status handler for better resource management and stability.


mkmer:
We are incrementally updating code to match the Asterisk coding guidelines - touching "old code" can (and does) require updates to match the standard.  
There is often spaces and tabs that are also not "obvious" so it is best to run the linter locally, but you can fight with it here too if you like.

PR #890, opened 2026-01-01
chan_usbradio.c: Remove unused configuration variable rxctcssadj

okcrum:
Ongoing code cleanup

Fixes: #751

## Summary by CodeRabbit

* **Refactor**
  * Removed RX CTCSS adjustment feature. CTCSS functionality now relies on standard configuration paths instead.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #899, opened 2026-01-05
chan_usbradio.c: fix backward comptibility for EEPROM r/w

okcrum:
PR #890 which removed the rxctcssadj variable from chan_usbradio.c changed the memory block which is read from and written to EEPROM. This change adds a dummy variable as a placeholder to restore backward compatibility.

DeveloperNote: This will change when rxvoiceadj status is addressed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RX CTCSS adjustment persistence to ensure settings are correctly saved and restored from device memory.

### DIFF
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -327,8 +327,6 @@ struct chan_usbradio_pvt {
 	int rxmixerset;
 	int txboost;
 	float rxvoiceadj;
-	float old_rxctcssadj; /* This variable is a placeholder for EEPROM r/w compatibility */
-						  /* This may change when the need for rxvoiceadj is addressed */
 	int txmixaset;
 	int txmixbset;
 	int txctcssadj;
@@ -761,7 +759,6 @@ static int load_tune_config(struct chan_usbradio_pvt *o, const struct ast_config
 	o->txmixaset = 500;
 	o->txmixbset = 500;
 	o->rxvoiceadj = 0.5;
-	o->old_rxctcssadj = 0.5;
 	o->txctcssadj = 200;
 	o->rxsquelchadj = 500;
 	o->txslimsp = DEFAULT_TX_SOFT_LIMITER_SETPOINT;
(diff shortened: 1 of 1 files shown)


channels/chan_usbradio.c
```diff
@@ -4555,6 +4558,7 @@ static void tune_write(struct chan_usbradio_pvt *o)
 		o->eeprom[EEPROM_USER_TXMIXASET] = o->txmixaset;
 		o->eeprom[EEPROM_USER_TXMIXBSET] = o->txmixbset;
 		memcpy(&o->eeprom[EEPROM_USER_RXVOICEADJ], &o->rxvoiceadj, sizeof(float));
+		memcpy(&o->eeprom[EEPROM_USER_RXCTCSSADJ], &o->rxdummy, sizeof(float));
```
Allan-N: If you make any changes for EEPROM compatibility then you might just zero out (or set the default value for) the float value.  Or, just leave whatever value is present alone.

channels/chan_usbradio.c
```diff
@@ -327,6 +327,8 @@ struct chan_usbradio_pvt {
 	int rxmixerset;
 	int txboost;
 	float rxvoiceadj;
+	float rxdummy; // This variable is a placeholder for EEPROM r/w compatibility; it was rxctcssadj
```
mkmer: As much as I hate these comments... we are using /* */ not //

channels/chan_usbradio.c
Allan-N:
```
const float old_rxctcssadj = 0.5;
memcpy(&o->eeprom[EEPROM_USER_RXCTCSSADJ], &old_rxctcssadj, sizeof(float));
```

Allan-N (changes requested):

channels/chan_usbradio.c
```diff
@@ -289,44 +289,46 @@ struct chan_usbradio_pvt {
...
+	int set_txfreq; /* in Hz */
 	int set_rxfreq;
 
-	//      end remote operation info
+	/*      end remote operation info */
 
 	int rxmixerset;
 	int txboost;
 	float rxvoiceadj;
+	float old_rxctcssadj; /* This variable is a placeholder for EEPROM r/w compatibility */
```
Allan-N: What are we using this variable for?  Why is it needed?

channels/chan_usbradio.c
```diff
@@ -759,6 +761,7 @@ static int load_tune_config(struct chan_usbradio_pvt *o, const struct ast_config
 	o->txmixaset = 500;
 	o->txmixbset = 500;
 	o->rxvoiceadj = 0.5;
+	o->old_rxctcssadj = 0.5;
```
Allan-N: Why do we need to set `o->old_rxctcssadj`?  Is that value ever used?  anywhere?  If not, this line can/should be removed.

channels/chan_usbradio.c
```diff
@@ -1211,6 +1214,7 @@ static void *hidthread(void *arg)
 							o->txmixaset = o->eeprom[EEPROM_USER_TXMIXASET];
 							o->txmixbset = o->eeprom[EEPROM_USER_TXMIXBSET];
 							memcpy(&o->rxvoiceadj, &o->eeprom[EEPROM_USER_RXVOICEADJ], sizeof(float));
+							memcpy(&o->old_rxctcssadj, &o->eeprom[EEPROM_USER_RXCTCSSADJ], sizeof(float));
```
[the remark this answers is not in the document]
okcrum: > Why do we need to copy the EEPROM value and store it in `o->old_rxctcssadj`? Is that value ever used? anywhere? If not, this line can/should be removed.

Just noticed and removed all that.

PR #918, opened 2026-01-21
app_rpt: Remove requirement for `ident` to be defined/configured

### LINKED ISSUE #381: Can we remove the requirement that each node have a non-null "idrecording"
Right now, app_rpt has a requirement that each node have a non-null "idrecording" variable. If not present, the main thread for nodes with a NULL setting is not started.  While that's OK for most configurations there are cases (e.g. radio-less nodes) where an ID is not required.

Can the "idrecording" requirement be removed?

mkmer:
Fixes #381 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability when repeater nodes are misconfigured: nodes with missing identifiers no longer cause the main repeater controller to exit. Affected repeaters are now skipped or disabled so other nodes continue operating.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -5853,11 +5853,6 @@ static void *rpt_master(void *ignore)
 		rpt_vars[i].offset = REM_SIMPLEX;
 		rpt_vars[i].powerlevel = REM_LOWPWR;
 		rpt_vars[i].splitkhz = 0;
-
-		if (rpt_vars[i].p.ident && (!*rpt_vars[i].p.ident)) {
-			ast_log(LOG_WARNING, "Did not specify ident for node %s\n", rpt_vars[i].name);
-			pthread_exit(NULL);
-		}
 		rpt_vars[i].ready = 0;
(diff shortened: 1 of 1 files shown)


Allan-N:
>I don't know... we would need to dig a bit deeper to see what happens if it's not configured.

and that's why we have yet to address Issue #381 :)

> I'd say we do need one if we have a non DAHDI (soon to be local) channel because it's "probably" connected to a radio.

A "local" channel could be a radio-less node.

> I suppose it's the responsibility of the operator to comply with the law.

Agreed.

> I was just "agreeing" with CodeRabbit that we shouldn't stop all repeaters if only 1 was missing and id :)

Sure.  But, now the configuration error will not be as pronounced.

I see 3 paths.
- address #381 now (removing the requirement)
- make the coderabbit suggestion now and look at #381 later
- punt this change and save #381 for later

Allan-N:
>Maybe just log a warning?

Why?  If we allow one to not specify an ID then a configuration without one is fine ... and it's up to the operator to check that an ID is being transmitted.

PR #941, opened 2026-02-21
app_rpt: Fix unused-but-set-variable errors with gcc 16

InterLinked1:
Disable the variables and code related to parking for now, since that isn't currently implemented.

Resolves: #940

## Summary by CodeRabbit

* **Chores**
  * Suppressed an unused-variable compiler warning in the Return/parking flow; no behavioral changes.
  * Maintains existing timeout logic and comments while cleaning up build-time noise.


mkmer:
If it's not used, why not remove it?

PR #973, opened 2026-03-13
app_rpt: Change default linkpost_time to 30 seconds

okcrum:
Testing has uncovered a previously unknown interoperability issue with HamVoIP-1.7.1. This requires linkpost_time to be in the range 10-40 seconds, instead of 60.

## Summary by CodeRabbit

* **Updates**
  * Changed default linkpost_time from 60 to 30 seconds
  * Reduced maximum allowed linkpost_time from 600 to 40 seconds (minimum remains 10 seconds)


mkmer (approved):
Probably should update our docs too...

PR #979, opened 2026-03-20
app_rpt: only downgrade nodes on a links list

mkmer:
Needed to "remember" that start each time we traverse the list, otherwise we incorrectly downgrade all of the links (including adjacent links)


apps/app_rpt/rpt_link.c
```diff
@@ -452,10 +452,9 @@ int __mklinklist(struct rpt *myrpt, struct rpt_link *mylink, struct ast_str **bu
 		return 0;
 	}
 
-	spos = ast_str_strlen(*buf); /* current buf size (before we add our stuff) */
-
 	/* go thru all links */
 	RPT_LIST_TRAVERSE(myrpt->links, l, l_it) {
+		spos = ast_str_strlen(*buf); /* current buf size (before we add our stuff) */
```
Allan-N: You "could" move this down, past line 474 (and all of the "continue" statements).

PR #988, opened 2026-03-31
app_rpt: a bit of logging and a bunch of cleanup

Allan-N:
Added a bit of [helpful?] logging.  But, mostly, over the past few months I've found a bunch of little things that could be "cleaned up".

## Release Notes

* **Bug Fixes**
  * Enhanced error reporting with detailed failure cause information
  * Corrected DTMF frame initialization timing
  * Improved node log writing accuracy

* **Documentation**
  * Updated configuration documentation with clarified DNS requirements and default behavior specifications

* **Improvements**
  * Refined error logging and telemetry mode reporting for better diagnostics
  * Improved protocol packet header handling and byte-order compliance

### DIFF
--- a/apps/app_rpt/rpt_config.c
+++ b/apps/app_rpt/rpt_config.c
@@ -87,8 +87,9 @@ int get_wait_interval(struct rpt *myrpt, enum rpt_delay type)
 
 	if (wait_times) {
 		wait_times_save = ast_strdup(wait_times);
-		if (!wait_times_save)
+		if (!wait_times_save) {
 			return 0;
+		}
 	}
 
 	switch (type) {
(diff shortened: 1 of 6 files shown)


apps/app_rpt/rpt_functions.c
```diff
@@ -296,20 +295,15 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
...
-		};
+	case 6: /* All Links Off, including permalinks */
 		rpt_mutex_lock(&myrpt->lock);
 		myrpt->savednodes[0] = 0;
 		/* loop through all links */
 		RPT_LIST_TRAVERSE(myrpt->links, l, l_it) {
 			char c1;
 
-			if ((l->name[0] <= '0') || (l->name[0] > '9')) { /* Skip any IAXRPT monitoring */
+			if ((l->name[0] <= '0') || (l->name[0] > '9')) {
```
mkmer: Wonder if we should make this a macro -> non numeric node id? (Or what ever the heck it means for the first digit to be non numeric)

Allan-N: Do we need an IS_VALID_NODE()?

or LOOKS_LIKE_A_VALID_NODE() that only looks at the 1st character :)

PR #1009, opened 2026-04-08
app_rpt.c:  Add RPT_CONFIG_EXPLODE_STRING macro

mkmer:
~~I couldn't catch all 4 of these - maybe not worth it???~~

## Summary by CodeRabbit

* **Refactor**
  * Consolidated repeated configuration-loading and splitting logic into a single pattern to reduce duplication.
  * Preserves existing defaults and behavior; uppercase transformation is still applied for the same specific settings.
  * No changes to exported/public interfaces or user-visible configuration semantics.

### DIFF
--- a/apps/app_rpt/rpt_config.c
+++ b/apps/app_rpt/rpt_config.c
@@ -881,8 +881,11 @@ void load_rpt_vars(int n, int init)
 		rpt_vars[n].p.var = default; \
 	}
 
-#define RPT_CONFIG_EXPLODE_STRING(buf_var, count_var, arr_var, name, transform) \
+#define RPT_CONFIG_EXPLODE_STRING(buf_var, count_var, arr_var, name, def_val, transform) \
 	val = ast_variable_retrieve(cfg, cat, name); \
+	if (!val) { \
+		val = def_val; \
+	} \
 	if (val) { \
 		tmp = ast_strdup(val); \
 		if (tmp) { \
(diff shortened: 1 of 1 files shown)


apps/app_rpt/rpt_config.c
```diff
@@ -881,6 +881,17 @@ void load_rpt_vars(int n, int init)
 		rpt_vars[n].p.var = default; \
 	}
 
+#define RPT_CONFIG_EXPLODE_STRING(buf_var, count_var, arr_var, name, transform) \
```
Allan-N: Maybe modify this to include a default value if the variable is not present in the configuration file ... like :
```
#define RPT_CONFIG_EXPLODE_STRING(buf_var, count_var, arr_var, name, def_val, transform) \
	val = ast_variable_retrieve(cfg, cat, name); \
	if (!val) { \
		val = def_val; \
	} \
	if (val) { \
		tmp = ast_strdup(val); \
		if (tmp) { \
			rpt_vars[n].p.buf_var = tmp; \
			rpt_vars[n].p.count_var = \
				explode_string(transform(tmp), (char **) rpt_vars[n].p.arr_var, ARRAY_LEN(rpt_vars[n].p.arr_var), ',', 0); \
		} \
	}
```

apps/app_rpt/rpt_config.c
Allan-N: With the above suggestion then this block ... and the other usages below would become :
```
	RPT_CONFIG_EXPLODE_STRING(extnodefiles_buf, extnodefilesn, extnodefiles, "extnodefile", EXTNODEFILE, )
	RPT_CONFIG_EXPLODE_STRING(locallinknodes_buf, locallinknodesn, locallinknodes, "locallinknodes", NULL, )
	RPT_CONFIG_EXPLODE_STRING(lconn_buf, nlconn, lconn, "lconn", NULL, strupr)
	RPT_CONFIG_EXPLODE_STRING(ldisc_buf, nldisc, ldisc, "ldisc", NULL, strupr)

```

apps/app_rpt/rpt_config.c
mkmer: I like that better than what I just whipped up...  No need for S_OR that way

apps/app_rpt/rpt_config.c
```diff
@@ -996,16 +1000,10 @@ void load_rpt_vars(int n, int init)
...
-	if (tmp) {
-		rpt_vars[n].p.extnodefilesn =
-			explode_string(tmp, (char **) rpt_vars[n].p.extnodefiles, ARRAY_LEN(rpt_vars[n].p.extnodefiles), ',', 0);
-		rpt_vars[n].p.extnodefiles_buf = tmp;
-	}
-	RPT_CONFIG_EXPLODE_STRING(locallinknodes_buf, locallinknodesn, locallinknodes, "locallinknodes", )
-	RPT_CONFIG_EXPLODE_STRING(lconn_buf, nlconn, lconn, "lconn", strupr)
-	RPT_CONFIG_EXPLODE_STRING(ldisc_buf, nldisc, ldisc, "ldisc", strupr)
+	RPT_CONFIG_EXPLODE_STRING(extnodefiles_buf, extnodefilesn, extnodefiles, "extnodefile", , EXTNODEFILE)
+	RPT_CONFIG_EXPLODE_STRING(locallinknodes_buf, locallinknodesn, locallinknodes, "locallinknodes", , 0)
```
Allan-N, line 1004: Can I ask that we please use `NULL` Instead of `0` (matching the [pointer] return value from ast_config_retrieve if the variable is not found)

mkmer: Yes, but I'm going with your macro - I think it's clearer and better than this one.

Allan-N: Our comments crossed paths.

Allan-N (approved):

PR #1020, opened 2026-04-13
app_rpt: no need to pass "size - 1" when API includes NUL term

Allan-N:
The ast_copy_string(), snprintf(), and strftime() APIs all ensure that the returned buffer will include the NUL termination character.  As such, there is no need to pass a buffer size of "xxx - 1".

## Summary by CodeRabbit

* **Chores**
  * Updated buffer handling across multiple application modules for improved consistency and code quality.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -892,12 +892,12 @@ void rpt_event_process(struct rpt *myrpt)
 				myrpt->cmdAction.functionNumber = thisAction;
 				myrpt->cmdAction.param[0] = 0;
 				if (argc > 1) {
-					ast_copy_string(myrpt->cmdAction.param, argv[1], MAX(sizeof(myrpt->cmdAction.param), MAXDTMF));
+					ast_copy_string(myrpt->cmdAction.param, argv[1], sizeof(myrpt->cmdAction.param));
 				}
 				myrpt->cmdAction.digits[0] = 0;
 				if (argc > 2) {
-					ast_copy_string(myrpt->cmdAction.digits, argv[2], MAX(sizeof(myrpt->cmdAction.digits), MAXDTMF));
-					snprintf(myrpt->cmdAction.param, MAX(sizeof(myrpt->cmdAction.param), MAXDTMF), "%s,%s", argv[1], argv[2]);
+					ast_copy_string(myrpt->cmdAction.digits, argv[2], sizeof(myrpt->cmdAction.digits));
+					snprintf(myrpt->cmdAction.param, sizeof(myrpt->cmdAction.param), "%s,%s", argv[1], argv[2]);
 				}
 				myrpt->cmdAction.command_source = SOURCE_RPT;
 				myrpt->cmdAction.state = CMD_STATE_READY;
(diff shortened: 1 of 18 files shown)


apps/app_rpt/rpt_channel.c
```diff
@@ -178,11 +178,17 @@ void do_dtmf_local(struct rpt *myrpt, char c)
 	char digit;
 
 	if (c) {
-		snprintf(myrpt->dtmf_local_str + strlen(myrpt->dtmf_local_str), sizeof(myrpt->dtmf_local_str) - 1, "%c", c);
+		i = strlen(myrpt->dtmf_local_str);
+		if (i < sizeof(myrpt->dtmf_local_str) - 1) {
+			myrpt->dtmf_local_str[i++] = c;
+			myrpt->dtmf_local_str[i] = '\0';
+		}
+
```
mkmer: Since we are tweaking here, should we change the scope of `i`?
at ln 180, `int i;`, or `int dmtf_len;` or ... , and then move `int i;` definition down inside the  `if (myrpt->dtmf_local_str[0]) {` around ln 191?

channels/chan_voter.c
```diff
@@ -5428,20 +5430,20 @@ static void *voter_reader(void *data)
 						timetv.tv_usec -= 1000000;
 					}
 					timestuff = (time_t) timetv.tv_sec;
-					strftime(timestr, sizeof(timestr) - 1, "%Y %T", localtime((time_t *) &timestuff));
+					strftime(timestr, sizeof(timestr), "%Y %T", localtime((time_t *) &timestuff));
```
mkmer: `strftime()` apparently can fail returning `0` -> at which point we are not guaranteed to be NULL terminated.  We probably should check of `0` and be sure to terminate these. (Probably in a new PR)  Maybe a macro?

apps/app_rpt.c
```diff
@@ -892,12 +892,12 @@ void rpt_event_process(struct rpt *myrpt)
 				myrpt->cmdAction.functionNumber = thisAction;
 				myrpt->cmdAction.param[0] = 0;
 				if (argc > 1) {
-					ast_copy_string(myrpt->cmdAction.param, argv[1], MAXDTMF - 1);
+					ast_copy_string(myrpt->cmdAction.param, argv[1], MAX(sizeof(myrpt->cmdAction.param), MAXDTMF));
```
mkmer: param is defined as `MAXMACRO` -> probably not MAXDTMF?
Maybe follow the nitpick and just use `sizeof()` for these 3 instances?

channels/chan_voter.c
Allan-N: Not something I'm worried about.  But, if you'd like to ensure the buffer is terminated then we can switch to `ast_strftime()` that starts off zero'ing buf[0].

mkmer: Probably "should" - Your call here.
I'm a little surprised CodeRabbit didn't complain, from the research I did, it can be in an "invalid" state (right?)

mkmer (approved):

channels/chan_voter.c
Allan-N: Well, it turns out that ast_strftime is not just a simple drop-in :-(

For now, I think I'm going to not worry about strftime() returning <= 0.

PR #1025, opened 2026-04-20
app_rpt: re-install lost ASL2 kerchunk timer

### LINKED ISSUE #90: Kerchunk filter
As mentioned at https://github.com/AllStarLink/ASL-Asterisk/issues/25.

A kerchunk filter prevents quick keying up of the repeater by providing a delay in the amount of time a signal must be present on the receiver (but not links) before allowing the signal to key the repeater and links, when repeater has been inactive for a selected period. Once the repeater is successfully keyed up the delay becomes inactive until the repeater is inactive again.

Configurable timers per node:
inactivity timer in minutes, default 10 
keyup delay in milliseconds, default 0 (defeats kerchunk filter)

mkmer:
https://github.com/AllStarLink/ASL-Asterisk/pull/69/changes

Fixes #90 

## Release Notes

* **New Features**
  * Enhanced repeater systems with kerchunk detection and state tracking capabilities. Added timing-based control mechanism for telemetry filtering.

* **Configuration**
  * New `kerchunk_time` configuration parameter for repeater nodes enables adjustable kerchunk sensitivity. Available range: 0–5 seconds (default: 0).

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -5039,7 +5039,7 @@ static void *rpt(void *this)
 	lastexttx = 0;
 	myrpt->keyed = 0;
 	myrpt->txkeyed = 0;
-	myrpt->kerchunked_ok = 0;
+	myrpt->kerchunked = 0;
 	time(&myrpt->lastkeyedtime);
 	myrpt->lastkeyedtime -= RPT_LOCKOUT_SECS;
 	time(&myrpt->lasttxkeyedtime);
@@ -5434,7 +5434,7 @@ static void *rpt(void *this)
 			myrpt->macropatch = 0;
 			channel_revert(myrpt);
 		}
-		if (!myrpt->totimer || (!myrpt->mustid && myrpt->p.beaconing) || !myrpt->kerchunked_ok) {
+		if (!myrpt->totimer || (!myrpt->mustid && myrpt->p.beaconing) || !myrpt->kerchunked) {
 			/* get rid of tail if timed out, beaconing, or kerchunked */
 			myrpt->tailtimer = 0;
 		}
@@ -5537,7 +5537,7 @@ static void *rpt(void *this)
 		/* Detect and log keyed to unkeyed transition point */
 		if (!totx && lasttx) {
 			lasttx = 0;
-			myrpt->kerchunked_ok = 0;
+			myrpt->kerchunked = 0;
 			log_unkeyed(myrpt);
 		}
 		time(&t);
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -5039,6 +5039,7 @@ static void *rpt(void *this)
 	lastexttx = 0;
 	myrpt->keyed = 0;
 	myrpt->txkeyed = 0;
+	myrpt->kerchunked = 0;
 	time(&myrpt->lastkeyedtime);
 	myrpt->lastkeyedtime -= RPT_LOCKOUT_SECS;
 	time(&myrpt->lasttxkeyedtime);
@@ -5218,16 +5219,25 @@ static void *rpt(void *this)
 				}
 			}
 
-			if (myrpt->sleep)
+			if (myrpt->sleep) {
 				myrpt->localtx = 0; /* No RX if asleep */
-			else
+			} else {
 				myrpt->localtx = myrpt->keyed; /* Set localtx to keyed state if awake */
+			}
 		} else {
 			myrpt->localtx = myrpt->keyed; /* If sleep disabled, just copy keyed state to localrx */
 		}
+
+		/* Set the kerchunk timer */
+		if ((myrpt->remrx || myrpt->keyed) && !myrpt->kerchunk_timer) {
+			time(&myrpt->kerchunk_timer);
+		}
+
 		/* Create a "must_id" flag for the cleanup ID */
-		if (myrpt->p.idtime) /* ID time must be non-zero */
+		if (myrpt->p.idtime) { /* ID time must be non-zero */
 			myrpt->mustid |= (myrpt->idtimer) && (myrpt->keyed || myrpt->remrx);
+		}
+
 		if (myrpt->keyed || myrpt->remrx) {
 			/* Set the inactivity was keyed flag and reset its timer */
 			myrpt->rptinactwaskeyedflag = 1;
@@ -5422,8 +5432,8 @@ static void *rpt(void *this)
 			myrpt->macropatch = 0;
 			channel_revert(myrpt);
 		}
-		if (!myrpt->totimer || (!myrpt->mustid && myrpt->p.beaconing)) {
-			/* get rid of tail if timed out or beaconing */
+		if (!myrpt->totimer || (!myrpt->mustid && myrpt->p.beaconing) || !myrpt->kerchunked) {
+			/* get rid of tail if timed out, beaconing, or kerchunked */
 			myrpt->tailtimer = 0;
 		}
 		if (myrpt->totimer) {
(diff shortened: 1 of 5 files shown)


mkmer:
I'm assuming this code actually worked in the first place?
Requesters: please test and confirm it's doing what you "thought" it did.
It looks to me like all it does is filter of CT, not keyup of the repeater.

apps/app_rpt.c
```diff
@@ -5039,6 +5039,7 @@ static void *rpt(void *this)
 	lastexttx = 0;
 	myrpt->keyed = 0;
 	myrpt->txkeyed = 0;
+	myrpt->kerchunked = 0;
 	time(&myrpt->lastkeyedtime);
 	myrpt->lastkeyedtime -= RPT_LOCKOUT_SECS;
```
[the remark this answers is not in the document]
mkmer: That's not going to work - we are setting the timer here on keyup one time based on a 0 value in timer.  IF we do the else part, we will just cycle through setting and unsetting the timer.

tsawyer:
Regarding the coderabbitai warning about the in/activity timer: This timer is a critical to the function of the Kerchunk filter. It is necessary to determine if a keyup delay should be in effect. 

- A repeater becomes locally active when COR is true long enough to satisfying the keyup delay timer.
- Once locally active the in/activity timer starts which prevents subsequent keyup delays.
- Each subsequent COR during the in/activity timer period resets the in/activity timer. 
- The in/activity timer times out when no COR for timer value. Keyup delay now in effect again.

configs/rpt/rpt.conf
```diff
@@ -140,6 +140,7 @@ time_out_reset_unkey_interval = 0   ; The minimum time for no carrier detect aft
 time_out_reset_kerchunk_interval = 250 ; Allows users to reset the time out timer when there is a signal coming in from a link (optional).
 									; A value of 0 disables this feature. Valid only if time_time_out_reset_unkey_interval is non zero. 
 									; Value is in milliseconds. Default is 250ms, minimum is 0ms, maximum is 3000ms
+kerchunk_time = 1                   ; kerchunk time (in seconds) (optional, default is 0, max 5)
```
Allan-N: Do we want to have this line start out in the .conf as a comment?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -3538,14 +3539,26 @@ void rpt_telemetry(struct rpt *myrpt, enum rpt_tele_mode mode, void *data)
 
 	case UNKEY:
 	case LOCUNKEY:
+		kerchunk_time_stamp = myrpt->kerchunk_timer;
```
Allan-N: Can/should these 2 lines be moved down a bit, to just before they are checked?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -3562,12 +3575,26 @@ void rpt_telemetry(struct rpt *myrpt, enum rpt_tele_mode mode, void *data)
 		break;
 
 	case LINKUNKEY:
-		mylink = (struct rpt_link *) data;
+		kerchunk_time_stamp = myrpt->kerchunk_timer;
```
Allan-N: Can/should these 2 lines be moved down a bit, to just before they are checked?

Allan-N:
FYI : it looks as if the https://github.com/AllStarLink/ASL-Asterisk/pull/69 changes were never merged into ASL2 ... and that's why they were also not in ASL3.

Allan-N:
I'm also wondering if the kerchunk time units should be changed from seconds to milliseconds (i.e. how many milliseconds vs. seconds have elapsed) ?

tsawyer:
> I'm also wondering if the kerchunk time units should be changed from seconds to milliseconds (i.e. how many milliseconds vs. seconds have elapsed) ?

Yeah, One second is too long. Milliseconds would be better.

mkmer:
Reiterating: I don't think this PR does what #90 is asking for.  It looks to me like it only suppresses CT "beeps" and does NOT prevent the repeater from keying up.

apps/app_rpt/rpt_config.c
```diff
@@ -937,6 +937,7 @@ void load_rpt_vars(int n, int init)
 
 	/* configure how we interact with "stats.allstarlink.org" */
 	RPT_CONFIG_VAR_INT_DEFAULT_MIN_MAX(statpost_time, "statpost_time", 60, 30, 600);
+	RPT_CONFIG_VAR_INT_DEFAULT_MIN_MAX(kerchunktime, "kerchunk_time", 0, 0, 5);
```
Allan-N, line 940: If we are changing the time from seconds to milliseconds then the "5" should become "5000"

apps/app_rpt/rpt_telemetry.c
```diff
@@ -3543,9 +3544,21 @@ void rpt_telemetry(struct rpt *myrpt, enum rpt_tele_mode mode, void *data)
 			return;
 		}
 
+		kerchunk_time_stamp = myrpt->kerchunk_timer;
+		myrpt->kerchunk_timer = 0;
+
+		if (myrpt->p.kerchunktime && (time(NULL) - kerchunk_time_stamp) < myrpt->p.kerchunktime) {
```
Allan-N, line 3550: If we are changing the time from seconds to milliseconds then does this test also need to change?

apps/app_rpt.c
```diff
@@ -5218,16 +5219,27 @@ static void *rpt(void *this)
...
 				myrpt->localtx = myrpt->keyed; /* Set localtx to keyed state if awake */
+			}
 		} else {
 			myrpt->localtx = myrpt->keyed; /* If sleep disabled, just copy keyed state to localrx */
 		}
+
+		/* Set the kerchunk timer */
+		if ((myrpt->remrx || myrpt->keyed)) {
+			if (!myrpt->kerchunk_timer) {
+				time(&myrpt->kerchunk_timer);
```
Allan-N, line 5234: If we are changing the time from seconds to milliseconds then do we need to save the current time (in milliseconds) ?

Allan-N (changes requested):

apps/app_rpt/rpt_config.c
mkmer: I missed backing out the config file change.  I think we need to verify it "does" what is wanted, then we can mess around with the duration.  Switching to ms is not "easy".

configs/rpt/rpt.conf
```diff
@@ -140,7 +140,7 @@ time_out_reset_unkey_interval = 0   ; The minimum time for no carrier detect aft
 time_out_reset_kerchunk_interval = 250 ; Allows users to reset the time out timer when there is a signal coming in from a link (optional).
 									; A value of 0 disables this feature. Valid only if time_time_out_reset_unkey_interval is non zero. 
 									; Value is in milliseconds. Default is 250ms, minimum is 0ms, maximum is 3000ms
-;kerchunk_time = 500                ; kerchunk time (in milliseconds) (optional, default is 0, max 5000)
+;kerchunk_time = 1                  ; kerchunk time (in seconds) (optional, default is 0, max 5000)
```
Allan-N: If you are [temporarily] switching back to seconds then 5000 -> 5

Allan-N:
Did https://github.com/AllStarLink/app_rpt/pull/1066 provided an alternate/workable solution?

mkmer:
I don't think so.  This one prevents Courtesy tones from "beeping" on short unkey/keyups essnetially durring picket fencing / noisy users.
It may be just as good or better though :)

PR #1037, opened 2026-04-25
rpt_serial: replace for() loops with usleep() for parallel port timing

### LINKED ISSUE #1016: Empty `for()` loops intended for time delay???
Ummm... Is this really an empty for() loop?

_Originally posted by @Allan-N in https://github.com/AllStarLink/app_rpt/pull/1014#discussion_r3060613527_

okcrum:
Fixes: #1016

## Release Notes

* **Refactor**
  * Improved timing efficiency by replacing CPU-intensive delay loops with system sleep functions, reducing unnecessary CPU usage while maintaining the same operational delays and functionality.

### DIFF
--- a/apps/app_rpt/rpt_serial.c
+++ b/apps/app_rpt/rpt_serial.c
@@ -430,6 +430,7 @@ static int rbi_pltocode(char *str)
 static void rbi_out_parallel(struct rpt *myrpt, unsigned char *data)
 {
 #ifdef HAVE_SYS_IO
+#ifdef __i386__
 	int i, j;
 	unsigned char od, d;
 	static volatile long long delayvar;
@@ -453,6 +454,7 @@ static void rbi_out_parallel(struct rpt *myrpt, unsigned char *data)
 	/* >= 50 us */
 	usleep(50);
 
+#endif /* __i386__ */
 #endif /* HAVE_SYS_IO */
 }
 
--- a/apps/app_rpt/rpt_serial.c
+++ b/apps/app_rpt/rpt_serial.c
@@ -430,7 +430,6 @@ static int rbi_pltocode(char *str)
 static void rbi_out_parallel(struct rpt *myrpt, unsigned char *data)
 {
 #ifdef HAVE_SYS_IO
-#ifdef __i386__
 	int i, j;
 	unsigned char od, d;
 	static volatile long long delayvar;
@@ -454,7 +453,6 @@ static void rbi_out_parallel(struct rpt *myrpt, unsigned char *data)
 	/* >= 50 us */
 	usleep(50);
 
-#endif /* __i386__ */
 #endif /* HAVE_SYS_IO */
 }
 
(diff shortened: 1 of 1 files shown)


apps/app_rpt/rpt_serial.c
Allan-N: I think we should remove this `#ifdef` line (and the matching `#endif` below.

okcrum: OK. Done. It was the only one in the codebase.

Allan-N (approved):

apps/app_rpt/rpt_serial.c
okcrum: Foo. The ubuntu pull request is failing because the \_\_i386\_\_ is gone, and there is no <sys/io.h>.

outb() seems to be a 32 bit function. So this is sort of a dead code block or needs a rewrite.

Allan-N: We also use outb() in res/res_usbradio.c

Allan-N: Looks like rpt_serial.c is missing :
```
#ifdef HAVE_SYS_IO
#include <sys/io.h>
#endif 
```

okcrum: Nope, that didn't do it either, with or without the conditional. So it's one of the other #includes in res_usbradio.c that has the function prototype.

I'm not getting a warning here on Debian 13.

Allan-N: Two source files with outb().  One compiles OK, the other not.  Gotta be a header file issue.

apps/app_rpt/rpt_serial.c
```diff
@@ -9,6 +9,10 @@
 
 #include <termios.h>
 
+#ifdef HAVE_SYS_IO
```
mkmer: If we take the ifdef out here, it compiles - even though clearly `HAVE_SYS_IO` is not defined, the compiler is looking to compile anyways?  I dunno.

apps/app_rpt/rpt_serial.c
```diff
@@ -430,7 +434,6 @@ static int rbi_pltocode(char *str)
 static void rbi_out_parallel(struct rpt *myrpt, unsigned char *data)
 {
 #ifdef HAVE_SYS_IO
-#ifdef __i386__
 	int i, j;
 	unsigned char od, d;
 	static volatile long long delayvar;
```
mkmer: This guy is unused now. `static volatile long long delayvar;`

apps/app_rpt/rpt_serial.c
mkmer: I pushed the fix in, just to see if it worked here too.  Not sure why it's broken with the ifdef in place.

apps/app_rpt/rpt_serial.c
```diff
@@ -9,6 +9,8 @@
 
 #include <termios.h>
 
+#include <sys/io.h>
```
Allan-N: Raspberry Pi does not have <sys/io.h>

apps/app_rpt/rpt_serial.c
Allan-N: The latest changes look good to me.

apps/app_rpt/rpt_serial.c
Allan-N: The latest changes look good to me.

apps/app_rpt/rpt_serial.c
Allan-N: The latest changes look good to me.

Allan-N (approved):

PR #1040, opened 2026-04-27
app_rpt: Correctly handle ast_dsp_process() frames

mkmer:
`ast_dsp_process()` returns the pointer passed in, or frees the pointer and returns a new pointer.  This requires a frame that can be freed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory management and error handling in DTMF signal processing across voice channels, enhancing system stability and reliability.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3848,14 +3848,16 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 				struct ast_frame *frame = NULL;
 				struct ast_frame *f_dup = ast_frdup(f);
 				/* leave f alone */
-				frame = ast_dsp_process(myrpt->rxchannel, myrpt->dsp, f_dup);
-				i = (frame->frametype == AST_FRAME_DTMF && frame->subclass.integer == 'q') ? 1 : 0; /* q indicates frequency hit */
+				if (f_dup) {
+					frame = ast_dsp_process(myrpt->rxchannel, myrpt->dsp, f_dup);
+					i = (frame->frametype == AST_FRAME_DTMF && frame->subclass.integer == 'q') ? 1 : 0; /* q indicates frequency hit */
 
-				if (f_dup != frame) {
-					ast_frfree(frame);
+					if (f_dup != frame) {
+						ast_frfree(frame);
+					}
+
+					ast_frfree(f_dup);
 				}
-
-				ast_frfree(f_dup);
 #else
 				i = tone_detect(&myrpt->burst_tone_state, f->data.ptr, f->samples);
 #endif
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3848,6 +3848,7 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 				struct ast_frame *frame = NULL;
 				struct ast_frame *f_dup;
 
+				i = 0;
 				/* leave f alone */
 				f_dup = ast_frdup(f);
 				if (f_dup) {
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3846,17 +3846,14 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 			} else {
 #ifdef NATIVE_DSP
 				struct ast_frame *frame = NULL;
-				struct ast_frame *f_dup = ast_frdup(f);
+				struct ast_frame *f_dup;
+
 				/* leave f alone */
+				f_dup = ast_frdup(f);
 				if (f_dup) {
 					frame = ast_dsp_process(myrpt->rxchannel, myrpt->dsp, f_dup);
 					i = (frame->frametype == AST_FRAME_DTMF && frame->subclass.integer == 'q') ? 1 : 0; /* q indicates frequency hit */
-
-					if (f_dup != frame) {
-						ast_frfree(frame);
-					}
-
-					ast_frfree(f_dup);
+					ast_frfree(frame);
 				}
 #else
 				i = tone_detect(&myrpt->burst_tone_state, f->data.ptr, f->samples);
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3846,11 +3846,16 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 			} else {
 #ifdef NATIVE_DSP
 				struct ast_frame *frame = NULL;
-
+				struct ast_frame *f_dup = ast_frdup(f);
 				/* leave f alone */
-				frame = ast_dsp_process(myrpt->rxchannel, myrpt->dsp, ast_frdup(f));
+				frame = ast_dsp_process(myrpt->rxchannel, myrpt->dsp, f_dup);
 				i = (frame->frametype == AST_FRAME_DTMF && frame->subclass.integer == 'q') ? 1 : 0; /* q indicates frequency hit */
-				ast_frfree(frame);
+
+				if (f_dup != frame) {
+					ast_frfree(frame);
+				}
+
+				ast_frfree(f_dup);
 #else
 				i = tone_detect(&myrpt->burst_tone_state, f->data.ptr, f->samples);
 #endif
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -2413,7 +2413,10 @@ static struct ast_frame *usbradio_read(struct ast_channel *c)
 				o->toneflag = 0;
 			} else {
 				if (o->toneflag) {
-					ast_frfree(f1);
+					if (f1 != f) {
+						ast_frfree(f1);
+					}
+
 					f1 = NULL;
 				} else {
 					o->tonetime = ast_radio_tvnow();
@@ -2421,7 +2424,15 @@ static struct ast_frame *usbradio_read(struct ast_channel *c)
 				}
 			}
 			if (f1) {
-				return f1;
+				struct ast_frame *f2;
+
+				f2 = ast_frisolate(f1);
+
+				if (f1 != f) {
+					ast_frfree(f1);
+				}
+
+				return f2;
 			}
 		}
 	}
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -2413,10 +2413,7 @@ static struct ast_frame *usbradio_read(struct ast_channel *c)
 				o->toneflag = 0;
 			} else {
 				if (o->toneflag) {
-					if (f1 != f) {
-						ast_frfree(f1);
-					}
-
+					ast_frfree(f1);
 					f1 = NULL;
 				} else {
 					o->tonetime = ast_radio_tvnow();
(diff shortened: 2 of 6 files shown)


Allan-N:
>These returned frames are either a pointer to the frame passed in, or an internal frame. We should not free internal frames and only assume ownership of the frame passed in.

I question this statement.  To me, it looks like `ast_dsp_process()` takes the frame passed in and returns a frame after it has finished processing.  The returned frame may be the same one that was passed in.  But, if a new/modified frame is returned then the one that was passed in will be free'd.

So, with this code :
```
				frame = ast_dsp_process(myrpt->rxchannel, myrpt->dsp, ast_frdup(f));
				i = (frame->frametype == AST_FRAME_DTMF && frame->subclass.integer == 'q') ? 1 : 0; /* q indicates frequency hit */
				ast_frfree(frame);
```
we dup the frame `f`, process it, and then free the returned frame.

With the update code :
```
				struct ast_frame *f_dup = ast_frdup(f);
				/* leave f alone */
				frame = ast_dsp_process(myrpt->rxchannel, myrpt->dsp, f_dup);
				i = (frame->frametype == AST_FRAME_DTMF && frame->subclass.integer == 'q') ? 1 : 0; /* q indicates frequency hit */
				ast_frfree(f_dup);
```
Here, we dup the frame, process it, and then free the dup'd frame that may have been free'd during processing ... and we leak the returned frame.

mkmer:
Should we not worry about these?

https://github.com/asterisk/asterisk/blob/d9b8066ec087535034afc0df7f0c1726c7f0c73d/main/dsp.c#L1746

https://github.com/asterisk/asterisk/blob/d9b8066ec087535034afc0df7f0c1726c7f0c73d/main/dsp.c#L1755

https://github.com/asterisk/asterisk/blob/d9b8066ec087535034afc0df7f0c1726c7f0c73d/main/dsp.c#L1892

In most cases, we still need to free the origianal frame as well.  If we only free the returned frame and it's not the same as the frame passed in, we leak the input frame....

Allan-N:
>Should we not worry about these?
...
In most cases, we still need to free the origianal frame as well. If we only free the returned frame and it's not the same as the frame passed in, we leak the input frame....

Remember, we're passing in a ast_frdup()'d frame to ast_dsp_process() so we should be good.

apps/app_rpt.c
```diff
@@ -3846,11 +3846,18 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
 			} else {
 #ifdef NATIVE_DSP
 				struct ast_frame *frame = NULL;
-
+				struct ast_frame *f_dup = ast_frdup(f);
```
Allan-N: Blank line between variables and code

apps/app_rpt.c
```diff
@@ -3846,11 +3846,18 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
...
-				ast_frfree(frame);
+				if (f_dup) {
+					frame = ast_dsp_process(myrpt->rxchannel, myrpt->dsp, f_dup);
+					i = (frame->frametype == AST_FRAME_DTMF && frame->subclass.integer == 'q') ? 1 : 0; /* q indicates frequency hit */
+
+					if (f_dup != frame) {
+						ast_frfree(frame);
+					}
+
+					ast_frfree(f_dup);
```
Allan-N: You've got too many calls to ast_frfree().  ast_dsp_process will either (a) return the passed in frame or (b) free the passed in frame and return a new one.

Allan-N (changes requested):

apps/app_rpt.c
mkmer: ~~It doesn't free the frame from what I can tell... maybe I'm missing the free in the DSP code?~~
I see it now...

Allan-N: Using one of the links you questioned, here's a link to where the passed in frame is free'd

https://github.com/asterisk/asterisk/blob/d9b8066ec087535034afc0df7f0c1726c7f0c73d/main/dsp.c#L1745

channels/chan_voter.c
```diff
@@ -2843,7 +2843,9 @@ static int voter_mix_and_send(struct voter_pvt *p, struct voter_client *maxclien
 				ast_queue_frame(p->owner, f2);
 				gettimeofday(&p->lastrxtime, NULL);
 			}
-			ast_frfree(f2);
+			if (f2 != &fr) {
```
mkmer: I think we have a technical issue with stack based frames being freed inside the `ast_dsp_process()` function.  When this condition is true, `ast_dsp_process` has already tried to free `fr`. 
Since `f2` is most often pointing the `fr`, we shouldn't be freeing it (right?)

channels/chan_usrp.c
```diff
@@ -622,7 +622,10 @@ static int usrp_xwrite(struct ast_channel *ast, struct ast_frame *frame)
 					}
 					ast_queue_frame(ast, f);
 				}
-				ast_frfree(f);
+
```
mkmer, line 626: Same stack based issue here....

channels/chan_usbradio.c
```diff
@@ -2413,15 +2413,26 @@ static struct ast_frame *usbradio_read(struct ast_channel *c)
...
 					o->toneflag = 1;
 				}
 			}
 			if (f1) {
-				return f1;
+				struct ast_frame *f2;
+
+				f2 = ast_frisolate(f1);
+
+				if (f1 != f) {
```
mkmer: Same stack issue here.  I think we NEED to frisolate() before we return. The original code was returning a pointer (`f1`) to the local stack...

Allan-N: First off, I would not say that `f1` is from the local stack.  While it is possible that the frame returned by ast_dsp_process() could be the same as the `struct ast_frame *f = &o->read_f` from the top of `usbradio_read()` that would not be on the stack.

In this case, I don't think we're dealing with the pvt data case here because `f1->frametype` was changed to a DTMF frame ... and that will be a new frame that can safely be returned.

mkmer: My mistake on the source there - mixing things up I guess.  
In any case, `o->read_f` is not something we should be freeing as it's part of the private structure (right?) Yet `ast_dsp_process()` "could" try to free it.

channels/chan_voter.c
mkmer: This one is a stack pointer (unless I'm full losing my mind).
https://github.com/AllStarLink/app_rpt/blob/a5c7a7670391c5c8671ec1d12b1bc18e6df18a79/channels/chan_voter.c#L2674

channels/chan_usbradio.c
Allan-N: A call to `ast_frfree()` only free's the malloc'd bits.  In the case of the `o-read_f` frame, that's effectively static data so free'ing would be a noop (just like free'ing `ast_null_frame`)

channels/chan_voter.c
Allan-N: Yes, `fr` is a frame on the stack ... but `f2 = ast_translate(p->toast1, &fr, 0);` would be a newly allocated frame.

mkmer: https://github.com/AllStarLink/app_rpt/blob/a5c7a7670391c5c8671ec1d12b1bc18e6df18a79/channels/chan_voter.c#L2833
in the dsp process - this is the one that I first found and did a bit too much assuming that the others were the same :(

apps/app_rpt.c
```diff
@@ -3846,11 +3846,15 @@ static inline int rxchannel_read(struct rpt *myrpt, const int lasttx)
...
 				/* leave f alone */
-				frame = ast_dsp_process(myrpt->rxchannel, myrpt->dsp, ast_frdup(f));
-				i = (frame->frametype == AST_FRAME_DTMF && frame->subclass.integer == 'q') ? 1 : 0; /* q indicates frequency hit */
-				ast_frfree(frame);
+				f_dup = ast_frdup(f);
+				if (f_dup) {
+					frame = ast_dsp_process(myrpt->rxchannel, myrpt->dsp, f_dup);
+					i = (frame->frametype == AST_FRAME_DTMF && frame->subclass.integer == 'q') ? 1 : 0; /* q indicates frequency hit */
+					ast_frfree(frame);
+				}
```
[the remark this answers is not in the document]
mkmer: No need to worry about `frame`, it's only NULL if the passed in pointer is NULL.

channels/chan_usrp.c
```diff
@@ -610,19 +610,22 @@ static int usrp_xwrite(struct ast_channel *ast, struct ast_frame *frame)
...
+					f = ast_dsp_process(ast, pvt->dsp, f1);
+					if ((f->frametype == AST_FRAME_DTMF_END) || (f->frametype == AST_FRAME_DTMF_BEGIN)) {
+						if ((f->subclass.integer == 'm') || (f->subclass.integer == 'u')) {
+							f->frametype = AST_FRAME_NULL;
+							f->subclass.integer = 0;
+						}
+						if (f->frametype == AST_FRAME_DTMF_END) {
+							ast_log(LOG_NOTICE, "Channel %s: Got DTMF char %c\n", ast_channel_name(ast), f->subclass.integer);
+						}
 						ast_queue_frame(ast, f);
```
Allan-N, line 624: Here's a behavior question.  Here, we are queuing the DTMF frame.  But, just a bit earlier (l.609) we have already queued the corresponding audio frame.  It seems like this is a break from our pattern of sending the audio frame or the DTMF begin/end but not both.

channels/chan_voter.c
```diff
@@ -2821,7 +2821,9 @@ static int voter_mix_and_send(struct voter_pvt *p, struct voter_client *maxclien
 			ast_frfree(f1);
 			return 0;
 		}
+
 		if (p->rxkey && p->dsp && p->usedtmf) {
+			struct ast_frame *f3;
```
Allan-N: Can we have a blank line between the variables and code?

channels/chan_voter.c
```diff
@@ -2843,8 +2853,10 @@ static int voter_mix_and_send(struct voter_pvt *p, struct voter_client *maxclien
 				ast_queue_frame(p->owner, f2);
 				gettimeofday(&p->lastrxtime, NULL);
 			}
+
 			ast_frfree(f2);
 		}
+
 		memset(silbuf, 0, sizeof(silbuf));
```
Allan-N, line 2825: Looks like the frame we are creating here (l.2848 ... 2855) is the same as the frame created earlier (l.2825 ... 2832).  Seems like we could create the frame one time and dup when/if needed.

As an aside, I'm also thinking that we can do a better job of naming the variables (e.g. f3 -> f_dtmf_check).

channels/chan_usrp.c
Allan-N: Note: I am OK with leaving the behavior AS-IS.  Just wanted to call it out.

mkmer: Maybe an issue so we remember to see if this is desired?

channels/chan_voter.c
mkmer: I think something deeper is "wrong" here -> why would we DSP process a silence frame at all???? I can be fairly sure there's no DTMF in it :)

channels/chan_usrp.c
Allan-N: or we wait until someone notices/cares...

Allan-N (approved):

Issue #1036, opened 2026-04-24
chan_echolink and chan_tlb connections seem to have channel audio directions reversed, compared to Allstar, when using res_muteaudio settings


wd5m:
While testing Allstar web interface enhancements to enable asterisk res_muteaudio features, the audio direction (in or out) used in AMI MuteAudio settings are reversed for echolink and tlb connections, as compared to Allstar connections. Allstar connection incoming audio can be muted/unmuted using the AMI MuteAudio action, and setting the direction to "in" with the appropriate state to "on" or "off". On echolink and tlb connections, to mute/unmute incoming audio the direction must use "out", instead of "in".  Same reversed operation with the "out" MuteAudio direction.  This seems to be backwards, or at least inconsistent with Allstar connections.

David, WD5M

> ********** AllStarLink [ASL] Version Info **********
> 
> OS            : Debian GNU/Linux 12 (bookworm)
> OS Kernel     : 6.1.0-44-amd64
> 
> Asterisk      : 22.8.2+asl3-3.8.3-1.deb12
> ASL [app_rpt] : 3.8.3
> 
> Installed ASL packages :
> 
>   Package                         Version
>   ==============================  ==============================
>   allmon3                         1.8.1-1.deb12
>   asl-apt-repos                   2.0-1.deb12
>   asl3                            3.17-1.deb12
>   asl3-asterisk                   2:22.8.2+asl3-3.8.3-1.deb12
>   asl3-asterisk-config            2:22.8.2+asl3-3.8.3-1.deb12
>   asl3-asterisk-modules           2:22.8.2+asl3-3.8.3-1.deb12
>   asl3-menu                       1.18-1.deb12
>   asl3-tts                        1.0.1-1.deb12
>   asl3-update-nodelist            2.0.0-1.deb12
>   dahdi                           1:3.1.0-2
>   dahdi-linux                     1:3.4.0-11.asl.deb12
>

mkmer:
This may be related to the architecture found in the channel drivers. 
Both drivers (EL and TLB) queue frames that arrive from the network, then when Asterisk writes a frame, the queue has 1 frame removed and "dispatched" to the Asterisk core. (Aka Writes trigger reads). If one mutes the outbound, I suspect this inhibits the write function which then actually prevents any inbound frames from being sent to the Asterisk core.
Does outbound mute on EL actually mute both ways at the same time? 
Does inbound do "anything at all?"

wd5m:
[...]

PR #1042, opened 2026-04-28
chan_echolink: Remove intermediate queue and general refactor for multi thread

### LINKED ISSUE #1039: Audio from ASL to Echolink Garbled after latest release
Since the latest update, audio being passed to the echolink network is garbled, however anyone who is connected on echolink sounds fine on the ASL network.

mkmer:
Use the asterisk audio queue for queuing frames.
Move timer logic out of `el_xwrite()` into `el_reader()` thread addressing mute "problem".
Addresses Echolink part of issue #1036 

## Summary by CodeRabbit

* **Refactor**
  * Simplified reader-thread processing and timing to use monotonic elapsed time, added per-connection audio buffering, and streamlined lifecycle/teardown handling for clearer resource ownership.

* **Bug Fixes**
  * More reliable key/unkey timing, aggregated pending unkey controls to avoid deadlocks, improved heartbeat/disconnect handling, and clearer node disconnect semantics to reduce missed disconnects and improve session stability.

### DIFF
--- a/channels/chan_echolink.c
+++ b/channels/chan_echolink.c
@@ -1364,7 +1364,6 @@ static int el_call(struct ast_channel *ast, const char *dest, int timeout)
 	do_new_call(instp, p, "OUTBOUND", "OUTBOUND");
 	process_cmd(buf, sizeof(buf), "127.0.0.1", instp);
 	ast_mutex_unlock(&instp->lock);
-	p->owner = ast;
 	ast_setstate(ast, AST_STATE_RINGING);
 
 	return 0;
@@ -2427,6 +2426,7 @@ static struct ast_channel *el_request(const char *type, struct ast_format_cap *c
 			el_destroy(p);
 		}
 	}
+	p->owner = tmp;
 	return tmp;
 }
 
@@ -3300,15 +3300,18 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *pvt, const char
 				ast_mutex_unlock(&el_db_lock);
 				return -1;
 			}
-			pvt->owner = chan;
+
 			ast_queue_frame(chan, &fr);
 			el_node_key->rx_ctrl_packets++;
 			ast_mutex_lock(&instp->lock);
 			time(&now);
+
 			if (instp->starttime < (now - EL_APRS_START_DELAY)) {
 				instp->aprstime = now;
 			}
+
 			ast_mutex_unlock(&instp->lock);
+
 		} else {
 			el_node_key->pvt = pvt;
 			ast_copy_string(el_node_key->pvt->ip, instp->el_node_test.ip, EL_IP_SIZE);
@@ -3317,9 +3320,11 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *pvt, const char
 			ast_mutex_lock(&instp->lock);
 			strcpy(instp->lastcall, mynode->callsign);
 			time(&now);
+
 			if (instp->starttime < (now - EL_APRS_START_DELAY)) {
 				instp->aprstime = now;
 			}
+
 			ast_mutex_unlock(&instp->lock);
 		}
 
--- a/channels/chan_echolink.c
+++ b/channels/chan_echolink.c
@@ -458,6 +458,7 @@ struct el_instance {
 	struct timeval current_talker_start_time;
 	struct timeval current_talker_last_time;
 	pthread_t el_reader_thread;
+	pthread_t el_register_thread;
 };
 
 /*!
@@ -521,7 +522,6 @@ static void *el_db_nodenum = NULL;
 static void *el_db_ipaddr = NULL;
 
 /* Echolink registration thread */
-static pthread_t el_register_thread = 0;
 static pthread_t el_directory_thread = 0;
 static int run_forever = 1;
 static int el_sleeptime = 0;
@@ -1414,6 +1414,7 @@ static void el_destroy(void *obj)
 	if (p->u) {
 		ast_module_user_remove(p->u);
 	}
+	ast_mutex_destroy(&p->lock);
 }
 
 /*!
@@ -1447,6 +1448,7 @@ static struct el_pvt *el_alloc(const char *data)
 		p->keepalive = KEEPALIVE_TIME;
 		p->instp = instances[n];
 		p->dsp = ast_dsp_new();
+		ast_mutex_init(&p->lock);
 
 		if (!p->dsp) {
 			ast_log(LOG_ERROR, "Cannot get DSP!\n");
@@ -1466,9 +1468,8 @@ static struct el_pvt *el_alloc(const char *data)
 			ao2_cleanup(p);
 			return NULL;
 		}
-
-		ast_mutex_init(&p->lock);
 	}
+
 	return p;
 }
 
@@ -1519,7 +1520,6 @@ static int el_hangup(struct ast_channel *chan)
 	}
 
 	ast_channel_tech_pvt_set(chan, NULL);
-	ast_mutex_destroy(&p->lock);
 	ao2_ref(p, -1);
 	ast_setstate(chan, AST_STATE_DOWN);
 
@@ -2000,8 +2000,9 @@ static void process_unkey_timers(const void *nodep, const VISIT which, void *clo
 		if (p->rxkey <= 0) {
 			/* The timer has expired, queue up an unkey for the channel */
 			struct ast_channel *chan;
-
+			ast_mutex_lock(&p->lock);
 			chan = p->owner ? ast_channel_ref(p->owner) : NULL;
+			ast_mutex_unlock(&p->lock);
 			if (chan) {
 				struct pending_ctrl item = {
 					.chan = chan,
@@ -2343,7 +2344,7 @@ static void process_cmd(char *buf, int buf_len, const char *fromip, struct el_in
 			pack_length = rtcp_make_bye(pack, sizeof(pack), "bye");
 			n = 20;
 			ast_copy_string(key.ip, arg1, sizeof(key.ip));
-			if (find_delete(&key, NULL)) {
+			if (find_delete(&key, instp)) {
 				for (i = 0; i < n; i++) {
 					sendto(instp->ctrl_sock, pack, pack_length, 0, (struct sockaddr *) &sin, sizeof(sin));
 				}
@@ -3506,7 +3507,7 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *p, const char *
 
 	ast_mutex_lock(&el_nodelist_lock);
 
-	if (tsearch(el_node_key, &el_node_list, compare_ip)) {
+	if (tfind(el_node_key, &el_node_list, compare_ip)) {
 		ast_debug(1, "New Call - Callsign %s, IP Address %s, Node %i, Name %s.\n", el_node_key->call, el_node_key->ip,
 			el_node_key->nodenum, el_node_key->name);
 
@@ -3571,7 +3572,7 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *p, const char *
 
 		ast_mutex_unlock(&el_nodelist_lock);
 		ast_mutex_unlock(&el_db_lock);
-
+		tsearch(el_node_key, &el_node_list, compare_ip); /* Add the node to the list  after everything is "happy" */
 		return 0;
 	}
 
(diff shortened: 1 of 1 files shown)


channels/chan_echolink.c
Allan-N, line 4066: Do we need to call `ast_mutex_unlock(&el_nodelist_lock);` before the `continue;` ?

Allan-N, line 4084: Do we need to call `ast_mutex_unlock(&el_nodelist_lock);` before the `continue;` ?

mkmer: We do, I had the `el_nodelist_lock` all messed up.

mkmer: We do, I had the `el_nodelist_lock` all messed up. 
I also reduced the scope of the lock -> it was not needed beyond  this `continue;`

channels/chan_echolink.c
```diff
@@ -3376,47 +3482,92 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *pvt, const char
...
 			ast_copy_string(el_node_key->pvt->ip, instp->el_node_test.ip, EL_IP_SIZE);
 			chan = el_new(el_node_key->pvt, AST_STATE_RINGING, el_node_key->nodenum, NULL, NULL);
+
 			if (!chan) {
-				el_destroy(el_node_key->pvt);
+				ao2_cleanup(el_node_key->pvt);
 				ast_free(el_node_key);
 				ast_mutex_unlock(&el_db_lock);
+				ast_mutex_unlock(&el_nodelist_lock);
 				return -1;
```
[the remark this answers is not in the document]
mkmer: ao2_cleanup(pvt); is the same as ao2_ref(pvt, -1); so this is matched

channels/chan_echolink.c
```diff
@@ -1892,6 +1897,113 @@ static void lookup_node_nodenum(const void *nodep, const VISIT which, void *clos
...
+				if (AST_VECTOR_APPEND(cl->pend, item)) {
+					ast_log(LOG_ERROR, "Unable to append pending control item.\n");
+					ast_channel_unref(chan);
+				}
+			}
+
+			pvt->rxkey = 0;
+		}
+
+		ao2_cleanup(pvt);
```
Allan-N: To match the earlier `ao2_ref(pvt, +1)` it seems like this should be `ao2_ref(pvt, -1)`.  Yes, it looks like the two are equivalent but using ao2_ref() before and after will have the code looking more balanced.

channels/chan_echolink.c
```diff
@@ -3374,49 +3482,99 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *pvt, const char
 			pvt = el_alloc(instp->name);
 			if (!pvt) {
 				ast_log(LOG_ERROR, "Cannot alloc el channel %s.\n", instp->name);
+				ao2_cleanup(el_node_key->pvt);
+				el_node_key->pvt = NULL;
 				ast_free(el_node_key);
 				ast_mutex_unlock(&el_db_lock);
```
Allan-N: I'd like to see these two calls flipped so that we unlock in the reverse order of how the locks were acquired (lockA, lockB, unlockB, unlockA).

channels/chan_echolink.c
```diff
@@ -3374,49 +3482,99 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *pvt, const char
...
 			el_node_key->pvt = pvt;
 			ast_copy_string(el_node_key->pvt->ip, instp->el_node_test.ip, EL_IP_SIZE);
 			chan = el_new(el_node_key->pvt, AST_STATE_RINGING, el_node_key->nodenum, NULL, NULL);
+
 			if (!chan) {
-				el_destroy(el_node_key->pvt);
+				ao2_cleanup(el_node_key->pvt);
+				el_node_key->pvt = NULL;
 				ast_free(el_node_key);
 				ast_mutex_unlock(&el_db_lock);
```
Allan-N: I'd like to see these two calls flipped so that we unlock in the reverse order of how the locks were acquired (lockA, lockB, unlockB, unlockA).

channels/chan_echolink.c
```diff
@@ -3374,49 +3482,99 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *pvt, const char
...
 			}
+
 			ast_mutex_unlock(&instp->lock);
 		}
-	} else {
-		ast_log(LOG_ERROR, "Failed to add new call, Callsign %s, IP Address %s, Name %s.\n", el_node_key->call, el_node_key->ip,
-			el_node_key->name);
-		ast_free(el_node_key);
+
 		ast_mutex_unlock(&el_db_lock);
```
Allan-N: I'd like to see these two calls flipped so that we unlock in the reverse order of how the locks were acquired (lockA, lockB, unlockB, unlockA).

channels/chan_echolink.c
```diff
@@ -3374,49 +3482,99 @@ static int do_new_call(struct el_instance *instp, struct el_pvt *pvt, const char
...
+
 			if (instp->starttime < (now - EL_APRS_START_DELAY)) {
 				instp->aprstime = now;
 			}
+
 			ast_mutex_unlock(&instp->lock);
+
 		} else {
+			ao2_ref(pvt, +1);
 			el_node_key->pvt = pvt;
```
Allan-N: Do we need to `ao2_cleanup(el_node_key->pvt)` before we assign the new `pvt` ?

channels/chan_echolink.c
```diff
@@ -3767,6 +3989,9 @@ static void *el_reader(void *data)
 									send_text_one(node, "You are doubling.");
 								}
 								node->isdoubling = 1;
+								ast_mutex_unlock(&el_nodelist_lock);
+								ast_channel_unref(ast);
```
Allan-N: From above, `ast` might be NULL

channels/chan_echolink.c
```diff
@@ -3778,34 +4003,91 @@ static void *el_reader(void *data)
...
 							continue;
 						}
+
+						ast_mutex_unlock(&el_nodelist_lock);
+
 						/* queue the gsm packets */
 						if (recvlen == sizeof(struct gsmVoice_t)) {
 							if (gsmPacket->version == 3 && gsmPacket->payt == 3) {
 								/* break them up for Asterisk */
+								char inbuf[GSM_FRAME_SIZE + AST_FRIENDLY_OFFSET];
```
Allan-N, line 4093: Blank line between variables and code.

channels/chan_echolink.c
mkmer: Cleanup is NULL tolerant, but mostly not required as you say

channels/chan_echolink.c
mkmer: I think so.  On that note, I don't think we needed to +1 as we already have a ref when creating the object.

channels/chan_echolink.c
```diff
@@ -1455,8 +1456,10 @@ static struct el_pvt *el_alloc(const char *data)
 		pvt->xpath = ast_translator_build_path(ast_format_slin, ast_format_gsm);
 		if (!pvt->xpath) {
 			ast_log(LOG_ERROR, "Cannot get translator!\n");
+			ao2_cleanup(pvt);
```
Allan-N: If we are giving up, should we also be doing something with `pvt->dsp`?

channels/chan_echolink.c
```diff
@@ -487,12 +479,13 @@ struct el_pvt {
...
 	int keepalive;
 	int txindex;
-	struct el_rxqast rxqast;
 	struct ast_dsp *dsp;
 	struct ast_module_user *u;
 	struct ast_trans_pvt *xpath;
 	unsigned int nodenum;
 	struct ast_str *linkstr;
+	struct gsmVoice_t audio;
+	struct el_node el_node_test;
```
Allan-N: To avoid some confusion between `instp->el_node_test` and `pvt->el_node_test`, can we rename one of the variables?  Also, in each case, what are we testing?  and might that answer help guide the variable names?

channels/chan_echolink.c
```diff
@@ -2051,24 +2171,41 @@ static void free_node(void *nodep) {}
...
+				instp->current_talker_start_time = (struct timeval) { 0 };
+				instp->current_talker_last_time = (struct timeval) { 0 };
+			}
+		}
+
 		ast_debug(3, "Removing from current node list Callsign %s, IP Address %s.\n", node->call, node->ip);
 		found = 1;
 		node->pvt->hangup = 1;
 		tdelete(node, &el_node_list, compare_ip);
+		node->pvt = NULL; /* We've dereferenced the node pvt, by setting it to null.  Hangup will clean up the reference */
```
Allan-N: This comment doesn't make any sense.  The following call to ast_free() releases the memory which nobody should be referencing.  Honestly, this entire statement can go because the memory you are zero'ing is about to be free'd.

channels/chan_echolink.c
```diff
@@ -2168,42 +2309,44 @@ static void process_cmd(char *buf, int buf_len, const char *fromip, struct el_in
...
 		memset(&sin, 0, sizeof(sin));
 		sin.sin_family = AF_INET;
 		sin.sin_port = htons(instp->ctrl_port);
 		sin.sin_addr.s_addr = inet_addr(arg1);
 
-		if (strcmp(cmd, "o.dconip") == 0) {
+		if (conn == 0) {
+			n = 1;
+			pack_length = rtcp_make_sdes(pack, sizeof(pack), instp->mycall, instp->myname, instp->astnode);
+			for (i = 0; i < n; i++) {
```
Allan-N: Knowing that `n = 1`, do we need the `for () {}` loop?

mkmer: Probably not, just copying code blindly again .

channels/chan_echolink.c
mkmer: One renamed -> nothing sacred about the name I picked.

channels/chan_echolink.c
mkmer: Reworked the reference counters -> now +1 when adding a pointer to the `el_node_key` and -1 when we set it to NULL.  It seems like this is the "right" way to count references.

channels/chan_echolink.c
mkmer: On further analysis, we really don't need either of these as globalish variables.  The `twalk()` really should be `twalk_r()` and a local "temporary" variable used to pass the lookup values in.
Refactored.

channels/chan_echolink.c
```diff
@@ -1359,67 +1363,68 @@ static int el_call(struct ast_channel *ast, const char *dest, int timeout)
...
-	}
-	if (pvt->linkstr) {
-		ast_free(pvt->linkstr);
+
+	if (p->linkstr) {
+		ast_free(p->linkstr);
 	}
-	pvt->linkstr = NULL;
+
+	ast_mutex_lock(&p->lock);
```
Allan-N: Why the lock/unlock here?

channels/chan_echolink.c
```diff
@@ -1892,6 +1912,115 @@ static void lookup_node_nodenum(const void *nodep, const VISIT which, void *clos
...
+		if (!p || !p->rxkey) {
+			return;
+		}
+
+		ao2_ref(p, +1);
+		update_timer(&p->rxkey, cl->elap, 0);
+
+		if (p->rxkey <= 0) {
+			/* The timer has expired, queue up an unkey for the channel */
+			struct ast_channel *chan;
```
Allan-N, line 2000: Add a blank line between the variables and the code statements.

channels/chan_echolink.c
```diff
@@ -1959,27 +2088,28 @@ static void send_info(const void *nodep, const VISIT which, const int depth)
...
 	struct el_instance *instp = node->instp;
 	struct sockaddr_in sin;
+	struct el_node *node_lookup = closure;
 	unsigned char sdes_packet[256];
 	int sdes_length;
 
 	if ((which == leaf) || (which == postorder)) {
-		if (node->countdown >= 0) {
-			node->countdown--;
+		if (node->heartbeat_countdown >= 0) {
```
Allan-N, line 2102: Is `heartbeat_countdown` a simple counter?  or a time-based value?  If the latter, are we counting down correctly?

channels/chan_echolink.c
mkmer: True -> if we've removed all references, this should be "safe" to do anything.
Also, really no need to clear `p->owner` since it's released elsewhere.

channels/chan_echolink.c
mkmer: I believe (assuming I've decoded this right) - it's a counter of heartbeats -> if we miss (currently) 15 heartbeats, we set the node to "dead" and disconnect it.
It is a bit weird how they are setting this up...
It sends heartbeats based on the `KEEPALIVE_TIME = 10000` aka 10 seconds.
It configures the `heartbeat_countdown` to `instp->rtcptimeout`
https://github.com/AllStarLink/app_rpt/blob/57e28c48ae3d33b295f817e99ba5c014651e429f/channels/chan_echolink.c#L3745

(sounds like a timer by the name but it's not)

https://github.com/AllStarLink/app_rpt/blob/57e28c48ae3d33b295f817e99ba5c014651e429f/channels/chan_echolink.c#L4140

Finally the setup:
https://github.com/AllStarLink/app_rpt/blob/57e28c48ae3d33b295f817e99ba5c014651e429f/channels/chan_echolink.c#L3885-L3890

Allan-N (approved):

PR #1051, opened 2026-05-19
chan_simpleusb, chan_usbradio: nonblocking pttkick pipe

mkmer:
## Summary by CodeRabbit

* **Bug Fixes**
  * Made internal signaling non-blocking to prevent threads from stalling and improve reliability.
  * Reduced spurious error messages by treating transient non-blocking I/O conditions as non-fatal.
  * Fixed event monitoring so signal reads are handled correctly, improving consistent interrupt delivery.

### DIFF
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -1137,7 +1137,7 @@ static void *hidthread(void *arg)
 		o->had_gpios_in = 0;
 
 		memset(&rfds, 0, sizeof(rfds));
-		rfds[0].fd = o->pttkick[1];
+		rfds[0].fd = o->pttkick[0];
 		rfds[0].events = POLLIN;
 
 		ast_radio_time(&o->lasthidtime);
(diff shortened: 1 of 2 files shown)


mkmer:
We "could" block on the write side if something slows down the read side.  Since we are polling, we really should never fail to read in the `hidthread()`.

Allan-N (approved):

Allan-N:
We do have another option here.  Instead of using a pipe to signal that the HID thread needs to wake up we can switch the code to use `ast_cond_timedwait()` and `ast_cond_signal()`.

and, of course, we could always look at this "later" :)

channels/chan_simpleusb.c
Allan-N: Why are we polling on `o->pttkick[1]` but inside the while loop we are reading from `o->pttkick[0]` ?

mkmer: Well.. that's a good question isn't it.  Been that way for a LONG time....
In both usb drivers too....

PR #1057, opened 2026-05-28
app_rpt: TOTP DTMF Authentication

SigmaScott:
# Add TOTP DTMF Authentication for app_rpt

## Summary

Adds per-user, per-node TOTP (RFC 6238) authentication to app_rpt, allowing administrators to gate privileged DTMF commands behind a one-time password verified via any standard
 authenticator app (Google Authenticator, Authy, FreeOTP, etc.).

## Motivation

Previously, every DTMF function in `rpt.conf` was accessible to anyone who could key the radio — there was no concept of user-level access control. This feature introduces an a
uthentication layer so that sensitive control commands (e.g., system key/unkey, restart scripts) can be restricted to authorized operators.

## What It Does

- Defines an **additive, privileged DTMF command set** that becomes available only after a user authenticates via a 4-digit user ID + 6-digit TOTP code keyed over DTMF.
- Sessions are **per-node, single-user, sliding-timeout** — active use refreshes the session; idle sessions expire automatically.
- Configurable **brute-force lockout** (threshold + duration) protects against OTP guessing over the air.
- **Security-conscious on-air feedback**: all failure modes (bad code, lockout, disabled) produce identical error tones — distinguishing details only appear in logs and CLI.
- Two new Asterisk CLI commands: `rpt auth show <node>` and `rpt auth logout <node>`.

## Configuration

- **`rpt.conf`**: New `auth_*` keys on the node stanza (`auth_users`, `auth_timeout`, `auth_lockout_threshold`, `auth_lockout_duration`, `auth_otp_step`, `auth_otp_window`), pl
us function table entries for login/status/logout sub-commands.
- **`rpt_auth.conf`** (new file): Maps 4-digit user IDs to base32 TOTP secrets and privileged function stanzas. Contains shared secrets — file permissions enforced (`0640 root:
asterisk`).

## Files Changed

| File | Change |
|------|--------|
| `apps/app_rpt/rpt_totp.{h,c}` | New — pure TOTP/HMAC-SHA1/base32 implementation. |
| `apps/app_rpt/rpt_auth.{h,c}` | New — per-node auth session state, user table loader, lockout logic. Opaque state hung off `struct rpt`. |
| `apps/app_rpt/rpt_functions.{h,c}` | New `function_auth` DTMF handler (login/status/logout sub-commands). |
| `apps/app_rpt/app_rpt.h` | New config fields on `struct rpt::p` + opaque `struct rpt_auth_state *auth` pointer. |
| `apps/app_rpt/rpt_config.c` | Config key registration (`auth_*`) and reload/free hooks. |
| `apps/app_rpt.c` | `function_table[]` registration + `collect_function_digits()` modified for authed-stanza-first dispatch and sliding timeout touch. |
| `apps/app_rpt/rpt_cli.c` | New `rpt auth show` / `rpt auth logout` CLI commands. |
| `configs/rpt/rpt.conf` | Documented `auth_*` keys + example admin stanza. |
| `configs/samples/rpt_auth.conf.sample` | New — user/secret file template. |

## Design Decisions

- **Opaque auth state** — all auth internals are behind `rpt_auth.c`; no ABI churn for future extensions.
- **Minimal `collect_function_digits` surgery** — authed-stanza walk is O(1) skip when no session is active; hot path impact is negligible for nodes that don't enable auth.
- **One session per node** — simplifies dispatch and avoids per-channel session tracking. Extensible to multiple sessions if needed later.
- **Separate secrets file** — `rpt.conf` is widely shared/version-controlled; secrets stay in a permission-protected file with no network attack surface.

## Testing

- `rpt_totp.c` includes `RPT_TOTP_SELFTEST` validating against RFC 6238 Appendix B canonical SHA-1 test vectors.
- End-to-end smoke test recipe provided in documentation (generate secret → configure → reload → DTMF login via CLI → verify session → privileged command → logout).

## Summary by CodeRabbit

* **New Features**
  * Per-user TOTP DTMF login (4-digit ID + 6-digit OTP), RFC-compliant verification with replay protection, configurable step/window, lockout controls, single active sliding session, session-refresh on successful dispatch, and auth-aware command routing with privileged command sets. OTP digits are never logged.
  * CLI commands to show node auth status and force logout.

* **Documentation**
  * Admin/developer guide, config templates, and a sample users file with setup, permissions guidance, and an end-to-end test recipe.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -331,6 +331,7 @@
 #include "app_rpt/rpt_telemetry.h"
 #include "app_rpt/rpt_link.h"
 #include "app_rpt/rpt_functions.h"
+#include "app_rpt/rpt_auth.h"
 #include "app_rpt/rpt_manager.h"
 #include "app_rpt/rpt_translate.h"
 #include "app_rpt/rpt_xcat.h"
@@ -660,6 +661,7 @@ const struct function_table_tag function_table[] = {
 	{ "meter", function_meter, 1 },
 	{ "userout", function_userout, 1 },
 	{ "cmd", function_cmd, 0 },
+	{ "auth", function_auth, 0 },
 };
 
 int rpt_function_lookup(const char *f)
@@ -1611,6 +1613,7 @@ static enum rpt_function_response collect_function_digits(struct rpt *myrpt, cha
 	int i, rv;
 	char *stringp, *action, *param, *functiondigits;
 	char function_table_name[30] = "";
+	char authed_function_table_name[30] = "";
 	char workstring[200];
 
 	struct ast_variable *vp;
@@ -1637,13 +1640,34 @@ static enum rpt_function_response collect_function_digits(struct rpt *myrpt, cha
 	} else {
 		ast_copy_string(function_table_name, myrpt->p.functions, sizeof(function_table_name));
 	}
-	/* find context for function table in rpt.conf file */
-	vp = ast_variable_browse(myrpt->cfg, function_table_name);
-	while (vp) {
-		if (!strncasecmp(vp->name, digits, strlen(vp->name))) {
-			break;
+
+	/*
+	 * Authed-stanza first-match-wins walk (additive privileged command set).
+	 * Per design: per-node session, additive, authed entries take priority over
+	 * the source stanza on prefix collision (admin's responsibility to avoid).
+	 * No-op when no active session.
+	 */
+	vp = NULL;
+	if (rpt_auth_get_active_stanza(myrpt, authed_function_table_name, sizeof(authed_function_table_name))) {
+		struct ast_variable *avp = ast_variable_browse(myrpt->cfg, authed_function_table_name);
+		while (avp) {
+			if (!strncasecmp(avp->name, digits, strlen(avp->name))) {
+				vp = avp;
+				break;
+			}
+			avp = avp->next;
 		}
-		vp = vp->next;
+	}
+
+	/* Fall through to source stanza if no authed match */
+	if (!vp) {
+		vp = ast_variable_browse(myrpt->cfg, function_table_name);
+		while (vp) {
+			if (!strncasecmp(vp->name, digits, strlen(vp->name))) {
+				break;
+			}
+			vp = vp->next;
+		}
 	}
 	/* if function context not found */
 	if (!vp) {
@@ -1658,6 +1682,14 @@ static enum rpt_function_response collect_function_digits(struct rpt *myrpt, cha
 			n = myrpt->alt_longestfunc;
 		} else if (command_source == SOURCE_DPHONE) {
 			n = myrpt->dphone_longestfunc;
+		}
+
+		/* Authed stanza may extend the "have we collected enough digits?" gate */
+		{
+			int an = rpt_auth_longestfunc(myrpt);
+			if (an > n) {
+				n = an;
+			}
 		}
 
 		if (strlen(digits) >= n) {
@@ -1681,6 +1713,12 @@ static enum rpt_function_response collect_function_digits(struct rpt *myrpt, cha
 	functiondigits = digits + strlen(vp->name);
 	rv = (*function_table[i].function)(myrpt, param, functiondigits, command_source, mylink);
 	ast_debug(7, "rv=%i\n", rv);
+
+	/* Refresh sliding session timeout on any successful function dispatch */
+	if (rv == DC_COMPLETE || rv == DC_COMPLETEQUIET || rv == DC_DOKEY) {
+		rpt_auth_touch(myrpt);
+	}
+
 	return rv;
 }
 
--- a/apps/app_rpt/rpt_auth.c
+++ b/apps/app_rpt/rpt_auth.c
@@ -1,0 +1,486 @@
+
+#include "asterisk.h"
+
+#include <ctype.h>
+#include <stdint.h>
+#include <string.h>
+#include <strings.h>
+#include <time.h>
+
+#include "asterisk/channel.h" /* required for AST_MAX_EXTENSION via app_rpt.h */
+#include "asterisk/config.h"
+#include "asterisk/lock.h"
+#include "asterisk/logger.h"
+#include "asterisk/strings.h"
+#include "asterisk/utils.h"
+
+#include "app_rpt.h"
+#include "rpt_auth.h"
+#include "rpt_lock.h"
+#include "rpt_totp.h"
+
+#define RPT_AUTH_USER_ID_LEN 4
+#define RPT_AUTH_OTP_LEN 6
+#define RPT_AUTH_USERS_SECTION "users"
+
+#define RPT_AUTH_DEFAULT_TIMEOUT 300
+#define RPT_AUTH_DEFAULT_LOCKOUT_THRESHOLD 5
+#define RPT_AUTH_DEFAULT_LOCKOUT_DURATION 60
+#define RPT_AUTH_DEFAULT_OTP_STEP 30
+#define RPT_AUTH_DEFAULT_OTP_WINDOW 1
+
+struct rpt_auth_user {
+	char id[RPT_AUTH_USER_ID_LEN + 1];
+	char *secret;
+	char *command_set;
+	uint64_t last_counter;
+	int fail_count;
+	time_t lockout_until;
+};
+
+struct rpt_auth_state {
+	struct rpt_auth_user *users;
+	int nusers;
+
+	int session_active;
+	char session_user[RPT_AUTH_USER_ID_LEN + 1];
+	char *session_command_set;
+	int session_longestfunc;
+	time_t session_expires;
+};
+
+static int effective_timeout(struct rpt *myrpt)
+{
+	int v = myrpt->p.auth_timeout;
+	if (v <= 0) {
+		v = RPT_AUTH_DEFAULT_TIMEOUT;
+	}
+	if (v < 30) {
+		v = 30;
+	}
+	if (v > 86400) {
+		v = 86400;
+	}
+	return v;
+}
+
+static int effective_lockout_threshold(struct rpt *myrpt)
+{
+	int v = myrpt->p.auth_lockout_threshold;
+	if (v < 0) {
+		return RPT_AUTH_DEFAULT_LOCKOUT_THRESHOLD;
+	}
+	if (v > 1000) {
+		v = 1000;
+	}
+	return v;
+}
+
+static int effective_lockout_duration(struct rpt *myrpt)
+{
+	int v = myrpt->p.auth_lockout_duration;
+	if (v <= 0) {
+		v = RPT_AUTH_DEFAULT_LOCKOUT_DURATION;
+	}
+	if (v > 86400) {
+		v = 86400;
+	}
+	return v;
+}
+
+static int effective_otp_step(struct rpt *myrpt)
+{
+	int v = myrpt->p.auth_otp_step;
+	if (v <= 0) {
+		v = RPT_AUTH_DEFAULT_OTP_STEP;
+	}
+	if (v < 10) {
+		v = 10;
+	}
+	if (v > 120) {
+		v = 120;
+	}
+	return v;
+}
+
+static int effective_otp_window(struct rpt *myrpt)
+{
+	int v = myrpt->p.auth_otp_window;
+	if (v < 0) {
+		v = RPT_AUTH_DEFAULT_OTP_WINDOW;
+	}
+	if (v > 3) {
+		v = 3;
+	}
+	return v;
+}
+
+static void clear_session_locked(struct rpt_auth_state *st)
+{
+	st->session_active = 0;
+	st->session_user[0] = '\0';
+	if (st->session_command_set) {
+		ast_free(st->session_command_set);
+		st->session_command_set = NULL;
+	}
+	st->session_longestfunc = 0;
+	st->session_expires = 0;
+}
+
+static void free_users_locked(struct rpt_auth_state *st)
+{
+	int i;
+	for (i = 0; i < st->nusers; i++) {
+		if (st->users[i].secret) {
+			ast_free(st->users[i].secret);
+			st->users[i].secret = NULL;
+		}
+		if (st->users[i].command_set) {
+			ast_free(st->users[i].command_set);
+			st->users[i].command_set = NULL;
+		}
+	}
+	ast_free(st->users);
+	st->users = NULL;
+	st->nusers = 0;
+}
+
+static struct rpt_auth_user *find_user_locked(struct rpt_auth_state *st, const char *user_id4)
+{
+	int i;
+	for (i = 0; i < st->nusers; i++) {
+		if (!strcmp(st->users[i].id, user_id4)) {
+			return &st->users[i];
+		}
+	}
+	return NULL;
+}
+
+static int parse_user_value(const char *value, char **out_secret, char **out_set)
+{
+	char *tmp, *secret, *set;
+
+	if (ast_strlen_zero(value)) {
+		return -1;
+	}
+	tmp = ast_strdupa(value);
+
+	secret = strsep(&tmp, ",");
+	set = tmp;
+
+	if (!set) {
+		return -1;
+	}
+
+	secret = ast_strip(secret);
+	set = ast_strip(set);
+
+	if (ast_strlen_zero(secret) || ast_strlen_zero(set)) {
+		return -1;
+	}
+
+	*out_secret = ast_strdup(secret);
+	*out_set = ast_strdup(set);
+	if (!*out_secret || !*out_set) {
+		ast_free(*out_secret);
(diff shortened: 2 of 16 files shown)


apps/app_rpt/rpt_totp.c
```diff
@@ -0,0 +1,282 @@
...
+#include "asterisk/utils.h"
+
+#include <openssl/sha.h>
+
+#include "rpt_totp.h"
+
+#define HMAC_BLOCK 64
+#define HMAC_HASH 20
+
+static int hmac_sha1(const uint8_t *key, size_t keylen,
```
Allan-N: How does this function differ from asterisk's `ast_sha1_hash_uint()` ?

apps/app_rpt/rpt_totp.c
```diff
@@ -0,0 +1,282 @@
...
+		SHA1_Update(&sha, opad, HMAC_BLOCK) != 1 ||
+		SHA1_Update(&sha, inner, HMAC_HASH) != 1 ||
+		SHA1_Final(out, &sha) != 1) {
+		return -1;
+	}
+
+	return 0;
+}
+
+int rpt_base32_decode(const char *in, uint8_t *out, size_t outlen)
```
Allan-N, line 59: Might we opt for asterisk's `ast_base64decode()` ?

and, yes, this is base64 vs base32 but I would think the former (base64) is more common/prevalent.

apps/app_rpt/rpt_totp.c
SigmaScott: we started looking at the  built in function but it is my understanding that the upstream function you reference does not support binary data since it is string based.  A RFC 6238-compliant TOTP implementation requires a hashing of binary data that contains nulls.  While OpenSSL is not used with in apt_rpt it is already included and used in the larger Asterisk project.

apps/app_rpt/rpt_totp.c
SigmaScott: TOTP requires the use of base 32 with 5 bits per character.  Base 64 uses 6 bits per character and would break all the existing authenticator apps that are in the market. (ie Google Authenticator, Authy, DUO, etc.)

apps/app_rpt/rpt_totp.c
Allan-N: OK

apps/app_rpt/rpt_totp.c
Allan-N: Agreed.  We most certainly want to be able to use the existing / available authenticator apps.

apps/app_rpt.c
```diff
@@ -1681,6 +1714,12 @@ static enum rpt_function_response collect_function_digits(struct rpt *myrpt, cha
 	functiondigits = digits + strlen(vp->name);
 	rv = (*function_table[i].function)(myrpt, param, functiondigits, command_source, mylink);
 	ast_debug(7, "rv=%i\n", rv);
+
+	/* Refresh sliding session timeout on any successful function dispatch */
+	if (rv == DC_COMPLETE || rv == DC_COMPLETEQUIET || rv == DC_DOKEY) {
+		rpt_auth_touch(myrpt);
+	}
```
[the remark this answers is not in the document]
SigmaScott: This was a design choice given that an authenticated user would also need to use unauthenticated commands during a particular session.  Yes a non authenticated user could keep the authenticated privileges alive but this is nature of shared medium of RF.  This would need to be a training issue to have the authenticated user specifically log out when they were done.

apps/app_rpt/rpt_auth.c
```diff
@@ -0,0 +1,501 @@
...
+		if (now >= st->session_expires) {
+			ast_log(LOG_NOTICE, "rpt_auth: session for user %s expired on node %s\n",
+				st->session_user, myrpt->name);
+			clear_session_locked(st);
+		} else {
+			ret = st->session_command_set;
+		}
+	}
+	rpt_mutex_unlock(&myrpt->lock);
+	return ret;
```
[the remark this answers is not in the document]
SigmaScott: This potential race condition requires two dtmf process simultaneously on a single node that is prevented in the existing app_rpt design.

apps/app_rpt/rpt_config.c
```diff
@@ -675,6 +676,8 @@ void rpt_free_config_vars(struct rpt *myrpt)
 		ast_free(myrpt->p.ldisc_buf);
 		myrpt->p.ldisc_buf = NULL;
 	}
+
+	rpt_auth_free(myrpt);
```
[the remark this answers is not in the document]
SigmaScott: Yes a user would be de-authenticated if the app_rpt was reloaded.  Reloading apt_rpt is an infrequent activity on a stable system and this behavior was design decision to require a user to re-auth if this did happen.

apps/app_rpt/rpt_functions.c
```diff
@@ -2005,3 +2006,76 @@ enum rpt_function_response function_cmd(struct rpt *myrpt, char *param, char *di
...
+{
+	const char *digits = digitbuf ? digitbuf : "";
+	size_t len = strlen(digits);
+	const char *subcmd = (param && *param) ? param : "";
+	size_t i;
+	int rc;
+
+	if (myrpt->remote) {
+		return DC_ERROR;
+	}
```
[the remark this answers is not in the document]
SigmaScott: This was a design decision to allow auth from all command sources for a node

apps/app_rpt/rpt_functions.c
```diff
@@ -2005,3 +2006,76 @@ enum rpt_function_response function_cmd(struct rpt *myrpt, char *param, char *di
...
+
+	ast_debug(1, "auth subcmd=%s digitbuf_len=%zu source=%d\n", subcmd, len, command_source);
+
+	if (!strcasecmp(subcmd, "s")) {
+		rpt_telem_select(myrpt, command_source, mylink);
+		rpt_telemetry(myrpt, COMPLETE, NULL);
+		return DC_COMPLETEQUIET;
```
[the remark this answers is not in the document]
SigmaScott: This was a design decision. The status sub-command is intentionally unauthenticated.

apps/app_rpt/rpt_auth.h
```diff
@@ -0,0 +1,61 @@
...
+
+/*! \brief Free all auth state for a node.  Called at module unload / node teardown. */
+void rpt_auth_free(struct rpt *myrpt);
+
+/*! \brief Copy the active command-set stanza name into buf, or return 0 if no session.
+ *  Returns 1 if an active session exists (buf populated), 0 otherwise.
+ *  Copies under lock so the returned string is caller-owned and safe from
+ *  concurrent rpt_auth_logout / rpt_auth_reload on the CLI thread.
+ *  Internally checks expiry and clears stale sessions. */
+int rpt_auth_active_set(struct rpt *myrpt, char *buf, size_t buflen);
```
Allan-N: I think this function is mis-named.  You are not "setting" the active set, you are returning the name of currently active command set (actually the section name).  Shouldn't this be `rpt_auth_active_get()` ?  or `rpt_auth_get_active_section()` ?  or `rpt_auth_get_section()` ?

apps/app_rpt/rpt_auth.c
```diff
@@ -0,0 +1,502 @@
...
+	char id[RPT_AUTH_USER_ID_LEN + 1];
+	char *secret;
+	char *command_set;
+	uint64_t last_counter;
+	int fail_count;
+	time_t lockout_until;
+};
+
+struct rpt_auth_state {
+	struct rpt_auth_user users[RPT_AUTH_MAX_USERS];
```
Allan-N: We've been trying to remove fixed arrays from `app_rpt` and instead using lists (e.g. AST_LIST, RPT_LIST, ...), queues, etc.

apps/app_rpt/rpt_auth.c
```diff
@@ -0,0 +1,502 @@
...
+	for (i = 0; i < st->nusers; i++) {
+		if (!strcmp(st->users[i].id, user_id4)) {
+			return &st->users[i];
+		}
+	}
+	return NULL;
+}
+
+static int parse_user_value(const char *value, char **out_secret, char **out_set)
+{
```
Allan-N, line 160: Have you looked at strsep() ?  and strdup() ?

apps/app_rpt/rpt_auth.c
```diff
@@ -0,0 +1,502 @@
...
+#include "asterisk/utils.h"
+
+#include "app_rpt.h"
+#include "rpt_auth.h"
+#include "rpt_lock.h"
+#include "rpt_totp.h"
+
+#define RPT_AUTH_USER_ID_LEN 4
+#define RPT_AUTH_OTP_LEN 6
+#define RPT_AUTH_MAX_USERS 64
```
Allan-N: We've been trying to move away from fixed arrays and MAX values.

apps/app_rpt/rpt_auth.h
SigmaScott: you are correct.  I have updated this to rpt_auth_get_active_stanza()

apps/app_rpt/rpt_auth.c
SigmaScott: yes strsep and ast_strip will make this cleaner.

apps/app_rpt/rpt_auth.c
SigmaScott: These were defined and set at a fixed size for a few reasons:
 * The id length having user ids between 0000 and 9999 should be more then adequate for any system operator.  It needed to be a fixed length to know where in the DTMF sequence is the user id and and where is the pass code.
 * The OTP length 6 is a standard length used by most apps.  The TOTP rfc can support 4 and 8 also that give you more or less security from a bang on the door and guess it perspective. It would require additional setup steps in an app to go to 4/8 pass code length.  
 * The 64 was chosen because having more then 64 unique users in a system seems like a low probability.  If an operator needed more then 64 the define could be adjust and the system could be recompiled.

All of these are hard coded because they have a low probability of being more or different.  Having to keep track of the size using if dynamic lists were used adds more code to maintain.  I could see the max users needing to be more if there was a large intertie system and they wanted to use this feature to allow members to have additional privileges while non members would not.

If the group thinks that needs to be accounted for or if they want to move to lists because that is current philosophy of the group that we can move them to lists for sure.

apps/app_rpt/rpt_auth.c
SigmaScott: see other comments on fixed arrays for max values.  The fixed arrays were used because they were defining fixed, known and not likely to change things.  This can of course be changed if the group thinks it would be better to move to the lists

apps/app_rpt/rpt_config.c
```diff
@@ -1010,6 +1013,12 @@ void load_rpt_vars(int n, int init)
 	RPT_CONFIG_VAR(eloutbound, "eloutbound");
 	RPT_CONFIG_VAR_DEFAULT(events, "events", "events");
 	RPT_CONFIG_VAR(timezone, "timezone");
+	RPT_CONFIG_VAR(auth_users, "auth_users");
```
mkmer, line 1016: Note: These (along with the "whole") will need to be added to the docs. With a PR here: https://github.com/AllStarLink/ASL3-Manual

mkmer (commented):
Very well done!

mkmer:
Most of these errors can be "fixed" locally by running the pre-commit hook scripts (and activating them while working).  Instructions are in the readme.

apps/app_rpt/rpt_auth.c
Allan-N: I am not concerned about RPT_AUTH_USER_ID_LEN or RPT_AUTH_OTP_LEN.  The issue is with RPT_AUTH_MAX_USERS where we are allocating unneeded/unused space that for most installs would not be used.  Also, while one can recompile it's not that simple and then you'd have the added burden of keeping in sync with all future updates.

In short, it would be nice to see :
```
struct rpt_auth_state {
        struct rpt_auth_user users[RPT_AUTH_MAX_USERS];
```
changed from a fixed array to a list.

SigmaScott: i was not thinking about too much space.  I will change this

apps/app_rpt/rpt_config.c
SigmaScott: i will get started on that.

mkmer: It's not a huge priority, just wanted to "remember" we need to update the manual :)
I think you have most of it in the `adminauth.md` file...

configs/rpt/rpt.conf
```diff
@@ -124,6 +124,20 @@ scheduler = schedule                ; Scheduler stanza
...
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; TOTP per-user authentication (optional). When enabled, a user can DTMF-
+; authenticate with a 4-digit user-id + 6-digit RFC 6238 TOTP, which unlocks
+; an additional command-set stanza for a sliding timeout. The "auth" function
+; must be added to the active functions stanza (see [functions-main] below).
+;
+;auth_users = /etc/asterisk/rpt_auth.conf  ; Path to user/secret file. Unset = feature disabled.
+;auth_timeout = 300                 ; Sliding session timeout, seconds. Range 30..86400, default 300.
+;auth_lockout_threshold = 5         ; Failed-login attempts before lockout. 0 disables lockout. Range 0..1000.
```
Allan-N: What's the default value?

configs/rpt/rpt.conf
```diff
@@ -502,6 +516,41 @@ parrottime = 1000                   ; Set the amount of time in milliseconds
 
 [functions](functions-main)
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
Allan-N: I'm thinking that this block may be a bit too much documentation for the rpt.conf file.  Seems like the ASL Manual would be a better place for the details.  Should we limit the rpt.conf content to just :

```
; Enable/disable functions requiring authentication
; A1 = auth,a        ; login:  *A1<id4><otp6>  e.g. *A11234849203
; A2 = auth,s        ; status: *A2             (courtesy tone only; details via "rpt auth show <node>")
; A3 = auth,l        ; logout: *A3
```

Also, do these function definitions belong a few lines up (in `[node-main](!)`)?

apps/app_rpt.c
```diff
@@ -1611,6 +1613,7 @@ static enum rpt_function_response collect_function_digits(struct rpt *myrpt, cha
 	int i, rv;
 	char *stringp, *action, *param, *functiondigits;
 	char function_table_name[30] = "";
+	char authed_stanza_buf[30] = "";
```
Allan-N: authed_table_name ?

apps/app_rpt/rpt_functions.c
```diff
@@ -2005,3 +2006,76 @@ enum rpt_function_response function_cmd(struct rpt *myrpt, char *param, char *di
...
+
+	/* Default / "a" — login: need exactly 10 decimal digits */
+	if (len < 10) {
+		return DC_INDETERMINATE;
+	}
+	if (len != 10) {
+		return DC_ERROR;
+	}
+	for (i = 0; i < 10; i++) {
+		if (digits[i] < '0' || digits[i] > '9') {
```
Allan-N: Could this be :
```
		if (!isdigit(digits[i])) {
```

apps/app_rpt/rpt_auth.c
```diff
@@ -0,0 +1,485 @@
...
+}
+
+static int valid_user_id(const char *name)
+{
+	int i;
+	if (strlen(name) != RPT_AUTH_USER_ID_LEN) {
+		return 0;
+	}
+	for (i = 0; i < RPT_AUTH_USER_ID_LEN; i++) {
+		if (name[i] < '0' || name[i] > '9') {
```
Allan-N: Can we use isdigit() ?

configs/rpt/rpt.conf
SigmaScott: 5 is the default.  I will update this.

configs/rpt/rpt.conf
```diff
@@ -500,8 +528,6 @@ parrottime = 1000                   ; Set the amount of time in milliseconds
 ; 85 = cop,6                        ; Remote base telephone key
 
 
-[functions](functions-main)
```
Allan-N, line 503: Please put back the [functions] section :

```
[functions](functions-main)

```

Allan-N (changes requested):

configs/rpt/rpt.conf
SigmaScott: oops... Fixed.

Allan-N (approved):

mkmer (approved):

PR #1061, opened 2026-06-01
app_rpt.c: hangup pchan if call fails

mkmer:
Hangup pchan if call fails. of return, which allows the channel to be hung up if the call fails. This is needed to prevent the channel from being left in a bad state if the call fails.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of link errors and hangups to ensure proper resource cleanup
  * Enhanced link setup failure handling with better channel management

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4342,7 +4342,7 @@ static inline void safe_hangup(struct ast_channel *chan)
 	}
 }
 
-/*! \note myrpt->lock must be held when calling */
+/*! \note myrpt->lock must not be held when calling */
 static inline void hangup_link_chan(struct rpt_link *l)
 {
 	if (l->chan) {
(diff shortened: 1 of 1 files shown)


apps/app_rpt.c
```diff
@@ -4342,7 +4342,7 @@ static inline void safe_hangup(struct ast_channel *chan)
 	}
 }
 
-/*! \note myrpt->lock must be held when calling */
+/*! \note myrpt->lock must not be held when calling */
```
Allan-N, line 4345: I think I understand why one should not be holding the myrpt->lock but how do we ensure that l->chan won't be modified while we're trying to hangup the channel?

mkmer: I believe everything is now in this one thread - did I miss some?

Allan-N (approved):

PR #1063, opened 2026-06-04
app_rpt: on shutdown, protect NULL `linklist` and `macrobuf`

### LINKED ISSUE #1064: rpt restart command crashes asterisk
In ASL 3.9.3 (and most likely any version after ao2 containers were implemented)
CLI command `rpt restart` results in:
``` 
[2026-06-05 20:04:12.703] ERROR[439738] app_rpt/rpt_cli.c: FRACK!, Failed assertion user_data is NULL (0)
[2026-06-05 20:04:12.703] ERROR[439738] : Got 8 backtrace records
# 0: /usr/sbin/asterisk(__ao2_container_clone+0x156) [0x5634a4497806]
# 1: /usr/lib/x86_64-linux-gnu/asterisk/modules/app_rpt.so(+0x29aae) [0x7f6cea429aae]
# 2: /usr/sbin/asterisk(ast_cli_command_full+0x1fd) [0x5634a450349d]
# 3: /usr/sbin/asterisk(ast_cli_command_multiple_full+0xa4) [0x5634a45037d4]
# 4: /usr/sbin/asterisk(+0x70ca9) [0x5634a448fca9]
# 5: /usr/sbin/asterisk(+0x1f6815) [0x5634a4615815]
# 6: /lib/x86_64-linux-gnu/libc.so.6(+0x92b7b) [0x7f6cebc9eb7b]
# 7: /lib/x86_64-linux-gnu/libc.so.6(+0x1107f8) [0x7f6cebd1c7f8]
```
The comand line showed line numbers pointing to https://github.com/AllStarLink/app_rpt/blob/ff107130198dda2b8ebc4fa9e9862d8bbd117629/apps/app_rpt/rpt_cli.c#L451 as one of the sources

mkmer:
Links were not being removed from the list when shutting down.  This causes a LONG delay before asterisk actually force shuts down without completing the cleanup logic.
Use `pthread_tryjoin_np()` to detect when `rpt()` has exited as `pthread_kill` does not detect zombie threads, which is what happens when a thread dies and has not been joined.  
Add null protection to `myrpt->links` protecting races at shutdown.
Add null protection to `myrpt->macrobuf`

Fixes #1064 

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented operations on missing link containers across CLI, manager, telemetry, link, and utility paths to avoid crashes and failed commands.
  * Commands now fail gracefully when a node or system isn't ready (e.g., send, playback, command paths).
  * Suppressed broadcast/hangup actions during module shutdown to avoid shutdown races.

* **Refactor**
  * Improved teardown so links drain cleanly and are cleaned up quietly.
  * Made thread monitoring/restart logic more robust to reliably detect exited threads and skip restarts during shutdown.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -6128,8 +6128,9 @@ static void *rpt_master(void *ignore)
 				thread_hung[i] = rpt_true;
 				ast_log(LOG_WARNING, "RPT thread on %s is hung for %ld seconds.\n", rpt_vars[i].name, current_loop_time);
 			}
-			rv = pthread_kill(rpt_vars[i].rpt_thread, 0); /* Check thread status by sending signal 0 */
-			if (rv) {
+
+			rv = pthread_tryjoin_np(rpt_vars[i].rpt_thread, 0); /* Check thread status by trying to join it */
+			if (rv != EBUSY) {
 				if (rpt_vars[i].deleted == RPT_DELETED_PENDING) {
 					rpt_vars[i].name[0] = 0;
 					if (rpt_vars[i].rpt_thread != AST_PTHREADT_NULL) {
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1755,8 +1755,14 @@ static int distribute_to_all_links(struct rpt *myrpt, struct rpt_link *mylink, c
 {
 	struct rpt_link *l;
 	struct ao2_iterator l_it;
+
 	/* see if this is one in list */
 	rpt_mutex_lock(&myrpt->lock);
+	if (!myrpt->links) {
+		rpt_mutex_unlock(&myrpt->lock);
+		return -1;
+	}
+
 	RPT_LIST_TRAVERSE(myrpt->links, l, l_it) {
 		if (l->name[0] == '0') {
 			continue;
@@ -3081,24 +3087,28 @@ static inline void dump_rpt(struct rpt *myrpt, const int lasttx, const int laste
 	ast_debug(2, "myrpt->p.parrotmode = %d\n", (int) myrpt->p.parrotmode);
 	ast_debug(2, "myrpt->parrotonce = %d\n", (int) myrpt->parrotonce);
 	ast_debug(2, "myrpt->rpt_newkey =%d\n", myrpt->rpt_newkey);
+
 	rpt_mutex_lock(&myrpt->lock);
-	RPT_LIST_TRAVERSE(myrpt->links, zl, l_it) {
-		ast_debug(2, "*** Link Name: %s ***\n", zl->name);
-		ast_debug(2, "        link->lasttx %d\n", zl->lasttx);
-		ast_debug(2, "        link->lastrx %d\n", zl->lastrx);
-		ast_debug(2, "        link->connected %d\n", zl->connected);
-		ast_debug(2, "        link->hasconnected %d\n", zl->hasconnected);
-		ast_debug(2, "        link->outbound %d\n", zl->outbound);
-		ast_debug(2, "        link->disced %d\n", zl->disced);
-		ast_debug(2, "        link->killme %d\n", zl->killme);
-		ast_debug(2, "        link->disctime %d\n", zl->disctime);
-		ast_debug(2, "        link->retrytimer %d\n", zl->retrytimer);
-		ast_debug(2, "        link->retries = %d\n", zl->retries);
-		ast_debug(2, "        link->reconnects = %d\n", zl->reconnects);
-		ast_debug(2, "        link->link_newkey = %d\n", zl->link_newkey);
+	if (myrpt->links) {
+		RPT_LIST_TRAVERSE(myrpt->links, zl, l_it) {
+			ast_debug(2, "*** Link Name: %s ***\n", zl->name);
+			ast_debug(2, "        link->lasttx %d\n", zl->lasttx);
+			ast_debug(2, "        link->lastrx %d\n", zl->lastrx);
+			ast_debug(2, "        link->connected %d\n", zl->connected);
+			ast_debug(2, "        link->hasconnected %d\n", zl->hasconnected);
+			ast_debug(2, "        link->outbound %d\n", zl->outbound);
+			ast_debug(2, "        link->disced %d\n", zl->disced);
+			ast_debug(2, "        link->killme %d\n", zl->killme);
+			ast_debug(2, "        link->disctime %d\n", zl->disctime);
+			ast_debug(2, "        link->retrytimer %d\n", zl->retrytimer);
+			ast_debug(2, "        link->retries = %d\n", zl->retries);
+			ast_debug(2, "        link->reconnects = %d\n", zl->reconnects);
+			ast_debug(2, "        link->link_newkey = %d\n", zl->link_newkey);
+		}
+		ao2_iterator_destroy(&l_it);
 	}
+
 	rpt_mutex_unlock(&myrpt->lock);
-	ao2_iterator_destroy(&l_it);
 	zt = myrpt->tele.next;
 	if (zt != &myrpt->tele) {
 		ast_debug(2, "*** Telemetry Queue ***\n");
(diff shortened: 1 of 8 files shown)


apps/app_rpt.c
```diff
@@ -5786,6 +5776,24 @@ static void *rpt(void *this)
...
+	}
+
+	if (myrpt->xlink == 1) {
+		myrpt->xlink = 2;
+	}
+
+	rpt_mutex_unlock(&myrpt->lock);
+
+	ao2_cleanup(myrpt->links);
+	myrpt->links = NULL;
```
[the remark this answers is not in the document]
mkmer: When we remove the link in hangup helper, myrpt is locked.  In order to complete the `container_count` test on exit it must acquire a lock.  Once the list is empty and unlocked there is no other access to `myrpt->links`.

mkmer:
@coderabbitai review

apps/app_rpt/rpt_manager.c
```diff
@@ -208,9 +208,16 @@ static int rpt_manager_do_xstat(struct mansession *ses, const struct message *m)
 				tel_mode = "3";
 			}
 
+			if (!myrpt->links) {
```
Allan-N: Do we move this block just after l.151 so we're not checking/setting a bunch of variables only to return early?

apps/app_rpt/rpt_manager.c
```diff
@@ -542,6 +549,12 @@ static int rpt_manager_do_stats(struct mansession *s, const struct message *m)
 			/* Traverse the list of connected nodes */
 			reverse_patch_state = "DOWN";
 
+			if (!myrpt->links) {
```
Allan-N: Do we move this block just after l.404 (when we set `myrpt =` ... and grab the lock earlier) so we're not checking/setting a bunch of variables only to return early?

apps/app_rpt/rpt_link.c
```diff
@@ -346,12 +348,20 @@ void send_link_dtmf(struct rpt *myrpt, char c)
...
 			ao2_ref(l, -1);
 			ao2_iterator_destroy(&l_it);
+			rpt_mutex_unlock(&myrpt->lock);
 			return;
 		}
 	}
 	ao2_iterator_destroy(&l_it);
+
 	/* if not, give it to everyone */
+	if (!myrpt->links) {
```
Allan-N: Seems like we can move the block up a bit and avoid both the RPT_LIST_TRAVERSE and the ao2_callback (if there are no links then there's nothing to do)

apps/app_rpt/rpt_cli.c
```diff
@@ -118,6 +118,12 @@ static int rpt_do_stats(int fd, int argc, const char *const *argv)
 
 			/* Traverse the list of connected nodes */
 			reverse_patch_state = "DOWN";
+
+			if (!myrpt->links) {
```
Allan-N: Move up to just we grab the lock (no need in checking/setting any variables if we're just going to return) ?

Allan-N (approved):

apps/app_rpt/rpt_cli.c
mkmer: All of these make sense.  I was thinking it was "clearer" next to the location it "matters" most, and it's only ever false in the VERY small window.

mkmer:
> ♻️ Duplicate comments (1)
> > apps/app_rpt.c (1)> `6149-6153`: _⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_
> > > **NULL thread handles must still reach the restart path.**
> > > Line 6150 turns an `ast_pthread_create()` failure into a permanent outage: the loop exits before the recreate block at Lines 6164-6195, so a repeater with `rpt_thread == AST_PTHREADT_NULL` never gets restarted. Only skip `pthread_tryjoin_np()` when the handle is null; do not skip the rest of the monitor logic.
> > > 
> > > 🛠️ Minimal fix
> > > ```diff
> > > -			if (!rpt_vars[i].rpt_thread || rpt_vars[i].rpt_thread == AST_PTHREADT_NULL) {
> > > -				continue; /* Thread is not running, nothing to do */
> > > -			}
> > > -
> > > -			rv = pthread_tryjoin_np(rpt_vars[i].rpt_thread, 0); /* Check thread status by trying to join it */
> > > -			if (rv == EBUSY) {
> > > -				continue; /* Thread is still running, nothing to do */
> > > -			}
> > > -			if (rv == 0) {
> > > -				rpt_vars[i].rpt_thread = AST_PTHREADT_NULL;
> > > -			} else {
> > > -				ast_log(LOG_WARNING, "Failed to query %s thread: %s\n", rpt_vars[i].name, strerror(rv));
> > > -				continue;
> > > +			if (rpt_vars[i].rpt_thread != AST_PTHREADT_NULL) {
> > > +				rv = pthread_tryjoin_np(rpt_vars[i].rpt_thread, NULL); /* Check thread status by trying to join it */
> > > +				if (rv == EBUSY) {
> > > +					continue; /* Thread is still running, nothing to do */
> > > +				}
> > > +				if (rv == 0) {
> > > +					rpt_vars[i].rpt_thread = AST_PTHREADT_NULL;
> > > +				} else {
> > > +					ast_log(LOG_WARNING, "Failed to query %s thread: %s\n", rpt_vars[i].name, strerror(rv));
> > > +					continue;
> > > +				}
> > >  			}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@apps/app_rpt.c` around lines 6149 - 6153, The current guard `if
> > > (!rpt_vars[i].rpt_thread || rpt_vars[i].rpt_thread == AST_PTHREADT_NULL) {
> > > continue; }` prevents repeaters with a NULL thread handle from reaching the
> > > restart logic; change the flow so only the call to
> > > `pthread_tryjoin_np(rpt_vars[i].rpt_thread, 0)` is skipped when `rpt_thread` is
> > > null/`AST_PTHREADT_NULL`, but do not `continue` out of the loop—allow the
> > > subsequent monitor/recreate block (the restart logic) to run for that
> > > `rpt_vars[i]`. Concretely, replace the single `if`+`continue` with a conditional
> > > that sets `rv` to a value indicating "not joined" (or skips calling
> > > `pthread_tryjoin_np`) when `rpt_thread` is NULL/`AST_PTHREADT_NULL`, otherwise
> > > call `pthread_tryjoin_np`, and keep the existing restart code that checks `rv`
> > > and restarts the repeater.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Duplicate comments:
> In `@apps/app_rpt.c`:
> - Around line 6149-6153: The current guard `if (!rpt_vars[i].rpt_thread ||
> rpt_vars[i].rpt_thread == AST_PTHREADT_NULL) { continue; }` prevents repeaters
> with a NULL thread handle from reaching the restart logic; change the flow so
> only the call to `pthread_tryjoin_np(rpt_vars[i].rpt_thread, 0)` is skipped when
> `rpt_thread` is null/`AST_PTHREADT_NULL`, but do not `continue` out of the
> loop—allow the subsequent monitor/recreate block (the restart logic) to run for
> that `rpt_vars[i]`. Concretely, replace the single `if`+`continue` with a
> conditional that sets `rv` to a value indicating "not joined" (or skips calling
> `pthread_tryjoin_np`) when `rpt_thread` is NULL/`AST_PTHREADT_NULL`, otherwise
> call `pthread_tryjoin_np`, and keep the existing restart code that checks `rv`
> and restarts the repeater.
> ```
> 
> ℹ️ Review info

This proposed solution allows a NULL `rpt_thread` to reach `pthread_tryjoin_np()` which is NOT valid.
IF the very rare event where the inital thread creation fails, we will log a hung thread message and it will require a restart or module unload/load to recover.

PR #1066, opened 2026-06-15
app_rpt: Add keyup delay via first_keyup_min_time and first_keyup_inactivity_time parameters

### LINKED ISSUE #90: Kerchunk filter

### LINKED ISSUE #734: app_rpt: Add Key-up Delay Configurable Option
Many analog controllers have an option to delay the assertion of a valid COS/COR signal until that signal has been present for a configurable amount of time.

This feature is used to reduce "kerchunking" of the repeater due to quick bursts of noise, or users "bumping their PTT" (intentionally, or unintentionally).

This is **not** the same as timers/delays used for squelch tail elimination.

For example, if `keyupdelay` is set to `250`(mS), then COS/COR has to be valid for that minimum time, before the repeater will be keyed. Any squelch openings shorter than that will just be ignored.

mkmer:
Add two new parameters:
keyupdelay_inactivity: time in ms of inactivity before keyupdelay becomes active.
keyupdelay: time in ms to wait for a "valid" keyup only after `keyupdelay_inactivity` has expired

This delay will prevent/remove/mute audio during the `keyupdelay` time.
Fixes #90 

## Summary by CodeRabbit

* **New Features**
  * Added “first keyup” timing controls to better manage when a repeater begins TX after initial RX key activity.
* **Configuration**
  * New `first_keyup_min_time` sets the minimum delay after inactivity before local RX key events trigger TX keying.
  * New `first_keyup_inactivity_time` sets the inactivity duration (seconds) before the extended “first keyup” behavior applies.
* **Improvements**
  * Key/unkey timing is now handled more consistently across multiple RX key-up sources, with improved timer resets during unkey events and thread startup initialization.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3608,7 +3608,7 @@ static inline void rpt_keyed(struct rpt *myrpt, int delay)
 			return;
 		}
 
-		if (delay && myrpt->p.keyupdelay_time > 0 && myrpt->p.keyupdelay_inactivity_time > 0 && myrpt->keyupinactivitytimer <= 0) {
+		if (myrpt->p.keyupdelay_time > 0 && myrpt->p.keyupdelay_inactivity_time > 0 && myrpt->keyupinactivitytimer <= 0) {
 			myrpt->keyupdelaytimer = myrpt->p.keyupdelay_time;
 			ast_debug(6, "[%s] delaying keyed state for %d ms\n", myrpt->name, myrpt->p.keyupdelay_time);
 			return;
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3608,9 +3608,9 @@ static inline void rpt_keyed(struct rpt *myrpt, int delay)
 			return;
 		}
 
-		if (myrpt->p.keyupdelay_time > 0 && myrpt->p.keyupdelay_inactivity_time > 0 && myrpt->keyupinactivitytimer <= 0) {
-			myrpt->keyupdelaytimer = myrpt->p.keyupdelay_time;
-			ast_debug(6, "[%s] delaying keyed state for %d ms\n", myrpt->name, myrpt->p.keyupdelay_time);
+		if (myrpt->p.first_keyup_min_time > 0 && myrpt->p.first_keyup_inactivity_time > 0 && myrpt->first_keyup_inactivity_timer <= 0) {
+			myrpt->first_keyup_timer = myrpt->p.first_keyup_min_time;
+			ast_debug(6, "[%s] delaying keyed state for %d ms\n", myrpt->name, myrpt->p.first_keyup_min_time);
 			return;
 		}
 	}
@@ -3619,8 +3619,8 @@ static inline void rpt_keyed(struct rpt *myrpt, int delay)
 	myrpt->keyed = 1;
 	time(&myrpt->lastkeyedtime);
 	myrpt->keypost = RPT_KEYPOST_ACTIVE;
-	myrpt->keyupdelaytimer = 0;
-	myrpt->keyupinactivitytimer = myrpt->p.keyupdelay_inactivity_time;
+	myrpt->first_keyup_timer = 0;
+	myrpt->first_keyup_inactivity_timer = myrpt->p.first_keyup_inactivity_time;
 }
 
 static inline int update_timers(struct rpt *myrpt, const int elap, const int totx)
@@ -3677,14 +3677,13 @@ static inline int update_timers(struct rpt *myrpt, const int elap, const int tot
 
 	update_timer(&myrpt->time_out_reset_unkey_interval_timer, elap, 0);
 
-	last_time = myrpt->keyupdelaytimer;
-	update_timer(&myrpt->keyupdelaytimer, elap, 0);
-
-	if (last_time && !myrpt->keyupdelaytimer && !myrpt->keyed) {
+	last_time = myrpt->first_keyup_timer;
+	update_timer(&myrpt->first_keyup_timer, elap, 0);
+	if (last_time && !myrpt->first_keyup_timer && !myrpt->keyed) {
 		rpt_keyed(myrpt, 0);
 	}
 
-	update_timer(&myrpt->keyupinactivitytimer, elap, 0);
+	update_timer(&myrpt->first_keyup_inactivity_timer, elap, 0);
 
 	update_timer(&myrpt->idtimer, elap, 0);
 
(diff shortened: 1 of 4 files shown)


apps/app_rpt.c
```diff
@@ -3594,6 +3594,30 @@ static inline int do_link_post(struct rpt *myrpt)
 	return 0;
 }
 
+static inline void rpt_keyed_set_actual(struct rpt *myrpt)
```
Allan-N: Just a thought...

I wonder if it would make sense to combine these two functions into one :
```
void rpt_set_keyed(struct rpt *myrpt, int delay)
```
If you pass a non-zero `delay` value for the _with_delay behavior.

apps/app_rpt.c
```diff
@@ -3594,6 +3594,30 @@ static inline int do_link_post(struct rpt *myrpt)
...
+		ast_debug(6, "[%s] delaying keyed state for %d ms\n", myrpt->name, myrpt->p.keyupdelay_time);
+		return;
+	}
+
+	rpt_keyed_set_actual(myrpt);
+}
+
 static inline int update_timers(struct rpt *myrpt, const int elap, const int totx)
 {
 	int i;
```
Allan-N: Seems like we could use a more descriptive variable name (and not one like `i` that tends to be used as a loop index) :-)

mkmer: Makes sense, really just reused what was already there :)

apps/app_rpt.c
mkmer: I like explicit names, but we really only use it 1 time outside of the "internal" call.  Let's see how this looks.

configs/rpt/rpt.conf
```diff
@@ -232,6 +232,10 @@ parrottime = 1000                   ; Set the amount of time in milliseconds
 ; reporting of key up/down changes.
 ;statpost_time = 60                 ; (optional) time (in seconds) (min 30, max 600, default 60)
 
+;keyupdelay = 0                    ; (optional) time in milliseconds to delay setting node keyed
```
Allan-N: Nit : can we add a few more spaces so that these semi-colon's line up with those above

Q? are there (or should there be any) bounds on `keyupdelay`?
Q? should the keyupdelay_inactivity be expressed in milliseconds?  or seconds?

Note: if we change the inactivity time to seconds we can still keep the internal variables as milliseconds (just multiply the value in rpt_config.c).

Also, I think the explanation / wording could use some help.  If I'm understanding the changes, the inactivity time controls when the keyup delay is needed (after the repeater/node has been idle for longer than the inactivity time then the first keyup must be longer) and the keyup delay is how long you must be keyed before you can talk.  And maybe we should comment on why one might want to change the values (e.g. mention the delay is to prevent short kerchunks and bursts of noise from keying the repeater/node) ?

apps/app_rpt.c
```diff
@@ -3595,9 +3595,37 @@ static inline int do_link_post(struct rpt *myrpt)
...
+ * \param delay Set the delay if not 0, keyup if 0
+ */
+static inline void rpt_keyed(struct rpt *myrpt, int delay)
+{
+	if (delay) {
+		if (myrpt->keyed) {
+			return;
+		}
+
+		if (delay && myrpt->p.keyupdelay_time > 0 && myrpt->p.keyupdelay_inactivity_time > 0 && myrpt->keyupinactivitytimer <= 0) {
```
Allan-N: No need to check `delay` again.

configs/rpt/rpt.conf
mkmer: TBH, I just left what the AI gave me :(
I've "tweaked" it a bit, feel free to wordsmith further.  
Also need to get on updating the documentation on the web site.

Allan-N (commented):
Q? are we setting up this delay so that it only applies to SimpleUSB/USBRadio/Voter channels?  Asking because I don't think we want to delay any IAX traffic; else we'd be cascading the keyup delay for downstream nodes.

mkmer:
> Q? are we setting up this delay so that it only applies to SimpleUSB/USBRadio/Voter channels? Asking because I don't think we want to delay any IAX traffic; else we'd be cascading the keyup delay for downstream nodes.

This only applies to the `rxchannel_read()` function, so it "should" only apply to the "hardware" channels (aka things connected to the `rxchannel`).  It appears to work that way, but I've missed things before...

apps/app_rpt.c
```diff
@@ -3649,6 +3677,15 @@ static inline int update_timers(struct rpt *myrpt, const int elap, const int tot
 
 	update_timer(&myrpt->time_out_reset_unkey_interval_timer, elap, 0);
 
+	last_time = myrpt->keyupdelaytimer;
+	update_timer(&myrpt->keyupdelaytimer, elap, 0);
+
```
Allan-N: Can we remove this blank line?

configs/rpt/rpt.conf
Allan-N: Do words like the following help ?

```
;first_keyup_min_time = 200         ; after a period of inactivity, the "first keyup"
                                    ; time (in ms) that a local RX key event must be
                                    ; present before the TX will be keyed. This setting
                                    ; can be used to suppress short noise bursts or
                                    ; quick keying (kerchunk'ing).
                                    ; Optional, range 0..1000, default 0.

;first_keyup_inactivity_time = 180  ; the time (in seconds) with no local RX activity
                                    ; before the longer "first keyup" event is needed
                                    ; for TX keying.  Longer intervals would allow
                                    ; quick conversations, shorter intervals would
                                    ; force key-wait-talk behavior.
                                    ; Optional, default 0.
```

and, yes, I did not use the default values for the .conf variables.  I put in place real/sample values as examples so that folks could just remove the leading `;`.

and, yes, the .conf variables names have also been tweaked :-)

mkmer: Done.

Allan-N (approved):

tsawyer:
A kerchunk filter is supposed keep users and noise from kerchunking the repeater by pressing their PTT for a short selectable period of time.

From the description, it sounds like these new timers will do that. Can't wait to try it.

PR #865, opened 2025-12-12
app_rpt: Remove dependency on DAHDI

mkmer:
- Convert all DAHDI/Pseudo channel references to local channels.
- Remove all DAHDI and Pseudo channel related code and configuration.
- Utilize softmix conference bridge for CONF and TXCONF functionality.
- Replace VOX channel with talker detection callback to control transmitter state on outbound autopatch.
- Replace PARROT channel placing file recording inline with txpchannel.

## Summary by CodeRabbit

* **New Features**
  * Bridge-based conferencing with Local-channel types and talker callback support.

* **Improvements**
  * Migrated from hardware-specific channels to Local/bridge topology across autopatch, link and telemetry.
  * Tone playback unified to indication-based API; improved logging and safer CDR handling for local channels.

* **Bug Fixes**
  * Parrot/monitor resources cleaned up automatically; more robust channel teardown.

* **Configuration Changes**
  * Default channel set to Local/pseudo; bridging components enabled and DAHDI timing/driver disabled.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1067, opened 2026-06-15
app_rpt: Restore ability to flush telemetry when a channel is not allocated

mkmer:
The refactored telemetry in PR #865 no longer allocates channels until the telemetry becomes active.  This broke the code which flushes telemetry for various reasons.
The "obvious" sign was PFXTONE not being canceled, specifically if one used `*xxx` in a startup macro.  
Multiple instances of 
```
[2026-06-15 13:50:40.104] DEBUG[220783] app_rpt/rpt_telemetry.c: cancel_pfxfone!!
[2026-06-15 13:50:40.204] DEBUG[220783] app_rpt/rpt_telemetry.c: cancel_pfxfone!!
[2026-06-15 13:50:40.305] DEBUG[220783] app_rpt/rpt_telemetry.c: cancel_pfxfone!!
``` 
Were logged yet the telemetry stayed on the list.

## Summary by CodeRabbit

**Bug Fixes**
- Centralized telemetry termination so entries are consistently marked as stopped, even when they don’t have an associated communication channel.
- Updated the telemetry worker behavior to immediately stop processing killed entries, avoiding continued handling and unnecessary channel allocation.
- Improved cleanup paths across status, paging, and connected-node scenarios to use consistent, safer termination handling.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -337,6 +337,7 @@
 #include "app_rpt/rpt_xcat.h"
 #include "app_rpt/rpt_rig.h"
 #include "app_rpt/rpt_radio.h"
+#include "app_rpt/rpt_telemetry.h"
 
 /*** DOCUMENTATION
 	<application name="Rpt" language="en_US">
@@ -4970,12 +4971,7 @@ static void *rpt(void *this)
 
 	telem = myrpt->tele.next;
 	while (telem != &myrpt->tele) {
-		if (telem->chan) {
-			ast_softhangup(telem->chan, AST_SOFTHANGUP_DEV);
-		} else {
-			telem->killed = 1;
-		}
-
+		rpt_kill_telem(telem);
 		telem = telem->next;
 	}
 	rpt_mutex_unlock(&myrpt->lock);
@@ -5493,23 +5489,21 @@ static void *rpt(void *this)
 			telem = myrpt->tele.next;
 			while (telem != &myrpt->tele) {
 				if (telem->mode == ID && !telem->killed) {
-					telem->killed = 1;
+					rpt_kill_telem(telem);
 					hasid = 1;
-					if (telem->chan) {
-						ast_softhangup(telem->chan, AST_SOFTHANGUP_DEV); /* Whoosh! */
-					}
 				}
+
 				if (telem->mode == TAILMSG && !telem->killed) {
-					telem->killed = 1;
-					if (telem->chan) {
-						ast_softhangup(telem->chan, AST_SOFTHANGUP_DEV); /* Whoosh! */
-					}
+					rpt_kill_telem(telem);
 				}
+
 				if (telem->mode == IDTALKOVER) {
 					hastalkover = 1;
 				}
+
 				telem = telem->next;
 			}
+
 			if (hasid && !hastalkover) {
 				ast_debug(6, "Tracepoint IDTALKOVER\n");
 				rpt_mutex_unlock(&myrpt->lock);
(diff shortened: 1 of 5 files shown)


mkmer:
@coderabbitai review

mkmer:
@coderabbitai review

apps/app_rpt.c
```diff
@@ -4970,7 +4970,12 @@ static void *rpt(void *this)
 
 	telem = myrpt->tele.next;
 	while (telem != &myrpt->tele) {
-		ast_softhangup(telem->chan, AST_SOFTHANGUP_DEV);
+		if (telem->chan) {
+			ast_softhangup(telem->chan, AST_SOFTHANGUP_DEV);
+		} else {
```
Allan-N: In other places you are conditionally calling `ast_softhangup()` but always setting `telem->killed = 1;`.  Why is this case different?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -623,12 +623,17 @@ void flush_telem(struct rpt *myrpt)
...
 
-			if (myrpt->active_telem == telem) {
-				/* If we are the active telemetry, we need to clean it up */
-				telem_done(myrpt, telem);
+				if (myrpt->active_telem == telem) {
+					/* If we are the active telemetry, we need to clean it up */
+					telem_done(myrpt, telem);
+				}
+
+			} else {
```
Allan-N: In other places you are conditionally calling ast_softhangup() but always setting telem->killed = 1;. Why is this case different?

apps/app_rpt/rpt_telemetry.c
```diff
@@ -646,12 +651,17 @@ void birdbath(struct rpt *myrpt)
...
```
Allan-N: In other places you are conditionally calling ast_softhangup() but always setting telem->killed = 1;. Why is this case different?

Allan-N (commented):
A general question.  Once you set `telem->killed = 1` we're done with telemetry, forever?  Is there ever a reason this `telem` would be un-killed?

or, asking another way, once the current telemetry is flushed can more be added?

apps/app_rpt/rpt_telemetry.c
mkmer: I guess we don't need the conditional here, just set killed no matter where we are in the process.
This was originally set up as a "one time" trigger mechanism.  It was easily repurposed to actually kill the telemetry message.

mkmer:
> A general question. Once you set `telem->killed = 1` we're done with telemetry, forever? Is there ever a reason this `telem` would be un-killed?
> 
> or, asking another way, once the current telemetry is flushed can more be added?

Each telemetry "message" creates a thread and a structure `struct rpt_tele` -> All of the threads hang around arbitrating who is next via `myrpt->active_telem` and "die" after they individually do their thing.  With the "new" functionality, we kill a specific message via the killed flag (in its specific structure).  Other than current "backed up" telemetry being flushed, this does not prevent further telemetry messages from being queued and/or activated.

Previously, hanging up on the "channel" associated with the telemetry would "kill" the message (No "unkilling" possible).  
When we switched to "Local", I refactored it to only create the channels when they go "active", which inadvertantly broke the ability to "kill" a message.  This change should restore the functionality as it previously worked.

Allan-N (approved):

mkmer:
Fixes https://community.allstarlink.org/t/unknown-unwanted-dial-tone-after-update-04-09/24492/4

PR #1073, opened 2026-06-18
app_rpt: Protect rxchannel calls causing core dump at shutdown

mkmer:
Resolves coredump / segfault during shutdown / restart.

## Bug Fixes
* Improved DTMF handling and related updates by safely handling missing channel context, using stable channel references, and ensuring manager events are triggered only when valid.  
* Strengthened link updates and repeater notifications so they cope with absent/changing channel state, preventing inconsistent manager event data and improving resource lifetime handling.  
* Updated manager event dispatch to consistently use explicit channel context, ensuring the “Channel” field is reliably populated across repeaters (including MDC1200 and other triggers).

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2593,7 +2593,7 @@ static void local_dtmf_helper(struct rpt *myrpt, char c_in)
 	}
 
 	chan = ast_channel_ref(myrpt->rxchannel);
-	rpt_manager_trigger(myrpt, "DTMF", tone);
+	rpt_manager_trigger(myrpt, chan, "DTMF", tone);
 	ast_channel_unref(chan);
 	donodelog_fmt(myrpt, "DTMF,MAIN,%c", c);
 	if (c == myrpt->p.endchar) {
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -7896,6 +7896,8 @@ static int stop_repeaters(void)
 
 	for (i = 0; i < nrpts; i++) {
 		struct rpt *myrpt = &rpt_vars[i];
+		struct ast_channel *chan = NULL;
+
 		if (!myrpt) {
 			ast_debug(1, "No RPT at index %d?\n", i);
 			continue;
@@ -7907,16 +7909,22 @@ static int stop_repeaters(void)
 		rpt_mutex_lock(&myrpt->lock);
 
 		if (myrpt->rxchannel) {
-			ast_verb(4, "Hanging up channel %s\n", ast_channel_name(myrpt->rxchannel));
-			ast_channel_lock(myrpt->rxchannel);
-			ast_softhangup(myrpt->rxchannel, AST_SOFTHANGUP_EXPLICIT); /* Hanging up one channel will signal the thread to abort */
-			ast_channel_unlock(myrpt->rxchannel);
+			chan = ast_channel_ref(myrpt->rxchannel);
 			myrpt->rxchannel = NULL; /* If we aborted the repeater but haven't unloaded, this channel handle is not valid anymore
 										in a future call to stop_repeaters() */
 		}
 
 		rpt_mutex_unlock(&myrpt->lock);
+
+		if (chan) {
+			ast_verb(4, "Hanging up channel %s\n", ast_channel_name(chan));
+			ast_channel_lock(chan);
+			ast_softhangup(chan, AST_SOFTHANGUP_EXPLICIT); /* Hanging up one channel will signal the thread to abort */
+			ast_channel_unlock(chan);
+			ast_channel_unref(chan);
+		}
 	}
+
 	return 0;
 }
 
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2592,7 +2592,13 @@ static void local_dtmf_helper(struct rpt *myrpt, char c_in)
 		return;
 	}
 
+	rpt_mutex_lock(&myrpt->lock);
 	chan = ast_channel_ref(myrpt->rxchannel);
+	rpt_mutex_unlock(&myrpt->lock);
+	if (!chan) {
+		return;
+	}
+
 	rpt_manager_trigger(myrpt, chan, "DTMF", tone);
 	ast_channel_unref(chan);
 	donodelog_fmt(myrpt, "DTMF,MAIN,%c", c);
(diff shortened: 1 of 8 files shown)


apps/app_rpt/rpt_link.c
```diff
@@ -657,25 +658,33 @@ void rpt_update_links(struct rpt *myrpt)
...
 	rpt_mutex_lock(&myrpt->lock);
 	n = __mklinklist(myrpt, NULL, &buf, USE_FORMAT_RPT_LINK);
 	rpt_mutex_unlock(&myrpt->lock);
 	if (n) {
 		ast_str_set(&obuf, 0, "%d,%s", n, ast_str_buffer(buf));
 	}
-	pbx_builtin_setvar_helper(myrpt->rxchannel, "RPT_LINKS", ast_str_buffer(obuf));
+	pbx_builtin_setvar_helper(chan, "RPT_LINKS", ast_str_buffer(obuf));
 	rpt_manager_trigger(myrpt, "RPT_LINKS", ast_str_buffer(obuf));
 	ast_str_set(&obuf, 0, "%d", n);
```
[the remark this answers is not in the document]
mkmer: These are not new calls, and while they could fail, we don't check them anywhere else in the code.

apps/app_rpt/rpt_link.c
```diff
@@ -657,25 +658,33 @@ void rpt_update_links(struct rpt *myrpt)
...
-	pbx_builtin_setvar_helper(myrpt->rxchannel, "RPT_NUMALINKS", ast_str_buffer(obuf));
-	rpt_manager_trigger(myrpt, "RPT_NUMALINKS", ast_str_buffer(obuf));
 
+	if (!myrpt->rxchannel) {
+		ast_free(buf);
+		ast_free(obuf);
+		return;
+	}
+
+	chan = ast_channel_ref(myrpt->rxchannel);
```
Allan-N, line 668: Should we be grabbing the rxchannel ref while we are holding myrpt->lock ?
```
	rpt_mutex_lock(&myrpt->lock);

	chan = myrpt->rxchannel ? ast_channel_ref(myrpt->rxchannel) : NULL;
	if (!chan) {
		/* if no rxchannel */
		ast_free(obuf);
		return;
	}

	n = __mklinklist(myrpt, NULL, &buf, USE_FORMAT_RPT_ALINK);

	/* parse em */
	if (n) {
		ast_str_set(&obuf, 0, "%d,%s", n, ast_str_buffer(buf));
	}

	pbx_builtin_setvar_helper(chan, "RPT_ALINKS", ast_str_buffer(obuf));
	...
```

mkmer: I "think so", we don't want other threads de-referencing the pointer held in `myrpt->rxchannel` while we look at it and create a copy (right?)  Once we have a copy + reference to the we don't care what happens to the ref pointer in `myrpt->rxchannel` and the structure has a ref so it won't disappear until we are done.
At least that's my take on it :)

apps/app_rpt/rpt_bridging.c
```diff
@@ -118,9 +119,11 @@ static struct ast_channel **rpt_chan_channel(struct rpt *myrpt, struct rpt_link
 void rpt_hangup(struct rpt *myrpt, enum rpt_chan_type chantype)
 {
 	struct ast_channel **chanptr = rpt_chan_channel(myrpt, NULL, chantype);
```
Allan-N: Should we be grabbing the lock before we call rpt_chan_channel() ?

apps/app_rpt/rpt_bridging.c
```diff
@@ -150,6 +153,7 @@ void rpt_hangup(struct rpt *myrpt, enum rpt_chan_type chantype)
 	ast_debug(2, "Hanging up channel %s\n", ast_channel_name(*chanptr));
 	ast_hangup(*chanptr);
 	*chanptr = NULL;
+	rpt_mutex_unlock(&myrpt->lock);
```
Allan-N, line 157: Is it OK (and preferred) to hold the lock when calling ast_hangup() ?

apps/app_rpt/app_rpt.h
```diff
@@ -1160,7 +1160,14 @@ void donodelog(struct rpt *myrpt, char *str);
...
 
-void rpt_event_process(struct rpt *myrpt);
+/*!
+ * \brief Process RPT events for a repeater using a caller-owned channel reference.
+ *
+ * \param myrpt Non-NULL reference to the repeater structure.
+ * \param chan Non-NULL channel reference that remains valid for the duration of this call.
+ */
+void rpt_event_process(struct rpt *myrpt, struct ast_channel *chan);
+void rpt_event_process(struct rpt *myrpt, struct ast_channel *chan);
```
Allan-N: This line looks to be a duplicate.

apps/app_rpt.c
```diff
@@ -789,6 +789,7 @@ void rpt_event_process(struct rpt *myrpt)
 		/* start indicating no command to do */
 		cmd = NULL;
 		c = toupper(*argv[1]);
+
```
Allan-N, line 791: Why are we adding this blank line?  and, if you want some separation then "I" would think that the `c = ...` goes with the `if (c == ...`

apps/app_rpt.c
```diff
@@ -2582,12 +2583,26 @@ static void local_dtmf_helper(struct rpt *myrpt, char c_in)
...
+	}
+
+	chan = ast_channel_ref(myrpt->rxchannel);
+	rpt_mutex_unlock(&myrpt->lock);
+	if (!chan) {
+		return;
+	}
+
+	rpt_manager_trigger(myrpt, chan, "DTMF", tone);
+	ast_channel_unref(chan);
```
Allan-N, line 2605: Here, you removed the blank line between the channel manipulation and the logging block.  "I" would have kept that blank line.

apps/app_rpt/rpt_bridging.c
mkmer: I don't think it hurts anything to be locked while hanging up.  We "could" grab a ref then set NULL, unlock and hangup the ref if it's not "ok".  The "protection" here is for the `rxchannel` reference (or any other channel reference we are about to "NULL")

Allan-N (approved):

PR #1081, opened 2026-06-25
app_rpt: Handle double disconnect messages

mkmer:
Prevent reporting a disconnect and executing the disconnect program twice.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link shutdown behavior by draining queued link text more cleanly during disconnects.
  * Ensured queued link text is processed earlier in each periodic cycle.
  * Reduced unnecessary outbound reconnect state changes, improving retry consistency.
  * Adjusted disconnect telemetry/logging so normal disconnects no longer produce failure-style reporting (while preserving silent behavior).

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4463,10 +4463,9 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 				hangup_link_chan(l);
 				return 1;
 			}
-			if (l->outbound && (l->retries++ < l->max_retries) && (l->hasconnected)) {
+			if (l->outbound && (l->retries++ < l->max_retries) && l->hasconnected) {
 				hangup_link_chan(l);
 				rpt_mutex_lock(&myrpt->lock);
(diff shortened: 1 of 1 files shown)


mkmer:
@coderabbitai review

apps/app_rpt.c
```diff
@@ -3520,9 +3531,11 @@ static inline void periodic_process_link(struct rpt *myrpt, struct rpt_link *l,
 					rpt_telemetry(myrpt, REMDISC, l);
 				}
 			}
-			if (l->hasconnected)
+			if (l->hasconnected) {
 				rpt_update_links(myrpt);
+			}
 			donodelog_fmt(myrpt, l->hasconnected ? "LINKDISC,%s" : "LINKFAIL,%s", l->name);
+			dodispgm(myrpt, l->name);
```
Allan-N: On maxretries, I don't think we've ever "connected".   If that's the case, should we be running the disconnect program?

mkmer: We "could" have connected, lost the connection and been in the retry loop (I believe so any way).
Maybe should check that `l->hasconnected` is true before running `dodispgm()`?

Allan-N (approved):

PR #1082, opened 2026-06-25
res_usbradio: remove code only used with ancient kernels

Allan-N:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USB radio device matching so the correct audio device is more reliably selected across Linux systems.
  * Simplified the device lookup behavior to reduce cases where radios could be associated with the wrong sound card.


mkmer (approved):
Makes sense.

PR #1084, opened 2026-06-29
chan_voter: VOTER Client Auth Refactor

### LINKED ISSUE #856: chan_voter: Doesn't detect loss of master client when it is the only client
The "master" client is the master timing source for an instance. If it isn't available, then nothing else matters.

In a configuration where we have a master client and >1 "other" clients, and the master client disappears, we detect that and throw a warning: VOTER lost master timing source!!". We should probably do something else to shut the channel down, but that is a different issue.

In a configuration where we only have a master client configured, and that client goes away (ie network disconnected), `chan_voter` does not detect that anything is wrong. There are timeout checks, but it appears that we never get to that part of the code.

It appears that as long as there is data coming in the UDP port from multiple clients, we continue to iterate on the `voter_reader` loop, and can detect the loss of the master client.

However, if the master client is the only client sending data, and it stops, we aren't handling that case, since there are no more packets to process.

### LINKED ISSUE #902: Voter: many client disconnects when believed "normal" network
```
[2026-01-06 04:19:18.614] NOTICE[1596827] chan_voter.c: Client: Ada is sending mix mode flag, setting client to mix mode
[2026-01-06 04:58:53.889] NOTICE[1596828] chan_voter.c: VOTER client Ada disconnect (timeout)
[2026-01-06 05:18:41.549] NOTICE[1596828] chan_voter.c: VOTER client Ada disconnect (timeout)
[2026-01-06 05:18:48.277] NOTICE[1596827] chan_voter.c: Client: Ada is sending mix mode flag, setting client to mix mode
[2026-01-06 05:38:32.929] NOTICE[1596828] chan_voter.c: VOTER client Ada disconnect (timeout)
[2026-01-06 05:38:38.341] NOTICE[1596827] chan_voter.c: Client: Ada is sending mix mode flag, setting client to mix mode
```

### LINKED ISSUE #1062: Potential issue in chan_voter.c
**Actionable comments posted: 1**

<details>
<summary>♻️ Duplicate comments (1)</summary><blockquote>

<details>
<summary>channels/chan_voter.c (1)</summary><blockquote>

`1230-1231`: _⚠️ Potential issue_ | _🔴 Critical_

**Fix frame-size padding math to avoid tail overread.**

`audio_samples += audio_samples % FRAME_SIZE` does not round up to a `FRAME_SIZE` boundary; it can still leave a partial tail before the fixed-size copy loop.

<details>
<summary>Proposed fix</summary>

```diff
-		audio_samples += audio_samples % FRAME_SIZE;
+		audio_samples += (FRAME_SIZE - (audio_samples % FRAME_SIZE)) % FRAME_SIZE;
```
</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@channels/chan_voter.c` around lines 1230 - 1231, The padding math is wrong:
replace the line that does audio_samples += audio_samples % FRAME_SIZE with
logic that rounds audio_samples up to the next multiple of FRAME_SIZE (compute
the padding as (FRAME_SIZE - (audio_samples % FRAME_SIZE)) % FRAME_SIZE and add
that), so audio_samples becomes a true multiple of FRAME_SIZE before the
fixed-size copy loop; update the code around the audio_samples variable in
channels/chan_voter.c accordingly to use the new padded value to avoid tail
overread.
```

</details>

</blockquote></details>

</blockquote></details>

<details>
<summary>🤖 Prompt for all review comments with AI agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

Inline comments:
In `@channels/chan_simpleusb.c`:
- Around line 1156-1159: When ast_radio_usb_device_from_alsa_card(o->devicenum)
returns NULL, fail the audio-device initialization instead of just logging: in
the function that contains the o->usb_dev assignment (the init_audio_device /
simpleusb init routine that sets o->usb_dev and then continues), change the
branch so that if o->usb_dev is NULL you log an error (not just ast_debug) and
immediately return -1 to abort initialization; this ensures the channel isn’t
marked and mixer probing isn’t attempted when the HID-backed USB device cannot
be resolved.

---

Duplicate comments:
In `@channels/chan_voter.c`:
- Around line 1230-1231: The padding math is wrong: replace the line that does
audio_samples += audio_samples % FRAME_SIZE with logic that rounds audio_samples
up to the next multiple of FRAME_SIZE (compute the padding as (FRAME_SIZE -
(audio_samples % FRAME_SIZE)) % FRAME_SIZE and add that), so audio_samples
becomes a true multiple of FRAME_SIZE before the fixed-size copy loop; update
the code around the audio_samples variable in channels/chan_voter.c accordingly
to use the new padded value to avoid tail overread.
```

</details>

<details>
<summary>🪄 Autofix (Beta)</summary>

Fix all unresolved CodeRabbit comments on this PR:

- [ ]  Push a commit to this branch (recommended)
- [ ]  Create a new PR with the fixes

</details>

---

<details>
<summary>ℹ️ Review info</summary>

<details>
<summary>⚙️ Run configuration</summary>

**Configuration used**: Path: .coderabbit.yaml

**Review profile**: CHILL

**Plan**: Pro

**Run ID**: `9fbe10a7-7392-475a-9bb8-9aae798d0e75`

</details>

<details>
<summary>📥 Commits</summary>

Reviewing files that changed from the base of the PR and between 5d9a4b81059fb7128e12e728227a0ef598d1ed65 and 0c5a53a807c7c31eda4fe8071ed420cbfd18fa82.

</details>

<details>
<summary>📒 Files selected for processing (6)</summary>

* `channels/Makefile.diff`
* `channels/chan_simpleusb.c`
* `channels/chan_usbradio.c`
* `channels/chan_voter.c`
* `include/asterisk/res_usbradio.h`
* `res/res_usbradio.c`

</details>

<details>
<summary>✅ Files skipped from review due to trivial changes (1)</summary>

* channels/chan_usbradio.c

</details>

<details>
<summary>🚧 Files skipped from review as they are similar to previous changes (1)</summary>

* channels/Makefile.diff

</details>

</details>

_Originally posted by @coderabbitai[bot] in https://github.com/AllStarLink/app_rpt/pull/1029#pullrequestreview-4429002723_

### LINKED ISSUE #1079: chan_voter: module reload prevents transmit
`module reload chan_voter.so` or `core reload` causes the voter/RTCM to no longer transmit.  Restarting Asterisk required to recover.
For sure >3.8.0 maybe older versions.

ve7fet:
Initial refactoring of the client authentication process.

Reorganize/rewrite how we process client connections to better follow the protocol specification.

Explicitly set the auth packet payload type during authentication.

Refactor master client connect and disconnect routines.

Move and revise where and how we detect master clients.

More gracefully fail over to backup master client (if multiple are defined and exist on the same network).

Move a ulaw debug block and fix to prevent buffer over-read.
 
Add additional debugging and status messages.

Implement consistent connect/disconnect messages for clients.

Force disconnection of all clients on a server when there is a master client defined (ie. voting applications), and the master client disconnects. Note this will also disconnect any mix mode clients on the server as well, since loss of the master timing source makes continued operation moot.

Block clients from connecting if there is a master client defined, but it hasn't connected yet. Can't do anything without our master timing source (when defined).

Add a method to track and log when a disconnected active master client becomes connected.

Add a number of comments on how the master client selection process works.

The voter_timer thread was using both time() and gettimeofday() for handling the evaluation of timed out clients. This could potentially be causing random timeouts and disconnects of clients, since the two time sources may not be identical. Converted to use gettimeofday() exclusively.

Re-factor the voter_timer thread to monitor all connected clients for loss of connectivity, and log a message when they disconnect.

Remove the duplicate code for checking for timed out clients from voter_reader, which only ran when there was a master client connected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved client authentication and disconnect behavior when no master is available.
  * Fixed handling of master loss so remaining clients are disconnected more reliably.
  * Tightened duplicate-client checks to avoid false matches and inconsistent connections.
  * Improved packet and audio handling to reduce invalid responses and dropped or malformed playback.
  * Added more robust timer and network processing for better stability during reloads and runtime changes.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -4121,6 +4121,7 @@ static int reload(void)
 			cp = ast_strdup(v->value);
 			if (!cp) {
 				close(udp_socket);
+				udp_socket = -1;
 				ast_config_destroy(cfg);
 				ast_mutex_unlock(&voter_lock);
 				return -1;
@@ -4148,6 +4149,7 @@ static int reload(void)
 				if (!client) {
 					ast_free(cp);
 					close(udp_socket);
+					udp_socket = -1;
 					ast_config_destroy(cfg);
 					ast_mutex_unlock(&voter_lock);
 					return -1;
@@ -4210,6 +4212,7 @@ static int reload(void)
 				client->audio = ast_realloc(client->audio, client->buflen);
 				if (!client->audio) {
 					close(udp_socket);
+					udp_socket = -1;
 					ast_config_destroy(cfg);
 					ast_mutex_unlock(&voter_lock);
 					return -1;
@@ -4219,6 +4222,7 @@ static int reload(void)
 				client->audio = ast_malloc(client->buflen);
 				if (!client->audio) {
 					close(udp_socket);
+					udp_socket = -1;
 					ast_config_destroy(cfg);
 					ast_mutex_unlock(&voter_lock);
 					return -1;
@@ -4229,6 +4233,7 @@ static int reload(void)
 				client->rssi = ast_realloc(client->rssi, client->buflen);
 				if (!client->rssi) {
 					close(udp_socket);
+					udp_socket = -1;
 					ast_config_destroy(cfg);
 					ast_mutex_unlock(&voter_lock);
 					return -1;
@@ -4238,6 +4243,7 @@ static int reload(void)
 				client->rssi = ast_calloc(1, client->buflen);
 				if (!client->rssi) {
 					close(udp_socket);
+					udp_socket = -1;
 					ast_config_destroy(cfg);
 					ast_mutex_unlock(&voter_lock);
 					return -1;
@@ -4265,6 +4271,8 @@ static int reload(void)
 		if (client->digest == 0) {
 			ast_log(LOG_ERROR, "Can not load chan_voter -- VOTER client %s has invalid authentication digest (can not be 0)!!!\n",
 				client->name);
+			close(udp_socket);
+			udp_socket = -1;
 			ast_mutex_unlock(&voter_lock);
 			return -1;
 		}
@@ -4278,6 +4286,8 @@ static int reload(void)
 			if (client->digest == client1->digest) {
 				ast_log(LOG_ERROR, "Can not load chan_voter -- VOTER clients %s and %s have same authentication digest!!!\n",
 					client->name, client1->name);
+				close(udp_socket);
+				udp_socket = -1;
 				ast_mutex_unlock(&voter_lock);
 				return -1;
 			}
@@ -6145,6 +6155,7 @@ static int load_module(void)
 	if (bind(udp_socket, &sin, sizeof(sin)) == -1) {
 		ast_log(LOG_ERROR, "Unable to bind port for VOTER audio connection: %s\n", strerror(errno));
 		close(udp_socket);
+		udp_socket = -1;
 		return AST_MODULE_LOAD_DECLINE;
 	}
 
@@ -6156,6 +6167,7 @@ static int load_module(void)
 		if (setsockopt(udp_socket, IPPROTO_IP, IP_TOS, &i, sizeof(i))) {
 			ast_log(LOG_ERROR, "Can't setsockopt: IP_TOS: %s\n", strerror(errno));
 			close(udp_socket);
+			udp_socket = -1;
 			return AST_MODULE_LOAD_DECLINE;
 		}
 	}
@@ -6197,6 +6209,7 @@ static int load_module(void)
 		ast_timer_close(voter_thread_timer);
 		voter_thread_timer = NULL;
 		close(udp_socket);
+		udp_socket = -1;
 		return AST_MODULE_LOAD_DECLINE;
 	}
 	ast_format_cap_append(voter_tech.capabilities, ast_format_slin, 0);
@@ -6207,6 +6220,7 @@ static int load_module(void)
 		ast_timer_close(voter_thread_timer);
 		voter_thread_timer = NULL;
 		close(udp_socket);
+		udp_socket = -1;
 		return AST_MODULE_LOAD_DECLINE;
 	}
 	nullfd = open("/dev/null", O_RDWR);
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -3851,6 +3851,8 @@ static int reload(void)
 
 	if (!(cfg = ast_config_load(config, zeroflag))) {
 		ast_log(LOG_ERROR, "Unable to load/reload config %s\n", config);
+		close(udp_socket);
+		udp_socket = -1;
 		ast_mutex_unlock(&voter_lock);
 		return -1;
 	} else {
@@ -6173,8 +6175,15 @@ static int load_module(void)
 	/* Set voter_thread_timer for 50 ticks/second (every 20ms). */
 	ast_timer_set_rate(voter_thread_timer, VOTER_TICKS);
 
-	/* Load the rest of the values from the config file by running reload. */
+	/* Load the rest of the values from the config file by running reload.
+	 * reload() will return 0 on success, or a non-zero value on failure.
+	 * If it returns non-zero, we will abort loading the module.
+	 * We won't close the UDP socket here, because reload() will do that if it fails,
+	 * avoiding a double-close of the socket.
+	 */
 	if (reload()) {
+		ast_timer_close(voter_thread_timer);
+		voter_thread_timer = NULL;
 		return AST_MODULE_LOAD_DECLINE;
 	}
 
@@ -6186,6 +6195,7 @@ static int load_module(void)
 
 	if (!(voter_tech.capabilities = ast_format_cap_alloc(AST_FORMAT_CAP_FLAG_DEFAULT))) {
 		ast_timer_close(voter_thread_timer);
+		voter_thread_timer = NULL;
 		close(udp_socket);
 		return AST_MODULE_LOAD_DECLINE;
 	}
@@ -6195,6 +6205,7 @@ static int load_module(void)
 	if (ast_channel_register(&voter_tech)) {
 		ast_log(LOG_ERROR, "Unable to register channel class %s\n", type);
 		ast_timer_close(voter_thread_timer);
+		voter_thread_timer = NULL;
 		close(udp_socket);
 		return AST_MODULE_LOAD_DECLINE;
 	}
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -4400,6 +4400,8 @@ static void *voter_timer(void *data)
 	struct voter_pvt *p;
 	struct voter_client *client, *client1;
 	struct timeval tv;
+	char *client_ip;
+	char *client1_ip;
 
 	while (run_forever && !ast_shutting_down()) {
 		/* Check and acknowledge our thread timer. This timer keeps our audio in sync (for IAX2). */
@@ -4517,8 +4519,8 @@ static void *voter_timer(void *data)
 							}
 
 							/* Save the IP string so we can call ast_inet_ntoa again */
-							char *client_ip = ast_strdupa(ast_inet_ntoa(client->sin.sin_addr));
-							char *client1_ip = ast_strdupa(ast_inet_ntoa(client1->sin.sin_addr));
+							client_ip = ast_strdupa(ast_inet_ntoa(client->sin.sin_addr));
+							client1_ip = ast_strdupa(ast_inet_ntoa(client1->sin.sin_addr));
 							ast_debug(2, "Client %s IP: %s:%d = client %s IP: %s:%d\r\n", client->name, client_ip,
 								ntohs(client->sin.sin_port), client1->name, client1_ip, ntohs(client1->sin.sin_port));
 							ast_log(LOG_ERROR, "Client %s and client %s have same IP and port! Resetting client connections (sanity)\n",
@@ -4557,6 +4559,8 @@ static void *voter_reader(void *data)
 {
 	char buf[4096], timestr[100], hasmastered, *cp, *cp1;
 	char gps1[300], gps2[300], isproxy;
+	char *client_ip;
+	char *client1_ip;
 	struct sockaddr_in sin, sin_stream, psin;
 	struct voter_pvt *p;
 	int fd, i, j, k, timeout_ms, maxrssi, master_port, no_ast_channel = 0, logged_no_ast_channel = 0, logged_buflen_too_small = 0;
@@ -5458,8 +5462,8 @@ static void *voter_reader(void *data)
 										}
 
 										/* Save the IP string so we can call ast_inet_ntoa again */
-										char *client_ip = ast_strdupa(ast_inet_ntoa(client->sin.sin_addr));
-										char *client1_ip = ast_strdupa(ast_inet_ntoa(client1->sin.sin_addr));
+										client_ip = ast_strdupa(ast_inet_ntoa(client->sin.sin_addr));
+										client1_ip = ast_strdupa(ast_inet_ntoa(client1->sin.sin_addr));
 										ast_debug(2, "Client %s IP: %s:%d = client %s IP: %s:%d\r\n", client->name, client_ip,
 											ntohs(client->sin.sin_port), client1->name, client1_ip, ntohs(client1->sin.sin_port));
 										ast_log(LOG_ERROR, "Client %s and client %s have same IP and port! Resetting client connections (sanity)\n",
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -4853,18 +4853,18 @@ static void *voter_reader(void *data)
 						if (!logged_buflen_too_small) {
 							ast_log(LOG_ERROR, "VOTER %u: Mix-mode client %s (proxy) rejected: buflen=%d (<160). Fix voter.conf.\n",
 								client->nodenum, client->name, client->buflen / 8);
-							logged_buflen_too_small = 1;
+							logged_buflen_too_small = 1; /* Only want to log this once */
 						}
 						client->mix = 0;
 						client->heardfrom = 0;
 						client->respdigest = 0;
 						continue;
-					} else {
-						client->mix = 1;
-						ast_log(LOG_NOTICE, "VOTER %u: Client %s is sending mix mode flag, setting client to mix mode\n",
-							client->nodenum, client->name);
-						logged_buflen_too_small = 0;
 					}
+					/* If client->buflen is sane, set the mix mode flag. */
+					client->mix = 1;
+					ast_log(LOG_NOTICE, "VOTER %u: Client %s is sending mix mode flag, setting client to mix mode\n",
+						client->nodenum, client->name);
+					logged_buflen_too_small = 0;
 				}
 			}
 
@@ -4881,27 +4881,26 @@ static void *voter_reader(void *data)
 				client->heardfrom = 0;
 				client->respdigest = 0;
 				continue;
-			} else {
-				/* Otherwise, we should be good to continue configuring the client.
-				 *
-				 * Set the flags we are going to send to the client for configuration.
-				 */
-				authpacket.flags = 0;
-				if (client->ismaster) {
-					authpacket.flags |= (FLAG_SENDALWAYS | FLAG_MASTERTIMING);
-				}
-				if (client->doadpcm) {
-					authpacket.flags |= FLAG_ADPCM;
-				}
-				if (client->mix) {
-					authpacket.flags |= FLAG_MIX;
-				}
-				if (client->nodeemp || (p && p->hostdeemp)) {
-					authpacket.flags |= FLAG_FLATAUDIO;
-				}
-				if (client->noplfilter) {
-					authpacket.flags |= FLAG_NOCTCSSFILTER;
-				}
+			}
+			/* Otherwise, we should be good to continue configuring the client.
+			 *
+			 * Set the flags we are going to send to the client for configuration.
+			 */
+			authpacket.flags = 0;
+			if (client->ismaster) {
+				authpacket.flags |= (FLAG_SENDALWAYS | FLAG_MASTERTIMING);
+			}
+			if (client->doadpcm) {
+				authpacket.flags |= FLAG_ADPCM;
+			}
+			if (client->mix) {
+				authpacket.flags |= FLAG_MIX;
+			}
+			if (client->nodeemp || (p && p->hostdeemp)) {
+				authpacket.flags |= FLAG_FLATAUDIO;
+			}
+			if (client->noplfilter) {
+				authpacket.flags |= FLAG_NOCTCSSFILTER;
 			}
 
 			/* The sin structure has the IP info received off the wire for the current packet from
@@ -4909,7 +4908,7 @@ static void *voter_reader(void *data)
 			 * update the client's sin structure with this information.
 			 */
 			client->sin = sin;
-			/*! \todo VE7FET not sure why this is here/needed, when it is dealing with proxy stuff, it probably needs to be moved */
+			/*Reset the client->proxy_sin IP structure to zero to prevent stale proxy IP information. */
 			memset(&client->proxy_sin, 0, sizeof(client->proxy_sin));
 
 			/* Print the address and port the client is connecting from */
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -6151,7 +6151,8 @@ static int load_module(void)
 	/* Do an initial configuration load from the config file. Note we also run reload
 	 * further down.
 	 */
-	if (!(cfg = ast_config_load(config, zeroflag))) {
+	cfg = ast_config_load(config, zeroflag);
+	if (!cfg || cfg == CONFIG_STATUS_FILEINVALID) {
 		ast_log(LOG_ERROR, "Unable to load config %s\n", config);
 		return AST_MODULE_LOAD_DECLINE;
 	}
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -4397,15 +4397,25 @@ static void *voter_timer(void *data)
 		int timer_fd = ast_timer_fd(voter_thread_timer);
 		int timeout_ms = -1; /* block until the timer fires */
 
+		if (timer_fd < 0) {
+			ast_log(LOG_ERROR, "Failed to get pollable thread timer fd.\n");
+			break;
+		}
+
 		if (ast_waitfor_n_fd(&timer_fd, 1, &timeout_ms, NULL) < 0) {
 			ast_log(LOG_ERROR, "Failed to wait on VOTER thread timer.\n");
 			break;
 		}
-		if (ast_timer_get_event(voter_thread_timer) == AST_TIMING_EVENT_EXPIRED) {
+		if (ast_timer_get_event(voter_thread_timer) != AST_TIMING_EVENT_EXPIRED) {
 			if (ast_timer_ack(voter_thread_timer, 1) < 0) {
 				ast_log(LOG_ERROR, "Failed to acknowledge timer.\n");
 				break;
 			}
+			continue;
+		}
+		if (ast_timer_ack(voter_thread_timer, 1) < 0) {
+			ast_log(LOG_ERROR, "Failed to acknowledge timer.\n");
+			break;
 		}
 
 		ast_mutex_lock(&voter_lock);
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -6235,9 +6235,33 @@ static int load_module(void)
 	return AST_MODULE_LOAD_SUCCESS;
 }
 
+/*!
+ * \brief Reload the VOTER channel module
+ *
+ * This function is called when the module is reloaded. It locks the voter_lock mutex,
+ * calls the reload() function to reload the configuration, and unlocks the mutex.
+ * If reload() fails, it logs an error and closes the UDP socket.
+ *
+ *	 \return 				0 on success; non-zero on failure (typically -1 if the configuration could not be reloaded).
+ */
+static int asterisk_reload(void)
+{
+	int res;
+
+	ast_mutex_lock(&voter_lock);
+	res = reload();
+	if (res) {
+		ast_log(LOG_ERROR, "Failed to reload configuration\n");
+		close(udp_socket);
+		udp_socket = -1;
+	}
+	ast_mutex_unlock(&voter_lock);
+	return res; /* Return the result of the reload operation back to Asterisk*/
+}
+
 AST_MODULE_INFO(ASTERISK_GPL_KEY, AST_MODFLAG_DEFAULT, "Voter Radio Channel Driver",
 	.support_level = AST_MODULE_SUPPORT_EXTENDED,
 	.load = load_module,
 	.unload = unload_module,
-	.reload = reload,
+	.reload = asterisk_reload,
 );
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -4658,9 +4658,10 @@ static void *voter_reader(void *data)
 		}
 		isproxy = 0;
 
-		/* No valid Asterisk channel, do not respond to the client. This prevents the client
-		 * from "looping" through online/offline mode when there is no valid channel in app_rpt
-		 * to connect to.
+		/* We check (further down) during authentication to see if a client can connect to a valid Asterisk channel,
+		 * and update no_ast_channel accordingly. If we didn't find a valid Asterisk channel, we do not want to
+		 * respond to the client. By not responding, we prevent the client from "looping" through online/offline
+		 * mode when there is no valid channel in app_rpt to connect to.
 		 */
 		if (no_ast_channel) {
 			continue;
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -424,18 +424,19 @@ char context[100];
 
 /* Define voter packet flags. */
 enum voter_auth_flags {
-    /*! \brief Send flat audio (nodeemp or hostdeemp) (aka Flag 1) */
-    FLAG_FLATAUDIO = (1 << 0),
-    /*! \brief Send audio always (master) (aka Flag 2) */
-    FLAG_SENDALWAYS = (1 << 1),
-    /*! \brief Do not filter CTCSS (noplfilter) (aka Flag 4) */
-    FLAG_NOCTCSSFILTER = (1 << 2),
-    /*! \brief Master timing source (do not delay sending audio packet) (master) (aka Flag 8) */
-    FLAG_MASTERTIMING = (1 << 3),
-    /*! \brief Use ADPCM rather than ulaw (adpcm) (aka Flag 16) */
-    FLAG_ADPCM = (1 << 4),
-    /*! \brief Request "mix" option to host (mixminus) (aka Flag 32) */
-    FLAG_MIX = (1 << 5),
+  /*! \brief Send flat audio (nodeemp or hostdeemp) (aka Flag 1) */
+  FLAG_FLATAUDIO = (1 << 0),
+  /*! \brief Send audio always (master) (aka Flag 2) */
+  FLAG_SENDALWAYS = (1 << 1),
+  /*! \brief Do not filter CTCSS (noplfilter) (aka Flag 4) */
+  FLAG_NOCTCSSFILTER = (1 << 2),
+  /*! \brief Master timing source (do not delay sending audio packet) (master)
+     (aka Flag 8) */
+  FLAG_MASTERTIMING = (1 << 3),
+  /*! \brief Use ADPCM rather than ulaw (adpcm) (aka Flag 16) */
+  FLAG_ADPCM = (1 << 4),
+  /*! \brief Request "mix" option to host (mixminus) (aka Flag 32) */
+  FLAG_MIX = (1 << 5),
 };
 
 /* vdesc and type are used when Asterisk interacts with our module. */
@@ -4517,9 +4518,8 @@ static void *voter_timer(void *data)
 
 							char *client_ip = ast_strdupa(ast_inet_ntoa(client->sin.sin_addr)); /* save the string so we can call ast_inet_ntoa again */
 							char *client1_ip = ast_strdupa(ast_inet_ntoa(client1->sin.sin_addr));
-							ast_debug(2, "Client %s IP: %s:%d = client %s IP: %s:%d\r\n",
-								client->name, client_ip, ntohs(client->sin.sin_port),
-								client1->name, client1_ip, ntohs(client1->sin.sin_port));
+							ast_debug(2, "Client %s IP: %s:%d = client %s IP: %s:%d\r\n", client->name, client_ip,
+								ntohs(client->sin.sin_port), client1->name, client1_ip, ntohs(client1->sin.sin_port));
 							ast_log(LOG_ERROR, "Client %s and client %s have same IP and port! Resetting client connections (sanity)\n",
 								client->name, client1->name);
 							client->respdigest = 0;
@@ -4605,7 +4605,7 @@ static void *voter_reader(void *data)
 
 	while (run_forever && !ast_shutting_down()) {
 		ast_mutex_unlock(&voter_lock);
-		timeout_ms = 50;										 /* 50ms timeout */
+		timeout_ms = 50;										/* 50ms timeout */
 		fd = ast_waitfor_n_fd(&udp_socket, 1, &timeout_ms, NULL); /* Poll the UDP socket, looking for data */
 		ast_mutex_lock(&voter_lock);
 		/* Check the returned fd and see if there is a datagram ready to process.
@@ -4636,7 +4636,7 @@ static void *voter_reader(void *data)
 				p->lastwon = NULL;
 			}
 		}
-		 /* Go get the datagram from the UDP port, and start processing. */
+		/* Go get the datagram from the UDP port, and start processing. */
 		fromlen = sizeof(struct sockaddr_in);
 		recvlen = recvfrom(udp_socket, buf, sizeof(buf) - 1, 0, (struct sockaddr *) &sin, &fromlen);
 		/* Handle recvfrom() errors */
@@ -5144,7 +5144,8 @@ static void *voter_reader(void *data)
 			if (client && client->heardfrom &&
 				(((ntohs(vph->payload_type) == VOTER_PAYLOAD_ULAW) && (recvlen == (sizeof(VOTER_PACKET_HEADER) + FRAME_SIZE + 1))) ||
 					((ntohs(vph->payload_type) == VOTER_PAYLOAD_ADPCM) && (recvlen == (sizeof(VOTER_PACKET_HEADER) + FRAME_SIZE + 4))) ||
-					((ntohs(vph->payload_type) == VOTER_PAYLOAD_PROXY) && ((size_t) recvlen >= (sizeof(VOTER_PACKET_HEADER) + sizeof(VOTER_PROXY_HEADER)))))) {
+					((ntohs(vph->payload_type) == VOTER_PAYLOAD_PROXY) &&
+						((size_t) recvlen >= (sizeof(VOTER_PACKET_HEADER) + sizeof(VOTER_PROXY_HEADER)))))) {
 				/* Find the matching Asterisk channel for this client. */
 				for (p = pvts; p; p = p->next) {
 					if (p->nodenum == client->nodenum) {
@@ -5455,11 +5456,11 @@ static void *voter_reader(void *data)
 											continue;
 										}
 
-										char *client_ip = ast_strdupa(ast_inet_ntoa(client->sin.sin_addr)); /* save the string so we can call ast_inet_ntoa again */
+										char *client_ip =
+											ast_strdupa(ast_inet_ntoa(client->sin.sin_addr)); /* save the string so we can call ast_inet_ntoa again */
 										char *client1_ip = ast_strdupa(ast_inet_ntoa(client1->sin.sin_addr));
-										ast_debug(2, "Client %s IP: %s:%d = client %s IP: %s:%d\r\n",
-											client->name, client_ip, ntohs(client->sin.sin_port),
-											client1->name, client1_ip, ntohs(client1->sin.sin_port));
+										ast_debug(2, "Client %s IP: %s:%d = client %s IP: %s:%d\r\n", client->name, client_ip,
+											ntohs(client->sin.sin_port), client1->name, client1_ip, ntohs(client1->sin.sin_port));
 										ast_log(LOG_ERROR, "Client %s and client %s have same IP and port! Resetting client connections (sanity)\n",
 											client->name, client1->name);
 										client->respdigest = 0;
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -442,6 +442,7 @@ struct ast_timer *voter_thread_timer = NULL;
 int voter_timing_count = 0;
 int last_master_count = 0;
 int hasmaster = 0;
+int masterconnected = 0;
 
 int maxpvtorder = 0;
 
@@ -4028,6 +4029,7 @@ static int reload(void)
 		}
 	}
 	hasmaster = 0;
+	masterconnected = 0;
 	ctg = NULL;
 	while ((ctg = ast_category_browse(cfg, ctg)) != NULL) {
 		if (ctg == NULL) {
@@ -4438,19 +4440,13 @@ static void *voter_timer(void *data)
 						ast_log(LOG_WARNING, "Lost master timing client, disconnecting remaining clients.\n");
 						/* Reset the current active master flag for this client, since it disconnected. */
 						client->curmaster = 0;
+						masterconnected = 0;
 						for (client1 = clients; client1; client1 = client1->next) {
-							/*! \todo VE7FET Not sure if we need to implement this... have to see if there actually can
-							 * be more than one master client. Un-comment this to only disconnect clients
-							 * belonging to this node number.
-							 */
-							// if (client1->nodenum != client->nodenum) {
-							//		continue;
-							//	}
 							/* Only drop connections we have heard from (not all configured clients
 							 * in voter.conf).
 							 */
 							if (client1->heardfrom) {
-								ast_log(LOG_NOTICE, "Forcing disconnect of client: %s\n", client1->name);
+								ast_log(LOG_WARNING, "Forcing disconnect of client: %s\n", client1->name);
 								client1->heardfrom = 0;
 								client1->respdigest = 0;
 								client1->lastheardtime = ast_tv(0, 0);
@@ -4466,11 +4462,30 @@ static void *voter_timer(void *data)
 						continue;
 					}
 					for (client1 = client->next; client1; client1 = client1->next) {
+						/* Check our original client against the other clients, and see
+						 * if there is a client with the same IP and Port. If the client
+						 * we're checking against isn't connected, skip it. Otherwise, if
+						 * we find a client with the same IP and Port, dump both clients,
+						 * as that is not sane. The IP's could match (if they are behind
+						 * the same NAT, but the UDP port better not).
+						 */
 						if ((client1->sin.sin_addr.s_addr == client->sin.sin_addr.s_addr) && (client1->sin.sin_port == client->sin.sin_port)) {
+							/* If the client isn't connected/authenticated, skip */
 							if (!client1->respdigest) {
 								continue;
 							}
-							ast_log(LOG_NOTICE, "VOTER %u: Client %s disconnect (sanity)\n", client1->nodenum, client1->name);
+							/* If the IP's of both clients are 0, skip. This can happen on initial
+							 * start, and is a not a problem until everyone is fully connected
+							 */
+							if (!client1->sin.sin_addr.s_addr && !client->sin.sin_addr.s_addr) {
+								continue;
+							}
+
+							ast_debug(2, "Client %s IP: %s:%d = client %s IP: %s:%d\r\n", client->name,
+								ast_inet_ntoa(client->sin.sin_addr), ntohs(client->sin.sin_port), client1->name,
+								ast_inet_ntoa(client1->sin.sin_addr), ntohs(client1->sin.sin_port));
+							ast_log(LOG_ERROR, "Client %s and client %s have same IP and port! Resetting client connections (sanity)\n",
+								client->name, client1->name);
 							client->respdigest = 0;
 							client->heardfrom = 0;
 							client1->respdigest = 0;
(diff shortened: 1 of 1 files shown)


channels/chan_voter.c
```diff
@@ -4364,40 +4364,99 @@ static void voter_xmit_master(void)
...
+		/* Add a tick to the voter_timing_count every time we pass though. */
+		voter_timing_count++;
+
+		/* Cycle through our Asterisk channels, checking the status of our clients,
+		 * to make sure they are still sending data to us. Disconnect them if we
+		 * haven't heard from them after the timeout period. Additionally, if we lost
+		 * our master timing client, force a disconnect of all remaining clients that
+		 * were connected to that instance (they can't do anything if there is no master).
+		 */
+		for (p = pvts; p; p = p->next) {
```
mkmer, line 4468: I "think" we need a lock on the privates here -> they are accessed in mutliple threads and this list "could" be changed on the fly breaking the `for` loop.

I'm thinking this: https://github.com/AllStarLink/app_rpt/blob/4688f14943cd5859f6dead546ef24a8be6b9f1de/channels/chan_voter.c#L982-L985 
Can/will happen in the middle of any of the for loops looking at the pvts table?
Same may actually be true for the `clients` structures too?  Seems like they can be added/removed in a different thread.

channels/chan_voter.c
```diff
@@ -4492,15 +4571,15 @@ static void *voter_reader(void *data)
 
 	while (run_forever && !ast_shutting_down()) {
 		ast_mutex_unlock(&voter_lock);
-		ms = 50;
-		i = ast_waitfor_n_fd(&udp_socket, 1, &ms, NULL);
+		ms = 50;										 /* 50ms timeout */
+		i = ast_waitfor_n_fd(&udp_socket, 1, &ms, NULL); /* Open UDP socket with 50ms timeout. */
```
mkmer: We really don't "open" the socket here, just listen for the fd and/or timeout (right?)

channels/chan_voter.c
```diff
@@ -4539,110 +4621,444 @@ static void *voter_reader(void *data)
...
+				client->heardfrom = 0;
+				client->respdigest = 0;
+				continue;
+			} else {
+				/* Otherwise, we should be good to continue configuring the client.
+				 *
+				 * Set the flags we are going to send to the client for configuration.
+				 */
+				if (client->ismaster) {
+					authpacket.flags |= 2 | 8;
```
mkmer: Not sure, but "maybe"  `(2 | 8)` is more clear? It works either way

channels/chan_voter.c
```diff
@@ -5705,13 +6149,20 @@ static int load_module(void)
...
 	voter_thread_timer = ast_timer_open();
 	if (!voter_thread_timer) {
 		ast_log(LOG_ERROR, "Failed to open timer\n");
 		close(udp_socket);
 		return AST_MODULE_LOAD_DECLINE;
 	}
-	ast_timer_set_rate(voter_thread_timer, 50); /* 50 ticks per second = every 20ms */
+
+	/* Set voter_thread_timer for 50 ticks/second (every 20ms). */
+	ast_timer_set_rate(voter_thread_timer, 50);
```
mkmer: Maybe a #define for the "magic number" 50?  Comments are clear enough, but it's still a magic number.

mkmer (commented):
There are some Rabbit comments that seem to make sense as well (maybe).

ve7fet:
Yeah, I'm going to have to sift through these and do some compiling/testing. Most of the changes seem reasonable... just need to make sure they work in practice with talking to the hardware. :)

channels/chan_voter.c
```diff
@@ -4539,110 +4621,444 @@ static void *voter_reader(void *data)
...
 		isproxy = 0;
-		if (vph->digest) {
+
+		/* No valid Asterisk channel, do not respond to the client. This prevents the client
+		 * from "looping" through online/offline mode when there is no valid channel in app_rpt
+		 * to connect to.
+		 */
+		if (no_ast_channel) {
+			continue;
+		}
```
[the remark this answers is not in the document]
ve7fet: This is intended behaviour.

channels/chan_voter.c
```diff
@@ -4539,110 +4621,444 @@ static void *voter_reader(void *data)
...
+				proxy_authpacket.flags = authpacket.flags;
+				proxy_authpacket.vprox.ipaddr = sin.sin_addr.s_addr;
+				proxy_authpacket.vprox.port = sin.sin_port;
+				proxy_authpacket.vp.payload_type = htons(VOTER_PAYLOAD_PROXY);
+				sendto(udp_socket, &proxy_authpacket, sizeof(proxy_authpacket), 0, (struct sockaddr *) &psin, sizeof(psin));
+			} else {
+				authpacket.vp.payload_type = htons(VOTER_PAYLOAD_AUTH);
+				ast_debug(2, "Sending initial packet payload %i challenge %s digest %08x password %s to client %s\n",
+					authpacket.vp.payload_type, authpacket.vp.challenge, ntohl(authpacket.vp.digest), password,
+					((client) ? client->name : "UNKNOWN"));
```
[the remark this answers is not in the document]
ve7fet: This is intended debugging behaviour.

channels/chan_voter.c
```diff
@@ -4364,40 +4364,99 @@ static void voter_xmit_master(void)
...
+		 * timer handle (timingfd) to become ready. It should happen on
+		 * every "tick" (every 20ms)?
+		 */
 		int timeout = -1;
 		ast_waitfor_n_fd(&timingfd, 1, &timeout, NULL);
+
+		/* When the timer is ready, acknowledge it (one event only). This
+		 * timer should be used to keep all our audio in sync (ie for IAX2).
+		 */
 		if (ast_timer_ack(voter_thread_timer, 1) < 0) {
```
Allan-N: The doc for ast_timer_ack() says "This function should only be called if timer_get_event() returned AST_TIMING_EVENT_EXPIRED."

channels/chan_voter.c
```diff
@@ -4364,40 +4364,99 @@ static void voter_xmit_master(void)
...
+		 * master_time.vtime_sec from the system clock here. Otherwise,
+		 * master_time.vtime_sec will be set in voter_reader from the
+		 * timestamp embedded in the packets from the master client.
+		 *
+		 * Changed from using time() to using gettimeofday() so that all
+		 * our timing is consistent. The two functions may not return the
+		 * "same time", which could lead to random disconnects of mix mode
+		 * clients. See Issue 902.
+		 */
+		gettimeofday(&tv, NULL);
```
Allan-N, line 4447: I'm going to ask a question here, one time.  Do we want/need to change all of the gettimeofday() references to a call that returns a monotonic time (one that always increases over time even if someone adjusts the system data/time) ?

Kinda like res_usbradio.[ch]'s ast_radio_time() and ast_radio_tvnow()

and the time/place to use a monotonic time value is when you are looking at elapsed time.

channels/chan_voter.c
```diff
@@ -4406,10 +4465,30 @@ static void *voter_timer(void *data)
...
 							}
+							/* If the IP's of both clients are 0, skip. This can happen on initial
+							 * start, and is a not a problem until everyone is fully connected
+							 */
+							if (!client1->sin.sin_addr.s_addr && !client->sin.sin_addr.s_addr) {
+								continue;
+							}
+
+							ast_debug(2, "Client %s IP: %s:%d = client %s IP: %s:%d\r\n", client->name,
+								ast_inet_ntoa(client->sin.sin_addr), ntohs(client->sin.sin_port), client1->name,
```
Allan-N: The string returned by ast_inet_ntoa() is a fixed buffer.  All is good when you are making a single call but when you have multiple calls (like you do here) you are likely to only see 1 of the IP addresses.  You probably need to do something like :
```
client_ip = ast_strdupa(ast_inet_ntoa(client->sin.sin_addr); /* save the string so we can call ast_inet_ntoa again */
client1_ip = ast_inet_ntoa(client1->sin.sin_addr);
ast_debug(2, "Client %s IP: %s:%d = client %s IP: %s:%d\r\n",
    client->name, client_ip, ntohs(client->sin.sin_port),
    client1->name, client1_ip, ntohs(client1->sin.sin_port));

File: channels/chan_voter.c [hunk 10]
```diff
@@ -4518,19 +4597,22 @@ static void *voter_reader(void *data)
 				p->lastwon = NULL;
 			}
 		}
+		/*! \todo VE7FET this should not be needed, we bail if i == -1 above? */
```
Allan-N: Yes, you bail above if `i == -1` but what if the value is -2, -3, -4?  Maybe change the test above?

But, per your comment, if we can't get here then it's just dead code and can be removed.

channels/chan_voter.c
Allan-N: Seems like this variable, `i`,  is poorly named.  If you are up to a few extra changes, maybe we should rename this to be `fd` ? :)

ve7fet: Yeah... this whole bit of code is a bit messy.

I added some comments as I was trying to figure out what was going on.

Indeed `i` is more of an index or yes, the `fd` when the socket has data (or it is set to `-1` on error or `0` on timeout).

There is definitely some cleanup that can happen here.

I'm open to better ways to clean it up and improve the error handling.. :)

channels/chan_voter.c
ve7fet: I think `i` can only be `-1` on error, `0` on timeout (no data received on the UDP port), or the `fd` if there is data waiting.

So the check for `< 0` seems to be the same as the check for `== -1`... but we never actually check for a timeout (and skip)?

Do we need to test for `== 0`? Do we care? Other than checking for an actual error (`-1`), we are just spinning our wheels until the `fd == udp_socket` to indicate we have data to process.

channels/chan_voter.c
ve7fet: This sounds like a future issue to resolve on the next iteration. :)

No doubt, it should probably be looked at. In most cases (I'd have to dig in and check), I think we're using the timestamp to keep track of when we've hear from or sent data to connected clients.

It MAY affect which packets we choose to vote on, have to look deeper at that.

channels/chan_voter.c
ve7fet: So, do we even need this test then? I was just adding comments around code blocks. :)

channels/chan_voter.c
ve7fet: Will revised this and clean it up.

channels/chan_voter.c
ve7fet: Will add this in and add some comments.

channels/chan_voter.c
ve7fet: Yeah... this code is a bit of a mess. Needs some more cleanup.

channels/chan_voter.c
ve7fet: Yes, I know a comprehensive review of the locks is an outstanding issue. It may make sense to review it under a different PR?

channels/chan_voter.c
Allan-N: The code we're using in chan_echolink is :
```
    if (ast_timer_get_event(p->timer) == AST_TIMING_EVENT_EXPIRED) {
        if ((ast_timer_ack(p->timer, 1) < 0)) {
            ast_log(LOG_WARNING, "Timer ack failed. \n");
            return NULL;
        }
    }
```
I would suggest something similar for chan_voter.

channels/chan_voter.c
Allan-N: What caught my attention was the use of gettimeofday().  Yes, this is probably something to be looked at.  If now, fine.  If later, that's OK too.  I just wanted to ask the question.

channels/chan_voter.c
Allan-N: But, `i` is not being used as an index.  It is really the returned file descriptor number (all valid FD's are >= 0 ... with negative values used for "not set" or "invalid").  Do the variable name need to be changed?  Of course not.  Would changing the name improve the readability of the code?  IMHO, yes (I was trying to understand what the value represented).

channels/chan_voter.c
Allan-N: If I'm reading the ast_waitfor_n_fd() code correctly, it looks like the return value will either be the "fd" or  -99999.

And, upon further digging, it's looking like the earlier `(i == -1)` block is the one that will never be executed and it's this block that is needed.  In fact, the "Unable to open VOTER UDP socket" error reported/logged by the earlier block is not accurate/relevant.

channels/chan_voter.c
Allan-N: Just a thought :
```
enum voter_auth_flags {
    /*! \brief Send flat audio (nodeemp or hostdeemp) */
    FLAG_FLATAUDIO = (1 << 0),
    /*! \brief Send audio always (master) */
    FLAG_SENDALWAYS = (1 << 1),
    /*! \brief Do not filter CTCSS (noplfilter) */
    FLAG_NOCTCSSFILTER = (1 << 2),
    /*! \brief Master timing source (do not delay sending audio packet) (master) */
    FLAG_MASTERTIMING = (1 << 3),
    /*! \brief Use ADPCM rather than ulaw (adpcm) */
    FLAG_ADPCM = (1 << 4),
    /*! \brief Request "mix" option to host (mixminus) */
    FLAG_MIX = (1 << 5),
};
```
and then replace all of the 1, 2, 4, 8, 16, 32 values with the FLAG_xxx names :)

channels/chan_voter.c
```diff
@@ -4518,19 +4597,22 @@ static void *voter_reader(void *data)
...
 		if (i < 0) {
 			continue;
 		}
-		/* Is there activity on our UDP socket? */
+		/* Is there activity on our UDP socket? Do nothing (continue) if the waitfor fd above returned 0. */
 		if (i != udp_socket) {
```
Allan-N: ast_waitfor_n_fd() is either going to return the "fd" you are only watching/waiting for data on (because data is available) or a negative value.  Since we've already checked for the "< 0" case then this test can be dropped as "fd" should always be "udp_socket".

Now, if you were watching multiple FD's the you'd need to see which FD was ready for processing ... but you're only watching the 1 file/socket.

channels/chan_voter.c
```diff
@@ -4539,110 +4621,444 @@ static void *voter_reader(void *data)
 			sin.sin_port = htons(master_port);
 		}
 		isproxy = 0;
-		if (vph->digest) {
+
+		/* No valid Asterisk channel, do not respond to the client. This prevents the client
```
Allan-N: This is a stylistic comment (ignore/adopt as you please).

You have the following where you are describing an action that is being taken before the condition is checked :
```
		/* No valid Asterisk channel, do not respond to the client. This prevents the client
		 * from "looping" through online/offline mode when there is no valid channel in app_rpt
		 * to connect to.
		 */
		if (no_ast_channel) {
			continue;
		}
```
Would it be a bit clearer to use :
```
		if (no_ast_channel) {
			/* No valid Asterisk channel, do not respond to the client. This prevents the client
			 * from "looping" through online/offline mode when there is no valid channel in app_rpt
			 * to connect to.
			 */
			continue;
		}
```
Here, we've made the test and the comment reflects the condition we HAVE observed and the action we ARE taking.

Another option would be for the pre "if ()" documentation to say something like :
```
		/* Check to see if we have a valid Asterisk channel.  If not, we do not want to respond
		 * to the client. By not responding we prevent the client from "looping" through online/offline
		 * mode when there is no valid channel in app_rpt to connect to.
		 */
```
Here, the comment describes the following block, both the test and the action

channels/chan_voter.c
```diff
@@ -4539,110 +4621,444 @@ static void *voter_reader(void *data)
...
+						if (!logged_buflen_too_small) {
+							ast_log(LOG_ERROR, "VOTER %u: Mix-mode client %s (proxy) rejected: buflen=%d (<160). Fix voter.conf.\n",
+								client->nodenum, client->name, client->buflen / 8);
+							logged_buflen_too_small = 1;
+						}
+						client->mix = 0;
+						client->heardfrom = 0;
+						client->respdigest = 0;
+						continue;
+					} else {
```
Allan-N: Can we drop the `else {` ?
```
					if (client->buflen < (FRAME_SIZE * 8)) {
						...
						continue;
					}

					client->mix = 1;
					ast_log(LOG_NOTICE, "VOTER %u: Client %s is sending mix mode flag, setting client to mix mode\n",
					client->nodenum, client->name);
					logged_buflen_too_small = 0;
```

channels/chan_voter.c
```diff
@@ -4539,110 +4621,444 @@ static void *voter_reader(void *data)
...
+			if (!client->mix && !hasmaster) {
+				ast_log(LOG_WARNING, "VOTER %u: Client %s attempting to authenticate as GPS-timing-based with no master timing source defined!!\n",
+					client->nodenum, client->name);
+				/* Reject the connection. */
+				ast_log(LOG_WARNING, "VOTER %u: Client %s disconnect (forced)\n", client->nodenum, client->name);
+				authpacket.vp.digest = 0;
+				client->heardfrom = 0;
+				client->respdigest = 0;
+				continue;
+			} else {
```
Allan-N: With the `continue;`, can we drop the `else {` ?

channels/chan_voter.c
```diff
@@ -4539,110 +4621,444 @@ static void *voter_reader(void *data)
...
+					authpacket.flags |= 4;
+				}
+			}
+
+			/* The sin structure has the IP info received off the wire for the current packet from
+			 * the client. Now that we've validated the client (since we have a valid vph->digest),
+			 * update the client's sin structure with this information.
+			 */
+			client->sin = sin;
+			/*! \todo VE7FET not sure why this is here/needed, when it is dealing with proxy stuff, it probably needs to be moved */
```
Allan-N: Is the memset() saying "we are not using a proxy so clear any proxy IP" ?

channels/chan_voter.c
```diff
@@ -4539,110 +4621,444 @@ static void *voter_reader(void *data)
...
+			/* Timestamp when we last heard this client (system time). */
+			client->lastheardtime = tv;
+
+			/* Make our response digest based on the challenge sent by the client, and our host password,
+			 * and put that in the packet header.
+			 */
+			authpacket.vp.digest = htonl(crc32_bufs((char *) vph->challenge, password));
+
+			/* Do the same for proxy authentication packets. */
+			proxy_authpacket.vp.curtime.vtime_sec = htonl(tv.tv_sec);
```
Allan-N, line 4968: Should these lines go into the `if (isproxy) {}` block below?

channels/chan_voter.c
```diff
@@ -4539,110 +4621,444 @@ static void *voter_reader(void *data)
...
+		/* Go through all the clients, and stop when we find a client the has client->digest set
+		 * (from voter.conf) that matches the digest we received on the wire (vph->digest).
+		 *
+		 * When we find a match, we'll also update the associated client->lastheard time for that
+		 * client with the current timestamp, and set client->heardfrom to true.
+		 *
+		 * Upon initial connect from a client, vph->digest will be 0, so this won't match any
+		 * configured client in voter.conf, triggering the authentication process.
+		 */
+		gettimeofday(&tv, NULL);
```
Allan-N, line 4697: We called gettimeofday() a few lines back (l.4583).  Has enough time elapsed that we need to capture the time again?  or is this call redundant?

channels/chan_voter.c
```diff
@@ -4943,44 +5367,62 @@ static void *voter_reader(void *data)
...
 											continue;
 										}
+										/* If the IP's of both clients are 0, skip. This can happen on initial
+										 * start, and is a not a problem until everyone is fully connected
+										 */
+										if (!client1->sin.sin_addr.s_addr && !client->sin.sin_addr.s_addr) {
+											continue;
+										}
+
+										ast_debug(2, "Client %s IP: %s:%d = client %s IP: %s:%d\r\n", client->name,
```
Allan-N: The string returned by ast_inet_ntoa() is a fixed buffer. All is good when you are making a single call but when you have multiple calls (like you do here) you are likely to only see 1 of the IP addresses.

channels/chan_voter.c
ve7fet: Yes, I like that. Will make some changes to improve code readability for flag usage.

channels/chan_voter.c
ve7fet: Yeah, something along those lines. Coderabbit indicates it is to clear out any potentially stale IP info in that proxy structure when a client connects/reconnects.

channels/chan_voter.c
ve7fet: The `authpacket.vp.digest` for sure, no.

The `proxy_authpacket.vp...` stuff, quite probably. However, I am not tackling that in this update. It currently compiles and seems to work "as-is", so I don't want to risk breaking it worse that it may already be broken.

TBT, the whole proxy server implementation is mind-bending, and I need to spend some more time trying to actually draw out the state machine for how it all (apparently) works. That said, I have also been contemplating removing all the proxy server stuff entirely. I haven't been able to come up with a good reason for it to even exist... every time I think about it, if your network is broken to the point where it needs to "fail over"... you probably have bigger issues to deal with first!

But, I digress... that is a conversation for another day, and some community feedback.

channels/chan_voter.c
ve7fet: Yes, it looks like it probably needs to stay, per coderabbit analysis:

"Line 4645 now records the earlier tv from Line 4583, before `recvfrom()` and client matching. Since `lastheardtime` drives client/master liveness checks, this can skew timeout behavior under processing delay."

channels/chan_voter.c
```diff
@@ -4518,19 +4597,22 @@ static void *voter_reader(void *data)
...
 		if (i != udp_socket) {
 			continue;
 		}
+		/* Get the data from the UDP socket. */
 		fromlen = sizeof(struct sockaddr_in);
 		recvlen = recvfrom(udp_socket, buf, sizeof(buf) - 1, 0, (struct sockaddr *) &sin, &fromlen);
-		/* If set got something worthwhile. */
+		/* Skip if we got less than a header's worth of data. */
 		if (recvlen < sizeof(VOTER_PACKET_HEADER)) {
 			continue;
```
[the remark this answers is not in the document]
ve7fet: Change applied in 6f15b6726429c1f1178a6e1734238eae7bd85e30

channels/chan_voter.c
```diff
@@ -4539,110 +4621,444 @@ static void *voter_reader(void *data)
...
+		 * This only runs once for each client, in response to an auth packet with a received digest of 0.
+		 */
+		if (!client && (ntohs(vph->payload_type) == VOTER_PAYLOAD_AUTH) && !ntohl(vph->digest)) {
+			memset(&authpacket, 0, sizeof(authpacket));
+			memset(&proxy_authpacket, 0, sizeof(proxy_authpacket));
+
+			/* Our unique challenge is created in load_module. Copy our challenge into
+			 * the packet header.
+			 */
+			strcpy((char *) authpacket.vp.challenge, challenge);
```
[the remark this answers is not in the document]
ve7fet: Change applied in 2374bd20959ef0caa44594a30cb2c26ae40ff2ae

channels/chan_voter.c
```diff
@@ -4674,23 +5090,31 @@ static void *voter_reader(void *data)
...
-			/* If we know the client, find the connection that the audio belongs to and send it there. */
+
+			/* If we have a valid (authenticated) client, have recently heard from it, and it sent
+			 * us a valid audio or proxy packet, find the corresponding Asterisk channel and
+			 * send it there.
+			 */
 			if (client && client->heardfrom &&
 				(((ntohs(vph->payload_type) == VOTER_PAYLOAD_ULAW) && (recvlen == (sizeof(VOTER_PACKET_HEADER) + FRAME_SIZE + 1))) ||
 					((ntohs(vph->payload_type) == VOTER_PAYLOAD_ADPCM) && (recvlen == (sizeof(VOTER_PACKET_HEADER) + FRAME_SIZE + 4))) ||
 					(ntohs(vph->payload_type) == VOTER_PAYLOAD_PROXY))) {
```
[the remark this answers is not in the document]
ve7fet: Change applied in f1ebd3d55673214598528eea7128d71863832ef9

channels/chan_voter.c
ve7fet: Change applied in 56fea3ccb85c69d334fd6e9612a6e4766b041de1

channels/chan_voter.c
ve7fet: Refactored in c31d0b1040e7d86eeb0888ea9a38881ef4d74a03

channels/chan_voter.c
```diff
@@ -4364,40 +4364,99 @@ static void voter_xmit_master(void)
...
 		int timeout = -1;
 		ast_waitfor_n_fd(&timingfd, 1, &timeout, NULL);
+
+		/* When the timer is ready, acknowledge it (one event only). This
+		 * timer should be used to keep all our audio in sync (ie for IAX2).
+		 */
 		if (ast_timer_ack(voter_thread_timer, 1) < 0) {
 			ast_log(LOG_ERROR, "Failed to acknowledge timer\n");
 			break;
 		}
```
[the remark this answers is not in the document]
ve7fet: Refactored in c21aa14733ff2d8b958f021c63f7dee6aede3a40

channels/chan_voter.c
ve7fet: Refactored in c21aa14733ff2d8b958f021c63f7dee6aede3a40

channels/chan_voter.c
ve7fet: Refactored in 728842cb415d6ba667d5313e22464b4bf7b7c990

channels/chan_voter.c
ve7fet: Refactored in 728842cb415d6ba667d5313e22464b4bf7b7c990

channels/chan_voter.c
ve7fet: Removed the test in 728842cb415d6ba667d5313e22464b4bf7b7c990 as the state is implicit

channels/chan_voter.c
ve7fet: Revised the comment with more context in cbb9082cf2ce825cbebe4079d39e329b517bcfd1

channels/chan_voter.c
ve7fet: Revised in 25a8aff76d6354c97664e0b97cb17c3f702098fa

channels/chan_voter.c
ve7fet: Revised in 25a8aff76d6354c97664e0b97cb17c3f702098fa

channels/chan_voter.c
ve7fet: Removed in 4e09c7250e46e53c22b16de3fab0303ae666939b

channels/chan_voter.c
ve7fet: Removed in 4e09c7250e46e53c22b16de3fab0303ae666939b

channels/chan_voter.c
```diff
@@ -3848,6 +3868,8 @@ static int reload(void)
 
 	if (!(cfg = ast_config_load(config, zeroflag))) {
 		ast_log(LOG_ERROR, "Unable to load/reload config %s\n", config);
+		close(udp_socket);
```
Allan-N: Looks like we have LOTS of places in reload() where we have:
```
		close(udp_socket);
		udp_socket = -1;

```
Seems like we can remove all of this up to the [only] caller of reload().

Essentially, go with the opposite of _We won't close the UDP socket here, because reload() will do that if it fails, avoiding a double-close of the socket_.

p.s. you might also consider calling ast_mutex_lock/unlock() before/after calling reload() and removing the one lock and many unlocks inside the function

channels/chan_voter.c
```diff
@@ -4364,40 +4397,100 @@ static void voter_xmit_master(void)
...
-		ast_waitfor_n_fd(&timingfd, 1, &timeout, NULL);
-		if (ast_timer_ack(voter_thread_timer, 1) < 0) {
-			ast_log(LOG_ERROR, "Failed to acknowledge timer\n");
-			break;
+		/* Check and acknowledge our thread timer. This timer keeps our audio in sync (for IAX2). */
+		{
+			int timer_fd = ast_timer_fd(voter_thread_timer);
+			int timeout_ms = -1; /* block until the timer fires */
+
+			if (ast_waitfor_n_fd(&timer_fd, 1, &timeout_ms, NULL) < 0) {
```
Allan-N: Please review all places in the code where ast_waitfor_n_fd() is used.  I know the asterisk header is not clear but the code shows this function will return one of the FD's you are waiting on (in this code that tends to be the one/only socket) or a timeout.  Also, if you look deeper it looks as if the timeout return value is -99999.  I don;t think you would get -1 some cases and -99999 in others.  As for testing the return value, I would suggest a simple `fd < 0` to be safe indication of "timeout or some other/unexpected error.

https://github.com/asterisk/asterisk/blob/5eded44fee3a3e33117d75f275ef84c53cf05249/include/asterisk/channel.h#L2032-L2067

https://github.com/asterisk/asterisk/blob/5eded44fee3a3e33117d75f275ef84c53cf05249/main/channel.c#L2984-L2989

https://github.com/asterisk/asterisk/blob/5eded44fee3a3e33117d75f275ef84c53cf05249/main/channel.c#L3009-L3011

channels/chan_voter.c
ve7fet: Issue #1094 created to track this.

channels/chan_voter.c
ve7fet: Issue #1095 created to track.

channels/chan_voter.c
ve7fet: The questions I have are (since I don't know enough about how this I/O path works):

1) What does `ast_waitfor_n_fd` return for a timeout, if -99999 is an "error"? Does it return `0`? Or, is a timeout also considered an error?

2) Do we need to do anything on a timeout? Wouldn't a timeout potentially be a normally occurring event (if nothing was connected)?

I agree, catching and handling an error (presumably with `fd < 0`) would be good... I just don't want to inadvertently bail on a timeout that could be part of normal operations.

The confusing part is that the original code was trying to catch `fd == -1` as an error... which shouldn't exist, and if it found `fd < 0` (which is presumably the actual error condition), it just skipped (continued) out of the loop. So, what is the right way to handle the error?

channels/chan_voter.c
ve7fet: Yes, I can review and revise how that is all being done.

channels/chan_voter.c
Allan-N: For this particular case, you've specified a timeout value of "-1" (block until the timer fires) so there's really no timeout.  The function is going to return the timer_fd (the only FD it's watching) or an error (if I read the asterisk code correctly, -99999).

Looking more at THIS case, I think I was just questioning whether "failed to wait" made sense. Thinking about it a bit more, it's reasonable.

But, I did see other places where ast_waitfor_n_fd() was used with different (and incorrect) handling of the returned value (e.g. rad_rxwait, voter_primary_client)

ve7fet: Sorry, yes... for the timer code you're going to block and wait for the timer to fire, or get a `-99999` back on an error. That is more straight forward.

I was thinking more of the situation in `voter_primary_client` and `voter_reader`, where you are waiting on the UDP port with a timeout.

channels/chan_voter.c
mkmer: I was just commenting about the comment :)  
Looking at this again, maybe we don't need `ast_waitfor_n_fd()`.
Should we either be using `ast_waitfor_nandfds()` or used the "base" fd polling api?  (I'm not sure)
Probably something to look at in a different PR...

channels/chan_voter.c
```diff
@@ -419,6 +422,23 @@ char context[100];
 #define VOTER_PAYLOAD_PING 5
 #define VOTER_PAYLOAD_PROXY 0xf000
 
+/* Define voter packet flags. */
+enum voter_auth_flags {
```
mkmer, line 426: Shouldn't the `.flags` change from char to enum as well?
https://github.com/AllStarLink/app_rpt/blob/64aa743ed7602aee659fde8dd712097750eff129/channels/chan_voter.c#L2944

mkmer (approved):
Looks good.

channels/chan_voter.c
Allan-N: >Should we either be using ast_waitfor_nandfds()

ast_waitfor_nandfds() is the do-it-all API that allows you to wait for channels, FDs, and timeouts.  ast_waitfor_n_fd() uses the same API but leaves out the channels.

channels/chan_voter.c
ve7fet: Revised all the locking in 991a3f80ed8235a452d333c6129c2f03cf9c1ad9 but forgot to clean up the locks... cleaned up the locks in de72184634b2e437312419366e880984aeb93eeb

channels/chan_voter.c
```diff
@@ -5701,22 +6168,41 @@ static int load_module(void)
...
+	/* Set voter_thread_timer for 50 ticks/second (every 20ms). */
+	ast_timer_set_rate(voter_thread_timer, VOTER_TICKS);
+
+	/* Load the rest of the values from the config file by running reload.
+	 * reload() will return 0 on success, or a non-zero value on failure.
+	 * If it returns non-zero, we will abort loading the module.
+	 */
+	ast_mutex_lock(&voter_lock);
 	if (reload()) {
+		ast_log(LOG_ERROR, "Failed to reload configuration\n");
```
Allan-N, line 6237: You also need to call ast_mutex_unlock() if reload() returns an error?

channels/chan_voter.c
Allan-N: In `voter_primary_client()` you currently have :
```
        i = ast_waitfor_n_fd(&pri_socket, 1, &ms, NULL);
        ast_mutex_lock(&voter_lock);
        if (i == -1) {
            ast_mutex_unlock(&voter_lock);
            ast_log(LOG_ERROR, "VOTER %i: Error in select()\n", p->nodenum);
            pthread_exit(NULL);
        }
```
Here, the function will either return with the `pri_socket` FD value or a negative number if there was a timeout.  THE CODE IS FINE AS-IS but (for clarity) "I" would rewrite this as :
```
        fd = ast_waitfor_n_fd(&pri_socket, 1, &ms, NULL);
        if (fd < 0) {
            ast_log(LOG_ERROR, "VOTER %i: no data from primary connection\n", p->nodenum);
            pthread_exit(NULL);
        }

        ast_mutex_lock(&voter_lock);
```

... and the `voter_reader()` code looks the same.

channels/chan_voter.c
```diff
@@ -3840,15 +3860,13 @@ static int reload(void)
 	struct ast_config *cfg = NULL;
 	struct ast_variable *v;
 
-	ast_mutex_lock(&voter_lock);
```
mkmer, line 3843: Why do we no longer need a lock here?  Reload is running in a separate thread while the "other" thread is running and could change the `client` or the `p` list (Or I'm lost again)

channels/chan_voter.c
```diff
@@ -4364,40 +4374,98 @@ static void voter_xmit_master(void)
...
+	char *client_ip;
+	char *client1_ip;
 
 	while (run_forever && !ast_shutting_down()) {
-		int timeout = -1;
-		ast_waitfor_n_fd(&timingfd, 1, &timeout, NULL);
-		if (ast_timer_ack(voter_thread_timer, 1) < 0) {
-			ast_log(LOG_ERROR, "Failed to acknowledge timer\n");
+		/* Check and acknowledge our thread timer. This timer keeps our audio in sync (for IAX2). */
+		int timer_fd = ast_timer_fd(voter_thread_timer);
```
mkmer, line 4416: ast_timer_fd() can fail with <0 in the `timer_fd` -> should we do something about it?
I think `ast_witfor_n_fd()` will crash if we give it a -1 fd

channels/chan_voter.c
```diff
@@ -4364,40 +4374,98 @@ static void voter_xmit_master(void)
...
+		int timeout_ms = -1; /* block until the timer fires */
+
+		if (ast_waitfor_n_fd(&timer_fd, 1, &timeout_ms, NULL) < 0) {
```
mkmer, line 4424: Not saying we need or should change this, but why `ast_waitfor_n_fd()` when we know it's only and fd?

channels/chan_voter.c
```diff
@@ -4492,15 +4585,20 @@ static void *voter_reader(void *data)
 
 	while (run_forever && !ast_shutting_down()) {
 		ast_mutex_unlock(&voter_lock);
-		ms = 50;
-		i = ast_waitfor_n_fd(&udp_socket, 1, &ms, NULL);
+		timeout_ms = 50;										  /* 50ms timeout */
+		fd = ast_waitfor_n_fd(&udp_socket, 1, &timeout_ms, NULL); /* Poll the UDP socket, looking for data */
```
mkmer, line 4625: same question here: we know it's an fd ...

channels/chan_voter.c
Allan-N: This was my kinda-bad.  When looking at the code I noticed that reload() was grabbing the lock and then letting go before every return.  I also only found a single call to `reload()`.  What I missed was the asterisk module load/unload reference at the end of the file :
```
    .reload = reload,
```

But, rather then putting back all of the lock/unlock calls I might suggest we add :
```
static int astersk_reload(void)
{           
    ast_mutex_lock(&voter_lock);
    if (reload()) { 
        ast_log(LOG_ERROR, "Failed to reload configuration\n");
        close(udp_socket);
        udp_socket = -1;
    }
    ast_mutex_unlock(&voter_lock);
}
```
and then we'd change :
```
    .reload = asterisk_reload,

```

channels/chan_voter.c
Allan-N: The ast_timer provides its indication using an FD, correct?  and, of the available APIs, ast_waitfor_n_fd() is the most focused on waiting for that indication.  or are we misunderstanding?

mkmer: I was thinking we would/should use the poll() function, since this is not channel related. https://github.com/AllStarLink/app_rpt/blob/66ce4054b57282e800540c538adbc6c10f789363/channels/chan_echolink.c#L3872
OR
```

int fd = ast_timer_fd(timer);

struct pollfd pfd = { .fd = fd, .events = POLLIN };

if (poll(&pfd, 1, timeout_ms) > 0) {
    ast_timer_ack(timer, 1);
}
```

channels/chan_voter.c
mkmer: That looks good to me, and easier to get it "right" and keep it "right" :)

channels/chan_voter.c
Allan-N: ast_waitfor_n_fd() ends up calling ast_waitfor_nandfds() that ends up calling ast_poll().

mkmer: I get that... but it's the long way to it with a bunch of code we don't need ;)
As I mentioned, maybe it's ok, but I don't think it's the "asterisk" way for this.
Or we can just "leave it"

mkmer: Oh and CoPilot thinks this:
⚠️ But here’s the important nuance
❗ 1. It’s overkill if you only have a timer
If you’re only waiting on a timer, don’t do this:
`waitfor_n_fd(NULL, 0, ...)`
 Use `poll()` instead — it’s simpler and clearer.
 
 :)

Allan-N: Well, I would use ast_poll() instead of poll() ... 'cause that would be the asterisk way :)

Allan-N: Oh, and `waitfor_n_fd(NULL, 0, ...)` isn't waiting on a timer.  That would just be waiting for a timeout.

channels/chan_voter.c
ve7fet: Added a reload wrapper in 2dd1023d3e2d001c1e27248ea7ecefb604c6fe41 with some tweaks suggested by the rabbit to send a defined return value back to Asterisk core.

channels/chan_voter.c
ve7fet: I'm all for a better way to do this... but we're reaching the limits of my C/Asterisk knowledge. :)

All the `ast_waitfor_n_fd()` were Jim's original implementation. If we can clean up and simplify the calls to (potentially) reduce overhead and complexity, great.

Allan-N: "I" think it's OK to continue using ast_waitfor_n_fd().  We're now past the confusion of `== -1` vs `< 0`.

channels/chan_voter.c
ve7fet: Missing unlock added in 1738b6afcea16bf9553eee008b07a14f174c5635

Allan-N (approved):

channels/chan_voter.c
ve7fet: > int fd = ast_timer_fd(timer);
> 
> struct pollfd pfd = { .fd = fd, .events = POLLIN };
> 
> if (poll(&pfd, 1, timeout_ms) > 0) {
>     ast_timer_ack(timer, 1);
> }

So, I tried implementing something along these lines, adding in some error handling, however coderabbit analysis indicated that `ast_poll()` (and `poll()` for that matter) don't handle an interrupt signal (`EINTR`) gracefully, leading to more complicated code testing for an retrying on `EINTR` and `EAGAIN`. Apparently, `ast_waitfor_n_fd()` handles `EINTR` more gracefully, by returning a simulated timeout instead.

channels/chan_voter.c
```diff
@@ -2939,22 +2966,30 @@ static void *voter_primary_client(void *data)
 	p->primary_challenge[0] = 0;
 	while (run_forever && !ast_shutting_down() && !p->kill_primary_thread) {
 		ast_mutex_unlock(&voter_lock);
-		ms = 100;
-		i = ast_waitfor_n_fd(&pri_socket, 1, &ms, NULL);
+		ms = 100; /* 100ms timeout */
```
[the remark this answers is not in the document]
ve7fet: Fixed.

ve7fet:
Ok, went back and worked on the `voter_primary_client` and `voter_reader` threads to better handle (and fix) the handling of `ast_waitfor_n_fd`. 

I think I have it all straight now, and am properly handling the error conditions.

I've also removed an extraneous `if` from the `voter_primary_client` function, to keep it consistent with `voter_reader`, and added some comments on the return values for the `rad_rxwait` function.

See the changes in eb199e5c3d5d0f5ce17f7976cf20b59f8ce38273

mkmer (approved):

channels/chan_voter.c
```diff
@@ -1393,25 +1413,32 @@ static struct ast_frame *ast_frcat(const struct ast_frame *restrict f1, const st
...
- * \brief Poll the specified fd for input for the specified milliseconds.
+ * \brief Poll the specified Asterisk CLI fd for input for the specified milliseconds.
  *
  * Used exclusively with the "voter display" CLI command to refresh the display.
  *
- * \param fd			File descriptor.
+ * \param fd			Asterisk CLI file descriptor.
  * \param ms			Milliseconds to wait.
- * \return  -1, 1, 0    Needs to be defined.
+ * \return  -1, 1, 0	Failure, key pressed (exit), timeout (refresh CLI display).
```
Allan-N, line 1427: No.  The possible return values are `>= 0` (the FD you were waiting on) or `< 0` (a timeout or unlikely error).

channels/chan_voter.c
```diff
@@ -1393,25 +1413,32 @@ static struct ast_frame *ast_frcat(const struct ast_frame *restrict f1, const st
...
 	x = ast_waitfor_n_fd(&fd, 1, &myms, NULL);
-	if (x == -1) {
+	/* This is the error path, if ast_waitfor_n_fd fails (it returns a negative value). */
+	if (x < 0) {
 		return -1;
 	}
+	/* When the fd is returned, a key has been pressed in the CLI, and we exit
+	 * the voter_display.
+	 */
 	if (x == fd) {
```
Allan-N: Here, there is no need for the `if (x == fd)`.  If ast_waitfor_n_fd() returns a positive value it WILL be the fd you were watching ... and "0" is a valid FD value.

channels/chan_voter.c
```diff
@@ -2939,22 +2966,30 @@ static void *voter_primary_client(void *data)
...
-			ast_log(LOG_ERROR, "VOTER %i: Error in select()\n", p->nodenum);
-			pthread_exit(NULL);
+		/* Check the returned fd and see if there is a datagram ready to process.
+		 * fd will be positive (and equal to pri_socket) if there is activity on the UDP socket.
+		 * If fd is negative, there was an error, timeout, or no activity, and we should skip (continue) and
+		 * wait (try again).
+		 *
+		 * Note: fd can never be 0, since we are only waiting on one socket (pri_socket), and that socket is always positive.
+		 */
+		/* ast_waitfor_n_fd() returns -1 on both timeout and error; use errno to tell them apart. */
```
Allan-N: I would probably leave out any mention of specific negative return values.  I would also stray away from mentions of errno because if you did want to follow that path you'd need to set errno before calling ast_waitfor_n_fd and then check the value after.

channels/chan_voter.c
```diff
@@ -2939,22 +2966,30 @@ static void *voter_primary_client(void *data)
...
-		if (i == -1) {
-			ast_mutex_unlock(&voter_lock);
-			ast_log(LOG_ERROR, "VOTER %i: Error in select()\n", p->nodenum);
-			pthread_exit(NULL);
+		/* Check the returned fd and see if there is a datagram ready to process.
+		 * fd will be positive (and equal to pri_socket) if there is activity on the UDP socket.
+		 * If fd is negative, there was an error, timeout, or no activity, and we should skip (continue) and
+		 * wait (try again).
+		 *
+		 * Note: fd can never be 0, since we are only waiting on one socket (pri_socket), and that socket is always positive.
```
Allan-N: NO.  FD values are >= 0 ... and, yes, 0 is a valid FD number.

channels/chan_voter.c
Allan-N: and that's likely one of the reasons that the ast_waitfor functions exist ... to hide the complexity.

But, when you hide the complexity you sometimes lose the details.  Our current usage of ast_waitfor_n_fd() does not pass the last "exception" argument ... and from the code it looks like this is how you'd find out if the < 0 case was a timeout or an error.

Given where we are, I'd keep using ast_waitfor_n_fd().  If we really care about the difference between "no data / timeout" and "error" then we should pass the `exception` variable and pay attention to the value passed back.

channels/chan_voter.c
mkmer: I don't think this guy returns an FD anyway, it returns 0 if it's a timeout (no fd's ready), or it returns a positive number that is the number of "ready" FD's -> since we only give it one, it should never be >1.  Seems like >0 would be fine here.  
Something llke this:
```
int res = ast_waitfor_n_fd(fds, nfds, &ms, NULL);
if (res > 0) {
// FD is ready
printf("Ready FD: %d\n", outfd);
} else if (res == 0) {
// timeout
} else {
 // error
}
```

Allan-N: Have I been reading this )poorly documented) API and code wrong this whole time?  Oops Sigh

channels/chan_voter.c
Allan-N: Ummm. I may have misread the api/code

channels/chan_voter.c
Allan-N: Ummm. I may have misread the api/code

channels/chan_voter.c
ve7fet: So, this `rad_rxwait()` function isn't all that intuitive in what it does and how it functions, so I threw some debug messages in it to print to the console when I ran it.

This gets invoked when you run `voter display <instance>` from the Asterisk CLI. It shows the live status of all the VOTER clients, with their signal strength, updating/refreshing the display every 100ms, until you hit "any" key, when it returns back to the normal console (and squirts out all the pending console logs).

When it is sitting there refreshing every 100ms, we hit the `return 0` (I had a console log entry every time it looped and hit line 1442). Presumably, `x` must have been `0`, or some other positive value, that wasn't the fd to get there to `return 0` back to `voter_display` and go for another spin at updating the display.

When I press "any" key to exit, it hit the `return 1`, confirmed by my console logging, so `x` must have evaluated to whatever the fd was passed in from `voter_display`. That triggers the `break` in `voter_display` to stop the display, since `return 1` now satisfied the `if` in `voter_display`.

That just left the `return -1` case, which we shouldn't reach under normal circumstances anyways, where `ast_waitfor_n_fd` would return a negative value. All that would do is also cause a break/exit in `voter_display`.

The confusing part in this function is not using the results from `ast_waitfor_n_fd` directly, and sending back custom returns.

channels/chan_voter.c
ve7fet: That's an leftover from some other testing... I'll clean that up (testing for errno conditions didn't work as intended).

channels/chan_voter.c
ve7fet: Yeah those return values are the return values of the function, not `ast_waitfor_n_fd`. :)

channels/chan_voter.c
ve7fet: Resolved in 126ab2e7c7ab16de138c7821580adcf793f1ec4b

channels/chan_voter.c
ve7fet: Error checks have been reviewed and revised.

channels/chan_voter.c
ve7fet: Resolved in c5585fe7fe7e8b43a3362ca4c237d1b30f13991e

channels/chan_voter.c
ve7fet: Apparently not?

```
CodeRabbit
enum flags; won't compile and breaks the wire format.

enum flags; declares an (illegal, anonymous) enum type with no member name, so authpacket.flags = FLAG_MIX; 
at Line 2983 references a member that doesn't exist. Even if corrected to enum voter_auth_flags flags;, an enum 
is int-sized (4 bytes) even under #pragma pack(1), which changes sizeof(authpacket) and shifts the on-the-wire 
layout. The flags value is protocol-defined as a single octet (see the "octet 24" note at Line 4830), so this field 
must stay 1 byte.
```

ve7fet:
Okay, I think I have resolved all the major issues that Coderabbit found. I am going to stop going down the rabbit holes it keeps finding for minor issues (mostly code robustness), which can be incorporated into another update.

So let's pause here, see if there was anything that still needs to be resolved in this update, then move towards merging.

ve7fet:
The plot thickens... I have discovered that I broke the `voter display <instance>` command.

`ast_waitfor_n_fd` is to blame, again.

It appears that the function may have changed over time. It used to rely on `poll()`, and returned `-1` on error. I think now it is a wrapper for `ast_waitfor_nandfds`, and returns `-99999` on error.

I changed one of the tests in `rad_rxwait` from `x == -1` to `x < 0`, to presumably catch the proper error conditions. Well, that broke it, because `ast_waitfor_n_fd` is returning `-99999`, and causing `rad_rxwait` to exit.

Apparently, I've stumbled on a bug that was masked by a bad (old?) test.

So, now to figure out why we aren't getting the proper `fd` returned, or a timeout... debugging commencing.

channels/chan_voter.c
```diff
@@ -5735,19 +6274,43 @@ static int load_module(void)
...
+	ast_mutex_unlock(&voter_lock);
+	return res; /* Return the result of the reload operation back to Asterisk*/
 }
 
 AST_MODULE_INFO(ASTERISK_GPL_KEY, AST_MODFLAG_DEFAULT, "Voter Radio Channel Driver",
 	.support_level = AST_MODULE_SUPPORT_EXTENDED,
 	.load = load_module,
 	.unload = unload_module,
-	.reload = reload,
+	.reload = asterisk_reload,
```
mkmer: Should we be changing the name of this function to "asterisk_*" -  Thought that was reserved for core functions.

ve7fet: Hmmm... I did a bit of searching and didn't find any references to reserved function names.

I could change it to something like `voter_reload`. 

LMK.

Allan-N: cli_reload ?

mkmer: ast_* was / has been (I guess unwritten) reserved for core functions best I could tell.
Along with that, most modules have the "standard" names as it "was"
https://github.com/AllStarLink/app_rpt/blob/76eb287b78b38f7e102cf45d494e3bdb5b263991/apps/app_rpt.c#L8129-L8134
https://github.com/AllStarLink/app_rpt/blob/76eb287b78b38f7e102cf45d494e3bdb5b263991/channels/chan_usbradio.c#L5593-L5598
https://github.com/AllStarLink/app_rpt/blob/76eb287b78b38f7e102cf45d494e3bdb5b263991/channels/chan_simpleusb.c#L4700-L4705

Should we change all of them or stick with the "standard?"
Reviewing it, `app_rpt` is a bit off too :)

Allan-N: In this case, it really doesn't matter.  The load_module(), unload_module(), reload_module() functions are all marked "static" so they are not exported nor exposed to asterisk.  I had originally proposed asterisk_reload() because it was the main asterisk code that was requesting the reload operation.  When the asterisk_ prefix was questioned I suggested cli_ because this would be a CLI (or AMI?) request.  Using reload_module() is also a fine name.  or do we want to stick with reload() as the "module" function and then rename the common/shared function something like reload_internal()?

ve7fet: I can change `asterisk_reload()` back to `reload()`, and the change the original `reload()` to something like `config_reload()` or `voter_reload()`.

Would that work?

Allan-N: I kinda like changing the AST_MODULE_INFO function to be `module_reload()` ... and keeping the internal function as `reload()`.

I'm not wild about `config_reload()` as the function appears to do a bit more than "config" processing.  `voter_reload()`'s OK.

In the end, pick your favorite combination.

ve7fet: Sure, `module_reload` works too, and we'll leave `reload()` as-is.

I will push that change later today.

Are we good on everything else? I've got some more changes inbound, but want to wrap this one up first, instead of continuing to add to it.

mkmer: The "others" are using `reload_module`.  While it doesn't really matter, I was thinking for sanity, if one is hunting across modules, it can make it "easier" (and consistent).
We "should" probably fix the `app_rpt.c` one to match at some point (not critical of course)

mkmer: Few more "examples" from the core:
https://github.com/asterisk/asterisk/blob/99fe892b66411edb1f9120c6919f079a641ed410/channels/chan_audiosocket.c#L342-L348
https://github.com/asterisk/asterisk/blob/99fe892b66411edb1f9120c6919f079a641ed410/channels/chan_dahdi.c#L20749-L20756
https://github.com/asterisk/asterisk/blob/99fe892b66411edb1f9120c6919f079a641ed410/channels/chan_iax2.c#L15130-L15137

Maybe just "reload" is the way?

ve7fet: Committed as 07e44733972b24f1065bbb99219805af3dde1956 and went with `reload_module` for consistency with `chan_simpleusb` and `chan_usbradio`.

channels/chan_voter.c
Allan-N: As I wrote, we are using the following block in chan_echolink :
```
    if (ast_timer_get_event(p->timer) == AST_TIMING_EVENT_EXPIRED) {
        if ((ast_timer_ack(p->timer, 1) < 0)) {
            ast_log(LOG_WARNING, "Timer ack failed. \n");
            return NULL;
        }
    }
```
and I think you want the same here (not 1 check for != and a 2nd for ==).

Allan-N (changes requested):

channels/chan_voter.c
```diff
@@ -4114,9 +4170,8 @@ static int reload(void)
 			}
 			cp = ast_strdup(v->value);
 			if (!cp) {
-				close(udp_socket);
+				ast_log(LOG_ERROR, "VOTER: Memory allocation failure while reloading configuration\n");
```
Allan-N: ast_strdup() logs the following message on memory allocation failures :

```
#define MALLOC_FAILURE_MSG \
    ast_log_safe(LOG_ERROR, "Memory Allocation Failure in function %s at line %d of %s\n", func, lineno, file)
```

... so we don't need to add calls to ast_log() ... (here and elsewhere)

... and the same is true for calls to ast_alloca, ast_calloc, ast_malloc, ast_realloc, ...

channels/chan_voter.c
ve7fet: So, I fought with this a bit on the weekend, and it was still spitting out some warnings when coderabbit was analyzing changes. I decided to pivot, and went down the `ast_poll()` rabbit hole... this is what I came up with as a potential replacement:

```
while (run_forever && !ast_shutting_down()) {
		/* Check and acknowledge our thread timer. This timer keeps our audio in sync (for IAX2). */
		struct pollfd fds[1];
		int res;
		int timeout_ms = -1; /* block until the timer fires */
		int timer_fd = ast_timer_fd(voter_thread_timer);

		if (timer_fd < 0) {
			ast_log(LOG_ERROR, "Failed to get pollable thread timer fd.\n");
			break;
		}

		fds[0].fd = timer_fd;
		fds[0].events = POLLIN; // Monitor for incoming data


		res = ast_poll(fds, 1, timeout_ms); // block until the timer fires

		if (res < 0) {
			if (errno == EINTR) {
				/* The poll was interrupted by a signal (e.g., a process hang-up or
				 * reload). This is normal operation, just simulate a timeout by returning 0.
				 */
				ast_log(LOG_NOTICE, "Poll interrupted by a signal, retrying...\n");
				continue;
			} else {
				/* Log the actual system-level error */
				ast_log(LOG_ERROR, "ast_poll failed: %s (%d)\n", strerror(errno), errno);
				break;
			}
		} else {
			if (fds[0].revents & (POLLERR | POLLHUP | POLLNVAL)) {
				/* Handle error conditions for a specific file descriptor */
				ast_log(LOG_WARNING, "Error or hangup on descriptor %d\n", fds[0].fd);
				break;
			}
			if (fds[0].revents & POLLIN) {
				enum ast_timer_event event = ast_timer_get_event(voter_thread_timer);
				if (event == AST_TIMING_EVENT_EXPIRED) {
					if (ast_timer_ack(voter_thread_timer, 1) < 0) {
						ast_log(LOG_ERROR, "Failed to acknowledge timer.\n");
						break;
					}	
					/* Fall through to run the periodic tasks below. */
				} else {
					continue;
				}
			}
		}

		ast_mutex_lock(&voter_lock);

...

```

channels/chan_voter.c
ve7fet: Redundant logging removed in 572044af3d5ba992e7379e130e5f2f7af5e1526d

channels/chan_voter.c
Allan-N: Yup.  The `flags` variable is included an over-the-wire struct so we need to use a type of a fixed size.  The most recent C language updates allow one to express/define an enum type who's value is of a known/fixed size ... but asterisk (and our app_rpt code) has yet to adopt/accept the C23 "strong" enums".

channels/chan_voter.c
Allan-N: What you have now is :
```
        if (ast_waitfor_n_fd(&timer_fd, 1, &timeout_ms, NULL) < 0) {
            ast_log(LOG_ERROR, "Failed to wait on VOTER thread timer.\n");
            break;
        }
        if (ast_timer_get_event(voter_thread_timer) != AST_TIMING_EVENT_EXPIRED) {
            if (ast_timer_ack(voter_thread_timer, 1) < 0) {
                ast_log(LOG_ERROR, "Failed to acknowledge timer.\n");
                break;
            }
            continue;
        }
        if (ast_timer_ack(voter_thread_timer, 1) < 0) {
            ast_log(LOG_ERROR, "Failed to acknowledge timer.\n");
            break; 
        }
```
I read this as first calling `ast_waitfor_n_fd()` to wait for the timer to fire.  Then, we check if the timer even did not expire ... and try to acknowledge something that didn't happen.  Then we ack the timer.

What I think this entire block should look like is :
```
        if (ast_waitfor_n_fd(&timer_fd, 1, &timeout_ms, NULL) < 0) {
            ast_log(LOG_ERROR, "Failed to wait on VOTER thread timer.\n");
            break;
        }
        if (ast_timer_get_event(voter_thread_timer) == AST_TIMING_EVENT_EXPIRED) {
            if (ast_timer_ack(voter_thread_timer, 1) < 0) {
                ast_log(LOG_ERROR, "Failed to acknowledge timer.\n");
                break;
            }
        }
```

channels/chan_voter.c
ve7fet: I believe changing `char flags;` to `uint8_t flags;` should resolve this to make it a fixed 8 bits?

Allan-N: The only difference between a `char` and a `uint8_t` is that one is signed and the other not.  That won't address Mike's suggestion of changing `char flags;` to `enum voter_auth_flags flags;` ... something we can't do because we don't have "strong enums".

ve7fet: Ok, what's the path forward then to resolve this?

Allan-N: I would simply leave the `char flags;` in place and move on.  The other option would be to add a local `enum` variable for all of the manipulation and then make a final assignment, possibly with a cast.  But, since the compiler's not complaining I'd skip going down that path.

ve7fet: Ok, I will mark this as resolved then. No, the compiler isn't complaining.

channels/chan_voter.c
ve7fet: Committed the revision above in 396ffda4c6dc1d4c9ff4ddead3e712c697008f21

Allan-N (approved):

PR #1085, opened 2026-06-30
dev: Add pre-commit test for protected branch

mkmer:
## Summary by CodeRabbit

* **Chores**
  * Added protection to prevent committing directly on key branches like `main`, `master`, and `dev`.
  * Improved the commit hook behavior when a branch name can’t be determined.
* **Documentation**
  * Corrected a spelling error in an application header comment.

### DIFF
--- a/.dev/pre-commit
+++ b/.dev/pre-commit
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Antigravity pre‑commit hook: warn if committing directly to a protected branch (master/main)
+# pre‑commit hook: warn if committing directly to a protected branch
 # This script should be symlinked or referenced by the actual Git hook.
 # Redirect output to stderr.
 exec 1>&2
--- a/.dev/pre-commit
+++ b/.dev/pre-commit
@@ -1,14 +1,25 @@
-#!/bin/sh
-#
-# An example hook script to verify what is about to be committed.
-# Called by "git commit" with no arguments.  The hook should
-# exit with non-zero status after issuing an appropriate message if
-# it wants to stop the commit.
-#
-# To enable this hook, rename this file to "pre-commit".
-
+#!/usr/bin/env bash
+# Antigravity pre‑commit hook: warn if committing directly to a protected branch (master/main)
+# This script should be symlinked or referenced by the actual Git hook.
 # Redirect output to stderr.
 exec 1>&2
+
+# Get the current branch name (fallback to HEAD if detached)
+branch=$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo "HEAD")
+
+# List of branch names that are typically protected. Add more if needed.
+protected_branches=("master" "main" "dev")
+
+for pb in "${protected_branches[@]}" 
+do
+  if [[ "$branch" == "$pb" ]] 
+  then
+    echo "          Warning: You are committing directly to the protected branch '$branch'."
+    echo "          Please create a new feature branch first (e.g., git checkout -b feature/your-task) and open a PR."
+    echo "          Commit aborted."
+    exit 1
+  fi
+done
 
 BEFORE_DIFF="$(git diff --no-ext-diff --binary)"
 # Test clang-format
@@ -49,4 +60,3 @@ then
    echo "Please review the changes and stage the files again."
    exit 1
 fi
-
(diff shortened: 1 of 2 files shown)


.dev/pre-commit
```diff
@@ -1,15 +1,26 @@
...
+# Antigravity pre‑commit hook: warn if committing directly to a protected branch (master/main)
+# This script should be symlinked or referenced by the actual Git hook.
 # Redirect output to stderr.
 exec 1>&2
 
+# Get the current branch name (fallback to HEAD if detached)
+branch=$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo "HEAD")
+
+# List of branch names that are typically protected. Add more if needed.
+protected_branches=("master" "main" "dev")
```
Allan-N: Did you really want the "develop" (and not "dev") branch?

... and I'm also going to ask if there's a way to determine the 'default" main branch?  If so, would that help us eliminate the hard-coded list of protected branches?

mkmer: I think I found a better way to detect if the current branch is protected.

.dev/pre-commit
```diff
@@ -1,15 +1,29 @@
...
+# This script should be symlinked or referenced by the actual Git hook.
 # Redirect output to stderr.
 exec 1>&2
 
+# Get the current branch name (fallback to HEAD if detached)
+branch=$(git rev-parse --abbrev-ref HEAD 2>/dev/null || echo "HEAD")
+
+# Detect protected branches via GitHub rulesets (requires gh CLI)
+# If gh fails or returns no rules, fall back to a static list.
+if command -v gh >/dev/null 2>&1; then
```
Allan-N, line 12: Any chance that we can do this without `gh` as the usage would add another command/package dependency? ... and possibly a dependency on our always using github.com?

apps/app_rpt.c
```diff
@@ -4,7 +4,7 @@
  * Copyright (C) 2002-2014, Jim Dixon, WB6NIL
  *
  * Jim Dixon, WB6NIL <jim@lambdatel.com>
- * Serious contributions by Steve RoDgers, WA6ZFT <hwstar@rodgers.sdcoxmail.com>
+ * Serious contributions by Steve Rodgers, WA6ZFT <hwstar@rodgers.sdcoxmail.com>
```
Allan-N: I'm not convinced that this change belongs with this branch/PR.

.dev/pre-commit
mkmer: They claim protected branches is part of github NOT git, so we can either use GH, or use the API (which requires a key).  We "could' just bypass GH if it's not there rather than complain and exit...

apps/app_rpt.c
mkmer: Yes, this was part of my test for commit failure, Definately doesn't need to be here.

.dev/pre-commit
Allan-N: Supposedly, one can get the default branch for a repo with :
```
git remote show origin | sed -n 's/  HEAD branch: //p'

File: .dev/pre-commit [hunk 4]
```diff
@@ -1,15 +1,28 @@
...
+	# gh ruleset check returns lines starting with '-' for each rule
+	PROTECTED_COUNT=$(gh ruleset check "$branch" 2>/dev/null | grep -c '^-')
+	if [ "$PROTECTED_COUNT" -gt 0 ]; then
+		echo "          Warning: You are committing directly to a protected branch '$branch' (detected via GitHub rules)."
```
Allan-N: You are trying to commit directly to protected branch '$branch' ...

.dev/pre-commit
```diff
@@ -1,15 +1,28 @@
...
+		echo "          Please create a new feature branch first (e.g., git checkout -b feature/your-task) and open a PR."
+		echo "          Commit aborted."
+		exit 1
+	fi
+else
+	echo "          Please install GitHub CLI for branch protection"
```
Allan-N: Branch protection rule checks not available.  Please install GitHub CLI (gh).

.dev/pre-commit
mkmer: This will check the default branch, not the protected status of a branch... I suppose we can just confirm I'm not trying to commit to the default, but IF we someday have a couple of protected branches (say "Dev" and "Master", or whatever) it would not work for that.

Allan-N (approved):

Issue #307, opened 2024-04-20
Exceptionally long voice queue length queuing to...


tsawyer:
Found Allmon attempting to connect to a regularly connected node. `asterisk -r` would immediately return system prompt.

Attached messages.log and coredump --running. 
[iTerm2 Session Apr 19, 2024 at 8:53:34 PM.txt](https://github.com/AllStarLink/app_rpt/files/15046731/iTerm2.Session.Apr.19.2024.at.8.53.34.PM.txt)
[core-asterisk-running-2024-04-20T02-50-36Z-full.txt](https://github.com/AllStarLink/app_rpt/files/15046733/core-asterisk-running-2024-04-20T02-50-36Z-full.txt)

InterLinked1:
Seems like a deadlock with threads trying to acquire `&myrpt->lock`, just after line 6573. There appear to be hundreds of threads hung at this point, probably all from different incoming connections executing `Rpt()`.

There is a way to easily determine the thread holding a mutex using gdb, but that requires access to the actual corefile. Probably can be determined from the backtrace itself with more analysis, though I'm not going to do that now, so just leaving a comment for the future.

okcrum:
We see this in testing

`[2025-12-15 01:01:05.862] WARNING[935463] channel.c: Exceptionally long voice queue length (97 voice / 97 total) queuing to Local/IAX-Link@default-00000004;1
`

but it does not seem to relate to this issue. There are no core dumps with dahdi or without.

mkmer:
I believe this: `base-bridge <2e8f6ded-0f40-4664-8df2-d1fc06beeb2d>` is the TXCONF bridge - can you look further up in your log and confirm the id matches the creation of that bridge?
It seems odd that writing to `<Local/Telem@default-00000217;1>` would be long when `<Local/Telem@default-00000217;2>` is in the bridge -> these ;1 and ;2 are both ends of a "single" link (assuming I understand Local channels correctly).

MAYBE this is audio coming back into the channel and we don't service it... I bet that's it.
We probably need to autoservice telemetry channels to cope with any audio coming back at us....

mkmer:
> > OK - I'm fairly sure this warning is because we do not read the telemetry channel, yet frames are being sent from the `softmix_bridge`. Since we don't care about these frames, it's not "really" a problem, but it's not right.
> 
[...]

PR #1086, opened 2026-06-30
app_rpt: Address hanging pchan on disconnect

mkmer:
An outbound link is NEVER reconnectable and should cleanup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of a specific remote disconnect path so call teardown and resource cleanup run reliably before returning.
  * Updated link-channel hangup/disconnect processing to consistently complete the shared “link is done” cleanup sequence, reducing the risk of leftover call state or resources after a hangup.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4603,7 +4603,7 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 					continue;
 				}
 				/* A reconnect is not possible */
-				return;
+				break;
 			}
 			if (f->frametype == AST_FRAME_VOICE) {
 				int ismuted, n1;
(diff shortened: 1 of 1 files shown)


mkmer:
Fixes at least one path to:
https://github.com/AllStarLink/app_rpt/issues/307

apps/app_rpt.c
Allan-N: If we remove this `return;` we will fall through to the `return;` at the end of the function.

apps/app_rpt.c
```diff
@@ -3557,6 +3557,7 @@ static inline void periodic_process_link(struct rpt *myrpt, struct rpt_link *l,
 			donodelog_fmt(myrpt, "LINKDISC,%s", l->name);
 			dodispgm(myrpt, l->name);
 			/* hang-up on call to device */
+
 			return;
```
Allan-N: If we remove this `return;` we will fall through to the `return;` at the end of the function.

Allan-N (approved):

PR #1029, opened 2026-04-22
chan_simpleusb: Use PortAudio for audio output

mkmer:
Add PortAudio thread
Refactor startup of HID thread
Allow use of hw:xxx to configure usb

This allows ASL to work on any modern kernel by eliminating the obsolete `/dev/dsp` interface.

Add a USB buffer "monitor" to handle the mysterious condition were the audio buffer stops draining causing the TX output to remain active.  When the buffer exceeds 200ms the stream is now reset.

## Summary by CodeRabbit

* **New Features**
  * PortAudio-based audio I/O support
  * New `audiodev` option to select target audio hardware
  * Public API to map an ALSA card to the underlying USB device
  * `ast_radio_check_audio` now explicitly accepts a mono/stereo indicator

* **Changes**
  * Replaces legacy OSS/DSP audio flow with threaded PortAudio streaming and improved underflow handling
  * Refines USB/ALSA device discovery and HID timing
  * Improves radio audio/statistics processing (including mono downsampling)

PR #1091, opened 2026-07-06
res_usbradio: Add shared PortAudio API for USB radio drivers

hardenedpenguin:
## Summary
- Add refcounted PortAudio lifecycle (`ast_radio_pa_startup` / `shutdown`)
- Add `ast_radio_pa_*` stream helpers shared by USB channel drivers
- Match ALSA devices via `hw:N` parsing and `/proc/asound/cardN/id` fallback
- Link `-lportaudio` on `res_usbradio.so`

## Stack
This is **PR 1 of 3** in the OSS → PortAudio migration. Merge in order:
1. This PR (res_usbradio foundation)
2. #TBD — chan_simpleusb
3. #TBD — chan_usbradio + packaging

Supersedes the monolithic approach in #1029.

## Test plan
- [ ] `res_usbradio.so` builds and links `libportaudio`
- [ ] Module loads on ASL3 22.9 without errors

## Summary by CodeRabbit

* **New Features**
  * Added enhanced PortAudio-based USB radio streaming with wider hardware/device matching support, including mapping an ALSA card to a connected device.
* **Bug Fixes**
  * Improved mono vs. stereo audio validation and statistics handling, including safer behavior when downsampling produces no samples.
  * More robust mixer level limiting for both microphone capture and speaker playback.
  * Strengthened mixer element error handling, USB/HID device matching, and PortAudio startup/shutdown behavior for more reliable unloads.

### DIFF
--- a/include/asterisk/res_usbradio.h
+++ b/include/asterisk/res_usbradio.h
@@ -84,14 +84,6 @@
 #define AUDIO_ADJUSTMENT 1000
 
 /*!
- * \brief Default ALSA mixer maximum when hardware lookup fails.
- *
- * CM108-class devices commonly expose a 0-31 capture range; use this
- * instead of propagating -1 into volume scaling math.
- */
-#define AST_RADIO_MIXER_MAX_DEFAULT 31
-
-/*!
  * \brief EEPROM memory layout
  *	The AT93C46 eeprom has 64 addresses that contain 2 bytes (one word).
  *	The CMxxx sound card device will use this eeprom to read manuafacturer
@@ -257,13 +249,21 @@ int ast_radio_make_spkr_playback_value(int spkrmax, int request_value, int devty
 int ast_radio_amixer_max(int devnum, char *param);
 
 /*!
- * \brief Return a usable mixer maximum for volume scaling.
+ * \brief Query required ALSA mixer maximums for a USB radio device.
  *
- * \param limit Value from ast_radio_amixer_max(), or another mixer limit.
+ * Reads mic capture, speaker playback (with alternate control name), and
+ * mic playback limits. Fails when any required limit is unavailable.
  *
- * \retval limit if positive, otherwise AST_RADIO_MIXER_MAX_DEFAULT.
+ * \param devnum ALSA card number.
+ * \param micmax Returned Mic Capture Volume maximum.
+ * \param spkrmax Returned Speaker Playback Volume maximum.
+ * \param micplaymax Returned Mic Playback Volume maximum.
+ * \param newname Set to 1 when the alternate speaker control name is used.
+ *
+ * \retval 0 on success.
+ * \retval -1 if any required mixer limit is unavailable.
  */
-int ast_radio_mixer_limit(int limit);
+int ast_radio_init_mixer_limits(int devnum, int *micmax, int *spkrmax, int *micplaymax, int *newname);
 
 /*!
  * \brief Set mixer
--- a/include/asterisk/res_usbradio.h
+++ b/include/asterisk/res_usbradio.h
@@ -80,6 +80,14 @@
 #define AUDIO_ADJUSTMENT 1000
 
 /*!
+ * \brief Default ALSA mixer maximum when hardware lookup fails.
+ *
+ * CM108-class devices commonly expose a 0-31 capture range; use this
+ * instead of propagating -1 into volume scaling math.
+ */
+#define AST_RADIO_MIXER_MAX_DEFAULT 31
+
+/*!
  * \brief EEPROM memory layout
  *	The AT93C46 eeprom has 64 addresses that contain 2 bytes (one word).
  *	The CMxxx sound card device will use this eeprom to read manuafacturer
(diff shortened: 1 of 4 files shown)


hardenedpenguin:
Addressed CodeRabbit feedback in 6ec8440:

- **ast_radio_pa_write**: reject `frames > AST_RADIO_PA_FRAMES_PER_BUFFER` before underflow recovery to avoid over-reading `null_buf`
- **USB device list**: replace `strcat` with `memcpy` for bounds-safe append
- **ast_radio_check_audio**: updated `chan_usbradio` call sites with `mono=0` so the header signature change builds cleanly on master until the full usbradio migration lands in #1093

#1092 and #1093 are now marked ready for review so CodeRabbit can run on the full stack.

hardenedpenguin:
Additional CodeRabbit fixes in 98ef274:

- Check `snd_hctl_elem_info()` return before reading mixer element type (both mixer helpers)
- Pass `devnum` to `ast_radio_parse_hw_anywhere()` in `pa_device_matches()` so card-id fallback works
- `ast_radio_pa_stop()` now closes streams when open but not yet started (failed-start cleanup path)

Allan-N:
Before we spend lots of extra time on this PR (and the others that are related) I'd like to ask how this implementation differs from the [PR # 1029](https://github.com/AllStarLink/app_rpt/pull/1029) branch?

hardenedpenguin:
> Before we spend lots of extra time on this PR (and the others that are related) I'd like to ask how this implementation differs from the [PR # 1029](https://github.com/AllStarLink/app_rpt/pull/1029) branch?

We've compared our PortAudio stack (#1091–#1093) with mkmer's #1029. Both aim to replace OSS /dev/dsp with PortAudio/ALSA, and the simpleusb side is similar in spirit (threaded audio read, audiodev=hw:N, buffer stall handling).

The main differences:

#1029 migrates simpleusb only; chan_usbradio still uses OSS. PortAudio lives inside chan_simpleusb.c, with Pa_Initialize/Pa_Terminate tied to that module.
Our stack migrates both chan_simpleusb and chan_usbradio, with a shared PortAudio API in res_usbradio.c (ast_radio_pa_*). PortAudio is refcounted there and linked only on res_usbradio.so.
We also include packaging/docs (rpt_install.sh, sample configs) and are based on current master, with test packages deployed on several nodes.

We see #1029 as useful prior art for simpleusb-specific fixes, but not a complete path to dropping OSS across the USB radio stack. Our stacked PRs are intended to supersede #1029 once merged

hardenedpenguin:
@coderabbitai review

Rebased stack with review fixes and build-tested packages (100.deb13 arm64/amd64). Ready for fresh review.

hardenedpenguin:
@coderabbitai review

mkmer:
Since we had already reached the point in testing on #1029 that was ready for a beta, we are merging it.
Let's keep working on this group of PR's as we do intend (and need) to update the `chan_usbradio` driver.  
The first step is confirming we have PortAudio working reliably across a larger test bed of machines as we had quite a bit of "fun" rooting out the "bad" behavior.

hardenedpenguin:
@coderabbitai review

hardenedpenguin:
@coderabbitai review

hardenedpenguin:
Pushed 1a041b1 addressing mkmer's review on #1092 that applies to shared `res_usbradio` code:

- Restore `HAVE_SYS_IO` alongside `#include <portaudio.h>` in `res_usbradio.h`
- NUL-terminate `readlink()` results and use a single `cplen` in USB HID list construction

#1092 has been rebased onto this commit.

include/asterisk/res_usbradio.h
```diff
@@ -81,6 +83,14 @@
  */
 #define AUDIO_ADJUSTMENT 1000
 
+/*!
+ * \brief Default ALSA mixer maximum when hardware lookup fails.
+ *
+ * CM108-class devices commonly expose a 0-31 capture range; use this
+ * instead of propagating -1 into volume scaling math.
+ */
+#define AST_RADIO_MIXER_MAX_DEFAULT 31
```
Allan-N: I'm hesitant to add a hard-coded MAX_DEFAULT based on a specific class/type of devices.  One of the reason we're moving towards PortAudio/ALSA is to break our dependency on the CM108's.  I suspect this #define and the new `ast_radio_mixer_limit()` function is to guard against the possibility of not being able to obtain the mixer max value.  I would counter that if we can't programmatically determine "the" max value for the device then we should not allow its use (at all).  We should not be guessing nor making assumptions.

To address the "what if we can't get a value" concern I'd be fine with checking that a value is available when the device is first opened/initialized.  If so, we're good.  If not, log a "mixer limit not available" message and fail.

Allan-N (changes requested):

include/asterisk/res_usbradio.h
hardenedpenguin: Agreed — we’ll drop AST_RADIO_MIXER_MAX_DEFAULT and ast_radio_mixer_limit(). Mixer limits will be queried at device association; if any required element can’t be read, we log an error and refuse to bring the device up rather than assuming CM108-style ranges.

hardenedpenguin:
Closing per author request — PortAudio stack work is being withdrawn.

PR #1098, opened 2026-07-10
app_rpt.c: Restore mute detection for local channel bridge

mkmer:
This allows AMI muting to work with the local channel as it did with DAHDI

## Summary by CodeRabbit

* **Bug Fixes**
  * Muted status is now reported accurately for bridged channels (previously it always reported unmuted).
  * Improved mute detection by correctly deriving and reading bridge-channel mute features when evaluating a channel’s mute state.

### DIFF
--- a/apps/app_rpt/rpt_bridging.c
+++ b/apps/app_rpt/rpt_bridging.c
@@ -415,7 +415,12 @@ int __rpt_conf_add(struct ast_channel *chan, struct rpt *myrpt, enum rpt_conf_ty
 	return res;
 }
 
-struct ast_bridge_channel *rpt_get_bridge_channel_from_chan(struct ast_channel *chan)
+/*!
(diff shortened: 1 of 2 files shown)


apps/app_rpt/rpt_bridging.c
```diff
@@ -415,10 +415,44 @@ int __rpt_conf_add(struct ast_channel *chan, struct rpt *myrpt, enum rpt_conf_ty
 	return res;
 }
 
+struct ast_bridge_channel *rpt_get_bridge_channel_from_chan(struct ast_channel *chan)
```
Allan-N: It looks like this function is only used by `rpt_conf_get_muted()`.  As such, should the function be "static" (and internal to rpt_bridging.c)?

Allan-N (approved):

PR #1100, opened 2026-07-11
app_rpt: drain link pchan during disconnect grace

hardenedpenguin:
## Summary
- Non-blockingly drain `l->pchan` while `l->disctime` is active
- Covers timer expiry and activity on `l->chan`, not only when waitfor selects pchan

## Problem
During link disconnect grace, pchan frames could back up when the wait loop timed out or serviced the link channel instead of the pseudo-channel.

## Test plan
- [ ] Build passes
- [ ] Disconnecting link does not leave backed-up frames on pchan

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of disconnected links by clearing pending channel frames before processing continues.
  * Prevented stale frames from interfering with subsequent channel activity after timer expirations or disconnect events.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4592,6 +4592,20 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 		if (!ms) {
 			/* No channels had activity before the timer expired,
 			 * so just continue to the next loop. */
+			if (l->disctime && l->pchan) {
+				struct ast_frame *f;
+				int ms_drain = 0;
+				struct ast_channel *cs_drain[1] = { l->pchan };
+
+				if (ast_waitfor_n(cs_drain, 1, &ms_drain) == l->pchan) {
+					f = ast_read(l->pchan);
+					if (!f) {
+						ast_debug(1, "@@@@ rpt:Hung Up\n");
+						break;
+					}
+					ast_frfree(f);
+				}
+			}
 			continue;
 		}
 		if (l->disctime) {
@@ -4606,6 +4620,20 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 				}
 				ast_frfree(f);
 				continue;
+			}
+			if (l->pchan) {
+				struct ast_frame *f;
+				int ms_drain = 0;
+				struct ast_channel *cs_drain[1] = { l->pchan };
+
+				if (ast_waitfor_n(cs_drain, 1, &ms_drain) == l->pchan) {
+					f = ast_read(l->pchan);
+					if (!f) {
+						ast_debug(1, "@@@@ rpt:Hung Up\n");
+						break;
+					}
+					ast_frfree(f);
+				}
 			}
 			continue;
 		}


apps/app_rpt.c
```diff
@@ -4592,6 +4592,20 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 		if (!ms) {
 			/* No channels had activity before the timer expired,
 			 * so just continue to the next loop. */
+			if (l->disctime && l->pchan) {
+				struct ast_frame *f;
+				int ms_drain = 0;
+				struct ast_channel *cs_drain[1] = { l->pchan };
+
+				if (ast_waitfor_n(cs_drain, 1, &ms_drain) == l->pchan) {
```
mkmer, line 4600: We shouldn't be waiting on the pchan "again", We already wait on the pchan and trigger this code in ln4950 -https://github.com/hardenedpenguin/app_rpt/blob/ffebfb353dcc85a0d1f83e33f3f22c20a0ae7d81/apps/app_rpt.c#L4586-L4589
We deal with pchan reads here:
https://github.com/hardenedpenguin/app_rpt/blob/ffebfb353dcc85a0d1f83e33f3f22c20a0ae7d81/apps/app_rpt.c#L4877

IF we are exiting this thread, we should hangup on the channels -> maybe that is broken?  Maybe we are not getting the other side of the pchan out of the conference in fast enough?

mkmer:
@coderabbitai review

hardenedpenguin:
@mkmer Good catch — I re-traced the `disctime` path and agree this is probably the wrong fix.

Once `remote_hangup_helper()` sets `disctime`, it also calls `hangup_link_chan()`, so during grace the poll set is just `l->pchan`. The existing block at 4597–4608 should already drain when data is ready, and the extra `ast_waitfor_n()` duplicates the main loop wait without adding much. #1086 likely already fixed the real problem (cleanup not always reaching `ast_hangup(l->pchan)`).

Unless you've seen a post-#1086 repro I'm missing, I'll plan to close this. Happy to look at conference teardown timing instead if there's still a hang case.

hardenedpenguin:
Closing — redundant with #1086 disconnect cleanup. During `disctime`, `hangup_link_chan()` already clears `l->chan`, so the poll set is only `l->pchan` and the existing drain path should suffice. The added `ast_waitfor_n()` calls duplicate the main loop without addressing a real gap.

Thanks @mkmer for the review.

mkmer:
> @mkmer Good catch — I re-traced the `disctime` path and agree this is probably the wrong fix.
> 
> Once `remote_hangup_helper()` sets `disctime`, it also calls `hangup_link_chan()`, so during grace the poll set is just `l->pchan`. The existing block at 4597–4608 should already drain when data is ready, and the extra `ast_waitfor_n()` duplicates the main loop wait without adding much. #1086 likely already fixed the real problem (cleanup not always reaching `ast_hangup(l->pchan)`).
> 
> Unless you've seen a post-#1086 repro I'm missing, I'll plan to close this. Happy to look at conference teardown timing instead if there's still a hang case.

#1086 did not 100% address the issue - we still have an unread channel hanging around.  My current line of thought is it's "probably" one of the telemetry channels not closing correctly... the latest comment on the issue is related to a node that has no radios and no connections, so I think that points to telemetry being the most likely location.

PR #1103, opened 2026-07-11
rpt_functions: fix function_macro digit validation loop

hardenedpenguin:
## Summary
- Validate macro digits with `for (i = 0; digitbuf[i]; i++)` instead of using ASCII value of first digit as loop bound

## Problem
`function_macro()` used `digitbuf[i]` as both bound and index, causing wrong validation and out-of-bounds reads.

## Test plan
- [ ] Build passes
- [ ] Macro dial strings with multiple digits validate correctly


mkmer:
@coderabbitai review

mkmer (approved):

Allan-N (approved):

PR #1111, opened 2026-07-17
res_usbradio: Add shared PortAudio API for USB radio drivers

hardenedpenguin:
## Summary
- Add shared PortAudio lifecycle and stream helpers in `res_usbradio`
- Add ALSA card-index fallback for `audiodev=hw:N` binding
- Add `ast_radio_init_mixer_limits()` fail-fast mixer lookup (no CM108 default guess)

Standalone piece 1/3 for OSS → PortAudio migration. No channel driver changes in this PR.

## Test plan
- [ ] CI passes (ubuntu build, clang-format, codespell)
- [ ] `res_usbradio.so` links against `-lportaudio`
- [ ] Merge before opening/reviewing simpleusb PR (2/3)

## Summary by CodeRabbit

* **New Features**
  * Added PortAudio-based USB radio audio streaming, including open/start/stop/read/write and audio availability support.
  * Added helper APIs to query ALSA mixer maximums for mic capture and playback, speaker playback (with alternate control fallback).
  * Added PortAudio 48kHz audio checks for RX and stereo monitoring.
* **Bug Fixes**
  * Improved ALSA mixer element info/error handling with safer failure behavior and cleanup.
  * Hardened USB audio device path resolution (correct readlink handling) and improved USB device list construction.

### DIFF
--- a/res/res_usbradio.c
+++ b/res/res_usbradio.c
@@ -1010,16 +1010,6 @@ static void pa_lib_release(void)
 	ast_mutex_unlock(&pa_lock);
 }
 
-static void pa_lib_release_all(void)
-{
-	ast_mutex_lock(&pa_lock);
-	if (pa_refcount > 0) {
-		Pa_Terminate();
-		pa_refcount = 0;
-	}
-	ast_mutex_unlock(&pa_lock);
-}
-
 #define AST_RADIO_HW_CARD_MAX 9999
 #define AST_RADIO_HW_DEV_MAX 255
 
@@ -1490,7 +1480,9 @@ static int load_module(void)
 static int unload_module(void)
 {
 	cleanup_user_devices();
-	pa_lib_release_all();
+	ast_mutex_lock(&pa_lock);
+	ast_assert(pa_refcount == 0);
+	ast_mutex_unlock(&pa_lock);
 
 	return 0;
 }
--- a/res/res_usbradio.c
+++ b/res/res_usbradio.c
@@ -28,6 +28,7 @@
 
 /*** MODULEINFO
 	<depend>alsa</depend>
+	<depend>portaudio</depend>
 	<support_level>extended</support_level>
  ***/
 
@@ -48,6 +49,7 @@
 #include <linux/parport.h>
 #include <linux/version.h>
 #include <alsa/asoundlib.h>
+#include <portaudio.h>
 
 #include "asterisk/res_usbradio.h"
 
@@ -117,7 +119,47 @@ long ast_radio_lround(double x)
 
 int ast_radio_make_spkr_playback_value(int spkrmax, int request_value, int devtype)
 {
+	if (spkrmax <= 0) {
+		return 0;
+	}
 	return (request_value * spkrmax) / AUDIO_ADJUSTMENT;
+}
+
+int ast_radio_init_mixer_limits(int devnum, int *micmax, int *spkrmax, int *micplaymax, int *newname)
+{
+	int rv;
+
+	if (!micmax || !spkrmax || !micplaymax || !newname) {
+		return -1;
+	}
+
+	rv = ast_radio_amixer_max(devnum, MIXER_PARAM_MIC_CAPTURE_VOL);
+	if (rv <= 0) {
+		ast_log(LOG_ERROR, "Mixer limit not available for hw:%d (%s)\n", devnum, MIXER_PARAM_MIC_CAPTURE_VOL);
+		return -1;
+	}
+	*micmax = rv;
+
+	*newname = 0;
+	rv = ast_radio_amixer_max(devnum, MIXER_PARAM_SPKR_PLAYBACK_VOL);
+	if (rv <= 0) {
+		rv = ast_radio_amixer_max(devnum, MIXER_PARAM_SPKR_PLAYBACK_VOL_NEW);
+		if (rv <= 0) {
+			ast_log(LOG_ERROR, "Mixer limit not available for hw:%d (speaker playback)\n", devnum);
+			return -1;
+		}
+		*newname = 1;
+	}
+	*spkrmax = rv;
+
+	rv = ast_radio_amixer_max(devnum, MIXER_PARAM_MIC_PLAYBACK_VOL);
+	if (rv <= 0) {
+		ast_log(LOG_ERROR, "Mixer limit not available for hw:%d (%s)\n", devnum, MIXER_PARAM_MIC_PLAYBACK_VOL);
+		return -1;
+	}
+	*micplaymax = rv;
+
+	return 0;
 }
 
 int ast_radio_amixer_max(int devnum, char *param)
@@ -151,7 +193,10 @@ int ast_radio_amixer_max(int devnum, char *param)
 	}
 
 	snd_ctl_elem_info_alloca(&info);
-	snd_hctl_elem_info(elem, info);
+	if (snd_hctl_elem_info(elem, info) < 0) {
+		snd_hctl_close(hctl);
+		return -1;
+	}
 	type = snd_ctl_elem_info_get_type(info);
 	rv = 0;
 
@@ -199,7 +244,10 @@ int ast_radio_setamixer(int devnum, char *param, int v1, int v2)
 	}
 
 	snd_ctl_elem_info_alloca(&info);
-	snd_hctl_elem_info(elem, info);
+	if (snd_hctl_elem_info(elem, info) < 0) {
+		snd_hctl_close(hctl);
+		return -1;
+	}
 	type = snd_ctl_elem_info_get_type(info);
 	snd_ctl_elem_value_alloca(&control);
 	snd_ctl_elem_value_set_id(control, id);
@@ -425,6 +473,8 @@ int ast_radio_hid_device_mklist(void)
 	struct usb_bus *usb_bus;
 	struct usb_device *dev;
 	char devstr[10000], str[200], desdev[200], *cp;
+	ssize_t linklen;
+	size_t cplen;
 	int i;
 
 	ast_mutex_lock(&usb_list_lock);
@@ -464,11 +514,11 @@ int ast_radio_hid_device_mklist(void)
 
 				snprintf(str, sizeof(str), "/sys/class/sound/card%d/device", i);
 				memset(desdev, 0, sizeof(desdev));
-
-				if (readlink(str, desdev, sizeof(desdev) - 1) == -1) {
+				linklen = readlink(str, desdev, sizeof(desdev) - 1);
+				if (linklen == -1) {
 					continue;
 				}
-
+				desdev[linklen] = '\0';
 				cp = strrchr(desdev, '/');
 				if (!cp) {
 					continue;
@@ -481,23 +531,23 @@ int ast_radio_hid_device_mklist(void)
 				continue;
 			}
 
-			new_list = ast_realloc(usb_device_list, usb_device_list_size + 2 + strlen(cp));
-
+			cplen = strlen(cp);
+			new_list = ast_realloc(usb_device_list, usb_device_list_size + 2 + cplen);
 			if (!new_list) {
 				ast_mutex_unlock(&usb_list_lock);
 				return -1;
 			}
 
 			usb_device_list = new_list;
-			usb_device_list_size += strlen(cp) + 2;
+			usb_device_list_size += cplen + 2;
 			i = 0;
 
 			while (usb_device_list[i]) {
 				i += strlen(usb_device_list + i) + 1;
 			}
 
-			strcat(usb_device_list + i, cp);
-			usb_device_list[strlen(cp) + i + 1] = 0;
+			memcpy(usb_device_list + i, cp, cplen + 1);
+			usb_device_list[i + cplen + 1] = 0;
 		}
 	}
 	ast_mutex_unlock(&usb_list_lock);
@@ -509,6 +559,7 @@ struct usb_device *ast_radio_hid_device_init(const char *desired_device)
 	struct usb_bus *usb_bus;
 	struct usb_device *dev;
 	char devstr[10000], str[200], desdev[200], *cp;
+	ssize_t linklen;
 	int i;
 
 	usb_init();
@@ -530,11 +581,14 @@ struct usb_device *ast_radio_hid_device_init(const char *desired_device)
 				if (strcasecmp(desdev, devstr)) {
 					continue;
 				}
+
 				snprintf(str, sizeof(str), "/sys/class/sound/card%d/device", i);
 				memset(desdev, 0, sizeof(desdev));
-				if (readlink(str, desdev, sizeof(desdev) - 1) == -1) {
+				linklen = readlink(str, desdev, sizeof(desdev) - 1);
+				if (linklen == -1) {
 					continue;
 				}
+				desdev[linklen] = '\0';
 				cp = strrchr(desdev, '/');
 				if (!cp) {
 					continue;
@@ -542,11 +596,9 @@ struct usb_device *ast_radio_hid_device_init(const char *desired_device)
 				cp++;
 				break;
 			}
-
 			if (i >= 32) {
 				continue;
 			}
-
 			if (!strcmp(cp, desired_device)) {
 				return dev;
 			}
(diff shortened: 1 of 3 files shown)


hardenedpenguin:
Addressed the three actionable CodeRabbit items from #1112 in `2afc526`:

1. **Mixer-limit log spam** — `ast_radio_init_mixer_limits()` no longer logs on every retry; it returns `-1` and lets the channel driver log once. The simpleusb caller-side gate is in #1112 (`2aaf9d4`); usbradio already used `device_error` the same way.
2. **`hw:` parse overflow** — Added bounded `parse_hw_field()` (card ≤ 9999, subdevice ≤ 255); invalid tokens are rejected instead of truncated.
3. **Underflow recovery** — `ast_radio_pa_write()` now propagates a failed recovery `Pa_WriteStream()` instead of always returning the original underflow.

Branches 2/3 were rebased onto this commit.

hardenedpenguin:
Follow-up in `b03769c`: reject trailing junk after `hw:C` / `hw:C,D` tokens.

`ast_radio_parse_hw_anywhere()` now requires a valid boundary (NUL, whitespace, or common PA name punctuation like `)`) after the parsed fields, so config values such as `hw:1junk` no longer silently bind to card 1. Embedded selectors like `(hw:2,0)` still match.

This completes the parser hardening called out in review (overflow bounds + token boundary). Branches 2/3 rebased.

res/res_usbradio.c
```diff
@@ -925,6 +974,408 @@ static void cleanup_user_devices(void)
...
+			ast_log(LOG_WARNING, "Failed to initialize PortAudio - (%d) %s\n", res, Pa_GetErrorText(res));
+			return -1;
+		}
+	}
+	pa_refcount++;
+	ast_mutex_unlock(&pa_lock);
+	return 0;
+}
+
+void ast_radio_pa_shutdown(void)
```
Allan-N: This function looks to be only used by res_usbradio.c.  Should it be marked `static` (and removed from the .h) ?

... and what are we shutting down?  and how does this function differentiate itself from `ast_radio_pa_shutdown_all()` ?  Are we adding/removing an instance?  adding/removing a reference?  ...?

res/res_usbradio.c
```diff
@@ -925,6 +974,408 @@ static void cleanup_user_devices(void)
...
+	if (pa_refcount > 0) {
+		pa_refcount--;
+		if (pa_refcount == 0) {
+			Pa_Terminate();
+		}
+	}
+	ast_mutex_unlock(&pa_lock);
+}
+
+void ast_radio_pa_shutdown_all(void)
```
Allan-N: This function looks to be only used by res_usbradio.c.  Should it be marked `static` (and removed from the .h) ?

res/res_usbradio.c
```diff
@@ -925,6 +974,408 @@ static void cleanup_user_devices(void)
...
+
+static int64_t pa_now_ms(void)
+{
+	struct timespec ts;
+
+	clock_gettime(CLOCK_MONOTONIC, &ts);
+	return (int64_t) ts.tv_sec * 1000 + ts.tv_nsec / 1000000;
+}
+
+int ast_radio_pa_startup(void)
```
Allan-N: This function looks to be only used by res_usbradio.c. Should it be marked static (and removed from the .h) ?

Also, I will comment here that I'm not wild about the function name. What are we starting up? Are we adding/removing an instance? adding/removing a reference? ...?

res/res_usbradio.c
```diff
@@ -1047,6 +1498,7 @@ static int load_module(void)
 static int unload_module(void)
 {
 	cleanup_user_devices();
+	ast_radio_pa_shutdown_all();
```
Allan-N: Seems like stopping PA when it's still being used by SimpleUSB (or USBRadio) would leave us in a state that we can never recover from.  Or does the module know that it is not currently in use?  And, if the module can only be unloaded when not in use then is there a need to shutdown any/all instances/references?

res/res_usbradio.c
```diff
@@ -117,9 +119,46 @@ long ast_radio_lround(double x)
 
 int ast_radio_make_spkr_playback_value(int spkrmax, int request_value, int devtype)
 {
+	if (spkrmax <= 0) {
```
Allan-N: Have we already checked that we have a valid [positive] value?  If so, can we remove the added check?

Note: if we remove the check it's OK to leave a comment stating the value is known to be good.

res/res_usbradio.c
```diff
@@ -117,9 +119,46 @@ long ast_radio_lround(double x)
...
+		return 0;
+	}
 	return (request_value * spkrmax) / AUDIO_ADJUSTMENT;
 }
 
+int ast_radio_init_mixer_limits(int devnum, int *micmax, int *spkrmax, int *micplaymax, int *newname)
+{
+	int rv;
+
+	if (!micmax || !spkrmax || !micplaymax || !newname) {
```
Allan-N: Checking for coding errors where a pointer has not been passed?  To me, this is just extra code that serves little purpose other than to satisfy those who believe that every possible error condition needs to be tested, that log messages should be emitted telling you exactly what's going on, and that the program should never crash ... even though the code won't behave as it should.  If --I--, as the developer making changes, pass a NULL value then --I-- think it's OK to have the code crash. But, those are just my thoughts.

res/res_usbradio.c
```diff
@@ -464,11 +511,11 @@ int ast_radio_hid_device_mklist(void)
...
 				snprintf(str, sizeof(str), "/sys/class/sound/card%d/device", i);
 				memset(desdev, 0, sizeof(desdev));
-
-				if (readlink(str, desdev, sizeof(desdev) - 1) == -1) {
+				linklen = readlink(str, desdev, sizeof(desdev) - 1);
+				if (linklen == -1) {
 					continue;
 				}
-
+				desdev[linklen] = '\0';
```
Allan-N, line 511: Earlier, we called `memset()`.  Is this needed?  or can we remove the earlier `memset()` call?

res/res_usbradio.c
```diff
@@ -530,23 +578,24 @@ struct usb_device *ast_radio_hid_device_init(const char *desired_device)
...
 				}
+
 				snprintf(str, sizeof(str), "/sys/class/sound/card%d/device", i);
 				memset(desdev, 0, sizeof(desdev));
-				if (readlink(str, desdev, sizeof(desdev) - 1) == -1) {
+				linklen = readlink(str, desdev, sizeof(desdev) - 1);
+				if (linklen == -1) {
 					continue;
 				}
+				desdev[linklen] = '\0';
```
Allan-N, line 580: Earlier, we called memset(). Is this needed? or can we remove the earlier memset() call?

hardenedpenguin:
Thanks for the review, Allan. Responses below follow your inline comments.

### `ast_radio_pa_startup()` / `ast_radio_pa_shutdown()` / `ast_radio_pa_shutdown_all()` (lines 991, 1009, 1021)

You're right that these are only called from inside `res_usbradio.c` today (`ast_radio_pa_open()` / `ast_radio_pa_stop()`). Nothing in the channel drivers calls them directly.

What they actually manage is **PortAudio library lifecycle** — a refcount around `Pa_Initialize()` / `Pa_Terminate()`, not individual streams. Stream open/close is `ast_radio_pa_open()` / `ast_radio_pa_stop()`.

- `ast_radio_pa_shutdown()` — balanced decrement; calls `Pa_Terminate()` only when refcount reaches 0 (normal stream teardown path)
- `ast_radio_pa_shutdown_all()` — unconditional `Pa_Terminate()` and refcount reset; intended only for `res_usbradio` module unload

Agree the names don't convey that. Plan:

1. **Make all three `static`** and remove them from `res_usbradio.h`
2. **Rename** to something like `pa_lib_acquire()` / `pa_lib_release()` / `pa_lib_release_all()` (internal names, since they'll be static)

If you'd rather collapse acquire/release entirely into `ast_radio_pa_open()` / `ast_radio_pa_stop()` and drop the separate helpers, that works for me too — say the word.

---

### `ast_radio_pa_shutdown_all()` on module unload (line 1501)

Your concern is valid in isolation: calling `Pa_Terminate()` while a channel still has an open stream would be bad.

In practice, both `chan_simpleusb` and `chan_usbradio` declare `.requires = "res_usbradio"`, so Asterisk unloads the channel drivers **before** `res_usbradio`. Each stream close goes through `ast_radio_pa_stop()` → balanced release, so refcount should already be 0 when `unload_module()` runs.

`shutdown_all()` is really a safety net for a refcount leak or an imbalanced open/fail path — not the primary teardown mechanism.

Options I'd take your steer on:

- **Keep it** as belt-and-suspenders on unload
- **Replace it** with a check: if `pa_refcount != 0`, log a warning and then terminate (or assert in debug builds)
- **Remove it** entirely if you feel unload-order guarantees are sufficient and we'd rather surface imbalance as a bug

Which do you prefer?

---

### `spkrmax <= 0` guard in `ast_radio_make_spkr_playback_value()` (line 122)

Fair point. With fail-fast `ast_radio_init_mixer_limits()`, callers should never reach this with `spkrmax <= 0` on an operational channel.

I'll **remove the check** and add a brief comment that `spkrmax` is validated at init time.

---

### NULL pointer checks in `ast_radio_init_mixer_limits()` (line 132)

Agree — these are caller-contract checks, not runtime recovery. If we pass NULL, crashing is acceptable.

I'll **remove the four output-pointer NULL checks**.

---

### Redundant `memset()` before `readlink()` (lines 518, 588)

You're correct. The buffer is reused after `read_card_usbbus()`, but `readlink()` overwrites it and we always null-terminate at `linklen`. The `memset()` adds nothing.

I'll **remove both** `memset()` calls.

---

### Summary of planned changes

| Item | Action |
|------|--------|
| PA startup/shutdown API | `static`, remove from `.h`, rename for clarity |
| `shutdown_all()` on unload | Awaiting your preference (keep / log-if-nonzero / remove) |
| `spkrmax <= 0` guard | Remove + comment |
| NULL checks on mixer init outputs | Remove |
| `memset()` before `readlink()` | Remove (both sites) |

Follow-up commit on this branch implements all items except the unload behavior, pending your preference on that one.

Happy to adjust further once you weigh in on the unload behavior and whether you want the PA refcount helpers inlined vs. kept as static internal functions.

res/res_usbradio.c
Allan-N: >In practice, both chan_simpleusb and chan_usbradio declare .requires = "res_usbradio", so Asterisk unloads the channel drivers before res_usbradio. Each stream close goes through ast_radio_pa_stop() → balanced release, so refcount should already be 0 when unload_module() runs.
>
>shutdown_all() is really a safety net for a refcount leak or an imbalanced open/fail path — not the primary teardown mechanism.
>
>Options I'd take your steer on:
>
>- Keep it as belt-and-suspenders on unload
>- Replace it with a check: if pa_refcount != 0, log a warning and then terminate (or assert in debug builds)
>- Remove it entirely if you feel unload-order guarantees are sufficient and we'd rather surface imbalance as a bug
>
>Which do you prefer?

As much as I'd like to trust that we won't get an unload request it's probably worth adding a simple check to let us know about a difference in our expectations.  Maybe replace the call to ast_radio_pa_shutdown_all() with :
```
    ast_assert(pa_refcount == 0);
```

hardenedpenguin:
Done — replaced `pa_lib_release_all()` in `unload_module()` with `ast_assert(pa_refcount == 0)` under `pa_lock`, and removed the now-unused `pa_lib_release_all()` helper.

When refcount is balanced (normal path), `Pa_Terminate()` already ran on the last stream close via `pa_lib_release()`. The assert catches imbalance during development rather than forcing teardown on unload.

hardenedpenguin:
**Update:** Added commit 1e5c822 — clamp input_channels to device maxInputChannels in ast_radio_pa_open() before Pa_OpenStream. Fixes mono SHARI capture devices that fail when opened as stereo.

res/res_usbradio.c
```diff
@@ -1241,6 +1241,15 @@ PaError ast_radio_pa_open(struct ast_radio_pa_stream *ps)
 			return paDeviceUnavailable;
 		}
 
+		{
+			const PaDeviceInfo *in_dev = Pa_GetDeviceInfo(input_params.device);
+
+			if (in_dev && ps->input_channels > (unsigned int) in_dev->maxInputChannels) {
```
mkmer: I think in "other" parts of the code, we are now expecting this to be a mono channel.  We should probably only request 1 channel, OR if we go with more than 1 we need to update the down/up conversions that expect 1 channel.
This is the one that jumps to mind:
https://github.com/AllStarLink/app_rpt/blob/658d6db3786b588065549684cb89ac5e1acf2e06/res/res_usbradio.c#L798-L814

hardenedpenguin: usbradio requests stereo where supported; ast_radio_pa_open() clamps to device max (including default path — fixing that in #1111). #1113 reads mono when clamped to 1ch, duplicates to stereo workspace, then all downstream (check_audio, PmrRx, scaling) uses the stereo path. simpleusb stays mono end-to-end (#1112). Adding API contract comments in res_usbradio.h.

res/res_usbradio.c
```diff
@@ -869,6 +910,20 @@ int ast_radio_check_audio(short *sbuf, struct audiostatistics *o, short len, sho
 	return (seq_clips >= CLIP_EVENT_MIN_SAMPLES);
 }
 
+int ast_radio_check_audio_pa_rx(short *sbuf, struct audiostatistics *o, unsigned int input_channels)
```
Allan-N: I'm questioning the value of this helper function.

Right now, you have the caller using :
```
                        if (ast_radio_check_audio_pa_rx(o->simpleusb_read_buf, &o->rxaudiostats, o->pa.input_channels)) {
```

Couldn't the caller use :
```
                        mono = (o->pa.input_channels == 1);
                        if (ast_radio_check_audio(o->simpleusb_read_buf,
                            &o->rxaudiostats,
                            mono ? AST_RADIO_PA_48K_MONO_SAMPLES : AST_RADIO_PA_48K_STEREO_SAMPLES,
                            mono)) {
```

res/res_usbradio.c
```diff
@@ -869,6 +910,20 @@ int ast_radio_check_audio(short *sbuf, struct audiostatistics *o, short len, sho
...
+{
+	if (input_channels == 1) {
+		return ast_radio_check_audio(sbuf, o, AST_RADIO_PA_48K_MONO_SAMPLES, 1);
+	}
+
+	return ast_radio_check_audio(sbuf, o, AST_RADIO_PA_48K_STEREO_SAMPLES, 0);
+}
+
+int ast_radio_check_audio_stereo_48k(short *sbuf, struct audiostatistics *o)
```
Allan-N: How does this function help?

res/res_usbradio.c
```diff
@@ -925,6 +980,418 @@ static void cleanup_user_devices(void)
...
+
+		*card = c;
+		*dev = d;
+		return 1;
+	}
+
+	return 0;
+}
+
+int ast_radio_hw_match(const char *haystack, const char *needle)
```
Allan-N: This function does not appear to be used by SimpleUSB or USBRadio.  Should this be 'static' ?

res/res_usbradio.c
```diff
@@ -925,6 +980,418 @@ static void cleanup_user_devices(void)
...
+	}
+	if (isspace((unsigned char) c)) {
+		return 1;
+	}
+
+	/* Trailing punctuation in PA/ALSA device name strings, not part of hw:C,D. */
+	return strchr("):],;.-", c) != NULL;
+}
+
+int ast_radio_parse_hw_anywhere(const char *s, int *card, int *dev)
```
Allan-N: I'm not wild about this function name.
- what string are we parsing?
- how does having "anywhere" in the name clarify the behavior?
- are we being passed a device string (ast_radio_parse_device_string)?

mkmer: Looks like this "was" from the simplusb version: (and lost the docs?)
```
/*!
 * \brief Parse "hw:<card>" or "hw:<card>,<dev>" from anywhere in s.
 * \retval 1 if found; sets *card; sets *dev to parsed value or -1 if absent.
 */
static int parse_hw_anywhere(const char *s, int *card, int *dev)
```

res/res_usbradio.c
hardenedpenguin: Yeah, fair. It’s basically a one-liner wrapper. Happy to drop it and have SimpleUSB call `ast_radio_check_audio()` directly with the mono/stereo sample counts like you sketched — less API noise for no real gain.

res/res_usbradio.c
hardenedpenguin: Same boat as the other helper — it just hard-wires the stereo 48k args. Usbradio (and SimpleUSB TX) use it in a few spots, but that’s convenience more than necessity. If we’re killing `check_audio_pa_rx`, I’d rather yank this one too and call `ast_radio_check_audio(..., AST_RADIO_PA_48K_STEREO_SAMPLES, 0)` at the call sites so we don’t keep a half-measure API.

res/res_usbradio.c
hardenedpenguin: It does get used outside this file — SimpleUSB and USBRadio both call it when they parse `hw_device` (on the follow-up branches). `ast_radio_hw_match()` also leans on it internally. So I’d keep it exported rather than making it static.

res/res_usbradio.c
hardenedpenguin: Agreed the name is a bit muddy. As Mike noted, it came over from the old SimpleUSB helper and we lost the brief comment along the way.

What it actually does: dig through a string (ALSA/PortAudio device name, config value, etc.) and pull out `hw:<card>` or `hw:<card>,<dev>` from wherever it sits in that string — including things like `(hw:2,0)`.

I’ll put that doc block back on the declaration. If you’ve got a preferred name I’m fine renaming it across the stack; otherwise I’m leaning toward keeping the name for now and just fixing the docs so it’s obvious.

res/res_usbradio.c
Allan-N: 'tis not a biggie but, IMHO, less API noise is better.

res/res_usbradio.c
Allan-N: Maybe `ast_radio_parse_alsa_device()`? or `ast_radio_parse_alsa_hw_device()`?

res/res_usbradio.c
Allan-N: SimpleUSB and USBRadio appear to use `ast_radio_parse_hw_anywhere()`, not `ast_radio_hw_match()`

hardenedpenguin: You are seeing the current tree, not the follow up where we handle simpleusb and usbradio ... this whole PR is about nailing down the shared api, it still requires the refactor for both channels which are draft at moment, so you all are not reviewing code twice.

Allan-N: I was looking at (not reviewing) the followup PRs (in their current "draft" state).  If more changes will be needed to those PRs and if the `ast_radio_hw_match()` API will be used then my observation is premature.

res/res_usbradio.c
hardenedpenguin: Went ahead and dropped both wrappers. Call sites now hit `ast_radio_check_audio()` directly.

res/res_usbradio.c
hardenedpenguin: Same — `check_audio_stereo_48k` is gone too. Stereo call sites just pass `AST_RADIO_PA_48K_STEREO_SAMPLES` themselves now.

res/res_usbradio.c
hardenedpenguin: You’re right that the channel drivers don’t call `hw_match` — only the parse helper. I made `hw_match` static (renamed internally to `pa_hw_match`) and pulled it out of the header. The exported piece for SimpleUSB/USBRadio is the parse function.

res/res_usbradio.c
hardenedpenguin: Went with `ast_radio_parse_alsa_hw_device()` and put the old doc block back on it. Updated the follow-up branches to match so we’re not renaming this twice.

mkmer (approved):

Allan-N (approved):

PR #1112, opened 2026-07-17
chan_simpleusb: Migrate to shared PortAudio API (2/3)

hardenedpenguin:
## Summary
- Migrate `chan_simpleusb` off OSS `/dev/dsp` to shared `ast_radio_pa_*` helpers
- Fail device init when ALSA mixer limits are unavailable
- Document optional `audiodev=hw:N` in `simpleusb.conf.sample`

Standalone piece **2/3**. Depends on #1111 — merge that first, then rebase this branch onto `master` before final merge.

**Review focus:** `channels/chan_simpleusb.c`, `channels/Makefile.diff`, `configs/samples/simpleusb.conf.sample` (compare against `portaudio/1-res-usbradio` for scope-only diff).

## Test plan
- [ ] CI passes
- [ ] `chan_simpleusb.so` loads; audio via PortAudio on test node
- [ ] Mixer init fails cleanly when ALSA limits unavailable

## Summary by CodeRabbit

- **New Features**
  - Added USB radio channel support using the shared USB audio system.
  - Added an `audiodev` configuration option for selecting audio devices by ALSA card or device number.
  - Improved USB audio device and mixer configuration.

- **Bug Fixes**
  - Improved audio stream startup, shutdown, and recovery when USB audio becomes unavailable.
  - Improved channel cleanup during hangup and module shutdown.
  - Removed the direct PortAudio dependency from the simple USB channel.

### DIFF
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -1647,17 +1647,7 @@ static int soundcard_writeframe(struct chan_simpleusb_pvt *o, short *data)
 	 */
 
 	res = ast_radio_pa_write(&o->pa, data, AST_RADIO_PA_FRAMES_PER_BUFFER);
-	if (res == paOutputUnderflowed) {
-		short null_buf[AST_RADIO_PA_FRAMES_PER_BUFFER * AST_RADIO_PA_OUTPUT_CHANNELS] = { 0 };
-
-		/*
-		 * Prime the stream with one silence frame so the USB buffer does not
-		 * stay empty (choppy TX). Same recovery as pre-PortAudio-API simpleusb;
-		 * move into ast_radio_pa_write() in PR #1113 so usbradio shares it.
-		 */
-		ast_debug(6, "PortAudio write stream underflow, writing a 0 frame");
-		ast_radio_pa_write(&o->pa, null_buf, AST_RADIO_PA_FRAMES_PER_BUFFER);
-	} else if (res < 0) {
+	if (res < 0 && res != paOutputUnderflowed) {
 		ast_debug(2, "Pa_WriteStream Error %s", Pa_GetErrorText(res));
 	}
 
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -1942,7 +1942,6 @@ static int simpleusb_call(struct ast_channel *c, const char *dest, int timeout)
 {
 	struct chan_simpleusb_pvt *o = ast_channel_tech_pvt(c);
 	int res;
-	int hid_started_here = 0;
 
 	o->stophidthread = 0;
 	o->stopaudiothread = 0;
@@ -1950,24 +1949,19 @@ static int simpleusb_call(struct ast_channel *c, const char *dest, int timeout)
 
 	ast_radio_time(&o->lasthidtime);
 
-	if (o->hidthread == AST_PTHREADT_NULL) {
-		res = ast_pthread_create(&o->hidthread, NULL, hidthread, o);
-		if (res) {
-			ast_log(LOG_ERROR, "Channel %s: Failed to create HID thread: %s\n", o->name, strerror(res));
-			return -1;
-		}
-		hid_started_here = 1;
+	res = ast_pthread_create(&o->hidthread, NULL, hidthread, o);
+	if (res) {
+		ast_log(LOG_ERROR, "Channel %s: Failed to create HID thread: %s\n", o->name, strerror(res));
+		return -1;
 	}
 
 	res = ast_pthread_create(&o->audiothread, NULL, simpleusb_audio_thread, o);
 	if (res) {
 		ast_log(LOG_ERROR, "Channel %s: Failed to create audio thread: %s\n", o->name, strerror(res));
-		if (hid_started_here) {
-			o->stophidthread = 1;
-			if (o->hidthread != AST_PTHREADT_NULL) {
-				pthread_join(o->hidthread, NULL);
-				o->hidthread = AST_PTHREADT_NULL;
-			}
+		o->stophidthread = 1;
+		if (o->hidthread != AST_PTHREADT_NULL) {
+			pthread_join(o->hidthread, NULL);
+			o->hidthread = AST_PTHREADT_NULL;
 		}
 		return -1;
 	}
@@ -1997,13 +1991,18 @@ static int simpleusb_hangup(struct ast_channel *c)
 	struct chan_simpleusb_pvt *o = ast_channel_tech_pvt(c);
 
 	o->stopaudiothread = 1;
-	/* Wake HID's poll so PTT drops promptly instead of waiting for HID_POLL_RATE. */
+	o->stophidthread = 1;
+	/* Wake HID's poll so it notices stophidthread / can drop PTT promptly. */
 	kickptt(o);
 	if (o->audiothread != AST_PTHREADT_NULL) {
 		pthread_join(o->audiothread, NULL);
 		o->audiothread = AST_PTHREADT_NULL;
 	}
 	ast_radio_pa_stop(&o->pa);
+	if (o->hidthread != AST_PTHREADT_NULL) {
+		pthread_join(o->hidthread, NULL);
+		o->hidthread = AST_PTHREADT_NULL;
+	}
 
 	o->owner = NULL;
 	ast_channel_tech_pvt_set(c, NULL);
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -414,6 +414,8 @@ static int start_stream(struct chan_simpleusb_pvt *pvt)
 		return -1;
 	}
 
+	ast_assert(pvt->pa.input_channels == 1);
+
 	res = ast_radio_pa_start(&pvt->pa);
 	if (res != paNoError) {
 		ast_log(LOG_WARNING, "Failed to start stream - (%d) %s\n", res, Pa_GetErrorText(res));
@@ -1637,11 +1639,11 @@ static void *hidthread(void *arg)
 
 /*!
  * \brief Write a full frame of audio data to the sound card device.
- * \note The input data must be formatted as stereo at 48000 samples per second.
- *		 FRAME_SIZE * 2 * 2 * 6 (2 bytes per sample, 2 channels, 6 for upsample to 48K)
+ * \note data is 48 kHz stereo interleaved. ast_radio_pa_write() takes frames
+ *       per channel (PA_NUM_FRAMES); the buffer holds PA_NUM_FRAMES * 2 samples.
  * \param o		Channel private data.
  * \param data	Audio data to write.
- * \returns		Number bytes written.
+ * \returns		PaError from ast_radio_pa_write().
  */
 static int soundcard_writeframe(struct chan_simpleusb_pvt *o, short *data)
 {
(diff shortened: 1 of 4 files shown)


channels/chan_simpleusb.c
```diff
@@ -107,29 +98,7 @@
...
- * This is 1920 bytes for stereo, or 960 bytes for mono
- */
-#define PA_NUM_FRAMES FRAME_SIZE * 6 /* Number of samples to read/write from PortAudio stream per Asterisk frame at 48k */
-
-/*! \brief Mono Input */
-#define PA_INPUT_CHANNELS 1 /* PortAudio: Mono input for Microphone / RX */
-
-/*! \brief Stereo Output */
-#define PA_OUTPUT_CHANNELS 2 /* PortAudio: Stereo output for Speaker / TX */
+#define PA_NUM_FRAMES AST_RADIO_PA_FRAMES_PER_BUFFER
```
mkmer: Should we just use `AST_RADIO_PA_FRAMES_PER_BUFFER`?

channels/chan_simpleusb.c
```diff
@@ -428,267 +396,39 @@ static struct ast_channel_tech simpleusb_tech = {
...
+		ast_radio_pa_stop(&pvt->pa);
 		return -1;
 	}
 
-	pvt->streamstate = 1;
 	return 0;
 }
 
-static int stop_stream(struct chan_simpleusb_pvt *pvt)
+static void stop_stream(struct chan_simpleusb_pvt *pvt)
```
mkmer: Do we need this function?  Or just call `ast_radio_pa_stop()`

channels/chan_simpleusb.c
```diff
@@ -4658,21 +4422,30 @@ static int unload_module(void)
...
+		stop_stream(o);
 		if (o->hidthread != AST_PTHREADT_NULL) {
 			pthread_join(o->hidthread, NULL);
 			o->hidthread = AST_PTHREADT_NULL;
 		}
 		if (o->dsp) {
 			ast_dsp_free(o->dsp);
 		}
+		for (i = 0; i < GPIO_PINCOUNT; i++) {
+			ast_free(o->gpios[i]);
```
mkmer: This brings up a "few" questions:
1 - we should check o->gpios[i] is not NULL before freeing?
2 - we don't appear to initialize this array when creating it - we should as we test it for `!o->gpios[i]` as well as `strcasecmp()` against not NULL.

channels/chan_simpleusb.c
```diff
@@ -4658,21 +4422,30 @@ static int unload_module(void)
...
+		}
+		for (i = 0; i < ARRAY_LEN(o->pps); i++) {
+			ast_free(o->pps[i]);
```
mkmer: Same issue here as with `gpios[]`?  
Init, and check for NULL

channels/chan_simpleusb.c
hardenedpenguin: Good call — dropped the local alias and just use `AST_RADIO_PA_FRAMES_PER_BUFFER` everywhere now.

channels/chan_simpleusb.c
hardenedpenguin: Yeah, it was just a one-liner wrapper. Removed it and call `ast_radio_pa_stop()` directly at the hangup/unload sites.

channels/chan_simpleusb.c
hardenedpenguin: On init: the pvt comes from `ast_calloc()`, so `gpios[]` starts as NULL already.

On free: fair point for clarity — unload now only frees slots that were actually set.

channels/chan_simpleusb.c
hardenedpenguin: Same treatment for `pps[]` — calloc zeroes them on create, and unload now checks before free.

configs/samples/simpleusb.conf.sample
```diff
@@ -117,6 +117,8 @@ legacyaudioscaling = no     ; If yes, continue to do raw audio sample scaling an
 ;;;;; Tune settings ;;;;;
 ;devstr=
 ;serial=
+;audiodev=hw:0             ; Optional ALSA device for PortAudio audio (hw:N or hw:N,M).
```
Allan-N: Needs 1 space before the ; to align with the other comments

channels/chan_simpleusb.c
```diff
@@ -2244,16 +1997,12 @@ static int simpleusb_hangup(struct ast_channel *c)
 	struct chan_simpleusb_pvt *o = ast_channel_tech_pvt(c);
 
 	o->stopaudiothread = 1;
+	kickptt(o);
 	if (o->audiothread != AST_PTHREADT_NULL) {
 		pthread_join(o->audiothread, NULL);
 		o->audiothread = AST_PTHREADT_NULL;
 	}
-
-	o->stophidthread = 1;
```
Allan-N, line 2252: We don't need to stop the HID thread?  or did this get moved somewhere else?

channels/chan_simpleusb.c
```diff
@@ -2244,16 +1997,12 @@ static int simpleusb_hangup(struct ast_channel *c)
 	struct chan_simpleusb_pvt *o = ast_channel_tech_pvt(c);
 
 	o->stopaudiothread = 1;
+	kickptt(o);
```
Allan-N, line 1991: Why are we making this call?  ... and maybe add a comment so others can understand why?

configs/samples/simpleusb.conf.sample
hardenedpenguin: Good catch — the inline comment sat one column left of the rest of the file. Added the extra space so it lines up with the other options.

channels/chan_simpleusb.c
hardenedpenguin: Intentional — HID stays up across channel hangups now. Stopping and restarting it on every hangup/call was heavier than it needed to be, and the USB/HID side doesn’t have to tear down just because the Asterisk channel did.

Audio still stops and joins here on hangup. HID is only stopped/joined on module unload (and if call setup fails after we just started a new HID thread). Same idea as the usbradio side of this stack.

channels/chan_simpleusb.c
hardenedpenguin: HID’s main loop blocks in `poll()` on the `pttkick` pipe (up to `HID_POLL_RATE`). Without a kick, it can sit there that long before it notices the call is gone and drops PTT.

So this isn’t about stopping HID — it’s just waking that poll so PTT clears promptly on hangup. Added a short comment at the call site.

channels/chan_simpleusb.c
Allan-N: What about a `devstr` change + reload?  or a reload where a device is no longer being used?

Allan-N: Also, how often do we see a channel hangup/call?  Once at startup, once at shutdown, and when we run into a USB failure where a reset is warranted?

hardenedpenguin: Fair pushback — after looking at the hangup/call frequency and the reload/devstr gaps, the keep-alive change wasn’t earning its keep.

Reverted it: hangup stops and joins HID again (same model as the prior PortAudio simpleusb path). Audio and HID both tear down with the Asterisk channel; unload still does the same.

hardenedpenguin: Good question. Reload doesn’t really rebind devices today either way — `load_tune_config(..., reload=1)` intentionally skips `devstr`/`serial`/`audiodev`, and `module reload` doesn’t tear down an existing pvt/HID for a removed section. Device changes still need an unload/restart (or similar).

That’s a separate, pre-existing limitation. For this PR I went back to stopping HID on hangup so we don’t leave a long-lived HID thread sitting on an old assignment across channel teardown.

hardenedpenguin: Pretty much that — for an app_rpt node it’s basically once at node start, once at node teardown, plus failure/reset paths that hang up the radio channel. Not per-QSO.

That undercuts the reason for keeping HID alive across hangup, so I reverted it. Cheap to stop/restart when hangups are that rare, and the lifetime model is easier to reason about.

channels/chan_simpleusb.c
hardenedpenguin: Still needed now that hangup stops HID again — `kickptt()` wakes the HID `poll()` so it notices `stophidthread` (and can drop PTT) instead of waiting up to `HID_POLL_RATE`. Updated the comment at the call site to say that.

channels/chan_simpleusb.c
```diff
@@ -1311,20 +1057,23 @@ static int init_audio_device(struct chan_simpleusb_pvt *o)
 		}
 	}
 
+	{
+		int use_newname = 0;
```
Allan-N: Can we move this variable up (with the other declarations) and remove the `{ }` ?

hardenedpenguin: Yep — moved `use_newname` up with the other locals and dropped the block.

channels/chan_simpleusb.c
mkmer: > Also, how often do we see a channel hangup/call? Once at startup, once at shutdown, and when we run into a USB failure where a reset is warranted?

ANY hangups on the "core" channels (aka the radio stuff) will cause the repeater exit.  The repeater only askes to hang them up on a thread exit.

Allan-N (approved):

mkmer (approved):

channels/chan_simpleusb.c
```diff
@@ -1893,23 +1646,14 @@ static int soundcard_writeframe(struct chan_simpleusb_pvt *o, short *data)
 	 * a number of failures, to restart the output chain.
 	 */
 
-	res = Pa_WriteStream(o->stream, data, PA_NUM_FRAMES);
+	res = ast_radio_pa_write(&o->pa, data, AST_RADIO_PA_FRAMES_PER_BUFFER);
 	if (res == paOutputUnderflowed) {
-		short null_buf[PA_NUM_FRAMES * 2] = { 0 };
 		ast_debug(6, "PortAudio write stream underflow, writing a 0 frame");
```
Allan-N: Here, we are logging a message about and underflow and writing a zero frame but we are no longer writing the frame.  Do we need to write an empty frame?  correct the log?  ...?

Note: the rabbit commented about the underflow

mkmer: If we underflow, we need to add 1 null frame.  IF this is removed, the choppy audio problem will return.

mkmer: https://github.com/AllStarLink/app_rpt/pull/593
https://github.com/AllStarLink/app_rpt/pull/598

hardenedpenguin: Good catch — the log was left behind when the write moved to `ast_radio_pa_write()`, but the silence-frame recovery did not come along. Restored the one null-frame write on underflow here (same as the old simpleusb path / #593).

We’ll move that recovery into `ast_radio_pa_write()` in #1113 so usbradio shares it instead of duplicating the call-site logic.

Allan-N: Curious as to why we're writing/logging the null frame here, as opposed to ast_radio_pa_write().  Yes, could defer the write to #1113 but if you moved the write to this PR/branch then it's in the right place from the start and #1113 gets the change for free.

hardenedpenguin: Fair point — no reason to defer it. Moved the silence-frame recovery into `ast_radio_pa_write()` on this branch and dropped the call-site copy in simpleusb. #1113 will pick it up from there (and can drop its local duplicate).

Allan-N (approved):

res/res_usbradio.c
```diff
@@ -1361,6 +1361,9 @@ PaError ast_radio_pa_read(struct ast_radio_pa_stream *ps, short *buf, unsigned l
 
 PaError ast_radio_pa_write(struct ast_radio_pa_stream *ps, const short *data, unsigned long frames)
 {
+	PaError res;
+	short null_buf[AST_RADIO_PA_FRAMES_PER_BUFFER * AST_RADIO_PA_OUTPUT_CHANNELS];
```
mkmer: This was reduced scope in the `if` statement, and is only used there.  Why move it out?
Why not use the original init method as well? No need for memset
```	
		short null_buf[PA_NUM_FRAMES * 2] = { 0 };
```
https://github.com/AllStarLink/app_rpt/blob/2c3a7b167b4e429f700f31984a9645f70af56b47/channels/chan_simpleusb.c#L1896-L1904

mkmer (approved):

PR #1113, opened 2026-07-17
chan_usbradio: PortAudio migration and install/docs (3/3)

### LINKED ISSUE #1024: txtoctype=notone and txtoctype=phase have no effect in chan_usbradio — TX CTCSS encodes through entire hang time
Subject: Bug Report: txtoctype=notone and txtoctype=phase have no effect in chan_usbradio — TX CTCSS encodes through entire hang time

Hello ASL3 Development Team,

I would like to report a bug in chan_usbradio where the txtoctype setting has no effect on TX CTCSS behavior.

System Information:
- OS: Debian GNU/Linux 13 (trixie)
- Asterisk: 22.8.2+asl3-3.8.3-1.deb13
- Hardware: Raspberry Pi (arm64)
- Radio Interface: Repeater Builder Max-Trac RM board (CM119 chip)
- TX Radio: Motorola CDM1550 (CSQ mode, CTCSS generated by USBRadio)
- RX Radio: Motorola Max-Trac (CTCSS output on pin 8, active high)

Bug Description:
The TX CTCSS tone continues encoding through the entire hang time until PTT drops, regardless of the txtoctype setting. Both notone and phase have been tested with no effect. The PL tone only stops when PTT drops.

Configuration:
carrierfrom = dsp
ctcssfrom = usb
rxdemod = flat
duplex = 2
txtoctype = notone (also tested phase — same result)
txctcssdefault commented out
txctcssfreqs = 114.8
txmixa = composite
txmixb = no
txprelim = yes

Expected Behavior:
With txtoctype=notone, the TX CTCSS tone should drop at the moment COS drops (when the user unkeys), before the hang time expires and before PTT drops.

With txtoctype=phase, a reverse burst should be sent before PTT drops.

Actual Behavior:
In both cases the CTCSS tone stays on through the entire hang time and only stops when PTT finally drops. Verified using a Motorola R2660C service monitor watching the TX output.

Impact:
This causes a squelch crash on all receiving radios using CTCSS squelch when the TX carrier drops. This is particularly severe for blind amateur radio operators who rely on the repeater.

Additional Notes:
COP 58 (TX CTCSS on input only enable) was also tested with no improvement.
txctcssdefault was commented out per forum suggestions — no improvement.
DSP carrier detect is confirmed working correctly — COS properly follows the received signal.

Thank you for looking into this issue.

73,
Dean
K9QB

hardenedpenguin:
## Summary
- Migrate `chan_usbradio` off OSS `/dev/dsp` to shared `ast_radio_pa_*` helpers
- Document optional `audiodev=hw:N` in `usbradio.conf.sample`
- Update `rpt_install.sh`: drop `snd-pcm-oss`, add PortAudio deps, `mktemp` for Makefile diffs
- Fix `txtoctype=notone` / `phase` so CTCSS turn-off starts on local COS drop when full duplex (does not ride `app_rpt` hangtime). Fixes #1024.

Standalone piece **3/3**. Depends on #1111 and #1112 (both merged); rebased onto current `master`.

**Review focus:** `channels/chan_usbradio.c`, `channels/xpmr/xpmr.c`, `configs/samples/usbradio.conf.sample`, `rpt_install.sh`.

## Test plan
- [x] CI passes
- [x] `chan_usbradio.so` loads; audio via PortAudio on test node (crazytrain / node 546050)
- [ ] `rpt_install.sh` on dev tree: no `snd-pcm-oss`, PortAudio packages installed
- [ ] With `duplex=1`, `txtoctype=notone` (or `phase`), and TX CTCSS encode: tone drops when local COS drops, before hangtime/PTT drop (service monitor or CTCSS RX)
- [ ] Half duplex still applies ~600ms burst when PTT is released

## Summary by CodeRabbit

* **New Features**
  * USB radio audio now streams via PortAudio for improved ALSA device selection and more reliable audio handling.
  * Added an `audiodev` option to target a specific ALSA audio device (instead of `devstr`).
  * Installation now includes USB radio and PortAudio development/runtime support.
* **Bug Fixes**
  * Enhanced PortAudio handling for underflow, timeout, and overflow scenarios.
  * Improved shutdown/teardown to stop streaming reliably and recover cleanly.
  * Fixed CTCSS transmit/TOC behavior during duplex carrier loss (“hang-muted” handling).
* **Documentation**
  * Updated the sample configuration with `audiodev` guidance and clearer tone burst behavior.

### DIFF
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -1890,7 +1890,12 @@ static int usbradio_call(struct ast_channel *c, const char *dest, int timeout)
 
 	o->stophid = 0;
 	ast_radio_time(&o->lasthidtime);
-	ast_pthread_create(&o->hidthread, NULL, hidthread, o);
+	if (o->hidthread == AST_PTHREADT_NULL) {
+		if (ast_pthread_create(&o->hidthread, NULL, hidthread, o)) {
+			ast_log(LOG_ERROR, "Channel %s: Failed to create HID thread\n", o->name);
+			return -1;
+		}
+	}
 	ast_setstate(c, AST_STATE_UP);
 	return 0;
 }
@@ -1918,12 +1923,7 @@ static int usbradio_hangup(struct ast_channel *c)
 	ast_channel_tech_pvt_set(c, NULL);
 	o->owner = NULL;
 	ast_module_unref(ast_module_info->self);
-	o->stophid = 1;
 	kickptt(o);
-	if (o->hidthread != AST_PTHREADT_NULL) {
-		pthread_join(o->hidthread, NULL);
-		o->hidthread = AST_PTHREADT_NULL;
-	}
 	if (o->hookstate) {
 		o->hookstate = 0;
 	}
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -939,13 +939,19 @@ static void *hidthread(void *arg)
 			int subdev;
 
 			if (ast_radio_parse_hw_anywhere(o->hw_device, &i, &subdev)) {
+				int card_in_use = 0;
+
 				for (ao = usbradio_default.next; ao && ao->name; ao = ao->next) {
 					if (ao != o && ao->usbass && ao->devicenum == i) {
 						ast_log(LOG_ERROR, "Channel %s: Audio device %s is already assigned to channel %s\n", o->name, o->hw_device, ao->name);
-						ast_mutex_unlock(&usb_dev_lock);
-						usleep(500000);
-						continue;
+						card_in_use = 1;
+						break;
 					}
+				}
+				if (card_in_use) {
+					ast_mutex_unlock(&usb_dev_lock);
+					usleep(500000);
+					continue;
 				}
 				usb_dev = ast_radio_usb_device_from_alsa_card(i);
 				if (!usb_dev) {
@@ -1647,7 +1653,7 @@ static int used_blocks(struct chan_usbradio_pvt *o)
 	}
 
 	if (avail < AST_RADIO_PA_FRAMES_PER_BUFFER) {
-		return o->total_blocks;
+		return o->total_blocks + 1;
 	}
 
 	return 0;
@@ -1692,6 +1698,7 @@ static int soundcard_writeframe(struct chan_usbradio_pvt *o, short *data)
 	res = ast_radio_pa_write(&o->pa, data, AST_RADIO_PA_FRAMES_PER_BUFFER);
 	if (res < 0 && res != paOutputUnderflowed) {
 		ast_log(LOG_ERROR, "Channel %s: PortAudio write error %s\n", o->name, Pa_GetErrorText(res));
+		usbradio_stop_audio(o);
 		return 0;
 	}
 
@@ -1913,7 +1920,10 @@ static int usbradio_hangup(struct ast_channel *c)
 	ast_module_unref(ast_module_info->self);
 	o->stophid = 1;
 	kickptt(o);
-	pthread_join(o->hidthread, NULL);
+	if (o->hidthread != AST_PTHREADT_NULL) {
+		pthread_join(o->hidthread, NULL);
+		o->hidthread = AST_PTHREADT_NULL;
+	}
 	if (o->hookstate) {
 		o->hookstate = 0;
 	}
@@ -2066,6 +2076,7 @@ static struct ast_frame *usbradio_read(struct ast_channel *c)
 		}
 		ast_log(LOG_ERROR, "Channel %s: PortAudio read error %s\n", o->name, Pa_GetErrorText(pa_res));
 		o->hasusb = 0;
+		usbradio_stop_audio(o);
 		return &ast_null_frame;
 	}
 
@@ -2599,11 +2610,6 @@ static struct ast_channel *usbradio_new(struct chan_usbradio_pvt *o, char *ext, 
 		return NULL;
 	}
 	ast_channel_tech_set(c, &usbradio_tech);
-	if (!o->pa.active && o->hasusb) {
-		if (usbradio_start_audio(o) < 0) {
-			return NULL;
-		}
-	}
 	ast_channel_nativeformats_set(c, usbradio_tech.capabilities);
 	ast_channel_set_readformat(c, ast_format_slin);
 	ast_channel_set_writeformat(c, ast_format_slin);
@@ -2612,6 +2618,12 @@ static struct ast_channel *usbradio_new(struct chan_usbradio_pvt *o, char *ext, 
 
 	o->owner = c;
 	ast_module_ref(ast_module_info->self);
+	if (!o->pa.active && o->hasusb) {
+		if (usbradio_start_audio(o) < 0) {
+			ast_hangup(c);
+			return NULL;
+		}
+	}
 	ast_jb_configure(c, &global_jbconf);
 	if (state != AST_STATE_DOWN) {
 		if (ast_pbx_start(c)) {
@@ -4886,6 +4898,7 @@ static struct chan_usbradio_pvt *store_config(struct ast_config *cfg, const char
 			o->name = ast_strdup(ctg);
 			o->pttkick[0] = -1;
 			o->pttkick[1] = -1;
+			o->hidthread = AST_PTHREADT_NULL;
 			o->hw_device[0] = '\0';
 			if (!usbradio_active) {
 				usbradio_active = o->name;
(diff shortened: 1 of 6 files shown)


hardenedpenguin:
**Update:** Added commit 2caacc2 — fixes USBRadio PTT/TX after PortAudio migration: tick pipe on channel FD, PTT processing on PA read timeout, mono RX expand, direct PortAudio TX write. Validated on crazytrain node 546050.

hardenedpenguin:
Rebased onto `master` after #1112 merged. Dropped the duplicated simpleusb commits (already on master) and kept the usbradio-only series.

Also aligned `chan_usbradio` with the Mike/Allan review outcomes from #1112:

- Underflow silence priming lives in `ast_radio_pa_write()` only (no second local `null_buf` write)
- Call `ast_radio_pa_stop()` directly instead of a thin wrapper
- Free `gpios[]` / `pps[]` on unload with NULL checks
- Hoist `use_newname` with other locals (no nested `{ }` blocks)
- Align `audiodev` sample-config comment spacing

HID stop on hangup matches the final #1112 behavior (Mike: hangups on core radio channels exit the repeater).

channels/chan_usbradio.c
```diff
@@ -96,20 +96,14 @@
 #define MS_PER_FRAME 20						   /* 20 ms frames */
 #define MS_TO_FRAMES(ms) ((ms) / MS_PER_FRAME) /* convert ms to frames */
 
+#define USBRADIO_48K_STEREO_BYTES (FRAME_SIZE * 2 * 2 * 6)
```
Allan-N: An observation.  In res_usbradio.h we have :
```
#define AST_RADIO_PA_48K_STEREO_SAMPLES (12 * FRAME_SIZE)
```
In this code we currently have :
```
#define USBRADIO_48K_STEREO_BYTES (FRAME_SIZE * 2 * 2 * 6)
...
        char usbradio_write_buf[USBRADIO_48K_STEREO_BYTES];
```
and, scattered in the code we find that most references to `usbradio_write_buf` are cast either to `(i16 *)` or `(short *)`.

Does it help to change the buffer variable declaration to :
```
        short usbradio_write_buf[AST_RADIO_PA_48K_STEREO_SAMPLES];
```
and remove the casts and `#define USBRADIO_48K_STEREO_BYTES`?

... and if this is too much change for this PR then maybe we'll look at it later :)

channels/chan_usbradio.c
```diff
@@ -182,11 +176,15 @@ static const char *const mixer_type[] = { "no", "voice", "tone", "composite", "a
...
-	char *name;		  /* the internal name of our channel */
-	int devtype;	  /* actual type of device */
-	int pttkick[2];	  /* ptt kick pipe */
-	int total_blocks; /* total blocks in the output device */
-	int sounddev;
+	char *name;			 /* the internal name of our channel */
+	char hw_device[100]; /* ALSA/PortAudio device (hw:N or hw:N,M) */
+	int devtype;		 /* actual type of device */
+	int pttkick[2];		 /* ptt kick pipe */
+	int tickpipe[2];	 /* read loop wakeup pipe */
```
Allan-N: Why are we using a pipe for ticks?  Why not use an ast_timer (see chan_echolink)?

channels/chan_usbradio.c
hardenedpenguin: Good call — done in b2d4ad6.

`usbradio_write_buf` is now `short[AST_RADIO_PA_48K_STEREO_SAMPLES]`, the local `USBRADIO_48K_STEREO_BYTES` macro is gone, and the `(i16 *)` / `(short *)` casts on that buffer are dropped.

channels/chan_usbradio.c
hardenedpenguin: Agreed — also in b2d4ad6.

Switched to `ast_timer` the same way EchoLink does it: open, `ast_timer_set_rate(1000 / MS_PER_FRAME)`, hang the FD on the channel, and `ast_timer_ack()` in `usbradio_read()`. The tick pipe/thread is gone.

hardenedpenguin:
Addressed Allan's two notes in b2d4ad6:

1. `usbradio_write_buf` → `short[AST_RADIO_PA_48K_STEREO_SAMPLES]` (shared sample count, no byte macro / casts)
2. 20 ms wakeup → `ast_timer` (EchoLink pattern) instead of tick pipe + thread

Will re-validate PTT/TX on crazytrain with this tip before merge.

channels/chan_usbradio.c
```diff
@@ -1507,206 +1651,68 @@ static void *hidthread(void *arg)
...
-	}
-	res = write(o->sounddev, ((void *) data), FRAME_SIZE * 2 * 2 * 6);
-	if (res < 0) {
-		ast_log(LOG_ERROR, "Channel %s: Sound card write error %s\n", o->name, strerror(errno));
-	} else if (res != FRAME_SIZE * 2 * 2 * 6) {
-		ast_log(LOG_ERROR, "Channel %s: Sound card wrote %d bytes of %d\n", o->name, res, (FRAME_SIZE * 2 * 2 * 6));
-	}
 
-	return res;
+	return AST_RADIO_PA_48K_STEREO_SAMPLES * (int) sizeof(short);
```
Allan-N: The function header mentions `AST_RADIO_PA_FRAMES_PER_BUFFER` but here we are using `AST_RADIO_PA_48K_STEREO_SAMPLES`.  How does one get from "a" to "b"?

One thought would be to use `AST_RADIO_PA_FRAMES_PER_BUFFER * AST_RADIO_PA_OUTPUT_CHANNELS * (int) sizeof(short)`.

Another idea would be to change the function prototype to be :
```
static int soundcard_writeframe(struct chan_usbradio_pvt *o, short *data, size_t datalen)
```
and then call the function with `soundcard_writeframe(o, o->usbradio_write_buf, sizeof(o->usbradio_write_buf);` and `return datalen;`

... and there are other places that are also using AST_RADIO_PA_48K_STEREO_SAMPLES that could use `sizeof(o->usbradio_write_buf)`.

or just leave the code as-is :)

channels/chan_usbradio.c
```diff
@@ -1089,6 +1220,12 @@ static void *hidthread(void *arg)
 		}
 		if (pipe2(o->pttkick, O_NONBLOCK) == -1) {
 			ast_log(LOG_ERROR, "Channel %s: Is not able to create a pipe\n", o->name);
+			usb_close(usb_handle);
+			usb_handle = NULL;
+			ast_mutex_lock(&usb_dev_lock);
+			o->usbass = 0;
+			o->hasusb = 0;
+			ast_mutex_unlock(&usb_dev_lock);
 			pthread_exit(NULL);
```
mkmer: We have been changing these to `return NULL` as we find them.

channels/chan_usbradio.c
```diff
@@ -1988,7 +2004,8 @@ static int usbradio_write(struct ast_channel *c, struct ast_frame *f)
  */
 static struct ast_frame *usbradio_read(struct ast_channel *c)
 {
-	int res, oldpttout;
+	PaError pa_res;
+	int oldpttout;
```
mkmer: Not saying in need to change, but most of these around the code are call "lastxxx" where old = last.

mkmer (approved):
Just nit's here.  Looks good.

channels/chan_usbradio.c
```diff
@@ -1514,32 +1627,30 @@ static void *hidthread(void *arg)
...
-		}
+		o->total_blocks = o->queuesize ? o->queuesize : QUEUE_SIZE;
+	}
+
+	if (avail < AST_RADIO_PA_FRAMES_PER_BUFFER) {
+		return o->total_blocks;
 	}
 
-	return o->total_blocks - info.fragments;
+	return 0;
```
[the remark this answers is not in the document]
Allan-N: Looks like a few changes have been made.  But, I will note that `int total_blocks;` and `#define WARN_used_blocks` are present in the source but are unreferenced.

Allan-N (approved):

channels/chan_usbradio.c
```diff
@@ -636,6 +625,41 @@ static void kickptt(const struct chan_usbradio_pvt *o)
...
+	if (ast_timer_set_rate(o->timer, rate)) {
+		ast_log(LOG_ERROR, "Channel %s: Unable to set timer rate to %d Hz.\n", o->name, rate);
+		ast_timer_close(o->timer);
+		o->timer = NULL;
+		return -1;
+	}
+	ast_channel_set_fd(c, 0, ast_timer_fd(o->timer));
+	return 0;
+}
+
```
mkmer, line 662: This timer as and FD tied to the asterisk channel is problematic.
Any call to the `.read` or `.write` functions can not be blocking.  
This architecture in the `.read` is calling a read from the USB hardware which can wait up to 40ms (on failure) and up to 19ms when the timing is "just right".
We need a thread to handle the USB hardware which puts inbound audio on the channel queue and processes outbound audio (all frames in an outbound queue, up to the hardware limit).
There will be a challenge with the XPMR "stuff" as it expects 1 frame for each read/write pair - Notice `xpmr_rx()` grabs a tx frame, except there could be 2 or 3 frames of tx.

SimpleUSB has this "configuration" (less xpmr).

PR #1116, opened 2026-07-20
app_gps: Add native gpsd JSON position source

hardenedpenguin:
## Summary

- Add optional `gpsd_host` / `gpsd_port` settings in `gps.conf` to read position from a local gpsd JSON stream (default port 2947)
- When `gpsd_host` is set, `comport` is ignored and a dedicated `gpsd_reader` thread parses gpsd TPV records
- Serial NMEA via `comport` remains unchanged when gpsd is not configured
- Improve module unload/reload: shutdown gpsd socket, join reader for either source, reset state on reload

## Test plan

- [ ] Build `app_gps.so` on target platform
- [ ] Configure `gpsd_host = 127.0.0.1` with gpsd serving a USB receiver; verify `GPS connected via gpsd` in logs
- [ ] `asterisk -rx 'gps show status'` reports locked position with valid lat/lon/elevation
- [ ] `GPS_READ()` returns valid coordinates while gpsd has a fix
- [ ] With only `comport` configured (no `gpsd_host`), serial NMEA path still works
- [ ] Module reload/unload does not hang with gpsd connected

## Summary by CodeRabbit

* **New Features**
  * Added an alternate GPS input path via a local gpsd JSON TCP stream (TPV-based).
  * GPS position sourcing can now be configured as serial, gpsd, or disabled; gpsd takes precedence when enabled.
  * GPS status/reporting reflects whether data is coming from gpsd or serial.

* **Bug Fixes**
  * Improved handling of gpsd connection failures by automatically reconnecting and marking fixes as unavailable when needed.

* **Documentation**
  * Updated GPS configuration with guidance and example `gpsd_host`/`gpsd_port` settings.

### DIFF
--- a/apps/app_gps.c
+++ b/apps/app_gps.c
@@ -40,6 +40,8 @@
  * to generate the password: https://n5dux.com/ham/aprs-passcode/
  *
  * app_gps can connect to a serial GPS receiver to get position information.
+ * If gpsd_host is configured, position is read from a local gpsd JSON stream
+ * (TCP port 2947 by default) instead of a serial NMEA device.
  * If a GPS receiver is not configured, it can provide default position information
  * entered in the gps.conf file.  It decodes NMEA-0183 GGA sentences ($GPGGA,
  * $GNGGA, etc.).
@@ -252,9 +254,12 @@
 #define TT_COMMON "/tmp/aprs_ttcommon"
 #define TT_SUB_COMMON "/tmp/aprs_ttcommon_%s"
 #define GPS_DEFAULT_BAUDRATE B4800
+#define GPS_DEFAULT_GPSD_PORT 2947
+#define GPS_MAX_GPSD_PORT 65535
 #define GPS_UPDATE_SECS 60
 #define GPS_VALID_SECS 60
 #define SERIAL_MAXMS 10000
+#define JSON_KEY(x) "\"" #x "\":"
 
 /*!
  * \brief APRS TT entry.
@@ -276,8 +281,19 @@ static int run_forever = 1;
 
 /* Global configuration information */
 static char *comport, *server, *port;
+static char *gpsd_host = NULL;
 static int baudrate;
+static int gpsd_port = GPS_DEFAULT_GPSD_PORT;
+static int gpsd_sockfd = -1;
 static int sockfd = -1;
+
+enum gps_source_type {
+	GPS_SOURCE_NONE = 0,
+	GPS_SOURCE_SERIAL,
+	GPS_SOURCE_GPSD,
+};
+
+static enum gps_source_type gps_source = GPS_SOURCE_NONE;
 
 /* Message flags */
 static int gps_unlock_shown = 0;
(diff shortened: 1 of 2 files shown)


apps/app_gps.c
```diff
@@ -888,6 +902,286 @@ static void lon_decimal_to_DMS(float dec, char *value, int len)
 	snprintf(value, len, "%03d%05.2f%c", londeg, lonmin, direction);
 }
 
+/*!
+ * \brief Extract a numeric JSON field value from a gpsd line.
+ */
+static int gpsd_json_get_double(const char *json, const char *key, double *out)
+{
+	char pattern[48];
```
Allan-N: Why "48" ?

Have you considered using a macro?
```
#define JSON_KEY(x) "\"" #x "\":"
```

and then you would use `gpsd_json_get_double(line, JSON_KEY("lat"), &lat)`

apps/app_gps.c
```diff
@@ -888,6 +902,286 @@ static void lon_decimal_to_DMS(float dec, char *value, int len)
...
+		}
+		if (c != '\r') {
+			if (i < len - 1) {
+				buf[i++] = c;
+			}
+		}
+	}
+
+	/* Line exceeded buffer; discard remainder until newline. */
+	while (run_forever) {
```
Allan-N, line 1012: Any chance we could combine this loop with the one above?  Maybe have the top loop keep going until `\n` but, as it is already, only adding what will fit into `buf[]` ?

apps/app_gps.c
```diff
@@ -888,6 +902,286 @@ static void lon_decimal_to_DMS(float dec, char *value, int len)
...
+	if (gpsd_json_get_double(line, "lat", &lat) || gpsd_json_get_double(line, "lon", &lon)) {
+		ast_debug(1, "GPS ignoring incomplete gpsd TPV: %s\n", line);
+		return;
+	}
+
+	if (gpsd_json_get_double(line, "altMSL", &elev)) {
+		gpsd_json_get_double(line, "alt", &elev);
+	}
+
+	gps_position_update((float) lat, (float) lon, (float) elev);
```
Allan-N: Seems like `lat`, `lon`, and `elev` should be `float` variables ... and we need a `gpsd_json_get_float()` function?  ... or maybe change `gpsd_json_get_double()` to `gpsd_json_get_float()`?

apps/app_gps.c
```diff
@@ -888,6 +902,286 @@ static void lon_decimal_to_DMS(float dec, char *value, int len)
...
+				sleep(2);
+				continue;
+			}
+		}
+
+		memset(&fds, 0, sizeof(fds));
+		fds[0].fd = fd;
+		fds[0].events = POLLIN;
+
+		if (poll(fds, 1, 1000) < 0) {
```
Allan-N: Do we want to use `ast_poll()` ?

apps/app_gps.c
```diff
@@ -888,6 +902,286 @@ static void lon_decimal_to_DMS(float dec, char *value, int len)
...
+			continue;
+		}
+
+		if (!(fds[0].revents & POLLIN)) {
+			time_t now_mono = time_monotonic();
+			time_t last;
+
+			ast_mutex_lock(&position_update_lock);
+			last = current_gps_position.last_updated_mono;
+			if (last + GPS_VALID_SECS < now_mono) {
```
Allan-N: Does the `last` variable help?

apps/app_gps.c
```diff
@@ -1635,8 +1929,13 @@ static char *handle_cli_status(struct ast_cli_entry *e, int cmd, struct ast_cli_
 	}
 
 	ast_mutex_lock(&position_update_lock);
-	ast_cli(a->fd, "GPS: %s, Signal: %s \n", ast_strlen_zero(comport) ? "Disconnected" : "Connected",
-		current_gps_position.is_valid ? "Locked" : "Unlocked");
+	if (gps_source == GPS_SOURCE_GPSD) {
```
Allan-N: Should we be using a `switch()` statement here in order to handle all 3 GPS_SOURCE_xxx options?

apps/app_gps.c
```diff
@@ -1725,6 +2037,32 @@ static int load_module(void)
...
 	}
+	val = ast_variable_retrieve(cfg, "general", "gpsd_host");
+	if (val && !ast_strlen_zero(val)) {
+		gpsd_host = ast_strdup(val);
+		gps_source = GPS_SOURCE_GPSD;
+	}
+	val = ast_variable_retrieve(cfg, "general", "gpsd_port");
+	if (val && !ast_strlen_zero(val)) {
+		char *endptr = NULL;
+		long parsed;
```
Allan-N: 'tis your call but "I" would change this variable to `port`

apps/app_gps.c
```diff
@@ -1725,6 +2037,32 @@ static int load_module(void)
...
+		} else {
+			gpsd_port = (int) parsed;
+		}
+	}
+	if (gps_source == GPS_SOURCE_GPSD && comport) {
+		ast_log(LOG_NOTICE, "gpsd_host configured; ignoring comport %s\n", comport);
+		ast_free(comport);
+		comport = NULL;
+	} else if (comport) {
+		gps_source = GPS_SOURCE_SERIAL;
```
Allan-N, line 2152: Seems like we should be setting `gps_source = GPS_SOURCE_SERIAL;` above, when we find/set `comport = ast_strdup(val);`.   Either that, or move where we set `gps_source = GPS_SOURCE_GPSD` into this general section were we validate both `gpsd_host` and `comport`.  Having the setting/checking in different places just looks off.

apps/app_gps.c
```diff
@@ -1725,6 +2037,32 @@ static int load_module(void)
 	} else {
 		comport = NULL;
 	}
+	val = ast_variable_retrieve(cfg, "general", "gpsd_host");
+	if (val && !ast_strlen_zero(val)) {
+		gpsd_host = ast_strdup(val);
+		gps_source = GPS_SOURCE_GPSD;
+	}
```
mkmer, line 2128: It is possible for gpsd_host to not be assigned with this... I think we need to init this guy on creation.

mkmer (changes requested):

apps/app_gps.c
hardenedpenguin: There is already a global static char *port (APRS server port). A local port would shadow it and confuse readers.

apps/app_gps.c
```diff
@@ -279,7 +279,7 @@ static int run_forever = 1;
 
 /* Global configuration information */
 static char *comport, *server, *port;
-static char *gpsd_host;
+static char *gpsd_host = NULL;
```
mkmer, line 284: That will "for sure" do it, but we only need 1 or the other (right?) Probably just ln 282?

apps/app_gps.c
```diff
@@ -2040,26 +2029,29 @@ static int load_module(void)
...
 	if (val && !ast_strlen_zero(val)) {
 		char *endptr = NULL;
-		long parsed;
+		long port_num;
 
 		errno = 0;
-		parsed = strtol(val, &endptr, 10);
-		if (errno || endptr == val || *endptr || parsed <= 0 || parsed > 65535) {
+		port_num = strtol(val, &endptr, 10);
+		if (errno || endptr == val || *endptr || port_num <= 0 || port_num > 65535) {
```
mkmer: Magic numbers? (~~INT_MIN/INT_MAX~~ UINT16_MAX?)

apps/app_gps.c
hardenedpenguin: I can switch to the declaration only if that is what you prefer, I choose else for consistency with comport.

mkmer: Is there a guarantee that we get to the `else` before any other code evaluates `gpsd_host`?  If not, we should initialize it at creation (right?)

Allan-N (approved):

mkmer (approved):

PR #1120, opened 2026-07-25
chan_voter: clean up voter client priority and lockout handling

### LINKED ISSUE #1119: chan_voter.c VOTER Disable (Lockout) Priority Incorrect in CLI
When using the `off` or `disabled` keywords when overriding the client priority, the wrong level is used:

https://github.com/AllStarLink/app_rpt/blob/2c3a7b167b4e429f700f31984a9645f70af56b47/channels/chan_voter.c#L2262-L2263

A level of `-2` is "normal". To disable or turn off a client, a level of `-1` should be used.

Need to look at this whole `voter_do_prio` function, and clean it up to make it make more sense.

ve7fet:
Properly convert voter.conf prio value to signed int

Fix override priority keywords off or disable to lockout client

Add override priority keywords on or enable to re-enable client to default priority

Improve CLI response messages for better readability

Define priority values and replace in code for better readability

Fix voter_display to show all locked out clients (override and configured)

Fix locked out client checks to look at configured priority from voter.conf, not just CLI override

Add comments

Resolves #1119

## Summary by CodeRabbit

* **New Features**
  * Added clearer priority controls for voter clients, including normal, locked-out, and default states.
  * Updated the `voter prio` command to support enabling, disabling, querying, and changing client priorities with clearer status output.
  * Improved priority-based selection for audio mixing and voting decisions.

* **Bug Fixes**
  * Corrected handling of locked-out clients in status displays, configuration reloads, audio selection, and voting.
  * Invalid configured priorities now fall back to normal priority.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -2228,8 +2228,8 @@ static int voter_do_prio(int fd, int argc, const char *const *argv)
 			 * Otherwise, just display the client(s) priority value (client->prio) from voter.conf.
 			 */
 			if (client->prio_override > PRIO_DEFAULT) {
-				ast_cli(fd, "Client %s: current priority: %d, configured priority: %d, override priority: %d\n", client->name, client->prio_override,
-					client->prio, client->prio_override);
+				ast_cli(fd, "Client %s: current priority: %d, configured priority: %d, override priority: %d\n", client->name,
+					client->prio_override, client->prio, client->prio_override);
 			} else {
 				ast_cli(fd, "Client %s: current/configured priority: %d (not overridden)\n", client->name, client->prio);
 			}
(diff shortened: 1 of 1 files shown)


mkmer (approved):

channels/chan_voter.c
```diff
@@ -422,6 +422,11 @@ char context[100];
 #define VOTER_PAYLOAD_PING 5
 #define VOTER_PAYLOAD_PROXY 0xf000
 
+/* Define voter priority levels. */
+#define PRIO_NORMAL 0	/* Clients with a priority of 0 are "normal" and have no special priority */
+#define PRIO_LOCKOUT -1 /* Clients with a priority of -1 are "locked out" from selection */
+#define PRIO_DEFAULT -2 /* A priority of -2 is uses the client's default priority (often 0) */
```
Allan-N: is uses the ?

Allan-N (approved):
Just a thought ...

Would it help to add a `voter_get_priority(struct voter_client *client)` helper function that would combine `client->prio` and `client->prio_override` into a single value?

and if knowing if the priority was overridden is common/important you could have a variant like `voter_get_priority(struct voter_client *client, int *overridden)`

channels/chan_voter.c
ve7fet: Fixed the grammar...

Yes, there are probably some optimizations to be made by refactoring the whole client priority logic. 

I'm not sure I see the original intent of having two different priority variables to begin with... when you could probably just modify `client->prio` on the fly anyways.

I've created a new issue to track this and revisit it.

PR #1123, opened 2026-07-25
chan_voter: fix VoterStatus to use correct selected transmit client

### LINKED ISSUE #1122: chan_voter.c VoterStatus AMI Command Does Not Report Selected Mix Client
Reference https://github.com/AllStarLink/Allmon3/issues/230

"Voting" clients properly report the "voted" client via VoterStatus AMI command so that Allmon3 can colorize the selected client as "green".

"Mix" clients are not sending the same status for the client that has been chosen for TX audio via VoterStatus. However, the CLI command `voter display <instance>` does in fact show the selected client correctly.

This appears to be an issue in how the selected client is being reported by VoterStatus, and not an Allmon3 issue.

ve7fet:
VoterStatus was using `p->lastwon` which only reported the selected voting client for transmit (not mix clients), as `p->lastwon` is only set for voting clients.

Changed to use `p->winner`, as that is set for both mix and voting clients.

Resolves #1122

## Summary by CodeRabbit

* **New Features**
  * Improved voter priority handling through clearer priority levels and override behavior.
  * Updated mixed-mode audio selection to favor eligible clients with the highest priority.
  * CLI and live status displays now more clearly indicate priority and lockout states.

* **Bug Fixes**
  * Locked-out clients are now consistently excluded from voting and audio selection.
  * Invalid configured priority values are safely reset to the normal priority.
  * Voter status reporting now reflects the current winning client.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -2228,8 +2228,8 @@ static int voter_do_prio(int fd, int argc, const char *const *argv)
 			 * Otherwise, just display the client(s) priority value (client->prio) from voter.conf.
 			 */
 			if (client->prio_override > PRIO_DEFAULT) {
-				ast_cli(fd, "Client %s: current priority: %d, configured priority: %d, override priority: %d\n", client->name, client->prio_override,
-					client->prio, client->prio_override);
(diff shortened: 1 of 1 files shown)


ve7fet:
Need more coffee... needed to merge the other PR first. Will resubmit.

PR #1129, opened 2026-07-26
chan_simpleusb, chan_usbradio: Log USB radio device recovery after audio/host faults

hardenedpenguin:
## Summary
- Emit a single `NOTICE` when simpleusb/usbradio recover from a USB/audio fault (URI unplug/replug), so logs show the failed→OK transition Allan noted was missing.
- Downgrade expected `Pa_AbortStream` / `Pa_CloseStream` failures during teardown after a yank from `WARNING` to `DEBUG`.
- Clarify PortAudio read/write hard-error messages as restart-oriented wording.

## Test plan
- [ ] On a simpleusb node, unplug the URI and confirm one clear read/start failure (no Abort/Close WARNING spam).
- [ ] Replug the URI and confirm a `USB radio device recovered (...)` NOTICE appears once.
- [ ] Confirm first boot / clean start does not emit a recovery NOTICE.
- [ ] Repeat briefly on a usbradio channel if available.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of USB device and audio failures across USB audio and radio channels.
  - Added clear recovery notifications when USB/audio connectivity returns after an interruption.
  - Reduced noisy warning messages for expected stream cleanup failures after a USB device is disconnected.

### DIFF
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -284,6 +284,7 @@ struct chan_simpleusb_pvt {
 	unsigned int txcapraw:1;			   /* indicator if transmit capture is enabled */
 	unsigned int measure_enabled:1;		   /* indicator if measure mode is enabled */
 	unsigned int device_error:1;		   /* indicator set when we cannot find the USB device */
+	unsigned int usb_faulted:1;			   /* set after USB/audio failure; cleared on recovery NOTICE */
 	unsigned int newname:1;				   /* indicator that we should use MIXER_PARAM_SPKR_PLAYBACK_VOL_NEW */
 	unsigned int hasusb:1;				   /* indicator for has a USB device */
 	unsigned int usbass:1;				   /* indicator for USB device assigned */
@@ -368,6 +369,21 @@ static void tune_menusupport(int fd, struct chan_simpleusb_pvt *o, const char *c
 static void tune_write(struct chan_simpleusb_pvt *o);
 static int _send_tx_test_tone(int fd, struct chan_simpleusb_pvt *o, int ms, int intflag);
 static void *simpleusb_audio_thread(void *arg);
+
+/*!
+ * \brief Log once when USB/audio returns after a prior failure.
+ */
+static void simpleusb_log_usb_recovered(struct chan_simpleusb_pvt *o)
+{
+	const char *dev;
+
+	if (!o->usb_faulted) {
+		return;
+	}
+	dev = !ast_strlen_zero(o->devstr) ? o->devstr : o->hw_device;
+	ast_log(LOG_NOTICE, "Channel %s: USB radio device recovered (%s)\n", o->name, !ast_strlen_zero(dev) ? dev : "unknown");
+	o->usb_faulted = 0;
+}
 
 static char *simpleusb_active; /* the active device */
 
@@ -968,6 +984,7 @@ static int init_audio_device(struct chan_simpleusb_pvt *o)
 					if (!o->device_error) {
 						ast_log(LOG_ERROR, "Channel %s: No USB devices are available for assignment.\n", o->name);
 						o->device_error = 1;
+						o->usb_faulted = 1;
 					}
 					usleep(DEVICE_RETRY);
 					break;
@@ -1008,6 +1025,7 @@ static int init_audio_device(struct chan_simpleusb_pvt *o)
 				if (!o->device_error) {
 					ast_log(LOG_ERROR, "Channel %s: Device string %s was not found.\n", o->name, o->devstr);
 					o->device_error = 1;
+					o->usb_faulted = 1;
 				}
 				ast_mutex_unlock(&usb_dev_lock);
 				return -1;
@@ -1067,6 +1085,7 @@ static int init_audio_device(struct chan_simpleusb_pvt *o)
 		if (!o->device_error) {
 			ast_log(LOG_ERROR, "Channel %s: Cannot use audio device %s without mixer limits\n", o->name, o->hw_device);
 			o->device_error = 1;
+			o->usb_faulted = 1;
 		}
 		ast_mutex_unlock(&usb_dev_lock);
 		return -1;
@@ -1148,6 +1167,7 @@ static void *hidthread(void *arg)
 			if (!init_audio_failed) {
 				ast_log(LOG_ERROR, "Channel %s: Failed initialize the audio device\n", o->name);
 				init_audio_failed = 1;
+				o->usb_faulted = 1;
 			}
 
 			usleep(DEVICE_RETRY);
@@ -1158,6 +1178,7 @@ static void *hidthread(void *arg)
 			if (!init_hid_failed) {
 				ast_log(LOG_ERROR, "Channel %s: Cannot initialize device %s\n", o->name, o->devstr);
 				init_hid_failed = 1;
+				o->usb_faulted = 1;
 			}
 
 			usleep(DEVICE_RETRY);
@@ -1169,6 +1190,7 @@ static void *hidthread(void *arg)
 			if (!open_device_failed) {
 				ast_log(LOG_ERROR, "Channel %s: Cannot open device %s\n", o->name, o->devstr);
 				open_device_failed = 1;
+				o->usb_faulted = 1;
 			}
 
 			usleep(DEVICE_RETRY);
@@ -1180,6 +1202,7 @@ static void *hidthread(void *arg)
 				if (!detach_failed) {
 					ast_log(LOG_ERROR, "Channel %s: Is not able to detach the USB device\n", o->name);
 					detach_failed = 1;
+					o->usb_faulted = 1;
 				}
 
 				libusb_close(usb_handle);
@@ -1191,6 +1214,7 @@ static void *hidthread(void *arg)
 				if (!claim_failed) {
 					ast_log(LOG_ERROR, "Channel %s: Is not able to claim the USB device\n", o->name);
 					claim_failed = 1;
+					o->usb_faulted = 1;
 				}
 
 				libusb_close(usb_handle);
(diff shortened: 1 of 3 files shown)


channels/chan_simpleusb.c
```diff
@@ -283,6 +283,7 @@ struct chan_simpleusb_pvt {
 	unsigned int txcapraw:1;			   /* indicator if transmit capture is enabled */
 	unsigned int measure_enabled:1;		   /* indicator if measure mode is enabled */
 	unsigned int device_error:1;		   /* indicator set when we cannot find the USB device */
+	unsigned int usb_faulted:1;			   /* set after USB/audio failure; cleared on recovery NOTICE */
```
mkmer: I guess I missed this change, the previous PA version had individual flags for each fault.  Either way, spamming the log isn't cool :)

hardenedpenguin: Good catch on the naming — the per-fault rate-limit flags are still there in `hidthread` (`init_audio_failed`, `init_hid_failed`, `open_device_failed`, `detach_failed`, `claim_failed`). Those continue to gate the ERROR spam on the setup path.

`usb_faulted` is only a single recovery latch so we can emit one NOTICE when the device comes back after any fault (including audio-thread PA read/write failures that the hidthread locals do not cover). It is not meant to replace the per-fault logging flags.

On the PA teardown side, the Abort/Close WARNINGs after a yank are downgraded to DEBUG for the same reason — expected noise when the host device is already gone.

Allan-N (commented):
I'm just thinking out loud.  Looking over the changes I see many blocks like :
```
					if (!o->device_error) {
						ast_log(LOG_ERROR, "Channel %s: No USB devices are available for assignment.\n", o->name);
						o->device_error = 1;
						o->usb_faulted = 1;
					}
```
and I'm wondering if we need a `simpleusb_log_fault()` function that includes the format string and variable args, updates `o->usb_faulted` as needed, and logs the event with either`ast_log()` for the triggering event or `ast_debug() for subsequent / contributing / ... events`.

I'm also thinking that if we log the 1st event with `ast_log(LOG_ERROR` that the recovery should also be logged the same priority (LOG_ERROR vs LOG_NOTICE).  That way, we can pair the fault with the recovery.  Here, I'm thinking about the priority as visibility vs severity.

hardenedpenguin:
@Allan-N Good call — I added `simpleusb_log_fault()` / `usbradio_log_fault()` that take the format string + args, set `usb_faulted`, and log ERROR on the first hit / DEBUG on repeats. The rate-limit latches (device_error, init_*, pipe_*, etc.) still feed into that.

Also bumped the recovery log from NOTICE to ERROR so it pairs with the fault at the same priority.

channels/chan_simpleusb.c
```diff
@@ -1276,7 +1298,7 @@ static void *hidthread(void *arg)
 				/* HID thread monitors audio thread */
 				ast_radio_time(&audio_time_now);
 				if ((audio_time_now - o->lastaudiotime) > 1) {
-					ast_log(LOG_ERROR, "Channel %s: Audio process has died or is not responding.\n", o->name);
+					simpleusb_log_fault(o, 0, "Channel %s: Audio process has died or is not responding.\n", o->name);
```
Allan-N, line 1315: Don't we need to assign the result to an `XXX_failed` variable?  and where/how do we recover/log from this fault?

channels/chan_simpleusb.c
```diff
@@ -2137,11 +2160,13 @@ static void *simpleusb_audio_thread(void *arg)
...
 
 		if (!o->pa.active && start_stream(o) < 0) {
-			ast_log(LOG_ERROR, "Channel %s: Failed to start audio stream %s\n", o->name, o->hw_device);
+			start_stream_failed =
+				simpleusb_log_fault(o, start_stream_failed, "Channel %s: Failed to start audio stream %s\n", o->name, o->hw_device);
 			o->hasusb = 0;
 			usleep(DEVICE_RETRY);
 			continue;
 		}
+		start_stream_failed = 0;
```
Allan-N, line 2186: Do we need to add a call to `simpleusb_log_usb_recovered(o);` here to recover from the fault and log that we're again running?

channels/chan_simpleusb.c
Allan-N, line 2174: Do we log "Audio ready"?

channels/chan_simpleusb.c
hardenedpenguin: Nah — this one is a one-shot. We set `hasusb = 0` and break out of the HID loop, so it isn't spinning and re-logging the same fault. The `XXX_failed` latches are for the retry-in-place setup paths.

Recovery is on the other side: when HID comes back up it hits `simpleusb_log_usb_recovered()` right before setting `hasusb = 1`. Same idea for the other `log_fault(o, 0, ...)` breakouts (HID died, PA read, etc.).

channels/chan_simpleusb.c
hardenedpenguin: I don't think so — that would either double-log or invent a second recovery path.

`start_stream_failed = 0` here is just clearing the rate-limit latch for the next failure. If start_stream actually faulted, we already cleared `hasusb`, so HID re-runs and owns the recovery log when the device is back.

One asymmetry worth noting: usbradio logs recovered after audio start succeeds (in HID), while simpleusb logs it when HID is ready, before the audio thread opens the stream. Happy to align those if you'd rather they match.

channels/chan_simpleusb.c
hardenedpenguin: Ha — no, we don't. Just the "not ready" debug. I can add a matching "Audio ready" when the stream comes up, or drop both. Either way it's just debug noise.

Allan-N: It may just be debug noise but anyone (like me) who's looking will see "not ready" and question when/whether the issue has been resolved.  My stance is that if error logs that are later recovered should include a log indication of the recovery.

channels/chan_simpleusb.c
Allan-N: Without good reason, I would like to avoid having simpleusb and usbradio "differ".  If it's easy enough, I would opt for alignment.

channels/chan_simpleusb.c
mkmer: Might be easy enough to add a log message just before we enter the inner while loop about "Audio has started" or "Audio is ready" (I don't know).

channels/chan_simpleusb.c
hardenedpenguin: Aligned them. I moved `simpleusb_log_usb_recovered()` out of `hidthread` (it was firing before `hasusb = 1`) and into the audio thread right after `audio_thread_ready` is set. So both drivers now log recovery at the same logical point — once the audio path is actually back up. usbradio already logs right after `usbradio_start_audio()` succeeds, so they match now.

channels/chan_simpleusb.c
hardenedpenguin: Good point on the visibility. Rather than add a standalone "Audio ready" line, the recovery log now fires at that same spot (right after `audio_thread_ready = 1`), and only when we were previously faulted. So if you saw "Audio not ready" during an outage, you'll now get a paired `USB radio device recovered (...)` at ERROR when it comes back — fault and recovery at the same priority, with no extra noise on a clean boot.

channels/chan_simpleusb.c
```diff
@@ -368,6 +369,21 @@ static void tune_write(struct chan_simpleusb_pvt *o);
...
+static void simpleusb_log_usb_recovered(struct chan_simpleusb_pvt *o)
+{
+	const char *dev;
+
+	if (!o->usb_faulted) {
+		return;
+	}
+	dev = !ast_strlen_zero(o->devstr) ? o->devstr : o->hw_device;
+	ast_log(LOG_NOTICE, "Channel %s: USB radio device recovered (%s)\n", o->name, !ast_strlen_zero(dev) ? dev : "unknown");
+	o->usb_faulted = 0;
```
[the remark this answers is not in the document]
hardenedpenguin: Addressed: `usb_faulted` is now a whole-word `volatile sig_atomic_t` in both drivers (no longer a bit-field), so the cross-thread set/clear can't corrupt neighboring bits. Resolving.

channels/chan_usbradio.c
```diff
@@ -1679,7 +1708,8 @@ static int soundcard_writeframe(struct chan_usbradio_pvt *o, short *data)
 	 */
 	res = ast_radio_pa_write(&o->pa, data, AST_RADIO_PA_FRAMES_PER_BUFFER);
 	if (res < 0 && res != paOutputUnderflowed) {
-		ast_log(LOG_ERROR, "Channel %s: PortAudio write error %s\n", o->name, Pa_GetErrorText(res));
+		ast_log(LOG_ERROR, "Channel %s: PortAudio write failed (%s); restarting audio stream\n", o->name, Pa_GetErrorText(res));
+		o->usb_faulted = 1;
 		ast_radio_pa_stop(&o->pa);
 		return 0;
```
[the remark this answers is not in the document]
hardenedpenguin: Addressed: the PortAudio write-fault path in `soundcard_writeframe()` now clears `o->hasusb` (and logs via `usbradio_log_fault()`), so the HID thread re-runs and `usbradio_log_usb_recovered()` clears the latch on the way back up. Resolving.

Allan-N (approved):

channels/chan_simpleusb.c
hardenedpenguin: Confirmed we kept the per-fault rate-limit latches (`init_audio_failed`, `init_hid_failed`, `open_device_failed`, `detach_failed`, `claim_failed`, `pipe_failed`, `start_audio_failed`), so no log spam. They now feed the shared `simpleusb_log_fault()` helper, which logs ERROR on the first hit and DEBUG on repeats. Resolving.

mkmer (approved):

PR #1146, opened 2026-08-03
chan_voter.c: Various CodeRabbit Fixes and Other Minor Enhancements

### LINKED ISSUE #1125: chan_voter.c voter test Can Cause Crash
Currently, `chan_voter` calls the standard `random()` function (which may not even be seeded...), which is NOT thread safe.

https://github.com/AllStarLink/app_rpt/blob/bdf5916ef17febf2fe1f9717a90c8e383424a03f/channels/chan_voter.c#L5777

We should be using `ast_random()` here instead, which is seeded by Asterisk at startup, and is thread safe.

We should probably test `i` as well, to ensure it isn't `0`, preventing a diveide-by-0 crash... and CodeRabbit agrees:

**Guard the zero eligible-client case before using `i` as a divisor.**

If this VOTER instance has no eligible non-mix clients, `i` remains zero and `p->voter_test == 1` executes `random() % i`, causing division by zero and potentially terminating the reader path. Leave `maxclient` unset and let the existing silence handling run when `i == 0`.

```
+if (i == 0) {
+	maxclient = NULL;
+	p->testcycle = 0;
+	p->testindex = 0;
+} else {
     if (p->voter_test == 1) {
         p->testindex = random() % i;
     } else {
         ...
     }
     /* Existing indexed-selection loop */
+}
```

CodeRabbit also found another issue;

**Exclude zero-RSSI clients from test-mode candidates.**

When no client has positive RSSI, the preceding logic leaves `maxclient == NULL` and `maxrssi == 0`. These loops then treat every matching client with `lastrssi == 0` as eligible, bypassing the existing silence path and potentially copying stale audio from an inactive client.

```
- if (client->lastrssi == maxrssi) {
+ if (client->lastrssi > 0 && client->lastrssi == maxrssi) {
And in the selection loop:

- if (client->lastrssi != maxrssi) {
+ if (client->lastrssi <= 0 || client->lastrssi != maxrssi) {
     continue;
 }
```
Also applies to: 5807-5819

### LINKED ISSUE #1127: chan_voter.c Fix the Wrapped Audio Destination Offset
Per CodeRabbit:
"When `i < 0`, the second copy should start after the first segment at `FRAME_SIZE + i`. Using `maxclient->buflen - i` writes outside the current output frame, so the DSP stage reads stale bytes."

https://github.com/AllStarLink/app_rpt/blob/bdf5916ef17febf2fe1f9717a90c8e383424a03f/channels/chan_voter.c#L5828-L5831

Suggested change:

```
-									memcpy(p->buf + AST_FRIENDLY_OFFSET + (maxclient->buflen - i), maxclient->audio, -i);
+									memcpy(p->buf + AST_FRIENDLY_OFFSET + (FRAME_SIZE + i), maxclient->audio, -i);
```

Likely also applies to:

https://github.com/AllStarLink/app_rpt/blob/bdf5916ef17febf2fe1f9717a90c8e383424a03f/channels/chan_voter.c#L2872

Test/validate/implement if confirmed.

### LINKED ISSUE #1131: chan_voter.c Reset the hysteresis state when clearing lastwon.
Per CodeRabbit:

These paths null `p->lastwon` but leave `p->threshold`, `p->threshcount`, and `p->lingercount` intact. On the next received frame, the `if (p->lingercount)` path can dereference `p->lastwon->name` at Line 5754 and crash. Clear the related selection state in both paths.

https://github.com/AllStarLink/app_rpt/blob/383d5950d8cbd0bdeed3b8e4fb8a306422e03aa7/channels/chan_voter.c#L5751-L5755

https://github.com/AllStarLink/app_rpt/blob/383d5950d8cbd0bdeed3b8e4fb8a306422e03aa7/channels/chan_voter.c#L4734-L4739

Suggested fix:

```
 p->lastwon = NULL;
+p->threshold = 0;
+p->threshcount = 0;
+p->lingercount = 0;
```

Also applies to:

https://github.com/AllStarLink/app_rpt/blob/383d5950d8cbd0bdeed3b8e4fb8a306422e03aa7/channels/chan_voter.c#L5222-L5230

### LINKED ISSUE #1134: chan_voter.c Calculate wrap length per client before clearing its buffer
Per CodeRabbit:

_🩺 Stability & Availability_ | _🔴 Critical_ | _⚡ Quick win_

**Calculate wrap length per client before clearing its buffer.**

https://github.com/AllStarLink/app_rpt/blob/383d5950d8cbd0bdeed3b8e4fb8a306422e03aa7/channels/chan_voter.c#L5824-L5827

`i` is derived from `maxclient`, then reused for every client’s `drainindex` and `buflen`. Differing client buffers can make the silence writes cross a ring boundary incorrectly or write past the allocation. Compute a client-local wrap value inside the loop; use it for both recording and clearing.

Also applies to: 5934-5940

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@channels/chan_voter.c` around lines 5889 - 5890, Update the client-processing
loop around maxclient so the wrap-length calculation uses the current client’s
buflen and drainindex rather than the previously selected maxclient. Recompute
this client-local value inside each iteration and reuse it for both the silence
recording and buffer-clearing operations, including the logic around the
referenced later range.
```

</details>

_Originally posted by @coderabbitai[bot] in https://github.com/AllStarLink/app_rpt/pull/1133#discussion_r3661911489_

### LINKED ISSUE #1144: chan_voter.c Re-calculate the ring-buffer boundary for each client
`i` here:

https://github.com/AllStarLink/app_rpt/blob/7089a0f4cd34ff04eb46abc2ee06dae5a4e2f2c0/channels/chan_voter.c#L5998-L6005

Uses `i` previously calculated outside of the client loop for `maxclient` here:

https://github.com/AllStarLink/app_rpt/blob/7089a0f4cd34ff04eb46abc2ee06dae5a4e2f2c0/channels/chan_voter.c#L5964

If a non-winning client wraps while `maxclient` does not, Lines 6009-6010 write past that client's ring buffer. The inverse condition records and clears incorrect byte ranges.

Recalculate `i` from the current client after the client filters and before recording or silencing.

ve7fet:
Guard the zero eligible-client case before using `i` as a divisor (#1125).

Exclude zero-RSSI clients from test-mode candidates (#1125).

Fix the Wrapped Audio Destination Offset. When `i < 0`, the second copy should start after the first segment at `FRAME_SIZE + i`. Using `maxclient->buflen - i` writes outside the current output frame, so the DSP stage reads stale bytes. (#1127)

Remove redundant integer casts (#1134).

Update some of the variable names to use friendly names (#1134).

Re-calculate the ring buffer boundary, instead of using a stale counter from outside the client loop. (#1134 #1144)

Reset the hysteresis state when clearing `lastwon`. Reset `threshold` and `linger` values when `p->lastwon` is reset. (#1131)

Revise Time Debug Messages. Mix clients don't need to send time. This results in GPS epoch time being displayed (from 1969) when no time is sent. This change revises the debug messages to report "No time sent" when there is no time coming from the mix client.

Resolves #1125
Resolves #1127
Resolves #1131
Resolves #1134
Resolves #1144 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of wrapped audio and RSSI data during channel voting.
  - Reset voting and test-cycle state when reception stops, timing is unavailable, or no client is selected.
  - Excluded clients with non-positive signal strength from test-mode selection.
  - Improved debug messages for missing mix timestamps.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -5898,8 +5898,8 @@ static void *voter_reader(void *data)
 										if (client->mix) {
 											continue;
 										}
-										/* Count the clients at maxrssi. */
-										if (client->lastrssi == maxrssi) {
+										/* Count the clients at maxrssi (clients also must have a positive RSSI). */
+										if (client->lastrssi > 0 && client->lastrssi == maxrssi) {
 											i++;
 										}
 									}
@@ -5938,8 +5938,8 @@ static void *voter_reader(void *data)
 										if (client->mix) {
 											continue;
 										}
-										/* If this is client isn't at maxrssi, skip it. */
-										if (client->lastrssi != maxrssi) {
+										/* If this is client isn't at maxrssi, or has a non-positive RSSI, skip it. */
+										if (client->lastrssi <= 0 || client->lastrssi != maxrssi) {
 											continue;
 										}
 										/* See if the current number of clients matches p->testindex,
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -2817,7 +2817,7 @@ static struct ast_cli_entry voter_cli[] = {
  */
 static int voter_mix_and_send(struct voter_pvt *p, struct voter_client *maxclient, int maxrssi)
 {
-	int i, j, k, x, maxprio, haslastaudio;
+	int i, j, x, maxprio, haslastaudio, buffer_bytes_avail, rssi_sum;
 	struct ast_frame fr, *f1, *f2;
 	struct voter_client *client;
 	short silbuf[FRAME_SIZE];
@@ -2920,40 +2920,40 @@ static int voter_mix_and_send(struct voter_pvt *p, struct voter_client *maxclien
 		 * to send the audio from this highest priority client.
 		 */
 		/* Calculate the bytes available before the ring-buffer wraps. */
-		i = client->buflen - (client->drainindex + FRAME_SIZE);
-		if (i >= 0) {
+		buffer_bytes_avail = client->buflen - (client->drainindex + FRAME_SIZE);
+		if (buffer_bytes_avail >= 0) {
 			memcpy(p->buf + AST_FRIENDLY_OFFSET, client->audio + client->drainindex, FRAME_SIZE);
 		} else {
-			memcpy(p->buf + AST_FRIENDLY_OFFSET, client->audio + client->drainindex, FRAME_SIZE + i);
-			memcpy(p->buf + AST_FRIENDLY_OFFSET + (FRAME_SIZE + i), client->audio, -i);
+			memcpy(p->buf + AST_FRIENDLY_OFFSET, client->audio + client->drainindex, FRAME_SIZE + buffer_bytes_avail);
+			memcpy(p->buf + AST_FRIENDLY_OFFSET + (FRAME_SIZE + buffer_bytes_avail), client->audio, -buffer_bytes_avail);
 		}
-		if (i >= 0) {
+		if (buffer_bytes_avail >= 0) {
 			memset(client->audio + client->drainindex, ULAW_SILENCE, FRAME_SIZE);
 		} else {
-			memset(client->audio + client->drainindex, ULAW_SILENCE, FRAME_SIZE + i);
-			memset(client->audio, ULAW_SILENCE, -i);
+			memset(client->audio + client->drainindex, ULAW_SILENCE, FRAME_SIZE + buffer_bytes_avail);
+			memset(client->audio, ULAW_SILENCE, -buffer_bytes_avail);
 		}
 		/* Calculate the RSSI based on any RSSI samples in the buffer */
-		k = 0;
-		if (i >= 0) {
+		rssi_sum = 0;
+		if (buffer_bytes_avail >= 0) {
 			for (j = client->drainindex; j < client->drainindex + FRAME_SIZE; j++) {
-				k += client->rssi[j];
+				rssi_sum += client->rssi[j];
 				client->rssi[j] = 0; /* After reading an RSSI value, reset the array to 0 */
 			}
 		} else {
-			for (j = client->drainindex; j < client->drainindex + (FRAME_SIZE + i); j++) {
-				k += client->rssi[j];
+			for (j = client->drainindex; j < client->drainindex + (FRAME_SIZE + buffer_bytes_avail); j++) {
+				rssi_sum += client->rssi[j];
 				client->rssi[j] = 0;
 			}
-			for (j = 0; j < -i; j++) {
-				k += client->rssi[j];
+			for (j = 0; j < -buffer_bytes_avail; j++) {
+				rssi_sum += client->rssi[j];
 				client->rssi[j] = 0;
 			}
 		}
 		/* Take the sum of all the RSSI samples we found, get the average, and set client->lastrssi
 		 * for this client based on the result.
 		 */
-		client->lastrssi = k / FRAME_SIZE;
+		client->lastrssi = rssi_sum / FRAME_SIZE;
 		/* If this client's RSSI is has the strongest RSSI, set maxrssi to this new value, and
 		 * mark this client as the strongest (maxclient).
 		 */
@@ -4759,7 +4759,7 @@ static void *voter_reader(void *data)
 	char client1_ip[INET_ADDRSTRLEN];
 	struct sockaddr_in sin, sin_stream, psin;
 	struct voter_pvt *p;
-	int fd, i, j, k, timeout_ms, maxrssi, master_port, no_ast_channel = 0, logged_no_ast_channel = 0, logged_buflen_too_small = 0;
+	int fd, i, j, timeout_ms, maxrssi, master_port, no_ast_channel = 0, logged_no_ast_channel = 0, logged_buflen_too_small = 0, buffer_bytes_avail;
 	struct ast_frame *f1, fr;
 	socklen_t fromlen;
 	ssize_t recvlen;
@@ -5701,7 +5701,7 @@ static void *voter_reader(void *data)
 							maxrssi = 0;
 							maxclient = NULL;
 							for (client = clients; client; client = (startagain) ? clients : client->next) {
-								int maxprio, thisprio;
+								int maxprio, thisprio, rssi_sum;
 
 								startagain = 0;
 								/* If the client doesn't belong to this VOTER instance, skip it. */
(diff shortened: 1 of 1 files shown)


channels/chan_voter.c
```diff
@@ -5602,7 +5616,7 @@ static void *voter_reader(void *data)
 							index = (index + client->drainindex_40ms) % client->buflen;
 						}
 						flen = (f1) ? f1->datalen : FRAME_SIZE;
-						i = (int) client->buflen - (index + flen);
+						i = client->buflen - (index + flen);
```
mkmer: Is this `buffer_bytes_avail` rather than `i`?

ve7fet: Yeah, I saw that one. 

In looking at it some more, yes, it does appear to be a counter for buffer bytes left?

I'll push a commit to update it.

channels/chan_voter.c
```diff
@@ -2920,40 +2920,40 @@ static int voter_mix_and_send(struct voter_pvt *p, struct voter_client *maxclien
...
+		if (buffer_bytes_avail >= 0) {
 			memcpy(p->buf + AST_FRIENDLY_OFFSET, client->audio + client->drainindex, FRAME_SIZE);
 		} else {
-			memcpy(p->buf + AST_FRIENDLY_OFFSET, client->audio + client->drainindex, FRAME_SIZE + i);
-			memcpy(p->buf + AST_FRIENDLY_OFFSET + (client->buflen - i), client->audio, -i);
+			memcpy(p->buf + AST_FRIENDLY_OFFSET, client->audio + client->drainindex, FRAME_SIZE + buffer_bytes_avail);
+			memcpy(p->buf + AST_FRIENDLY_OFFSET + (FRAME_SIZE + buffer_bytes_avail), client->audio, -buffer_bytes_avail);
 		}
-		if (i >= 0) {
+		if (buffer_bytes_avail >= 0) {
```
Allan-N: Should we combine this `if (buffer_bytes_avail >= 0) {` with the one above?

Also, should the `/* Calculate ...` comment be rev'd to note that we're calculating, copying, and clearing?

channels/chan_voter.c
```diff
@@ -2920,40 +2920,40 @@ static int voter_mix_and_send(struct voter_pvt *p, struct voter_client *maxclien
 		 * to send the audio from this highest priority client.
 		 */
 		/* Calculate the bytes available before the ring-buffer wraps. */
-		i = (int) client->buflen - ((int) client->drainindex + FRAME_SIZE);
-		if (i >= 0) {
+		buffer_bytes_avail = client->buflen - (client->drainindex + FRAME_SIZE);
```
Allan-N, line 2929: Circular buffers are soooooo much fun :)

Is this really the bytes available in the [audio?] buffer?  or is the value more about how many bytes of the frame are present at the end of the [audio?] buffer before being wrapped back to the start of the buffer?

channels/chan_voter.c
```diff
@@ -2920,40 +2920,40 @@ static int voter_mix_and_send(struct voter_pvt *p, struct voter_client *maxclien
...
-			memset(client->audio + client->drainindex, ULAW_SILENCE, FRAME_SIZE + i);
-			memset(client->audio, ULAW_SILENCE, -i);
+			memset(client->audio + client->drainindex, ULAW_SILENCE, FRAME_SIZE + buffer_bytes_avail);
+			memset(client->audio, ULAW_SILENCE, -buffer_bytes_avail);
 		}
 		/* Calculate the RSSI based on any RSSI samples in the buffer */
-		k = 0;
-		if (i >= 0) {
+		rssi_sum = 0;
+		if (buffer_bytes_avail >= 0) {
```
Allan-N, line 2946: Just a thought :
```
for (j = 0; j < FRAME_SIZE; j++) {
    offset = (client->drainindex + j) % client->buflen;
    rssi_sum += client->rssi[offset];
}
```

channels/chan_voter.c
mkmer: Thinking about this a bit further, we do this math several times... should we have a set of functions that add / remove from a circular buffer and potentially replace this stuff? (Maybe in a different PR?)

ve7fet: @Allan-N Yeah, no kidding. There is an explanation at the top of the source that Jim gave, to which I think I've added some comments to during some previous code explorations.

To answer your question, it isn't the audio buffer in particular, since there are two buffers, one for audio and one for RSSI. From my understanding, it looks like `buffer_bytes_avail` is checking to see if there are still frames of audio and RSSI to process. 

`client->buflen` is the reference calculated from the config file. `client->drainindex` is the running index counter of stuff in the circular buffers (making sure there is at least a full frame left). So, if there is still audio and RSSI left to process, `buffer_bytes_avail` will be true (the number doesn't really matter, as long as it is positive). 

The "processing" involves sending the audio to the Asterisk channel (`p->buf`), calculating the running average RSSI for the client, using that average RSSI to update the selected/voted client (`maxclient`), and then clearing the audio and RSSI buffers just processed (setting silence in the audio buffer, and zeroing the RSSI).

If `buffer_bytes_avail` is false (`<0`), we're going to fall through to the `else`, which I *think* might be grabbing the last full frame (remember, `buffer_bytes_avail` will be negative at this point, so the addition is really a subtraction)? I'd have to run some numbers to see if that is really what it is doing... it is confusing, for sure.

@mkmer To your point, yes, it may make sense to put this in a function. The thing to keep in mind is that there is a lot of duplicate code right now with both the proxy stuff as well as ADPCM. If one or both of those sets of broken code goes away, there may be a whole bunch of simplification to realize that may or may not make another function a good optimization.

channels/chan_voter.c
ve7fet: Yeah, I can clean that up and will add some more comments around this.

channels/chan_voter.c
ve7fet: CodeRabbit helped figure out the other part of the mystery...

```
} else {
			memcpy(p->buf + AST_FRIENDLY_OFFSET, client->audio + client->drainindex, FRAME_SIZE + buffer_bytes_avail);
			memcpy(p->buf + AST_FRIENDLY_OFFSET + (FRAME_SIZE + buffer_bytes_avail), client->audio, -buffer_bytes_avail);
		}
```

When `buffer_bytes_avail` is negative, the buffer has wrapped around, so we go get the data from the end of the buffer, and the rest of the frame from the start of the buffer (apparently).

channels/chan_voter.c
ve7fet: I will ponder re-writing this in a future PR.

channels/chan_voter.c
Allan-N: Circular buffers aren't that difficult to work with.  On the plus side, you don't need to worry about memory management.  Typically, you have 1 pointer/offset that indicates where new data should be added and another pointer/offset indicating where data should be removed from.  If you have/need more data then space after the pointer/offset then you wrap back to the beginning of the buffer.  The actual space available looks at both of the pointers/offsets.

Also, I agree with @mkmer that a set of simple helper unctions to read-from / write-to the circular buffer would be good.

channels/chan_voter.c
Allan-N: Note: my `offset` calculation may be off ... but the idea is to end up with an offset between `0` and `buflen - 1`.

channels/chan_voter.c
ve7fet: Comments added in dcb380ee52f13d60d738bac19fe3655ee7037eb8

channels/chan_voter.c
ve7fet: Issue #1153 opened to look at this in the future.

Allan-N (approved):

mkmer (approved):

PR #1148, opened 2026-08-03
chan_simpleusb: address null buffer write thread blocks

mkmer:
Unfortunately, `ast_radio_pa_write_available();` never returns `paOutputUnderflowed` - remove "dead" code.
Additionally, there is a condition where `Pa_WriteStream` returns `paOutputUnderflowed` yet there is not enough room for the additional null frame. 
Adding a check to confirm room in the buffer before attempting to add the null frame.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USB radio stream error handling so output failures are detected consistently.
  * Prevented transmission attempts when insufficient audio capacity is available.
  * Improved stream cleanup and device availability handling after output errors.
  * Added clearer logging when audio output cannot be safely initialized.
* **Documentation**
  * Clarified audio frame and sample sizing to improve maintainability and troubleshooting.

### DIFF
--- a/res/res_usbradio.c
+++ b/res/res_usbradio.c
@@ -1368,7 +1368,7 @@ PaError ast_radio_pa_write(struct ast_radio_pa_stream *ps, const short *data, un
 	}
 
 	if (frames > AST_RADIO_PA_FRAMES_PER_BUFFER) {
-		ast_log(LOG_WARNING, "ast_radio_pa_write: frames %lu exceeds buffer capacity\n", frames);
+		ast_log(LOG_WARNING, "ast_radio_pa_write: PortAudio frames %lu exceeds buffer capacity\n", frames);
 		return paBufferTooBig;
 	}
 
--- a/res/res_usbradio.c
+++ b/res/res_usbradio.c
@@ -1359,7 +1359,7 @@ PaError ast_radio_pa_read(struct ast_radio_pa_stream *ps, short *buf, unsigned l
 	return paTimedOut;
 }
 
-PaError ast_radio_pa_write(struct ast_radio_pa_stream *ps, const short *data, unsigned long frames)
+PaError ast_radio_pa_write(struct ast_radio_pa_stream *ps, const short *data, unsigned long pa_frames)
 {
 	PaError res;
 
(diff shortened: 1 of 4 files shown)


res/res_usbradio.c
```diff
@@ -1375,17 +1375,28 @@ PaError ast_radio_pa_write(struct ast_radio_pa_stream *ps, const short *data, un
...
 		 */
-		ast_debug(6, "PortAudio write stream underflow, writing a 0 frame");
-		prime_res = Pa_WriteStream(ps->stream, null_buf, frames);
-		if (prime_res != paNoError) {
-			return prime_res;
+		frames_available = ast_radio_pa_write_available(ps);
+
+		if ((frames_available > 0) && (frames_available >= frames)) {
+			ast_debug(6, "PortAudio write stream underflow, priming with %ld silence frames\n", frames);
+			prime_res = Pa_WriteStream(ps->stream, null_buf, frames);
```
Allan-N: Seems like we should be passing how many frames are in `null_buf[]` and not the `frames` value passed to the function.

Also, how many null frames are we writing?  Exactly "1" to write something?  or more than 1 frame that might end up using space that we need on the next pass?

Allan-N (changes requested):

res/res_usbradio.c
mkmer: It's a bit strange here, the frames are the same, but the size is different depending on how the channel is set up (Stereo/2 or Mono/1) -> probably don't need to pass the frames at all, but stereo or not would be more appropriate?

Allan-N: What caught my eye was that we allocated `null_buf[AST_RADIO_PA_FRAMES_PER_BUFFER * AST_RADIO_PA_OUTPUT_CHANNELS]` but we are claiming the size to be (the passed in) `frames` that "could" be much larger.  Specifically, we should be stating the size of (# of frames in) null_buf.

Seeing `frames`, plural, raised the 2nd question/comment ... would we want to be writing multiple frames or would a single null frame be sufficient?

Allan-N (changes requested):

Allan-N (approved):
Looking a bit more at the changes I see enough tests to ensure that we are not going to be writing more data than available.

res/res_usbradio.c
mkmer: Documenting some offline conversations: (This reuse of the word "frame" and what it means in different contexts)

Asterisk Frames = 160 sample or 20ms of audio at 8khz
PA frames are actually samples -> so 1 Asterisk frame = 160 PA frames (at 8khz).
Our audio device is working at 48khz -> 48/8 = 6 -> thus we need to add (160 * 6) samples to the port audio (in frames) -> 960 Port Audio Frames.

Following on, the number of shorts the PA function expects depends on the configuration of the channel.
`Pa_WriteStream(ps->stream, null_buf, frames);` 

- on a mono channel it will index 960 shorts into `null_buf`
- on a stereo channel, it will index 1920 shorts into the `null_buff` 
Either for the same 960 value in "frames".

While not ideal here, having a null buffer sized for 2 channels should be "safe" as 1 channel will only index into 1/2 of it, while 2 channels will use the entire buffer.

PR #1155, opened 2026-08-04
chan_voter.c Refactor buffer allocations/reallocations in reload()

### LINKED ISSUE #1150: chan_voter.c Optimize if Statements
Review and look at optimizing these `if` statements:

https://github.com/AllStarLink/app_rpt/blob/1e81fcb3b5f73c59706a8ffafaa9cd7629fa06bb/channels/chan_voter.c#L4434-L4466

ve7fet:
Refactor audio and RSSI buffer allocations on reload to avoid overwriting live pointers and potentially leaving them unreachable if allocation/reallocation fails.

Update various/additional return values to use Asterisk module load status constants.

If a new client is added during reload, and buffer allocation fails, free the memory allocations as needed to clean up upon failure.

Resolves #1150

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved module reload handling by reporting accurate success, configuration, authentication, and allocation statuses.
  * Improved client audio and signal buffer management when configured buffer sizes change.
  * Preserved existing buffers where possible when memory allocation fails, reducing disruption.
  * Reset resized buffers cleanly to prevent stale audio or signal data.
* **Documentation**
  * Clarified the possible module load status results for administrators and integrators.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -4067,14 +4067,16 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
  * Note that on initial start, load_module runs this one time, before voter_request
  * loads the rest of the config file.
  *
- * \retval  			0 Success — configuration loaded and applied.
- * \retval 				-1 Failure — configuration load or allocation error;
- *						existing state is left unchanged where possible.
+ * \retval  			AST_MODULE_LOAD_SUCCESS (0) — configuration loaded and applied.
+ * \retval 				AST_MODULE_LOAD_FAILURE (-1) — configuration, validation,
+ *						authentication, or allocation error; existing state is left
+ *						unchanged, where possible.
  */
 static int reload(void)
 {
 	struct ast_flags zeroflag = { 0 };
 	int i, n, instance_buflen, buflen, oldtoctype, oldlevel;
+	uint8_t *newaudio, *newrssi;
 	char *ctg, *cp, *cp1, *cp2, *strs[40], newclient, data[100], oldctcss[100];
 	const char *val;
 	struct voter_pvt *p;
@@ -4089,7 +4091,7 @@ static int reload(void)
 
 	if (!(cfg = ast_config_load(config, zeroflag))) {
 		ast_log(LOG_ERROR, "Unable to load/reload config %s\n", config);
-		return -1;
+		return AST_MODULE_LOAD_FAILURE;
 	} else {
 		ast_log(LOG_NOTICE, "Config load/reload from %s\n", config);
 	}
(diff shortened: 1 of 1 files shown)


ve7fet:
Hmmm... stepped into a landmine of logic... Going to have to see if this is the best way to do this. Once again `reload()` proves to be "messy".

channels/chan_voter.c
```diff
@@ -4443,37 +4446,71 @@ static int reload(void)
...
-				if (!client->rssi) {
-					ast_config_destroy(cfg);
-					return -1;
+				/* If there is an existing RSSI buffer, reallocate in case the size changed,
+				 * otherwise allocate a new one. */
+				if (client->rssi) {
+					newrssi = ast_realloc(client->rssi, client->buflen);
+					/* Abort if allocation fails.*/
+					if (!newrssi) {
+						if (newclient) {
```
mkmer, line 4489: Maybe confused here... why do we care if it's a `newclient` when the realloc fails?  If we exit with a module failure don't we leave `client` and `client-audio` lost in space for "old" clients?

channels/chan_voter.c
```diff
@@ -4443,37 +4446,71 @@ static int reload(void)
...
+					}
+					client->rssi = newrssi;
+					/* Clear/reset the existing RSSI buffer. */
+					memset(client->rssi, 0, client->buflen);
+				} else {
+					/* Create a new RSSI buffer, note ast_calloc() will initialize to zero. */
+					client->rssi = ast_calloc(1, client->buflen);
+					/* Abort if allocation fails.*/
+					if (!client->rssi) {
+						if (newclient) {
```
mkmer, line 4504: Same Q about `newclient` here

channels/chan_voter.c
```diff
@@ -4443,37 +4446,71 @@ static int reload(void)
...
+			 * If any of the allocations fail, we abort the reload and return AST_MODULE_LOAD_FAILURE,
+			 * leaving the existing buffers intact.
+			 */
+			if (!client->audio || !client->rssi || (client->old_buflen && (client->buflen != client->old_buflen))) {
+				/* If there is an existing audio buffer, reallocate in case the size changed,
+				 * otherwise allocate a new one. */
+				if (client->audio) {
+					newaudio = ast_realloc(client->audio, client->buflen);
+					/* Abort if allocation fails.*/
+					if (!newaudio) {
```
Allan-N, line 4461: From the realloc(3) man page:
_... If these functions fail, the original block is left untouched; it is not freed or moved._

and since you are returning AST_MODULE_LOAD_FAILURE don't you also need to add :
```
						ast_free(client->audio); /* release the original audio buffer */

channels/chan_voter.c
Allan-N: Would a `voter_client_free()` function help?

ve7fet:
Yeah, I think I am going to go in a different direction with this. Looking at the whole function more closely, there are some things that should probably be moved around, and other changes made instead. 

The first step will be to add a bunch of comments to figure out what it all does, and why...

PR #1158, opened 2026-08-05
app_rpt: Use libusb-1.0

mkmer:
## Summary by CodeRabbit

* **Improvements**
  * Updated USB radio and SimpleUSB device support to use the modern libusb-1.0 interface.
  * Improved USB device detection, descriptor handling, error reporting, and resource cleanup.
  * Added more reliable USB initialization and device reference management.
  * Preserved existing ALSA integration and device-specific behavior.
* **Build**
  * Updated affected components to link with libusb-1.0 for improved compatibility with current systems.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -267,6 +267,7 @@
 
 #include "asterisk.h"
 
+#include <signal.h>
 #include <stdio.h>
 #include <unistd.h>
 #include <string.h>
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -267,7 +267,6 @@
 
 #include "asterisk.h"
 
-#include <signal.h>
 #include <stdio.h>
 #include <unistd.h>
 #include <string.h>
--- a/apps/app_rpt/app_rpt.h
+++ b/apps/app_rpt/app_rpt.h
@@ -1,7 +1,7 @@
 #define VERSION_MAJOR 3
 #define VERSION_MINOR 10
 #define VERSION_PATCH 1
-
+#include <signal.h>
 #include "asterisk/audiohook.h"
 
 /* 99% of the DSP code in app_rpt exists in dsp.c as private functions. This code can mostly be
--- a/apps/app_rpt/app_rpt.h
+++ b/apps/app_rpt/app_rpt.h
@@ -1,6 +1,7 @@
 #define VERSION_MAJOR 3
 #define VERSION_MINOR 10
 #define VERSION_PATCH 1
+
 #include "asterisk/audiohook.h"
 
 /* 99% of the DSP code in app_rpt exists in dsp.c as private functions. This code can mostly be
--- a/apps/app_rpt/app_rpt.h
+++ b/apps/app_rpt/app_rpt.h
@@ -1,7 +1,6 @@
 #define VERSION_MAJOR 3
 #define VERSION_MINOR 10
 #define VERSION_PATCH 1
-#include <signal.h>
 #include "asterisk/audiohook.h"
 
 /* 99% of the DSP code in app_rpt exists in dsp.c as private functions. This code can mostly be
--- a/channels/chan_simpleusb.c
+++ b/channels/chan_simpleusb.c
@@ -879,6 +879,10 @@ static int init_audio_device(struct chan_simpleusb_pvt *o)
 	o->hasusb = 0;
 	o->usbass = 0;
 	o->devicenum = 0;
+
+	if (o->usb_dev) {
+		libusb_unref_device(o->usb_dev);
+	}
 	o->usb_dev = NULL;
 
 	if (o->hw_device[0]) {
@@ -1218,6 +1222,9 @@ static void *hidthread(void *arg)
 			ast_mutex_lock(&usb_dev_lock);
 			o->usbass = 0;
 			o->hasusb = 0;
+			if (o->usb_dev) {
+				libusb_unref_device(o->usb_dev);
+			}
 			o->usb_dev = NULL;
 			ast_mutex_unlock(&usb_dev_lock);
 			libusb_close(usb_handle);
@@ -2004,7 +2011,9 @@ static int simpleusb_hangup(struct ast_channel *c)
 		pthread_join(o->hidthread, NULL);
 		o->hidthread = AST_PTHREADT_NULL;
 	}
-
+	if (o->usb_dev) {
+		libusb_unref_device(o->usb_dev);
+	}
 	o->owner = NULL;
 	ast_channel_tech_pvt_set(c, NULL);
 	ast_module_unref(ast_module_info->self);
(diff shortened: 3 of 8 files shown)


apps/app_rpt/app_rpt.h
```diff
@@ -1,7 +1,7 @@
 #define VERSION_MAJOR 3
 #define VERSION_MINOR 10
 #define VERSION_PATCH 1
-
+#include <signal.h>
```
Allan-N: What does this include pull in that needs something from <signal.h> ?

apps/app_rpt.c
```diff
@@ -267,7 +267,6 @@
 
 #include "asterisk.h"
 
-#include <signal.h>
```
Allan-N, line 270: Why are we moving the #include to `app_rpt.h` ?

include/asterisk/res_usbradio.h
```diff
@@ -28,8 +28,9 @@
 #define HAVE_SYS_IO
 #endif
 
-#include <usb.h>
+#include <libusb-1.0/libusb.h>
 #include <portaudio.h>
+#include <signal.h>
```
Allan-N, line 33: What does this include pull in that needs something from <signal.h> ?

res/res_usbradio.c
```diff
@@ -1418,44 +1478,62 @@ long ast_radio_pa_write_available(struct ast_radio_pa_stream *ps)
...
 	}
 
-	usb_init();
-	usb_find_busses();
-	usb_find_devices();
+	if (!usb_ctx) {
+		return NULL;
+	}
+
+	{
```
Allan-N, line 1499: Do we need the `{ ... }` ?

res/res_usbradio.c
```diff
@@ -1519,6 +1597,16 @@ static int load_config(int reload)
 
 static int reload_module(void)
 {
+	if (!usb_ctx) {
```
Allan-N: Do we need this test?  or just call `ast_radio_libusb_init()` ?

channels/chan_simpleusb.c
```diff
@@ -879,8 +879,12 @@ static int init_audio_device(struct chan_simpleusb_pvt *o)
 	o->hasusb = 0;
 	o->usbass = 0;
 	o->devicenum = 0;
-	o->usb_dev = NULL;
 
+	if (o->usb_dev) {
+		libusb_unref_device(o->usb_dev);
+	}
+
+	o->usb_dev = NULL;
```
Allan-N: This statement can go into the above `if () { }`

channels/chan_simpleusb.c
```diff
@@ -1170,25 +1173,25 @@ static void *hidthread(void *arg)
 			continue;
 		}
 		/* attempt to claim the usb hid interface and detach from the kernel */
-		if (usb_claim_interface(usb_handle, C108_HID_INTERFACE) < 0) {
-			if (usb_detach_kernel_driver_np(usb_handle, C108_HID_INTERFACE) < 0) {
+		if (libusb_claim_interface(usb_handle, C108_HID_INTERFACE) < 0) {
+			if (libusb_detach_kernel_driver(usb_handle, C108_HID_INTERFACE) < 0) {
```
Allan-N, line 1177: I'm not suggesting a change here but the code seems a bit odd.  Here, it looks like we are saying "if we can't claim the interface then try detaching and retry".  Are we running into an error and forgetting that we need to detach?  Are we simply missing a (or a few) detaches?

channels/chan_simpleusb.c
```diff
@@ -1219,21 +1222,30 @@ static void *hidthread(void *arg)
 			ast_mutex_lock(&usb_dev_lock);
 			o->usbass = 0;
 			o->hasusb = 0;
+			if (o->usb_dev) {
+				libusb_unref_device(o->usb_dev);
+			}
 			o->usb_dev = NULL;
```
Allan-N: Since we added the `if () {}` above we can move up setting the value to NULL.

channels/chan_simpleusb.c
```diff
@@ -1219,21 +1222,30 @@ static void *hidthread(void *arg)
...
 			usleep(DEVICE_RETRY);
 			continue;
 		}
 
-		if ((o->usb_dev->descriptor.idProduct & 0xfffc) == C108_PRODUCT_ID) {
-			o->devtype = C108_PRODUCT_ID;
-		} else {
-			o->devtype = o->usb_dev->descriptor.idProduct;
-		}
+		{
```
Allan-N, line 1238: Do we need the `{ }` ?

channels/chan_usbradio.c
```diff
@@ -1223,12 +1228,21 @@ static void *hidthread(void *arg)
 			usleep(500000);
 			continue;
 		}
-		if ((usb_dev->descriptor.idProduct & 0xfffc) == C108_PRODUCT_ID) {
-			o->devtype = C108_PRODUCT_ID;
-		} else {
-			o->devtype = usb_dev->descriptor.idProduct;
+
+		{
```
Allan-N, line 1232: Do we need the { } ?

apps/app_rpt.c
mkmer: There are "other" places using <signal.h> - moving it here got them all.  Maybe that's not "right"?

include/asterisk/res_usbradio.h
mkmer: `sig_atomic_t `

res/res_usbradio.c
mkmer: If we want to leave the "inline"
```		struct libusb_device **list = NULL;
		ssize_t count;
		size_t dev_index;
		unsigned int busnum;
		unsigned int devaddr;
```
Which limits the scope to just this chunk of code, we need brackets or the compiler will complain.
Other option is to put them further up and increase the scope...

channels/chan_simpleusb.c
mkmer: I'm not sure here... really just copying what was in there, assuming it has a reason.
I suspect, if it's not attached, the detach will fail and we don't know anything, so "first" try to grab it, if that fails, try to detach, if that "passes", try to grab it again.... (Maybe?)

channels/chan_simpleusb.c
mkmer: Same story about the inline variable creation..

channels/chan_usbradio.c
mkmer: Same

apps/app_rpt.c
mkmer: AND, it didn't need to move.

channels/chan_simpleusb.c
Allan-N: I suspect that there's some place where we're not detaching and this is just a "let's fix our oops" workaround.

channels/chan_simpleusb.c
Allan-N: "I" would move the variable and drop the braces ... but also OK as-is.

channels/chan_usbradio.c
Allan-N: "I" would move the variable and drop the braces ... but also OK as-is.

res/res_usbradio.c
Allan-N: "I" would move the variables and drop the braces ... but also OK as-is.

Allan-N (approved):

PR #1160, opened 2026-08-06
chan_voter: Refactor Portions of reload(), Add Comments, Resolve Identified Bugs

### LINKED ISSUE #1157: chan_voter.c Per-client `buflen` is Not Supported
Documentation and sample `voter.conf` file claim we support setting `buflen` on a per-client basis. We don't.

We never look for a per-client variable, when reading in the rest of the per-client variables:

https://github.com/AllStarLink/app_rpt/blob/6e1f95d7a0845aa35754a7cf8c64e73d76579904/channels/chan_voter.c#L4402-L4440

We only use the per-instance `buflen` to override any globally set `buflen`:

https://github.com/AllStarLink/app_rpt/blob/6e1f95d7a0845aa35754a7cf8c64e73d76579904/channels/chan_voter.c#L4392

Need to remove references of the per-client variable from the sample `voter.conf` and update the manual (see https://github.com/AllStarLink/ASL3-Manual/issues/241).

ve7fet:
Add comments in `reload()`.

Add comments about Asterisk channel request in `ast_channel()`.

Use friendly variable name for the Asterisk channel pointer that is returned when a new Asterisk VOTER channel is created.

Remove the un-documented `voter.conf` option `context` that can override the initial context the channel driver uses (none of our other channel drivers do that).

Remove documentation in sample `voter.conf` for un-supported per-client `buflen` variable. We only support global (in `[general]`) and per-instance (per-node) `buflen` variables.

Add additional comments and debug messages to reload function.

Use Asterisk module status constants in `reload()`.

Preserve the next client before removing the current client to prevent crash. When the outdated client is the list head, the existing else branch sets `clients = NULL` and discards the remaining clients. Line 4674 then assigns `client = clients`. The `for` increment dereferences `client->next` with `client == NULL`. Save `client->next` before unlinking. Set clients to that saved pointer for head removal, then advance without dereferencing the freed or NULL client.

Add `voter_client_free` helper function to free memory allocations when clients are removed.

Revise failure path when `reload()` fails to leave resources alone and just return AST_MODULE_LOAD_FAILURE back to Asterisk (was causing a coredump/crash with the original code when we tried to execute it on a failure).

Update debug message to match `voter.conf` settings.

Add comments in reload() around per-instance buflen variable loading.

Resolves #1157

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring buffer length globally or per channel instance.
  * Improved channel reload handling, including configuration validation and stale-client cleanup.

* **Bug Fixes**
  * Improved client resource cleanup and connection lifecycle handling.
  * Enhanced CTCSS configuration updates and diagnostic reporting during reloads.

* **Documentation**
  * Updated configuration guidance to reflect instance-level buffer-length overrides.
  * Removed the deprecated per-client buffer-length option.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -4328,8 +4328,9 @@ static int reload(void)
 		 * we need to recreate the PMR channel.
 		 */
 		if (strcmp(oldctcss, p->txctcssfreq) || (oldtoctype != p->txtoctype) || (oldlevel != p->txctcsslevel)) {
-			ast_debug(1, "VOTER %i: CTCSS frequency, level, or turn off code type changed, recreating PMR channel\n", p->nodenum);
 			t_pmr_chan tChan;
+			
+			ast_debug(1, "VOTER %i: CTCSS frequency, level, or turn off code type changed, recreating PMR channel\n", p->nodenum);
 
 			if (p->pmrChan) {
 				destroyPmrChannel(p->pmrChan);
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -3788,7 +3788,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 {
 	int i, j;
 	struct voter_pvt *p, *p1;
-	struct ast_channel *astchanptr = NULL;
+	struct ast_channel *chan = NULL;
 	char *cp, *cp1, *cp2, *strs[MAXTHRESHOLDS], *ctg;
 	const char *val;
 	struct ast_config *cfg = NULL;
@@ -3866,9 +3866,8 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 	 * endpoint: channel prefix "voter/%s" (where %s is the node number)
 	 * __FILE__: source file name (set to to the node identifier string, aka node number, for debugging)
 	 */
-	astchanptr =
-		ast_channel_alloc(1, AST_STATE_DOWN, 0, 0, "", (char *) data, context, assignedids, requestor, 0, "voter/%s", (char *) data);
-	if (!astchanptr) {
+	chan = ast_channel_alloc(1, AST_STATE_DOWN, 0, 0, "", (char *) data, context, assignedids, requestor, 0, "voter/%s", (char *) data);
+	if (!chan) {
 		ast_log(LOG_ERROR, "VOTER %i: Cannot alloc new Asterisk channel\n", p->nodenum);
 		ast_free(p);
 		return NULL;
@@ -3881,17 +3880,17 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 	}
 	pvts = p;
 	ast_mutex_unlock(&voter_lock);
-	ast_channel_tech_set(astchanptr, &voter_tech);
-	ast_channel_set_rawwriteformat(astchanptr, ast_format_slin);
-	ast_channel_set_writeformat(astchanptr, ast_format_slin);
-	ast_channel_set_rawreadformat(astchanptr, ast_format_slin);
-	ast_channel_set_readformat(astchanptr, ast_format_slin);
-	ast_channel_nativeformats_set(astchanptr, voter_tech.capabilities);
-	ast_channel_tech_pvt_set(astchanptr, p);
-	ast_channel_unlock(astchanptr);
-	ast_channel_language_set(astchanptr, "");
-	p->owner = astchanptr;
-	p->u = ast_module_user_add(astchanptr);
+	ast_channel_tech_set(chan, &voter_tech);
+	ast_channel_set_rawwriteformat(chan, ast_format_slin);
+	ast_channel_set_writeformat(chan, ast_format_slin);
+	ast_channel_set_rawreadformat(chan, ast_format_slin);
+	ast_channel_set_readformat(chan, ast_format_slin);
+	ast_channel_nativeformats_set(chan, voter_tech.capabilities);
+	ast_channel_tech_pvt_set(chan, p);
+	ast_channel_unlock(chan);
+	ast_channel_language_set(chan, "");
+	p->owner = chan;
+	p->u = ast_module_user_add(chan);
 	/* Load the configuration for this node. Note that not all variables are loaded here,
 	 * some are loaded in the reload function, which is also executed on initial start.
 	 */
@@ -3955,7 +3954,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 			}
 			j = finddelim(cp, strs, ARRAY_LEN(strs));
 			if (j < 2) {
-				ast_log(LOG_ERROR, "Channel %s: primary config not specified properly in %s\n", ast_channel_name(astchanptr), config);
+				ast_log(LOG_ERROR, "Channel %s: primary config not specified properly in %s\n", ast_channel_name(chan), config);
 			} else {
 				cp1 = strchr(strs[0], ':');
 				if (cp1) {
@@ -3963,8 +3962,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 					j = atoi(cp1 + 1);
 				} else {
 					j = listen_port;
-					ast_log(LOG_NOTICE, "Channel %s: Primary UDP port not configured, using default port %i\n",
-						ast_channel_name(astchanptr), j);
+					ast_log(LOG_NOTICE, "Channel %s: Primary UDP port not configured, using default port %i\n", ast_channel_name(chan), j);
 				}
 				p->primary.sin_family = AF_INET;
 				p->primary.sin_addr.s_addr = inet_addr(strs[0]);
@@ -3976,8 +3974,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 		val = ast_variable_retrieve(cfg, (char *) data, "isprimary");
 		if (val) {
 			p->isprimary = ast_true(val);
-			ast_log(LOG_NOTICE, "Channel %s: Found isprimary directive, this instance will be the primary server\n",
-				ast_channel_name(astchanptr));
+			ast_log(LOG_NOTICE, "Channel %s: Found isprimary directive, this instance will be the primary server\n", ast_channel_name(chan));
 		} else {
 			p->isprimary = 0;
 		}
@@ -4071,7 +4068,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 	if (SEND_PRIMARY(p)) {
 		ast_pthread_create(&p->primary_thread, NULL, voter_primary_client, p);
 	}
-	return astchanptr;
+	return chan;
 }
 
 /*!
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -4075,6 +4075,35 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 }
 
 /*!
+ * \brief Helper function to free memory allocations when clients are removed.
+ *
+ * \param client       Pointer to the voter_client structure to be freed.
+ */
+static void voter_client_free(struct voter_client *client)
+{
+	struct voter_client **target;
+
+	for (target = &clients; *target && *target != client; target = &(*target)->next) {
+		;
+	}
+	if (!*target) {
+		return;
+	}
+
+	*target = client->next;
+	if (client->audio) {
+		ast_free(client->audio);
+	}
+	if (client->rssi) {
+		ast_free(client->rssi);
+	}
+	if (client->gpsid) {
+		ast_free(client->gpsid);
+	}
+	ast_free(client);
+}
+
+/*!
  * \brief Reload VOTER driver configuration from disk and apply changes.
  *
  * Parses the configured voter.conf, updates per-instance (pvts) and per-client
@@ -4626,6 +4655,7 @@ static int reload(void)
 		if (client->digest == 0) {
 			ast_log(LOG_ERROR, "Can not load chan_voter -- VOTER client %s has invalid authentication digest (can not be 0)!!!\n",
 				client->name);
+			voter_client_free(client);
 			return AST_MODULE_LOAD_FAILURE;
 		}
 		for (client1 = clients; client1; client1 = client1->next) {
@@ -4638,6 +4668,8 @@ static int reload(void)
 			if (client->digest == client1->digest) {
 				ast_log(LOG_ERROR, "Can not load chan_voter -- VOTER clients %s and %s have same authentication digest!!!\n",
 					client->name, client1->name);
+				voter_client_free(client);
+				voter_client_free(client1);
 				return AST_MODULE_LOAD_FAILURE;
 			}
 		}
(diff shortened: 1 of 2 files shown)


channels/chan_voter.c
```diff
@@ -3788,7 +3788,7 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 {
 	int i, j;
 	struct voter_pvt *p, *p1;
-	struct ast_channel *tmp = NULL;
+	struct ast_channel *astchanptr = NULL;
```
mkmer: Not saying it needs to change, but "typically" they use `chan` for these guys (around the asterisk code base).

channels/chan_voter.c
```diff
@@ -4053,7 +4071,36 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
...
 	}
-	return tmp;
+	return astchanptr;
+}
+
+/*!
+ * \brief Helper function to free memory allocations when clients are removed.
+ *
+ * \param client       Pointer to the voter_client structure to be freed.
+ */
```
mkmer, line 4081: Traversing a list here... do we need to call it locked or should it be locked in here?

channels/chan_voter.c
```diff
@@ -4136,34 +4193,49 @@ static int reload(void)
 		puckit = 0;
 	}
 
+	/* Load/reload the following options that are associated with each defined
+	 * VOTER instance from voter.conf:
+	 * linger, plfilter, hostdeemp, mixminus, streams, txctcss, txctcsslevel,
+	 * txtoctype, thresholds, gtxgain
+	 */
+	ast_debug(1, "Loading per-instance options for each VOTER instance from voter.conf\n");
 	for (p = pvts; p; p = p->next) {
```
mkmer, line 4244: List without lock again?

channels/chan_voter.c
```diff
@@ -4353,44 +4471,72 @@ static int reload(void)
...
 			n = finddelim(cp, strs, ARRAY_LEN(strs));
 			if (n < 1) {
 				continue;
 			}
-			/* See if we "know" this client already. */
+			/* See if we "know" this client already. The first element in the strs array is the
+			 * client secret (password). We can feed that into crc32_bufs to get a digest, and see
+			 * if it matches any existing client.
+			 */
 			for (client = clients; client; client = client->next) {
```
mkmer, line 4533: Another lock candidate?

channels/chan_voter.c
```diff
@@ -4509,38 +4680,39 @@ static int reload(void)
...
-			ast_free(client->audio);
-		}
-		if (client->rssi) {
-			ast_free(client->rssi);
-		}
-		if (client->gpsid) {
-			ast_free(client->gpsid);
-		}
+		voter_client_free(client);
 		for (client1 = clients; client1; client1 = client1->next) {
```
mkmer: yet another lock question ...

mkmer (commented):
"Maybe" we can just lock for the entire reload?

channels/chan_voter.c
ve7fet: Sure, changing to `chan` is an easy enough change, for consistency.

channels/chan_voter.c
ve7fet: `reload()` is only called while locked:

https://github.com/AllStarLink/app_rpt/blob/bba5c47685fe5f2ecbe73b2b7c2bb0798340e32b/channels/chan_voter.c#L6796-L6802

https://github.com/AllStarLink/app_rpt/blob/bba5c47685fe5f2ecbe73b2b7c2bb0798340e32b/channels/chan_voter.c#L6850-L6857

channels/chan_voter.c
ve7fet: Resolved in 1251a49d9b1dac6edb77691569303eb651fc48da

channels/chan_voter.c
```diff
@@ -4053,7 +4071,36 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
...
+	if (client->audio) {
+		ast_free(client->audio);
+	}
+	if (client->rssi) {
+		ast_free(client->rssi);
+	}
+	if (client->gpsid) {
+		ast_free(client->gpsid);
+	}
+	ast_free(client);
```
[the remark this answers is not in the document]
ve7fet: Change committed in 6f4e03e

channels/chan_voter.c
```diff
@@ -4509,38 +4679,39 @@ static int reload(void)
...
-			clients = NULL;
+			/* Removed head; set list head to saved next */
+			clients = next;
 		}
-		ast_free(client);
-		client = clients;
+		ast_debug(1, "Removing outdated VOTER client %s from VOTER instance %i\n", client->name, client->nodenum);
+		voter_client_free(client);
+		/* Continue from saved next without dereferencing freed client */
+		client = next;
```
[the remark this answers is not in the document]
ve7fet: Change committed in a306ddd

channels/chan_voter.c
```diff
@@ -4053,7 +4068,46 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
...
+ * \brief Helper function to free memory allocations when clients are removed.
+ *
+ * \param client       Pointer to the voter_client structure to be freed.
+ */
+static void voter_client_free(struct voter_client *client)
+{
+	struct voter_client **target;
+	struct voter_pvt *p;
+
+	for (target = &clients; *target && *target != client; target = &(*target)->next) {
```
Allan-N: Your use of `target`, and especially this loop, is confusing.  I think what you are trying to do is remove the passed `client` from the `clients` list.  Does the following code feel better?
```
structure voter_client *prev;

if (client == clients) {
    /* if we are removing the client at the start of the list */
    clients = client->next;
} else {
    for (prev = clients; prev; prev = prev->next) {
        if (prev->next == client) {
            /* remove the client from the middle (or end) of the list */
            prev->next = client->next;
            break;
        }
    }
}
```

channels/chan_voter.c
```diff
@@ -4179,21 +4258,38 @@ static int reload(void)
 			cp = ast_strdup(val);
 			p->nstreams = finddelim(cp, p->streams, ARRAY_LEN(p->streams));
 		}
+		/* Backup the old CTCSS frequency, so we can tell if it changed when
+		 * we read in the txctcss variable.
+		 */
+		oldctcss[0] = 0;
```
Allan-N: Given the `ast_copy_string()` below, there's no need for this statement.

channels/chan_voter.c
```diff
@@ -4269,27 +4370,51 @@ static int reload(void)
 			}
 		}
 	}
+	/* Reset the hasmaster and masterconnected flags, so they can be re-evaluated later. */
 	hasmaster = 0;
 	masterconnected = 0;
+	/* Reset the config file category pointer. */
 	ctg = NULL;
```
Allan-N: Not needed.

channels/chan_voter.c
```diff
@@ -4269,27 +4370,51 @@ static int reload(void)
...
+	/* Passing ast_category_browse a second arg of NULL tells it to start from the
+	 * first category, and pass the current category on subsequent loop iterations.
+	 *
+	 * This is going to go through the instances again, this time looking for client
+	 * definitions, and loading them (and their options) into the clients list.
+	 */
 	while ((ctg = ast_category_browse(cfg, ctg)) != NULL) {
+		/* If there are no more categories (instances), skip. */
 		if (ctg == NULL) {
```
Allan-N: Dead code (the while loop ensures that ctg != NULL)

channels/chan_voter.c
```diff
@@ -4353,44 +4478,72 @@ static int reload(void)
...
 			newclient = 0;
-			/* If a new one, alloc its space. */
+			/* If we don't know this client, treat it as new, and alloc its space. Abort if that fails. */
 			if (!client) {
 				client = ast_calloc(1, sizeof(struct voter_client));
 				if (!client) {
 					ast_free(cp);
 					ast_config_destroy(cfg);
-					return -1;
+					return AST_MODULE_LOAD_FAILURE;
```
Allan-N, line 4558: We are returning here ... but what about all of the other clients?

channels/chan_voter.c
```diff
@@ -4438,45 +4595,63 @@ static int reload(void)
...
+					client->old_buflen / 8, client->buflen / 8);
 				client->drainindex = 0;
 			}
+			/* If the audio buffer exists and the buflen has changed, reallocate it. */
 			if (client->audio && client->old_buflen && (client->buflen != client->old_buflen)) {
 				client->audio = ast_realloc(client->audio, client->buflen);
 				if (!client->audio) {
 					ast_config_destroy(cfg);
-					return -1;
+					return AST_MODULE_LOAD_FAILURE;
```
Allan-N, line 4656: We are returning here ... but what about all of the other clients?

channels/chan_voter.c
```diff
@@ -4438,45 +4595,63 @@ static int reload(void)
...
 				}
+				/* Fill the new buffer with silence. */
 				memset(client->audio, ULAW_SILENCE, client->buflen);
+				/* If the audio buffer doesn't exist, allocate it. */
 			} else if (!client->audio) {
 				client->audio = ast_malloc(client->buflen);
 				if (!client->audio) {
 					ast_config_destroy(cfg);
-					return -1;
+					return AST_MODULE_LOAD_FAILURE;
```
Allan-N, line 4673: We are returning here ... but what about all of the other clients?

channels/chan_voter.c
```diff
@@ -4438,45 +4595,63 @@ static int reload(void)
...
+				/* Fill the new buffer with silence. */
 				memset(client->audio, ULAW_SILENCE, client->buflen);
 			}
+			/* If the RSSI buffer exists and the buflen has changed, reallocate it. */
 			if (client->rssi && client->old_buflen && (client->buflen != client->old_buflen)) {
 				client->rssi = ast_realloc(client->rssi, client->buflen);
 				if (!client->rssi) {
 					ast_config_destroy(cfg);
-					return -1;
+					return AST_MODULE_LOAD_FAILURE;
```
Allan-N, line 4688: We are returning here ... but what about all of the other clients?

channels/chan_voter.c
```diff
@@ -4509,38 +4687,23 @@ static int reload(void)
 			if (client->digest == client1->digest) {
 				ast_log(LOG_ERROR, "Can not load chan_voter -- VOTER clients %s and %s have same authentication digest!!!\n",
 					client->name, client1->name);
-				return -1;
+				voter_client_free(client);
+				voter_client_free(client1);
+				return AST_MODULE_LOAD_FAILURE;
```
Allan-N, line 4749: We are returning here after freeing both of the duplicate clients ... but what about all of the others? I could see you skipping the duplicate but keeping the original OR dropping all because the configuration is invalid but something here seems off.

channels/chan_voter.c
mkmer: Sorry, I was thinking of the front end to this was "this" function...
https://github.com/AllStarLink/app_rpt/blob/a306ddd046a3c92a4717ab39f49d130998c197ac/channels/chan_voter.c#L6837
Tis all good :)

channels/chan_voter.c
ve7fet: Removed in 49e9ad6

channels/chan_voter.c
ve7fet: Removed in 49e9ad6

channels/chan_voter.c
ve7fet: This is most definitely needed.

Without it, when starting Asterisk, the config file doesn't load completely. We never load the actual clients. I can see this in debug, because we just get authentication requests from "UNKNOWN" clients, and never answer them. Performing a manual reload from the console at that point completes the load, and clients start to authenticate.

Passing `ctg=NULL` to `ast_category_browse()` starts the iteration through the config file categories.

channels/chan_voter.c
ve7fet: So, I believe all the `return AST_MODULE_LOAD_FAILURE` that are noted are Jim's intentional behaviour.

I agree with the question, "what about all the others?". However, the point is, if we failed any part of the load/reload, we have an inconsistent state (perhaps a configuration error, or other resource issue). Therefore, the intention is to fail the load/reload process intentionally, as to not leave it in that inconsistent/broken state.

In regards to this specific case where you have two clients with the same digest, that is an easy mistake to make if you put the same client string in two different instances (I tested it). The digest calculates the same for the client in both instances, which is an an irrational state. That is why it blows away both clients and fails the load (because the config file is likely broken).

channels/chan_voter.c
Allan-N: You are absolutely correct.  I was looking at the `while ((ctg = ast_category_browse(cfg, ctg)) != NULL)` statement and focused in on the `ctg =` assignment.  I did not notice that `ctg` was also passed as an argument ... and that the argument must be initialized.

channels/chan_voter.c
Allan-N: I have no issues with returning AST_MODULE_LOAD_FAILURE if the configuration is not correct.  But, if we're going to make any attempt to free up previously allocated memory (i.e. each client) then we should do exactly that.  We shouldn't be free'ing just the offenders, we should be free'ing the space allocated for all clients that have been configured but are not going to be referenced.

Maybe another helper that could be used prior to the return like :
```
/* free any already allocated clients */
void voter_client_free_all()
{
    while (clients != NULL) {
        voter_client_free(clients);
    }
}
```

channels/chan_voter.c
ve7fet: All good. Marking this as resolved.

channels/chan_voter.c
ve7fet: Ok, I'll work on that.

channels/chan_voter.c
ve7fet: Refactored based on suggestion, and committed in 64f5f38

channels/chan_voter.c
```diff
@@ -4438,45 +4587,63 @@ static int reload(void)
...
+				/* If the RSSI buffer doesn't exist, allocate it. Note that ast_calloc will
+				 * automatically initialize the memory to zero.
+				 */
 			} else if (!client->rssi) {
 				client->rssi = ast_calloc(1, client->buflen);
 				if (!client->rssi) {
 					ast_config_destroy(cfg);
-					return -1;
+					return AST_MODULE_LOAD_FAILURE;
 				}
```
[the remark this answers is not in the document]
ve7fet: Resolved in 9b06077

channels/chan_voter.c
ve7fet: Ok, it looks like I have a working solution, committed at 9b06077 for review.

channels/chan_voter.c
```diff
@@ -4082,31 +4132,38 @@ static int reload(void)
...
 		client->old_buflen = client->buflen;
 	}
 
+	/* Attempt to load/reload voter.conf. */
 	if (!(cfg = ast_config_load(config, zeroflag))) {
 		ast_log(LOG_ERROR, "Unable to load/reload config %s\n", config);
-		return -1;
+		return AST_MODULE_LOAD_FAILURE;
 	} else {
 		ast_log(LOG_NOTICE, "Config load/reload from %s\n", config);
```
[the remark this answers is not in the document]
ve7fet: Issue #1165 created for tracking and resolution in a future PR.

channels/chan_voter.c
```diff
@@ -4691,8 +4746,9 @@ static int reload(void)
 			if (client->digest == client1->digest) {
 				ast_log(LOG_ERROR, "Can not load chan_voter -- VOTER clients %s and %s have same authentication digest!!!\n",
 					client->name, client1->name);
-				voter_client_free(client);
-				voter_client_free(client1);
+				// voter_client_free(client);
+				// voter_client_free(client1);
```
mkmer: Comments (of the illegal type too :) )

mkmer (approved):

Allan-N (approved):

channels/chan_voter.c
ve7fet: Oopsie... forgot to clean that up after testing. Done! (and some other ones too).

PR #1162, opened 2026-08-06
chan_usbradio: move PortAudio I/O onto a dedicated audio thread

### LINKED ISSUE #1142: chan_usbradio: USB hardware blocking issue.
> This timer as and FD tied to the asterisk channel is problematic.
> Any call to the `.read` or `.write` functions can not be blocking.  
> This architecture in the `.read` is calling a read from the USB hardware which can wait up to 40ms (on failure) and up to 19ms when the timing is "just right".
> We need a thread to handle the USB hardware which puts inbound audio on the channel queue and processes outbound audio (all frames in an outbound queue, up to the hardware limit).
> There will be a challenge with the XPMR "stuff" as it expects 1 frame for each read/write pair - Notice `xpmr_rx()` grabs a tx frame, except there could be 2 or 3 frames of tx.
> 
> SimpleUSB has this "configuration" (less xpmr). 

 _Originally posted by @mkmer in [#1113](https://github.com/AllStarLink/app_rpt/pull/1113/changes#r3695573893)_

hardenedpenguin:
## Summary
- Move blocking PortAudio read/write out of channel `.read`/`.write` into a dedicated audio thread (same shape as simpleusb)
- Queue outbound 8 kHz frames into XPMR via `PmrTx` on each audio tick; push inbound voice/control frames with `ast_queue_frame`
- Remove the channel `ast_timer` FD wakeup that was papering over USB waits on the Asterisk channel thread

Addresses Mike’s post-merge note on #1113 / issue #1142. XPMR still runs one `PmrRx` per RX period; multiple queued TX frames are fed through `PmrTx` into the existing dedrift buffer beforehand.

## Test plan
- [ ] Build/load `chan_usbradio` on a PortAudio USB radio node
- [ ] Confirm RX audio, COS/CTCSS key/unkey, and DTMF still work
- [ ] Confirm TX/PTT and linked audio still work (including duplex)
- [ ] Yank/replug USB briefly and confirm HID + audio threads recover without hanging Asterisk
- [ ] Confirm this PR does not include #1129 recovery-logging changes

Fixes #1142

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USB radio audio reliability and responsiveness during transmission and reception.
  * Added more resilient handling for temporary audio interruptions and device retries.
  * Improved shutdown behavior to help prevent stuck audio or HID activity.
  * Added monitoring to detect stalled audio processing and support recovery.

* **Performance**
  * Audio handling now runs asynchronously for smoother operation and reduced blocking.
  * Improved recovery from transmission backlogs to help maintain uninterrupted audio.

### DIFF
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -1954,19 +1954,18 @@ static int usbradio_call(struct ast_channel *c, const char *dest, int timeout)
 {
 	struct chan_usbradio_pvt *o = ast_channel_tech_pvt(c);
 
-	o->stophid = 0;
-	o->stopaudiothread = 0;
-	o->audio_thread_ready = 0;
-	ast_radio_time(&o->lasthidtime);
-	ast_radio_time(&o->lastaudiotime);
-
 	if (o->hidthread == AST_PTHREADT_NULL) {
+		o->stophid = 0;
+		ast_radio_time(&o->lasthidtime);
 		if (ast_pthread_create(&o->hidthread, NULL, hidthread, o)) {
 			ast_log(LOG_ERROR, "Channel %s: Failed to create HID thread\n", o->name);
 			return -1;
 		}
 	}
 	if (o->audiothread == AST_PTHREADT_NULL) {
+		o->stopaudiothread = 0;
+		o->audio_thread_ready = 0;
+		ast_radio_time(&o->lastaudiotime);
 		if (ast_pthread_create(&o->audiothread, NULL, usbradio_audio_thread, o)) {
 			ast_log(LOG_ERROR, "Channel %s: Failed to create audio thread\n", o->name);
 			o->stophid = 1;
@@ -2624,8 +2623,8 @@ static void *usbradio_audio_thread(void *arg)
 					}
 					if (f1) {
 						ast_queue_frame(o->owner, f1);
+						continue;
 					}
-					continue;
 				}
 			}
 
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -160,6 +160,9 @@ pthread_t pulserid;
 static const char *const cd_signal_type[] = { "no", "dsp", "vox", "usb", "usbinvert", "pp", "ppinvert" };
 static const char *const sd_signal_type[] = { "no", "usb", "usbinvert", "dsp", "pp", "ppinvert" };
 
+/* Keep the PortAudio TX buffer fed when no outbound audio is available. */
+static short silence_buf[AST_RADIO_PA_FRAMES_PER_BUFFER * AST_RADIO_PA_OUTPUT_CHANNELS] = { 0 };
+
 /*! \brief demodulation type */
 static const char *const demodulation_type[] = { "no", "speaker", "flat" };
 
@@ -1723,8 +1726,13 @@ static int soundcard_writeframe(struct chan_usbradio_pvt *o, short *data)
 		}
 	}
 
+	/*
+	 * Always write something so the USB TX buffer does not intentionally
+	 * underrun (same idea as simpleusb #1161). Radio PTT is HID-gated;
+	 * when unkeyed, feed silence instead of the XPMR TX buffer.
+	 */
 	if (!o->pmrChan->txPttIn && !o->pmrChan->txPttOut) {
-		return 0;
+		data = silence_buf;
 	}
 
 	/*
@@ -2171,6 +2179,8 @@ static void *usbradio_audio_thread(void *arg)
 		}
 
 		flush_tx_queue(o);
+		/* Prime one silence frame so a late first wake-up does not underrun. */
+		ast_radio_pa_write(&o->pa, silence_buf, AST_RADIO_PA_FRAMES_PER_BUFFER);
 		o->audio_thread_ready = 1;
 		last_frame_time = ast_radio_tvnow();
 
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -95,6 +95,7 @@
 #define TX_OFF_DELAY_MAX 60000				   /* in ms 60000ms, 60 seconds, 1 minute */
 #define MS_PER_FRAME 20						   /* 20 ms frames */
 #define MS_TO_FRAMES(ms) ((ms) / MS_PER_FRAME) /* convert ms to frames */
+#define DEVICE_RETRY 500000					   /* Retry time in uS when USB device is missing */
 
 #include "./xpmr/xpmr.h"
 #ifdef HAVE_XPMRX
@@ -119,7 +120,6 @@
 #include "asterisk/format.h"
 #include "asterisk/format_cache.h"
 #include "asterisk/format_compatibility.h"
-#include "asterisk/timing.h"
 
 /*! \brief Global jitterbuffer configuration - by default, jb is disabled */
 static struct ast_jb_conf default_jbconf = {
@@ -179,7 +179,6 @@ struct chan_usbradio_pvt {
 	char hw_device[100];	 /* ALSA/PortAudio device (hw:N or hw:N,M) */
 	int devtype;			 /* actual type of device */
 	int pttkick[2];			 /* ptt kick pipe */
-	struct ast_timer *timer; /* 20 ms channel wakeup (PTT / DSP tick) */
 	struct ast_radio_pa_stream pa;
 	enum {
 		M_UNSET,
@@ -199,20 +198,28 @@ struct chan_usbradio_pvt {
 	int micplaymax;
 
 	pthread_t hidthread;
+	pthread_t audiothread;
 	int stophid;
+	volatile sig_atomic_t stopaudiothread;
+	char audio_thread_ready;
+	time_t lastaudiotime;
 
 	struct ast_channel *owner;
+
+	/* Outbound 8 kHz frames from Asterisk, drained by the audio thread into PmrTx. */
+	AST_LIST_HEAD_NOLOCK(, ast_frame) txq;
+	ast_mutex_t txqlock;
 
 	/* TX workspace: 48 kHz stereo interleaved samples (PortAudio / PmrTx) */
 	short usbradio_write_buf[AST_RADIO_PA_48K_STEREO_SAMPLES];
 
-	/* buffers used in usbradio_read - AST_FRIENDLY_OFFSET space for headers
+	/* buffers used in the audio thread - AST_FRIENDLY_OFFSET space for headers
 	 * plus enough room for a full 48 kHz stereo PortAudio frame
 	 */
 	char usbradio_read_buf[AST_RADIO_PA_48K_STEREO_SAMPLES * (int) sizeof(short) + AST_FRIENDLY_OFFSET];
 	char usbradio_read_buf_8k[FRAME_SIZE * 2 + AST_FRIENDLY_OFFSET];
 	int readpos;			 /* read position above */
-	struct ast_frame read_f; /* returned by usbradio_read */
+	struct ast_frame read_f; /* scratch frame used by the audio thread */
 
 	char lastrx;
 	char rxhidsq;
@@ -442,6 +449,9 @@ static int hidhdwconfig(struct chan_usbradio_pvt *o);
 static void mixer_write(struct chan_usbradio_pvt *o);
 static int usbradio_start_audio(struct chan_usbradio_pvt *o);
 static PaError usbradio_read_pa_stereo(struct chan_usbradio_pvt *o);
+static void *usbradio_audio_thread(void *arg);
+static void flush_tx_queue(struct chan_usbradio_pvt *o);
+static void stream_cleanup(struct chan_usbradio_pvt *o);
 static struct ast_channel *usbradio_request(const char *type, struct ast_format_cap *cap,
 	const struct ast_assigned_ids *assignedids, const struct ast_channel *requestor, const char *data, int *cause);
 static int usbradio_digit_begin(struct ast_channel *c, char digit);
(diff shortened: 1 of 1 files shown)


channels/chan_usbradio.c
```diff
@@ -1363,7 +1338,18 @@ static void *hidthread(void *arg)
...
 			then = ast_radio_tvnow();
+			if (o->lastaudiotime) {
+				/* HID thread monitors audio thread */
+				ast_radio_time(&audio_time_now);
+				if ((audio_time_now - o->lastaudiotime) > 3) {
+					ast_log(LOG_ERROR, "Channel %s: Audio process has died or is not responding.\n", o->name);
+					o->hasusb = 0;
+					break;
+				}
+			}
```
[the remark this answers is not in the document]
hardenedpenguin: Fixed in f06e60e — `hasusb` is now `volatile sig_atomic_t` (kept the name, dropped the bitfield) so HID and audio can write it without RMW on the packed word.

channels/chan_usbradio.c
```diff
@@ -1987,493 +1983,552 @@ static int usbradio_write(struct ast_channel *c, struct ast_frame *f)
...
 	}
 
+	f1 = ast_frdup(f);
+	if (!f1) {
+		return 0;
+	}
+	memset(&f1->frame_list, 0, sizeof(f1->frame_list));
+	ast_mutex_lock(&o->txqlock);
+	AST_LIST_INSERT_TAIL(&o->txq, f1, frame_list);
+	ast_mutex_unlock(&o->txqlock);
```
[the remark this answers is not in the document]
hardenedpenguin: Fixed in f06e60e — `usbradio_write` now rejects non-voice / wrong-sized / null frames before `ast_frdup`. Also capped `txq` at `QUEUE_SIZE` and call `stream_cleanup()` on the `!hasusb` break so readiness clears and the queue can’t grow during retry sleeps.

hardenedpenguin:
Addressed the CodeRabbit notes in f06e60e:

- `hasusb` promoted off the packed bitfield to `volatile sig_atomic_t`
- Frame validation before queueing for `PmrTx`
- `txq` bounded at `QUEUE_SIZE`; `stream_cleanup()` on the `!hasusb` inner-loop break so `audio_thread_ready` clears
- `usbradio_read` debug line now has a terminating newline

mkmer:
Maybe I've missed it, but there should be a 200ms timer (or high water mark) on the tx queue resets the usb audio device - this handles the crazy condition where the usb device is reporting "OK" yet the hardware tx buffer is not draining :(

https://github.com/AllStarLink/app_rpt/blob/433863bfcaf52981296e1b653c198fd4eaf3d696/channels/chan_simpleusb.c#L2417-L2421

and all of the support for this chunk of code...

channels/chan_usbradio.c
```diff
@@ -1987,493 +1984,566 @@ static int usbradio_write(struct ast_channel *c, struct ast_frame *f)
...
+				if (o->clipledgpio) {
+					if (!o->hid_gpio_pulsetimer[o->clipledgpio - 1]) {
+						o->hid_gpio_pulsetimer[o->clipledgpio - 1] = CLIP_LED_HOLD_TIME_MS;
+					}
+				}
 			}
-		}
-	}
 
-	/* Below is an attempt to match levels to the original CM108 IC which has been
```
Allan-N, line 2199: Can we keep this comment (that explains the `legacyaudioscaling`) ?

channels/chan_usbradio.c
```diff
@@ -1987,493 +1984,566 @@ static int usbradio_write(struct ast_channel *c, struct ast_frame *f)
...
 #if DEBUG_CAPTURES == 1 && XPMR_DEBUG0 == 1
-	if (o->txcap2 && ftxcaptrace) {
-		fwrite((o->pmrChan->ptxDebug), 1, FRAME_SIZE * 2 * 16, ftxcaptrace);
-	}
+			if (o->txcap2 && ftxcaptrace) {
+				fwrite((o->pmrChan->ptxDebug), 1, FRAME_SIZE * 2 * 16, ftxcaptrace);
+			}
 #endif
 
-	/* Below is an attempt to match levels to the original CM108 IC which has been
```
Allan-N, line 2257: Can we keep this comment (that explains the legacyaudioscaling) ?

channels/chan_usbradio.c
hardenedpenguin: Yep — restored in 3a79450.

channels/chan_usbradio.c
hardenedpenguin: Same here — restored in 3a79450.

hardenedpenguin:
@mkmer Good catch — that was missing. In 3a79450 I mirrored the simpleusb pattern: only feed TX when `ast_radio_pa_write_available()` shows room, keep backlog in `txq`, and if the queue sits for more than `MAX_FRAME_DELAY` (200 ms) we log, clear `hasusb`, and `stream_cleanup()` so the stream restarts.

channels/chan_usbradio.c
```diff
@@ -1899,13 +1886,29 @@ static int usbradio_call(struct ast_channel *c, const char *dest, int timeout)
 	struct chan_usbradio_pvt *o = ast_channel_tech_pvt(c);
 
 	o->stophid = 0;
+	o->stopaudiothread = 0;
```
Allan-N: I'm a little confused here.  I thought that `stopaudiothread` was an indication to a newly started thread that it should stop ... and initializing that state makes sense.  Below, we check to see if we already have a thread and we don't create a new one if one doesn't already exist.  I'm thinking that we should only be initializing the `stopaudiothread` state IF we are creating a new thread.

and the same for `lastaudiotime` and the audio thread.

channels/chan_usbradio.c
```diff
@@ -1987,493 +1985,621 @@ static int usbradio_write(struct ast_channel *c, struct ast_frame *f)
...
+							f1 = NULL;
+						} else {
+							o->tonetime = ast_radio_tvnow();
+							o->toneflag = 1;
+						}
+					}
+					if (f1) {
+						ast_queue_frame(o->owner, f1);
+					}
+					continue;
```
Allan-N: Does this `continue` belong inside the `if (f1) {}` block above?

channels/chan_usbradio.c
hardenedpenguin: Good catch — that was wrong for the reuse path. In ac6b989 those inits only run when we actually create the HID/audio thread, so a live audio thread keeps `audio_thread_ready` and its watchdog time.

channels/chan_usbradio.c
hardenedpenguin: Yep — `continue` belongs with the queued DTMF frame. If BEGIN was discarded (`f1 == NULL`) we fall through and still deliver voice, matching the old `usbradio_read` behavior. Fixed in ac6b989.

channels/chan_usbradio.c
```diff
@@ -1987,493 +1984,621 @@ static int usbradio_write(struct ast_channel *c, struct ast_frame *f)
...
+	memset(&f1->frame_list, 0, sizeof(f1->frame_list));
+	ast_mutex_lock(&o->txqlock);
+	AST_LIST_TRAVERSE(&o->txq, f2, frame_list) {
+		qlen++;
 	}
+	if (qlen >= QUEUE_SIZE) {
+		ast_mutex_unlock(&o->txqlock);
+		ast_frfree(f1);
+		return 0;
+	}
```
mkmer: These are linked lists that can be infinitely long, why check for this?
We manage the "length" with the activity timer.  We "could" switch to a queue size to force the reset, or just stick with the timer.  Since everything is clocked on 20ms "clicks" I don't think it makes a difference if we use time or size.

channels/chan_usbradio.c
```diff
@@ -1987,493 +1984,621 @@ static int usbradio_write(struct ast_channel *c, struct ast_frame *f)
...
 			}
-			*sp++ = (int) v;
-		}
-	}
-
-	/* Write the received audio to the sound card */
-	soundcard_writeframe(o, o->usbradio_write_buf);
 
-	ast_radio_check_audio(o->usbradio_write_buf, &o->txaudiostats, AST_RADIO_PA_48K_STEREO_SAMPLES, 0);
+			soundcard_writeframe(o, o->usbradio_write_buf);
```
mkmer, line 2413: We recently modified simpleusb to always write silence frames IF there are no outbound frames.  By no longer intentionally underrunning the hardware buffer, the restart bug stays away :)
We should add it, but maybe in a different PR?
https://github.com/AllStarLink/app_rpt/pull/1161

channels/chan_usbradio.c
hardenedpenguin: Agreed — with the 200 ms drain timer in place the `QUEUE_SIZE` reject was just a second backstop. Dropped it in 0de1f16 and left the timer as the one that manages backlog.

channels/chan_usbradio.c
hardenedpenguin: Yeah, that belongs in a follow-up once this lands — keep #1142 scoped to getting I/O off the channel thread. Happy to mirror #1161’s silence writes for usbradio in a separate PR.

Allan-N (approved):

channels/chan_usbradio.c
```diff
@@ -1987,493 +1983,613 @@ static int usbradio_write(struct ast_channel *c, struct ast_frame *f)
...
 			}
-			*sp++ = (int) v;
-		}
-	}
 
-	/* Write the received audio to the sound card */
-	soundcard_writeframe(o, o->usbradio_write_buf);
-
-	ast_radio_check_audio(o->usbradio_write_buf, &o->txaudiostats, AST_RADIO_PA_48K_STEREO_SAMPLES, 0);
+			soundcard_writeframe(o, o->usbradio_write_buf);
```
mkmer, line 2413: Looking this over one last time, I think we have an architecture issue here: 
`PmrTx(o->pmrChan, u->data);` puts things on a second buffer inside the Pmr logic.  Maybe we could just use that buffer rather than adding a new one?
How do we reset the PmrTx buffer?
Then, we only read one of them off of the PmrTx buffer for every read of the usb hardware.  What we did over in simple USB is write as many as we can onto the hardware buffer so that we have some jitter time to deal with "waking up" from a read call.  
As this is right now, we can only load one frame per read to the hardware (right?)
Maybe if we prime it with an extra frame that will cover us for the jitter?
Then, adding the silence frames all the time may also keep use from starving?

hardenedpenguin: Yeah, fair point — looking at it again the TX path is a bit layered.

`txq` is only there so channel `.write` stays non-blocking and doesn’t touch XPMR from the Asterisk thread. `PmrTx`/dedrift is the old XPMR skew buffer. Different jobs, but you’re right that we now have two places audio can sit, and today we only push one PA frame per USB read. SimpleUSB filling the hardware buffer for wake-up jitter is clearly the better pattern.

Also noticed `stream_cleanup()` doesn’t reset the dedrift buffer on restart — that feels like a real gap.

A few ways we could go:

1. **Keep `txq`**, but when PA has room run multiple feed → `PmrRx` → write cycles (silence on the RX side for the extras if we need them), and clear/reset dedrift on cleanup. Closest to what simpleusb does for jitter without ripping out the thread handoff.

2. **Drop `txq`** and have `.write` take a lock and call `PmrTx` again (like before the thread). Audio thread only does read / `PmrRx` / PA fill. One less buffer, but the channel thread touches XPMR again under a mutex.

3. **Silence-always** (#1161 style) / priming an extra frame — I’d still rather do that as a follow-up so this PR stays “get blocking PA off the channel thread.”

I’m leaning toward (1) for this PR unless you’d rather bite the bullet on (2). What do you prefer?

mkmer: (1) seems sane, is it "ok" to call PmrRx() with a silence frame multiple times per loop (I have not dug into it).  
General thoughts follow:
Maybe better to separate the `PmrRX` and `TX` remove part into 2 functions?  Something like `PmrTX_` unload / remove / send??? (Not sure of a "good" name here). I suppose if we start tweaking `Pmr*` we could just use the one PmrTX buffer and add the support we need?
We do need to pump as many frames out of the buffer and into the USB TX buffer as we can for any 20ms cycle.  We will never read more than 1 frame out of the USB buffer as we have it setup that way.

hardenedpenguin: I dug into `PmrRx` and you're right to be wary. Calling it multiple times per loop isn't safe: each call is wired to be exactly one 20 ms tick. It bumps `frameCountRx`, decrements the blanking/VOX/smode timers by `MS_PER_FRAME`, runs the full RX sps chain + CTCSS detect, and calls `dedrift()` once — which is what pulls a single TX frame off the buffer. So draining N TX frames by calling `PmrRx` N× would advance all the RX-side timing N times per real 20 ms and throw off RX.

That's exactly why your idea of splitting the TX unload out of `PmrRx` is the right shape — a dedicated `PmrGetTx()`-style call that pulls from the dedrift buffer and runs only the TX sps chain, callable on its own. That also lets us drop the extra `txq` and lean on the one dedrift buffer like you're suggesting. But that's real XPMR surgery and I'd rather not bury it inside this threading PR.

So my proposal: for this PR, do the minimal cushion fix — prime an extra frame at stream start and write silence when the queue is empty (same thing #1161 did for simpleusb). Still one real TX frame per 20 ms so RX timing stays honest, but the hardware TX buffer keeps a cushion so a late wake-up doesn't underflow. Then do the `PmrRx`/TX split as its own follow-up where it can get proper review.

Would you rather land the prime+silence cushion here and tackle the XPMR TX split separately, or hold this one and do the split now? I'm good either way.

Allan-N: A question/thought?  Taking the frames and adding them to txq isn't really moving the data around.  Isn't it just adding the frame to a list?  We're not really putting the frame into a second buffer, we're just moving where we handle the buffered data from one place to another.

hardenedpenguin: Agreed — `txq` isn't a second audio buffer, just a hand-off list. `usbradio_write` (channel thread) `ast_frdup`s the frame and appends it; the audio thread pops each one and calls `PmrTx()`, which is the only real buffer (the dedrift ring). We've just moved *where* `PmrTx()` runs so the blocking PA write doesn't stall the channel thread. Only nuance: the dup is a small transient copy, freed as soon as `PmrTx()` consumes it.

mkmer: I think it's an "extra" step having the `txq` when the `PmrTx()` call will buffer them up anyway.  We CAN call `PmrTx()` multiple times and all is fine (I believe that's right anyway).  So, if the "final" answer is to modify `Pmr*` to separate the tx/rx logic, maybe we should head that way now replace `txq` with `PmrTx()`?  Or maybe not, as we don't have a way to flush the queue hiding inside `PmrTx()`
This is a bit ugly for sure... 
I suppose, as long as we get PA non blocking and "working" -> we have a base that isn't perfect, but works and recovers... 
I'm leaning toward your "minimal" option.

hardenedpenguin: Sounds good — we'll stick with the minimal path for this PR: keep `txq` as the hand-off, prime an extra frame at stream start, and write silence when the queue is empty (same idea as #1161). That gets PA non-blocking and recovers without touching XPMR.

One clarification for later: calling `PmrTx()` multiple times to enqueue is fine; the unsafe part is calling `PmrRx()` multiple times to unload TX (that advances RX timers). The `Pmr*` TX/RX split + dropping `txq` can be a follow-up.

mkmer (approved):

PR #1164, opened 2026-08-06
app_rpt.c: autoservice l->pchan before node_lookup blocking call

mkmer:
Turns out `node_lookup()` blocks us, leaving a few long voice queue warnings.  

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reconnection handling by keeping the parent channel serviced during connection recovery.
  * Added consistent retry behavior for lookup, unsupported-node, capability, and call-request failures.
  * Reduced the risk of blocked channel activity while reconnect attempts or call setup are in progress.
  * Ensured servicing stops appropriately after successful setup or unrecoverable call-placement failures.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2509,6 +2509,10 @@ static void *attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	struct ast_format_cap *cap;
 
 	ast_debug(1, "Attempting Reconnect");
+	/* rpt_make_call is blocking, long dns lookups result in exceptionally long queue warnings
+	 * autoservice handles "eating" the frames and eliminating the warning.
+	 */
+	ast_autoservice_start(l->pchan);
 	if (node_lookup(myrpt, l->name, tmp, sizeof(tmp), 1)) {
 		ast_log(LOG_WARNING, "attempt_reconnect: cannot find node %s\n", l->name);
 		rpt_mutex_lock(&myrpt->lock);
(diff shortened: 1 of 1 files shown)


apps/app_rpt.c
```diff
@@ -2509,6 +2509,10 @@ static void *attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	struct ast_format_cap *cap;
 
 	ast_debug(1, "Attempting Reconnect");
+	/* rpt_make_call is blocking, long dns lookups result in exceptionally long queue warnings
```
Allan-N: This comment needs a bit of help.  Is it `rpt_make_call()` that is blocking (or taking an extended amount of time)?  or `node_lookup()`?  or both?

Also, you now have a few places where the function returns early.  Do we need to call `ast_autoservice_stop()` before each of the early returns?

mkmer: True true

apps/app_rpt.c
```diff
@@ -2509,34 +2509,26 @@ static void *attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
...
-		return NULL;
+		goto retry;
 	}
 	/* cannot apply to tlb */
 	if (!strncasecmp(tmp, "tlb", 3)) {
-		rpt_mutex_lock(&myrpt->lock);
-		l->retrytimer = RETRY_TIMER_MS;
-		rpt_mutex_unlock(&myrpt->lock);
-		return NULL;
+		goto retry;
```
mkmer, line 2526: Should we be retrying on these or just giving up? Seems like they will not change "type".

Allan-N (approved):

Issue #1166, opened 2026-08-07
app_rpt: attempt_reconnect review tlb and echolink "failure" situation


mkmer:
> Should we be retrying on these or just giving up? Seems like they will not change "type". 

 _Originally posted by @mkmer in [#1164](https://github.com/AllStarLink/app_rpt/pull/1164/changes#r3735778498)_

PR #1171, opened 2026-08-07
app_rpt: Prevent echolink and tlb retry loop without call

mkmer:
Addresses Issue #1166 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reconnect handling for EchoLink and TLB connections by preventing unsupported retry attempts.
  * Ensured these connection types are cleaned up promptly when reconnection is unavailable.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2519,10 +2519,14 @@ static void *attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 	}
 	/* cannot apply to echolink */
 	if (!strncasecmp(tmp, "echolink", 8)) {
+		/* We should not retry this node type */
+		l->retries = l->max_retries + 1;
 		goto retry;
 	}
 	/* cannot apply to tlb */
 	if (!strncasecmp(tmp, "tlb", 3)) {
+		/* We should not retry this node type */
+		l->retries = l->max_retries + 1;
 		goto retry;
 	}
 


Allan-N (commented):
When/how does this retry code come into play?  What is the current behavior if an EchoLink server is not responding?  or if the network is unavailable / flakey?

... and I'm asking here because I think these are cases where a retry makes sense.

... or does EL/TLB already take care of retrying?

mkmer:
https://github.com/AllStarLink/app_rpt/pull/1171#pullrequestreview-4885288111
Maybe it makes sense?  I was just looking at what is does today, which is NOT retry the node, just add to the counter and eventually give up assuming it's not a permanent connection... then it just sucks up a connection to the conference and never dies.

More research would be necessary to see if we broke this moving from the old code base.

Allan-N (approved):

mkmer:
Looks like we "always" jumped out
https://github.com/AllStarLink/ASL-Asterisk/blob/d6a4b48d58288978bf2e5bf22e7e9cfcdea8b2f3/asterisk/apps/app_rpt.c#L18754-L18757

In general, I don't think we are supposed to trigger a retry here, so this is more of a "protection" in case it happens?
https://github.com/AllStarLink/ASL-Asterisk/blob/d6a4b48d58288978bf2e5bf22e7e9cfcdea8b2f3/asterisk/apps/app_rpt.c#L21691-L21725

mkmer:
Requires #1217 to be merged first and rebased....

PR #1176, opened 2026-08-09
chan_voter.c Remove Proxy/Redundant Server Code

### LINKED ISSUE #915: chan_voter: Redundant Proxy Server Authentication Broken
In reviewing the client authentication process (which has its own issues), it appears that the authentication for the "Redundant Proxy Server" mode of operation in `chan_voter` is likely broken.

## Overview of the Authentication Process
Authentication is handled in the `voter_reader` thread.

Per the VOTER Protocol Specification, a client initially sends a packet with a Payload Type of 0 (authentication), a "challenge", and a blank "digest" to the host.

The host then responds with a Payload Type 0 (authentication) packet, containing its own unique "challenge", as well as a "digest" that is computed from the client's "challenge" (with some other data). 

If the client agrees with the "digest" that the host sent, it responds back with another Payload Type 0 (authentication) packet, containing its "challenge", but now with a "digest" that is computed using its "client password". 

Upon receipt of this authentication packet from the client, the host can then use the "digest" that the client sent to identify which client this is by comparing the client password encoded in the digest with the clients configured in `voter.conf`. If there is a matching client found in `voter.conf`, the client is allowed to connect, and we are off to the races.

## The Problem
As noted above, the first packet the host should receive from a client is a Payload Type 0 (authentication) packet, with a blank "digest".

At this point, we don't know anything about the client, other than there is a client that wants to authenticate.

In the `voter_reader` thread, we start by resetting the flag for `isproxy` here:

https://github.com/AllStarLink/app_rpt/blob/afd4cba132bce89b3bfc6683bf538b7e8bbf2d13/channels/chan_voter.c#L4570

In order to set `isproxy` to true, we need to get to here:

https://github.com/AllStarLink/app_rpt/blob/afd4cba132bce89b3bfc6683bf538b7e8bbf2d13/channels/chan_voter.c#L4757

However, in order to get to that, we need to enter the (large) main packet processing loop, which is bounded by:

https://github.com/AllStarLink/app_rpt/blob/afd4cba132bce89b3bfc6683bf538b7e8bbf2d13/channels/chan_voter.c#L4571

Now, since in the first authentication packet sent to the host by the client the "digest" is empty, `vph->digest` is therefore empty, so we never enter the main packet processing loop. So, on that first authentication packet received, we can never set `isproxy`.

Since we don't enter the main packet processing loop, we fall through to the main packet authentication code below it, which should be fine. 

The problem is, when we hit the code for the host to figure out what type of reply to send the client for this initial packet, `isproxy` is unset, therefore it appears we can never send the proper response packet:

https://github.com/AllStarLink/app_rpt/blob/afd4cba132bce89b3bfc6683bf538b7e8bbf2d13/channels/chan_voter.c#L5609-L5622

Since `isproxy` is unset, we're going to default to sending a normal response packet to the client (probably not intended).

The initial response packet for a proxy client is different than for a normal client, as it appears to contain IP information for the proxy server, and also uses a specific proxy Payload Type in its response.

Therefore, it would appear that the authentication process for proxy clients is broken.

We never hit the authentication code again (in normal operation), because once the client sends its "digest", `vph->digest` would now always have something in it, so we are going to enter the main packet processing loop.

Further investigation is needed to see how a proxy client ever gets around to sending a proxy Payload Type, because otherwise we can never get to:

https://github.com/AllStarLink/app_rpt/blob/afd4cba132bce89b3bfc6683bf538b7e8bbf2d13/channels/chan_voter.c#L4749

## To-do
Probably need to debug and maybe Wireshark packets in a test environment with a proxy server and some clients, to figure out how/if the authentication is working as intended, and resolve outstanding issues.

Perhaps we need to do something in `voter_request` when we read `voter.conf` and the associated variables related to this server configuration to set it up for proxy authentication?

ve7fet:
The redundant server/proxy mode of operation was intended to provide a method of redundancy for voting clients, in the event of loss of communication with the primary voting server. The method by which this was envisaged can be found in the original VOTER System documentation written by Jim Dixon, as well as in the top of the existing `chan_voter.c` source.

While redundancy and resiliency is always a laudable goal in networks, the implementation in this case seems to have some issues.

1) There doesn't seem to be a suitable use case where this would actually provide the stated benefit(s). In simple terms, if your voting clients (VOTER or RTCM) can't talk to their ASL3 host, it is likely that you have bigger network or hardware issues that would prevent them from talking to a redundant server.

2) While this method provides a way for voting clients to communicate with two different servers, it is still left to the user to figure out how to configure the two ASL (`app_rpt`) servers to manage some sort of link to handle the audio and signaling between them (this isn't really spelled out in the VOTER System documentation).

3) It doesn't appear that this code even really functioned fully. Going back at least a few years to when there was a major cleanup of the code (for readability) in https://github.com/AllStarLink/app_rpt/commit/2d55a4e870b3bca87592d28e25534ad4197ed646, and reviewing the authentication process, it appears broken. 

Specifically, the flag `isproxy` used to designate a proxy server is required here to enable sending the initial authentication request to a (primary) proxy server:

https://github.com/AllStarLink/app_rpt/blob/2d55a4e870b3bca87592d28e25534ad4197ed646/channels/chan_voter.c#L4939-L4949

However, `isproxy` is only set here:

https://github.com/AllStarLink/app_rpt/blob/2d55a4e870b3bca87592d28e25534ad4197ed646/channels/chan_voter.c#L4079

Which is nested inside a check that `vph->digest` exists:

https://github.com/AllStarLink/app_rpt/blob/2d55a4e870b3bca87592d28e25534ad4197ed646/channels/chan_voter.c#L3856-L3857

When a client (in this case, the "secondary" server) sends an initial authentication request, the first packet is sent with a `digest` of `0`... so we can never set `isproxy`.

In addition, there appears to be no code that specifically sends a proxy authentication packet that would overcome this obstacle.

So, we seem to have a "chicken and egg" problem? Issue #915 has a more in-depth explanation of the authentication process.

<br>
Suffice to say, it appears that this code is "broken", and has been for a long time (ever?). It also would seem that no one is using this code, as no one has complained that it is "broken". An attempt was made to solicit feedback from the Community to see if there are any active users (https://community.allstarlink.org/t/does-anyone-use-the-voter-rtcm-proxy-server-configuration/24778), with no responses.

The proxy/redundant server code adds a significant amount of complexity and duplication to the source code, for little to no gain. 

Therefore, this PR has been created to obsolete this code, and remove it from the source.

Resolves #915

## Summary by CodeRabbit

* **Refactor**
  * Simplified channel connectivity to use direct client connections only.
  * Removed support for primary/secondary server proxying and related connection management.
  * Updated status, authentication, discovery, audio transmission, and channel teardown behavior to reflect the streamlined connection model.
  * Removed obsolete redundant-server configuration options.

### DIFF
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -4767,8 +4767,9 @@ static void *voter_reader(void *data)
 			 */
 
 			authpacket.vp.payload_type = htons(VOTER_PAYLOAD_AUTH);
-			ast_debug(2, "Sending initial packet payload %i challenge %s digest %08x password %s to client %s\n", authpacket.vp.payload_type,
-				ntohs(authpacket.vp.challenge), ntohl(authpacket.vp.digest), password, ((client) ? client->name : "UNKNOWN"));
+			ast_debug(2, "Sending initial packet payload %i challenge %s digest %08x password %s to client %s\n",
+				ntohs(authpacket.vp.payload_type), authpacket.vp.challenge, ntohl(authpacket.vp.digest), password,
+				((client) ? client->name : "UNKNOWN"));
 			sendto(udp_socket, &authpacket, sizeof(authpacket), 0, (struct sockaddr *) &sin, sizeof(sin));
 			continue;
 		}
--- a/channels/chan_voter.c
+++ b/channels/chan_voter.c
@@ -3256,12 +3256,11 @@ static void *voter_xmit(void *data)
 				audiopacket.vp.curtime.vtime_nsec = client->mix ? htonl(client->txseqno) : htonl(master_time.vtime_nsec);
 				/* Check to see if this client is a transmitter (transmit set in voter.conf) AND is NOT locked out from transmitting. */
 				if (client->totransmit && !client->txlockout) {
-					ast_debug(6, "VOTER %i: Sending ulaw TX audio packet to client %s digest %08x\n", p->nodenum,
-						client->name, client->respdigest);
+					ast_debug(6, "VOTER %i: Sending ulaw TX audio packet to client %s digest %08x\n", p->nodenum, client->name,
+						client->respdigest);
 					/* Send the ulaw audio packet over the wire to the client for transmitting */
-					sendto(udp_socket, &audiopacket, sizeof(audiopacket) - 3, 0, (struct sockaddr *) &client->sin,
-						sizeof(client->sin));
-					
+					sendto(udp_socket, &audiopacket, sizeof(audiopacket) - 3, 0, (struct sockaddr *) &client->sin, sizeof(client->sin));
+
 					/* Update when this client last sent an audio packet */
 					client->lastsenttime = ast_radio_tvnow();
 				}
@@ -3319,9 +3318,8 @@ static void *voter_xmit(void *data)
 						ast_debug(6, "VOTER %i: Sending ADPCM TX audio packet to client %s digest %08x\n", p->nodenum,
 							client->name, client->respdigest);
 						/* Send the ADPCM audio packet over the wire to the client for transmitting */
-						sendto(udp_socket, &audiopacket, sizeof(audiopacket), 0, (struct sockaddr *) &client->sin,
-							sizeof(client->sin));
-						
+						sendto(udp_socket, &audiopacket, sizeof(audiopacket), 0, (struct sockaddr *) &client->sin, sizeof(client->sin));
+
 						/* Update when this client last sent an audio packet */
 						client->lastsenttime = ast_radio_tvnow();
 					}
@@ -3420,9 +3418,8 @@ static void *voter_xmit(void *data)
 				audiopacket.vp.digest = htonl(client->respdigest);
 				audiopacket.vp.curtime.vtime_nsec = client->mix ? htonl(client->txseqno) : htonl(master_time.vtime_nsec);
 				ast_debug(5, "VOTER %i: Sending keepalive packet to client %s digest %08x\n", p->nodenum, client->name, client->respdigest);
-				sendto(udp_socket, &audiopacket, sizeof(VOTER_PACKET_HEADER), 0, (struct sockaddr *) &client->sin,
-					sizeof(client->sin));
-				
+				sendto(udp_socket, &audiopacket, sizeof(VOTER_PACKET_HEADER), 0, (struct sockaddr *) &client->sin, sizeof(client->sin));
+
 				/* Update when this client last sent a keepalive packet */
 				client->lastsenttime = ast_radio_tvnow();
 			}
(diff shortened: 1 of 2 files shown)


channels/chan_voter.c
```diff
@@ -5162,25 +4768,13 @@ static void *voter_reader(void *data)
...
-					authpacket.vp.payload_type, authpacket.vp.challenge, ntohl(authpacket.vp.digest), password,
+
+			authpacket.vp.payload_type = htons(VOTER_PAYLOAD_AUTH);
+			ast_debug(2, "Sending initial packet payload %i challenge %s digest %08x password %s to client %s\n",
+				authpacket.vp.payload_type, authpacket.vp.challenge, ntohl(authpacket.vp.digest), password,
 					((client) ? client->name : "UNKNOWN"));
-				sendto(udp_socket, &authpacket, sizeof(authpacket), 0, (struct sockaddr *) &sin, sizeof(sin));
-			}
+			sendto(udp_socket, &authpacket, sizeof(authpacket), 0, (struct sockaddr *) &sin, sizeof(sin));
+
```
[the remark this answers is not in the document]
ve7fet: Resolved in ee9584a. Not really an issue, since the payload type is always 0 here anyways, but good for code quality.

mkmer (approved):
Nice code diet :)

channels/chan_voter.c
```diff
@@ -424,7 +348,6 @@ static char context[AST_MAX_EXTENSION] = "default";
 #define VOTER_PAYLOAD_ADPCM 3
 #define VOTER_PAYLOAD_FUTURE 4 /* Reserved for future use */
 #define VOTER_PAYLOAD_PING 5
-#define VOTER_PAYLOAD_PROXY 0xf000
```
Allan-N, line 427: Is this a protocol constant (one that might show up over the network)?  If so, I'd suggest keeping the #define but add a `/* No longer supported */` comment.  But, this is only an internal thing then it can most certainly be removed.

channels/chan_voter.c
```diff
@@ -515,14 +438,6 @@ typedef struct {
 	uint8_t audio[FRAME_SIZE];
 	uint8_t rssi;
 } VOTER_REC;
-
-typedef struct {
```
Allan-N, line 519: As above, if this structure reflects an over-the-wire packet then maybe leave this here, commented out, as documentation.

channels/chan_voter.c
```diff
@@ -3272,12 +3007,6 @@ static void *voter_xmit(void *data)
 		char rssi;
 		char audio[FRAME_SIZE + 3];
 	} audiopacket;
-	struct {
```
Allan-N, line 3275: Keep, commented out, as documentation?

channels/chan_voter.c
```diff
@@ -5000,11 +4619,6 @@ static void *voter_reader(void *data)
 		VOTER_PACKET_HEADER vp;
 		char flags;
 	} authpacket;
-	struct {
```
Allan-N, line 5003: Keep, commented out, as documentation?

channels/chan_voter.c
ve7fet: No, it won't show up on the network. That payload was only used between proxy servers, not between clients and the server. So, with the proxy code removed, it will never be seen again.

channels/chan_voter.c
ve7fet: VOTER_PROXY_HEADER was only for inter-server proxy communications. All the related code that used it is being removed, so there is nothing to handle it.

channels/chan_voter.c
ve7fet: The code that used proxy_audiopacket was for inter-server proxy communications, and is all being removed.

channels/chan_voter.c
ve7fet: Same here, there is no proxy authentication to do anymore, as all the proxy code is being removed. This was just for inter-server proxy communication, which will no longer exist.

Allan-N (approved):

PR #1178, opened 2026-08-12
chan_usbradio: reset XPMR dedrift and keep PA TX full when idle

hardenedpenguin:
## Summary
- Follow-up to #1162 / Mike’s review: reset the XPMR dedrift ring on stream restart so stale TX does not replay after a USB hiccup
- When unkeyed, fill remaining PortAudio room with silence (same idea as #1161) so a late USB read does not underrun
- Keyed TX stays one XPMR frame per 20 ms tick; pumping multiple real TX frames still needs a later `PmrGetTx` split

## Test plan
- [ ] Build/load `chan_usbradio` on a PortAudio USB radio node
- [ ] Confirm RX, COS/CTCSS, DTMF, and TX/PTT still work (including duplex)
- [ ] Yank/replug USB and confirm audio restarts without a burst of old TX
- [ ] Leave the node unkeyed and confirm the stream stays alive (no underflow/chop)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USB radio audio stream recovery when playback or transmission writes fail.
  * Reset audio drift correction state during stream startup and cleanup for more reliable operation.
  * Ensured unkeyed audio output is filled with silence, reducing potential playback gaps or stale audio.

### DIFF
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -2416,43 +2416,16 @@ static void *usbradio_audio_thread(void *arg)
 				}
 			}
 
+			/*
+			 * One frame per 20 ms tick. When unkeyed, soundcard_writeframe()
+			 * substitutes silence; ast_radio_pa_write() primes one more on
+			 * underrun. Do not fill remaining PA room; that adds TX delay.
+			 */
 			if (!soundcard_writeframe(o, o->usbradio_write_buf)) {
 				stream_cleanup(o);
 				break;
 			}
 			ast_radio_check_audio(o->usbradio_write_buf, &o->txaudiostats, AST_RADIO_PA_48K_STEREO_SAMPLES, 0);
-
-			/*
-			 * Unkeyed: fill remaining PortAudio room with silence so a late USB
-			 * read does not underrun (simpleusb #1161). Do not call PmrRx again;
-			 * that would advance RX timers. Keyed TX stays one XPMR frame per tick
-			 * until PmrGetTx exists.
-			 */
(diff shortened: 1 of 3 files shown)


Allan-N (approved):

channels/chan_usbradio.c
```diff
@@ -2410,9 +2416,44 @@ static void *usbradio_audio_thread(void *arg)
...
 
-			soundcard_writeframe(o, o->usbradio_write_buf);
+			if (!soundcard_writeframe(o, o->usbradio_write_buf)) {
+				stream_cleanup(o);
+				break;
+			}
 			ast_radio_check_audio(o->usbradio_write_buf, &o->txaudiostats, AST_RADIO_PA_48K_STEREO_SAMPLES, 0);
 
+			/*
+			 * Unkeyed: fill remaining PortAudio room with silence so a late USB
```
mkmer: The problem with filling the hardware buffer here is that we have indeterminate amount of additional delay on the audio loop.  Once the additional silence frames are in there, they essentially add delay.  I think it's best to just put 1 frame in, if we under run, that code will add 1 additional silence frame.  At that point we should stay ahead based on the 20ms inbound audio "timing" - we only need 1 frame to "cover" the jitter /lateness to responding to a "hardware' based message.

hardenedpenguin: Agreed — packing the remaining PortAudio room was adding indeterminate TX delay. Dropped that fill loop; we now stay at one frame per 20 ms tick, with the existing underrun silence prime as the cushion.

mkmer (approved):

PR #1179, opened 2026-08-13
res_usbradio: support mono PortAudio TX for AIOC-class URIs

hardenedpenguin:
## Summary
- All-In-One-Cable (and similar URIs) advertise mono playback to PortAudio (`maxOutputChannels=1`); forcing stereo open fails where OSS used to accept stereo and convert
- Clamp TX channel count to the device max on `ast_radio_pa_open()`, and downmix stereo interleaved TX to mono in `ast_radio_pa_write()` when the stream is mono
- CM108-class stereo devices are unchanged (still open and write as 2 channels)

## Test plan
- [ ] Build/load `res_usbradio` + `chan_simpleusb` (or `chan_usbradio`) on an AIOC node (`AllInOneCable` / `hw:N`)
- [ ] Confirm PortAudio opens with 1 output channel (debug: `PortAudio opened ... output channel(s)`)
- [ ] Confirm RX/TX/PTT and tune levels still work
- [ ] Regression: CM108/CM119 stereo URI still opens with 2 output channels and TX is unaffected

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USB radio compatibility with devices that support different numbers of audio channels.
  * Correctly reports the actual transmit channel count after opening a radio stream.
  * Automatically downmixes stereo transmit audio for mono-output devices.
  * Preserves existing underflow handling during audio transmission.

### DIFF
--- a/res/res_usbradio.c
+++ b/res/res_usbradio.c
@@ -1271,6 +1271,24 @@ static void pa_clamp_input_channels(PaDeviceIndex device, unsigned int *input_ch
 	}
 }
 
+/*!
+ * \brief Cap requested TX channels to what PortAudio reports for the device.
+ * Mono URIs (AIOC) advertise maxOutputChannels=1; CM108-class devices report 2.
+ */
+static void pa_clamp_output_channels(PaDeviceIndex device, unsigned int *output_channels)
+{
+	const PaDeviceInfo *out_dev;
+
+	if (!output_channels || device == paNoDevice) {
+		return;
+	}
+
+	out_dev = Pa_GetDeviceInfo(device);
+	if (out_dev && *output_channels > (unsigned int) out_dev->maxOutputChannels) {
+		*output_channels = out_dev->maxOutputChannels ? (unsigned int) out_dev->maxOutputChannels : 1;
+	}
+}
+
 PaError ast_radio_pa_open(struct ast_radio_pa_stream *ps)
 {
 	PaError res = paInternalError;
@@ -1282,6 +1300,7 @@ PaError ast_radio_pa_open(struct ast_radio_pa_stream *ps)
 
 	ps->stream = NULL;
 	ps->active = 0;
+	ps->output_channels = AST_RADIO_PA_OUTPUT_CHANNELS;
 
 	if (pa_lib_acquire() < 0) {
 		return paInternalError;
@@ -1289,8 +1308,9 @@ PaError ast_radio_pa_open(struct ast_radio_pa_stream *ps)
 
 	if (!strcasecmp(ps->hw_device, "default")) {
 		pa_clamp_input_channels(Pa_GetDefaultInputDevice(), &ps->input_channels);
-		res = Pa_OpenDefaultStream(&ps->stream, ps->input_channels, AST_RADIO_PA_OUTPUT_CHANNELS, paInt16,
-			AST_RADIO_PA_SAMPLE_RATE, AST_RADIO_PA_FRAMES_PER_BUFFER, NULL, NULL);
+		pa_clamp_output_channels(Pa_GetDefaultOutputDevice(), &ps->output_channels);
+		res = Pa_OpenDefaultStream(&ps->stream, ps->input_channels, ps->output_channels, paInt16, AST_RADIO_PA_SAMPLE_RATE,
+			AST_RADIO_PA_FRAMES_PER_BUFFER, NULL, NULL);
 	} else {
 		PaStreamParameters input_params = {
 			.channelCount = ps->input_channels,
@@ -1299,7 +1319,7 @@ PaError ast_radio_pa_open(struct ast_radio_pa_stream *ps)
 			.device = paNoDevice,
 		};
 		PaStreamParameters output_params = {
-			.channelCount = AST_RADIO_PA_OUTPUT_CHANNELS,
+			.channelCount = ps->output_channels,
 			.sampleFormat = paInt16,
 			.suggestedLatency = (1.0 / 50.0),
 			.device = paNoDevice,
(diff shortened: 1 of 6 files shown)


res/res_usbradio.c
```diff
@@ -1433,12 +1457,23 @@ PaError ast_radio_pa_write(struct ast_radio_pa_stream *ps, const short *data, un
 		return paBufferTooBig;
 	}
 
-	res = Pa_WriteStream(ps->stream, data, pa_frames);
+	/*
```
mkmer, line 1460: While this fixes the 1 channel problem, what do we do with a configuration that is using stereo?
In simpleusb, we have a "pager config" where we put the pager on channel a or b and the repeater audio goes to the "other" channel.
The change here will cause some trouble if the pager is set to A, the repeater audio will be on B and silence will most likely be the result.

https://github.com/AllStarLink/app_rpt/blob/4571bbc455d278f0bf443108d31f0f8b9b8d1d90/channels/chan_simpleusb.c#L2361-L2392

YES, it's a fringe config problem, but would certainly be a head scratcher.
Maybe we need to do something in the channels preventing any stereo action if there is only 1 channel?

res/res_usbradio.c
```diff
@@ -1433,12 +1457,23 @@ PaError ast_radio_pa_write(struct ast_radio_pa_stream *ps, const short *data, un
...
-	res = Pa_WriteStream(ps->stream, data, pa_frames);
+	/*
+	 * Callers always pass stereo interleaved samples. Mono URIs (AIOC) open
+	 * with one output channel; average L/R like OSS/ALSA plug conversion did.
+	 */
+	if (ps->output_channels == 1) {
+		unsigned long i;
+
+		for (i = 0; i < pa_frames; i++) {
+			mono_buf[i] = (short) (((int) data[i * 2] + (int) data[i * 2 + 1]) / 2);
```
Allan-N, line 1468: In `usbradio.conf` we have `txmixa =` and `txmixb =` variables with values of `no`, `voice`, `tone`, `composite`, and `auxvoice`.  How does this stereo -> mono mixing align with the .conf settings?  Do the settings determine what goes into the left/right values?  or how the left/right values need to be mixed?  Do we need to change the above logic to be aware of the `txmixa`/`txmixb` settings?

res/res_usbradio.c
hardenedpenguin: Good catch. The res_usbradio downmix averages L/R rather than taking left-only, so `pager=a` would not mute the repeater — it would drop ~6 dB because the unused side is zeros. Stereo CM108/CM119 is unchanged (`output_channels` stays 2).

Agreed the split belongs in the channel. `5d40902` skips the pager A/B routing when PortAudio opened with 1 TX channel, so both sides stay live and the average is full level. It also warns at stream open if `pager` is set on a mono URI.

`txmixa`/`txmixb` in usbradio is the same class of problem (voice vs CTCSS vs aux on A vs B). A mono URI cannot split those either; left that alone here since AIOC is usually a single radio TX.

res/res_usbradio.c
hardenedpenguin: `txmixa`/`txmixb` choose what XPMR puts on left vs right. They do not tell PortAudio how to mix. The average here is after that, with no knowledge of the conf.

Default `txmixa=composite`, `txmixb=no` zeros B, so a 1-channel URI would drop ~6 dB the same way `pager=a` did. If A and B are different sources (`voice` vs `tone`), the average turns them into composite. If they already match, it is a no-op.

I would keep this downmix mixer-unaware. Same pattern as the pager guard: `chan_usbradio` should not leave one side silent when `output_channels==1` (treat `no` as same as the other mix; XPMR already has `monoOut` when A==B) and warn if A and B request different sources. AIOC is usually simpleusb, so I left that out of this PR.

Allan-N: >AIOC is usually simpleusb

That is not true.  I believe AIOC is usually setup with usbradio.

mkmer: https://na6d.com/pages/aioc-all-in-one-cable-setup-guide-for-allstarlink?srsltid=AfmBOor0nwWlEwkeFMPkCMqIfl68vhn-MCKxPR1JNPgCDPFqEId4o-uL
Seems to tell people to use `chan_usbradio`...

hardenedpenguin: > > AIOC is usually simpleusb
> 
> That is not true. I believe AIOC is usually setup with usbradio.

We are gonna have to agree to disagree, his original releases for a couple of years had everyone setup emulation which made it a simpleusb setup. I will close the request, I  will let you all decide how you want to handle it.

Allan-N:
I am re-opening this pull request as I think we still need to have some support for 1-channel (mono) audio interfaces.

As for the AIOC, this was just the first case of a common interface that was working for me with the pre-ALSA/PortAudio code that no longer functions.  I'd really like to avoid any regressions.

Are the changes currently on the branch sufficient to get us going again (with both SimpleUSB and USBRadio) ?

hardenedpenguin:
Thanks for reopening.

Yes — with the latest commits the branch covers both drivers for mono TX:

- **res_usbradio**: clamp output channels to the device max and average L/R on write (OSS/ALSA plug behavior)
- **simpleusb**: skip `pager=a/b` A/B split when PortAudio opened mono (`5d40902`)
- **usbradio**: after PA open, mirror an inactive `txmix` side so the average stays full level; voice+tone is kept and averaged into composite; unsupported differing pairs keep A and disable B (`de41807`, `cbfe9d1`)

Stereo CM108/CM119 is unchanged. Default AIOC-style `txmixa=composite` / `txmixb=no` should no longer come up ~6 dB quiet on a 1-channel URI.

res/res_usbradio.c
hardenedpenguin: Follow-up landed for USBRadio as well (`de41807`, `cbfe9d1`):

- inactive `txmix` side is mirrored on mono so the PA average stays full level
- voice+tone (either order) keeps both sources and averages into composite
- any other differing pair keeps A and disables B (explicit warning)

`res_usbradio` stays mixer-unaware; the channel owns the policy.

Allan-N (approved):

PR #1180, opened 2026-08-13
app_rpt: park link pchan out of CONF while IAX is down

hardenedpenguin:
## Summary
- When an outbound IAX link drops, `Announcer/IAXLink` stayed in CONF and mix audio filled that pchan, then backpressured every conference member (long voice-queue warnings, Allmon CONNECTING, registrations failing)
- Pull `l->pchan` out of CONF on link hangup
- Restore it only after `AST_CONTROL_ANSWER` (not after `rpt_make_call()` / while still CONNECTING)
- Wait until `ast_bridge_remove()` actually finishes before treating the channel as gone; restore only if the channel is in the requested conference (not some other bridge)

Fixes the cascade discussed in https://community.allstarlink.org/t/exceptionally-long-voice-queue-length-error-message-causing-node-to-fail/24807 and complements #1147 / #1164 (autoservice during blocking DNS).

## Test plan
- [ ] Outbound IAX link: yank network / kill peer, confirm no `Exceptionally long voice queue` on `Announcer/IAXLink`, `PChan`, `TXPChan`, or `LocalTX` while reconnecting
- [ ] Confirm reconnect succeeds and audio both ways returns only after answer
- [ ] Inbound link disconnect: pchan parked, thread exits cleanly without restoring
- [ ] Hub with several links: one flapping peer does not stall AMI / IAX registration
- [ ] Permanent (`lconn`) retry still redials after DNS recovery

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4787,6 +4787,20 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 				if (f->subclass.integer == AST_CONTROL_ANSWER) {
 					char lconnected = l->connected;
 
+					/*
+					 * Restore parked pchan before committing connected state.
+					 * On failure leave retries untouched so remote_hangup_helper
+					 * still honors max_retries instead of resetting the budget.
+					 */
+					if (l->pchan && rpt_conf_restore(l->pchan, myrpt, RPT_CONF)) {
+						ast_log(LOG_WARNING, "Unable to restore %s to conference after answer from %s\n", ast_channel_name(l->pchan), l->name);
+						ast_frfree(f);
+						if (remote_hangup_helper(myrpt, l)) {
+							continue;
+						}
+						break;
+					}
+
 					__kickshort(myrpt);
 					myrpt->rxlingertimer = RX_LINGER_TIME;
 					l->connected = 1;
@@ -4811,15 +4825,6 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 						doconpgm(myrpt, l->name);
 					} else
 						l->reconnects++;
-					/* Re-add parked pchan only after the peer has answered. */
-					if (l->pchan && rpt_conf_restore(l->pchan, myrpt, RPT_CONF)) {
-						ast_log(LOG_WARNING, "Unable to restore %s to conference after answer from %s\n", ast_channel_name(l->pchan), l->name);
-						ast_frfree(f);
-						if (remote_hangup_helper(myrpt, l)) {
-							continue;
-						}
-						break;
-					}
 				}
 				/* if RX key */
 				if ((f->subclass.integer == AST_CONTROL_RADIO_KEY) && (l->link_newkey != RADIO_KEY_NOT_ALLOWED)) {
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2569,6 +2569,17 @@ static void *attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
 		goto retry;
 	}
 
+	if (rpt_conf_restore(l->pchan, myrpt, RPT_CONF)) {
+		ast_log(LOG_WARNING, "Unable to restore %s to conference after reconnect to %s\n", ast_channel_name(l->pchan), l->name);
+		ast_hangup(l->chan);
+		rpt_mutex_lock(&myrpt->lock);
+		l->retrytimer = RETRY_TIMER_MS;
+		l->chan = NULL;
+		rpt_mutex_unlock(&myrpt->lock);
+		ast_autoservice_stop(l->pchan);
+		return NULL;
+	}
+
 	ast_autoservice_stop(l->pchan);
 	ast_log(LOG_NOTICE, "Reconnect Attempt to %s in progress\n", l->name);
 	return NULL;
@@ -4489,8 +4500,12 @@ static inline void safe_hangup(struct ast_channel *chan)
 }
 
 /*! \note myrpt->lock must not be held when calling */
-static inline void hangup_link_chan(struct rpt_link *l)
+static inline void hangup_link_chan(struct rpt *myrpt, struct rpt_link *l)
 {
+	/* Park pchan out of CONF so mix audio cannot fill Announcer/IAXLink while IAX is down. */
+	if (l->pchan) {
+		rpt_conf_remove(l->pchan, myrpt, RPT_CONF);
+	}
 	if (l->chan) {
 		safe_hangup(l->chan);
 		l->chan = NULL;
@@ -4520,16 +4535,16 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 					/* An allstar link node */
 					l->disctime = DISC_TIME;
 				}
-				hangup_link_chan(l);
+				hangup_link_chan(myrpt, l);
 				return 1;
 			}
 
 			if (l->retrytimer) {
-				hangup_link_chan(l);
+				hangup_link_chan(myrpt, l);
 				return 1;
 			}
 			if (l->outbound && (l->retries++ < l->max_retries) && l->hasconnected) {
-				hangup_link_chan(l);
+				hangup_link_chan(myrpt, l);
 				rpt_mutex_lock(&myrpt->lock);
 				l->retrytimer = RETRY_TIMER_MS;
 				l->elaptime = 0;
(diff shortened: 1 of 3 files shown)


apps/app_rpt/rpt_bridging.c
```diff
@@ -378,6 +381,59 @@ int __rpt_conf_create(struct rpt *myrpt, enum rpt_conf_type type, const char *fi
...
+static struct ast_bridge_channel *rpt_get_bridge_channel_from_chan(struct ast_channel *chan)
+{
+	struct ast_unreal_pvt *p;
+	struct ast_channel *pchan;
+	struct ast_bridge_channel *bridge_channel = NULL;
+
+	if (!chan) {
+		return NULL;
+	}
+
```
mkmer, line 434: Not sure why this is showing up as new code here - it's already in the base?!?

mkmer (commented):
This adds a level of complication we probably don't need.
When the `l->chan` is dead, we still read `l->pchan` and toss the frames 
https://github.com/AllStarLink/app_rpt/blob/4571bbc455d278f0bf443108d31f0f8b9b8d1d90/apps/app_rpt.c#L4830-L4832

The root problem was reconnections called a DNS lookup that hangs for a "long time".
I added autoservice routines to handle the backup:
https://github.com/AllStarLink/app_rpt/blob/4571bbc455d278f0bf443108d31f0f8b9b8d1d90/apps/app_rpt.c#L2503-L2519

Of course my first try set autoservice up a little too late in the `attempt reconnect()` 

Are we still seeing long queues after the last PR?

apps/app_rpt/rpt_bridging.c
hardenedpenguin: You’re right — that helper was already on master. GitHub was showing it as new because an earlier revision of this PR moved it above `__rpt_conf_add`.

Reordered so `rpt_get_bridge_channel_from_chan()` stays where it already lived; only the wait / remove / restore helpers are new below it.

apps/app_rpt.c
```diff
@@ -2569,6 +2569,17 @@ static void *attempt_reconnect(struct rpt *myrpt, struct rpt_link *l)
...
+	if (rpt_conf_restore(l->pchan, myrpt, RPT_CONF)) {
+		ast_log(LOG_WARNING, "Unable to restore %s to conference after reconnect to %s\n", ast_channel_name(l->pchan), l->name);
+		ast_hangup(l->chan);
+		rpt_mutex_lock(&myrpt->lock);
+		l->retrytimer = RETRY_TIMER_MS;
+		l->chan = NULL;
+		rpt_mutex_unlock(&myrpt->lock);
+		ast_autoservice_stop(l->pchan);
+		return NULL;
+	}
```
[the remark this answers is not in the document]
hardenedpenguin: Agreed — fixed. `attempt_reconnect()` no longer calls `rpt_conf_restore()`. Restoration now happens in `process_link_channel` on `AST_CONTROL_ANSWER`, with the prior restore-failure path preserved via `remote_hangup_helper()`.

hardenedpenguin:
@mkmer thanks for the review — addressing both points:

**“Already in the base” helper**  
Correct. That was a bad move in the earlier diff. The bridge-channel helper is back in its original place; only remove/wait/restore are new.

**Do we still need park after autoservice?**  
Yes — we’re still seeing the cascade on mustang after the autoservice reconnect work (ASL 3.10.1 / current master lineage). Reading `l->pchan` and discarding frames helps the *reader* side, but while `Announcer/IAXLink` remains a CONF member the mixer keeps *writing* into that channel. When reconnect / load stalls the reader even briefly, that queue hits 97 and backpressures every other conference member (`PChan`, `TXPChan`, `LocalTX`, other links). Mustang logs show thousands of those warnings with live `Announcer/IAXLink-*` still Up while IAX remotes are flapping.

Autoservice (#1147/#1164) fixed the blocking DNS hang during `attempt_reconnect()`. This PR covers the complementary case: keep mix audio off the dead link’s `pchan` until the peer answers.

**CodeRabbit restore timing**  
Also fixed: restore only on `AST_CONTROL_ANSWER`, not right after `rpt_make_call()`.

Happy to simplify further if you see a lighter way to keep CONF from writing into a parked link, but toss-on-read alone hasn’t been enough on the hubs that are still hitting this.

apps/app_rpt/rpt_bridging.h
```diff
@@ -90,6 +90,22 @@ int __rpt_conf_add(struct ast_channel *chan, struct rpt *myrpt, enum rpt_conf_ty
  */
 #define rpt_conf_add(chan, myrpt, type) __rpt_conf_add(chan, myrpt, type, __FILE__, __LINE__)
 
+int __rpt_conf_remove(struct ast_channel *chan, struct rpt *myrpt, enum rpt_conf_type type, const char *file, int line);
```
Allan-N: Looks like this header has the function prototypes (above) and then the #define's.  Can/should we move these prototypes up?

Nope.  The file is confused re: any convention (_local w/local, then _create+_add / create/add.  Maybe pick one style and fix what's off?

mkmer:
I guess I'd like to understand what is causing the backups when we should be reading and tossing the frames. While taking the channel out of the conference prevents frames from entering, why are we not removing the frames in the thread?  Besides the redial, what is hanging the thread preventing read and toss from keeping up as it should?

hardenedpenguin:
@mkmer good question — read/toss only runs while `process_link_channel` is in its `waitfor` / `ast_read(l->pchan)` loop. That **same** thread is also what blocks elsewhere, so the mixer can keep writing into `Announcer/IAXLink` while the reader is not draining.

Besides redial / DNS (which autoservice covers inside `attempt_reconnect()`):

1. **`remote_hangup_helper()`** still does `ast_safe_sleep(l->chan, MSWAIT * 10)` **before** hangup, with `pchan` still in `RPT_CONF` and autoservice **not** started. That alone is ~200 ms of mix writes with no toss on that thread.
2. After hangup, any further stall of that link thread (locks in `periodic_process_link`, other bookkeeping) has the same effect: softmix keeps feeding the member.
3. Once that member hits the high watermark, writers stall and you get the cascade (`PChan` / `TXPChan` / `LocalTX` / other links all reporting 97/97) — even if this one reader later catches up.

So toss-on-read is correct when the loop is running; it is reactive. Autoservice fixed the blocking DNS gap. Park stops the **write** side so brief non-autoserviced stalls cannot refill that member. Mustang on 3.10.1 (autoservice already present) still shows thousands of those warnings with live `Announcer/IAXLink-*` while IAX remotes flap.

If you have a lighter way to keep CONF from writing into a dead link’s `pchan` without remove/restore, I’m happy to try it — but tossing alone has not been enough on the hubs still hitting this.

apps/app_rpt/rpt_bridging.h
hardenedpenguin: Agreed — matched `create`/`add`: all four `__rpt_conf_*` prototypes first, then the public macros. Pushed in the latest commit.

mkmer:
I think we need to understand the blocks because in general that thread is expected to not block more than 19ms (ish).  
I'm not saying this pr is not the answer, but this block may be breaking many other things in the process that also expect to be handled (like key/unkey logic and other things in the periodic_process_link() function)
If we do address the hangs that we're not intended to be there, the issue will also go away (I think so anyway).
In general, any thread doing a waitfornandfd should be non blocking, so I'm thinking it's a bit of an architecture mistake on my part when these were separated in to independent threads (Or it's been there a long time and now understanding it a bit more?)

hardenedpenguin:
@mkmer after digging into the waitfor stalls, agree the right fix is to keep `process_link_channel` near the ~20 ms cadence — not park `pchan` out of CONF.

**Hang inventory (same thread as waitfor/read):**
1. `remote_hangup_helper()` — `ast_safe_sleep(l->chan, MSWAIT*10)` (~200 ms) with `pchan` still in CONF and no autoservice (uncovered softmix fill + skipped periodic key/unkey).
2. `attempt_reconnect()` from `periodic_process_link()` — blocking DNS (`node_lookup` SRV+A) + `rpt_make_call` for seconds (autoservice covers queues, not periodic work).

**Direction (hang-first, park deferred):**
- PR slice 1: remove that hangup sleep (textq already flushed) — branch `fix/link-hangup-no-waitfor-sleep` on the fork.
- PR slice 2: move reconnect DNS/dial off the waitfor thread (async worker + handoff).
- This PR (#1180 park/remove/restore): converting to **draft** and holding merge. If 1+2 clear mustang-class flaps, we close park instead of landing the extra bridge complexity.

Happy to adjust if you’d rather keep a short autoserviced sleep instead of removing it.

hardenedpenguin:
Hang-first slice 2 is up: https://github.com/AllStarLink/app_rpt/pull/1190 — outbound reconnect DNS/dial moved off the waitfor thread onto a detached worker so `process_link_channel` stays on `waitfor`/read.

With #1189 (hangup sleep) + #1190, park/remove (#1180) should no longer be needed once field-validated.

PR #1184, opened 2026-08-14
xpmr: preserve CTCSS through repeater hangtime

### LINKED ISSUE #1182: chan_usbradio CTCSS can remain muted after local COR drops during hang or linked audio
## Summary

The CTCSS turn-off logic added in commit bc4c4ed5 / PR #1113 can mute transmit CTCSS at the wrong time on a full-duplex repeater.

After upgrading from app_rpt 3.9.3 to 3.10.x, CTCSS encode intermittently stops while the transmitter remains keyed. It happens consistently when local COR drops and can also occur during hang time with linked audio. This appears to be a regression from the fix for #1024.

## Environment

- Debian 13.6 ASL3 appliance on Raspberry Pi 5/aarch64
- Kernel 6.18.39+rpt-rpi-2712
- asl3 3.18.2-2.deb13
- Asterisk/app_rpt 22.10.1+asl3-3.10.1-1.deb13
- chan_usbradio with CM119-family interface
- app_rpt duplex=2; usbradio duplex=1
- txmixa=voice; txmixb=tone
- Original txtoctype=phase

## Observed behavior

When local COR drops, CTCSS immediately phase-terminates/stops even though PTT remains asserted during hang time. If linked audio holds or begins holding the transmitter, voice can continue without CTCSS.

The behavior nearly works when CTCSS-on-input-only is enabled, but CTCSS does not reliably return when either:

1. A linked signal arrives during hang time.
2. The local receiver becomes active again during hang time.

When CTCSS-on-input-only is disabled, dropping CTCSS at local COR release is inappropriate. CTCSS should continue throughout hang time and reverse-burst immediately before the transmitter drops.

## Code finding

The new xpmr.c logic watches only local rxCarrierDetect while txPttIn remains active:

- Local carrier assertion sets txHadRxCarrier.
- Local carrier drop sets txCtcssHangMuted and starts notone/phase termination.
- Local carrier reassertion clears txCtcssHangMuted and restores the generator.
- The normal frequency-update path refuses to re-enable the generator while txCtcssHangMuted is set.

This does not distinguish why txPttIn remains asserted. app_rpt may be holding PTT for hang time, linked audio, telemetry, or another source. The restore path observes only rxCarrierDetect, so linked activity during hang cannot restore CTCSS.

Relevant condition:

```c
if (pChan->radioDuplex &&
    pChan->b.txHadRxCarrier &&
    !pChan->rxCarrierDetect &&
    !pChan->b.txCtcssHangMuted &&
    pChan->smode == SMODE_CTCSS &&
    !pChan->b.txCtcssInhibit &&
    pChan->b.ctcssTxEnable &&
    pChan->txTocType != TOC_NONE) {
        pChan->b.txCtcssHangMuted = 1;
        ...
}
```

The generator update is also gated by `!pChan->b.txCtcssHangMuted`.

### CTCSS-on-input-only disabled

- Encode CTCSS for the entire keyed period, including hang time.
- Apply phase/notone termination immediately before PTT drops.
- Local COR dropping must not terminate CTCSS while PTT is held for hang time, linked audio, telemetry, or another source.

### CTCSS-on-input-only enabled

- It is reasonable to terminate CTCSS when the qualifying input ends and nothing else requires encoded transmission.
- Restore CTCSS immediately if either the local receiver or a linked signal becomes active during hang time.
- Account for all active transmit sources, not only local rxCarrierDetect.

## Workaround

Changing:

```
txtoctype = phase
```

to:

```
txtoctype = no
```

prevents the new COS-drop block because txTocType is TOC_NONE. CTCSS remains present through hang time and stops with PTT. This restores reliable CTCSS but sacrifices reverse burst, so it is only a workaround.

## Suggested direction

Make early COR-based turn-off conditional on CTCSS-on-input-only mode and knowledge that no other source is holding or feeding TX. Restore CTCSS during hang for both local and linked activity. When input-only mode is disabled, preserve classic behavior and perform phase/notone termination at actual TX release.

PennyForward:
## Summary

Remove the XPMR state that terminates transmit CTCSS when local COR drops while PTT remains asserted.

With CTCSS-on-input-only disabled, CTCSS should remain encoded for the entire keyed period, including repeater hang time, and phase/notone termination should occur only at actual transmitter release.

Fixes #1182.

## Root cause

The COR-drop logic used only `rxCarrierDetect` to decide when to terminate CTCSS. It could not distinguish ordinary repeater hang time from linked audio, telemetry, or another source holding the transmitter. Once `txCtcssHangMuted` was set, linked activity could not restore CTCSS because the restore path also depended on local receiver carrier.

## Changes

- Remove the local-COR-driven CTCSS termination block.
- Remove `txHadRxCarrier` and `txCtcssHangMuted`.
- Restore the normal frequency update and transmitter-release paths.
- Preserve configured phase/notone termination immediately before PTT drops.

## Validation

Built against Asterisk/app_rpt 22.10.1+asl3-3.10.1 on aarch64 and installed on a full-duplex ASL3 repeater using chan_usbradio.

Validated with CTCSS-on-input-only disabled:

- CTCSS remains present after local COR drops and throughout hang time.
- Courtesy tone audio remains decodable during hang time.
- Linked and local receiver activity continue to transmit with CTCSS.
- Configured reverse burst occurs immediately before transmitter release.
- Multi-day on-air testing has not reproduced the prior intermittent CTCSS loss.

Tested by Chris Peterson, KG0BP.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CTCSS tone handling when releasing and re-keying the transmitter.
  * Ensured configured no-tone and phase-shift shutdown behavior is applied consistently.
  * Improved transmitter restart behavior during tone-off periods.
  * Prevented unnecessary interruption of frequency updates during CTCSS operation.

### DIFF
--- a/channels/xpmr/xpmr.c
+++ b/channels/xpmr/xpmr.c
@@ -2850,7 +2850,7 @@ i16 PmrRx(t_pmr_chan *pChan, i16 *input, i16 *outputrx, i16 *outputtx)
 			} else {
 				f = pChan->txctcssdefault_value;
 			}
-			if (f && pChan->spsSigGen0->freq != f * 10 && !pChan->b.txCtcssHangMuted) {
+			if (f && pChan->spsSigGen0->freq != f * 10) {
 				pChan->spsSigGen0->freq = f * 10;
 				pChan->spsSigGen0->option = 1;
 			}
@@ -2871,8 +2871,6 @@ i16 PmrRx(t_pmr_chan *pChan, i16 *input, i16 *outputrx, i16 *outputtx)
 				pChan->rxCtcssMap[pChan->rxCtcss->decode]);
 			pChan->dd.b.doitnow = 1;
 			pChan->spsSigGen0->freq = 0;
-			pChan->b.txHadRxCarrier = 0;
-			pChan->b.txCtcssHangMuted = 0;
 			if (pChan->smode == SMODE_CTCSS && !pChan->b.txCtcssInhibit) {
 				if (pChan->rxCtcss->decode > CTCSS_NULL) {
 					if (pChan->rxCtcssMap[pChan->rxCtcss->decode] != CTCSS_RXONLY) {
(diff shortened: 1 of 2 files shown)


hardenedpenguin:
Useful context for reviewers: this removes the #1113 duplex COS hang-mute path (`txHadRxCarrier` / `txCtcssHangMuted`). That logic ended CTCSS when local COR dropped while PTT was still held, which could not distinguish hangtime from linked/telemetry hold, and is what #1182 hit.

After this change, with CTCSS-on-input-only disabled, CTCSS stays encoded for the whole keyed period (including hangtime), and phase/notone runs at actual TX release.

Note that this also undoes the early-TOC-on-COS behavior from #1024/#1113. Operators who wanted reverse burst at the *start* of hang will get classic ride-through-hang again. Worth a short release-note callout; a narrower input-only early-mute path could be a later follow-up if anyone still needs it.

hardenedpenguin:
In our view, a full revert of the #1113 COS hang-mute path (#1184) fixes the #1182 regression for classic hang+CTCSS, but it also drops the early TOC behavior #1024 was after.

We opened an alternative that we think is the better long-term fix: https://github.com/AllStarLink/app_rpt/pull/1186

- Still removes local-COS hang mute (same root-cause fix as this PR)
- Classic (`itxctcss` off): CTCSS rides hangtime; phase/notone at real PTT drop
- Early TOC / restore is driven by app_rpt `itxctcss` → `TXCTCSS`, which already tracks remrx / localtx / autopatch / parrot — so linked hold during hang is handled correctly

If that approach looks right, #1184 could close in favor of #1186 (or land only if a minimal revert is preferred short-term).

Allan-N:
This PR looks to be a focused rollback of the XPMR portion of the #1113 fix.  That will revert the unwanted change in behavior but also takes away what was corrected.  We really want a fix that corrects #1024 but without any regressions ... and it's looking like #1186 will do this.

PennyForward:
I haven’t reviewed all of #1186 but I’m glad the issue is recognized and a fix is in the works. I’ll review it and see if I can poke any holes in it.

73 DE KG0BP

Chris Peterson, AFC®
Founder and CEO, Penny Forward
Blind People Building Bright Futures One Penny At A Time
Join Penny Forward<https://www.pennyforward.com/membership/> Schedule a Meeting<https://calendly.com/chris-peterson-pfwd/>
Email: ***@***.******@***.***>
Phone: 1 (888) 332-5558<tel:+18883325558>
Web site: www.pennyforward.com<http://www.pennyforward.com/>
[Penny Forward logo with an arrow pointing forward separating the two words. The word Penny and arrow are in black, with Forward in green.]

mkmer (commented):
Not 100% sure here, but I believe we want CTCSS to drop some amount of time before unkeying to deal with the squelch crash (right?)  Is this change breaking that behavior?

Allan-N:
The #1186 fix has been included in the 3.10.3 packages.  Please check if that resolves your issue.

PR #1189, opened 2026-08-15
app_rpt: autoservice pchan while draining hangup text

hardenedpenguin:
## Summary
- `remote_hangup_helper()` flushes the link text queue then waits ~200 ms so IAX can send those TEXT frames before hangup dumps the channel (Mike’s concern on this PR)
- `ast_safe_sleep` already autoservices `l->chan`; also autoservice `l->pchan` during that wait so softmix does not fill `Announcer/IAXLink` unread while the waitfor thread is blocked
- Complements #1190 (async reconnect) for the hang-first approach; park/remove of `pchan` (#1180) stays deferred

## Test plan
- [ ] Outbound permanent link: yank peer / kill network; confirm disconnect/reconnect and link TEXT (NEWKEY / L lists) still behave
- [ ] Confirm hangup no longer produces `Exceptionally long voice queue` on `Announcer/IAXLink` / `PChan` from this ~200 ms window
- [ ] If TEXT is still lost on hangup, next option is hang up `pchan` first then wait on the IAX chan (per Mike)

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4513,8 +4513,21 @@ static inline void hangup_link_chan(struct rpt_link *l)
 static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 {
 	if (l->chan) {
+		/*
+		 * Flush pending link text, then give IAX a short drain window.
+		 * ast_safe_sleep already autoservices l->chan; also autoservice
+		 * pchan so softmix does not fill Announcer/IAXLink unread while
+		 * this waitfor thread is blocked (key/unkey / periodic work still
+		 * stall for the sleep duration — reconnect DNS is handled elsewhere).
+		 */
 		link_process_textq(myrpt, l);
-		ast_safe_sleep(l->chan, MSWAIT * 10);  /* Allow the channel to send the text messages */
+		if (l->pchan) {
+			ast_autoservice_start(l->pchan);
+		}
+		ast_safe_sleep(l->chan, MSWAIT * 10); /* Allow the channel to send the text messages */
+		if (l->pchan) {
+			ast_autoservice_stop(l->pchan);
+		}
 	}
 
 	if (l->chan && !CHAN_TECH(l->chan, "echolink") && !CHAN_TECH(l->chan, "tlb")) {


mkmer:
I don't think we can be sure the text messages make it out the door ont he iax link channel, that's why we needed to wait here (if I remember the issues correctly) after we copy frames to the iax channel, issuing a hangup can dump (and did) the channel before the messages go out. Even 200ms was a bit sketchy, but I thought I found something similar in the core.  Maybe we just need auto service when waiting to hangup an iax link? Or maybe we just need to hangup the pchan before waiting on the iax chan to finish sending text frames?
I'm not in front of a computer for a few days, doing this from memory :)

hardenedpenguin:
Thanks Mike — that tracks. Dropping the sleep alone doesn’t guarantee the TEXT frames actually leave before hangup dumps the IAX channel.

We’ll try autoservice on that path so IAX can drain after `link_process_textq` without parking the waitfor thread (and starving `pchan` / softmix). We’ll test that locally and follow up here with what we see. If autoservice isn’t enough, hanging up `pchan` first then waiting on the IAX chan is the next thing we’ll look at.

Enjoy the time away from the keyboard.

hardenedpenguin:
Field testing with a build that includes both #1189 and #1190 (`3.10.2-92`) still shows the `Exceptionally long voice queue length (97 voice / 97 total)` warnings. On that run we did see a clean async reconnect to a peer (no retry storm), but the 97-frame bursts continued — including cases with no reconnect in the log and with an IAX jitterbuffer resync on the order of ~2.6 s.

So these changes do not resolve the warnings we were chasing. Closing #1189 and #1190 for now until we pin down the exact stall path. Happy to reopen (or carve out a narrower change) once we have a clearer root cause.

hardenedpenguin:
Closing as noted above; will revisit once the stall is identified.

mkmer:
I didn't see this PR, and essentially recreated it here: https://github.com/AllStarLink/app_rpt/pull/1194
I think we need to service `pchan` while waiting on `chan`.  
There are two or three long voice queue issues floating around -> 1 is the IAX pchan/chan pair which I believe these are addressing.  The other has to do with other channel pairs related to the main `rpt()` thread.

hardenedpenguin:
> I didn't see this PR, and essentially recreated it here: #1194 I think we need to service `pchan` while waiting on `chan`. There are two or three long voice queue issues floating around -> 1 is the IAX pchan/chan pair which I believe these are addressing. The other has to do with other channel pairs related to the main `rpt()` thread.

You are fine Mike, I want to continue my debugging to understand why the scheduler is causing the issues on a hub device. 

https://sm.w5gle.net has three hubs that run on it only so no radio attached of course. This is where I was seeing the issue with the scheduler. I will keep you posted as to my findings as I dig in deeper and confirm things.

PR #1190, opened 2026-08-15
app_rpt: run outbound link reconnect off the waitfor thread

hardenedpenguin:
## Summary
- Outbound reconnect (`node_lookup` DNS + `rpt_make_call`) no longer runs inline on `process_link_channel`
- A detached worker does the blocking work; the link waitfor thread keeps reading/tossing `l->pchan` (no autoservice needed on this path)
- `l->reconnecting` prevents duplicate workers and the max-retries teardown race; link exit waits for the worker and sets `killme` so dial aborts cleanly

Complements #1189 (hangup sleep removal). Together these are the hang-first alternative to parking `pchan` out of CONF (#1180, draft).

## Test plan
- [ ] Permanent outbound IAX link: yank peer / block network; confirm reconnect succeeds and audio returns
- [ ] During DNS/reconnect, no `Exceptionally long voice queue` on `Announcer/IAXLink` / `PChan` / `LocalTX` from that stall
- [ ] Key/unkey / link timers continue during reconnect (waitfor cadence intact)
- [ ] Tear down a link while reconnect is in flight (no crash / orphaned IAX channel)
- [ ] Non-IAX / echolink / tlb names still skip reconnect as before

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved outbound reconnection handling so link processing remains responsive during connection attempts.
  * Prevented duplicate reconnect attempts and improved cancellation, retry, and cleanup behavior.
  * Improved synchronization during reconnect activity and key-response transmission.
* **Reliability**
  * Ensured active reconnect operations are properly completed during link cleanup, repeater reloads, and module shutdown.
  * Improved deferred text-message queue handling during link operations.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -2065,9 +2065,11 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 			snprintf(tmp1, sizeof(tmp1), "K %s %s %d %d", src, myrpt->name, myrpt->keyed, n);
 			wf.data.ptr = tmp1;
 			wf.datalen = strlen(tmp1) + 1;
+			rpt_mutex_lock(&myrpt->lock);
 			if (mylink->chan) {
 				rpt_qwrite(mylink, &wf);
 			}
+			rpt_mutex_unlock(&myrpt->lock);
 			return;
 		}
 		if (myrpt->topkeystate != 1) {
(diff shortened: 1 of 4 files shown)


hardenedpenguin:
Field testing with a build that includes both #1189 and #1190 (`3.10.2-92`) still shows the `Exceptionally long voice queue length (97 voice / 97 total)` warnings. On that run we did see a clean async reconnect to a peer (no retry storm), but the 97-frame bursts continued — including cases with no reconnect in the log and with an IAX jitterbuffer resync on the order of ~2.6 s.

So these changes do not resolve the warnings we were chasing. Closing #1189 and #1190 for now until we pin down the exact stall path. Happy to reopen (or carve out a narrower change) once we have a clearer root cause.

hardenedpenguin:
Closing as noted above; will revisit once the stall is identified.

mkmer:
I had considered mutli threading the reconnect but decided against it since autoservice was the "easy" way (No need to noodle out all of the failure cases and how to feed that back into the main loop).
Once we get the actual problem noodled out, I think we should come back to this one and mutli thread the redial as it is really the right way, allowing the "frame processing" and timer processing to continue on the link thread.

PR #1198, opened 2026-08-18
app_rpt: Clear linklist on disconnect

### LINKED ISSUE #1197: Received node lists (`l->linklist`) never expire — stale entries persist indefinitely on permanently-linked multi-node systems
### Summary

On a system with **permanent** inter-node links, entries in `RPT_LINKS` / the "connected
nodes" list accumulate indefinitely and are never removed. Nodes that disconnected days
earlier are still advertised as connected, and nothing short of restarting Asterisk on
every node in the mesh clears them.

The received-list staleness timestamp appears to be unimplemented:

```
apps/app_rpt/app_rpt.h:603:    time_t linklistreceived;
```

`grep -rn linklistreceived .` over the whole tree returns **exactly that one line**
(checked at `d67f351`). The field is declared but never written and never read, so a
received node list is never aged out.

### Mechanism as I understand it

1. **Receive replaces wholesale** — `apps/app_rpt.c:1945`:
   ```c
   if (*str == 'L') { ...
       ast_str_set(&mylink->linklist, 0, "%s", str + 2);
   ```

2. **Advertise concatenates the neighbours' lists** — `apps/app_rpt/rpt_link.c:530-532`:
   ```c
   /* and append any nodes linked to the adjacent node */
   str = ast_str_buffer(l->linklist);
   ```

3. Each node re-posts that composite on `linkpost_time` (`app_rpt.c:3472-3479`).

So a stale entry on any node is folded into that node's advertisement, stored by its
neighbour, folded into *their* composite, and advertised back. The wholesale replace in
(1) would clear it, but the neighbour keeps re-sending it. With nothing aging the received
list, the entry is mutually sustained and never dies.

`RPT_ALINKS` is unaffected — `__mklinklist()` under `USE_FORMAT_RPT_ALINK` emits adjacent
nodes only and never reads `l->linklist`. In my testing ALINKS was accurate at all times
while `RPT_LINKS` was not.

### Why this is rarely seen

Ghost storage is per-link, and `rpt_link_destroy()` (`rpt_link.c:126-131`) frees
`l->linklist` on teardown. So:

- **Single-node systems** have no neighbour to learn a list from — unaffected entirely.
- **Normal transient links** (connect, ragchew, disconnect) free the storage on teardown,
  so any stale entries learned over that link die with it and the problem self-clears.
- **Permanently-linked multi-node systems** never destroy the link, so the received list
  is never freed and accumulates without bound.

The combination needed to notice this — permanent multi-node linking *and* actually
displaying `RPT_LINKS` to users — is probably uncommon, which may be why I could not find
an existing report.

### The configuration that triggers it

Three nodes:

```
      RF (band 1)                                 RF (band 2)
         |                                            |
    +---------+                                  +---------+
    |   <A>   |                                  |   <B>   |
    | ASL3    |                                  | ASL3    |
    +---------+                                  +---------+
          \                                          /
           \    permanent ilink 13 (transceive)      /
            \                                       /
             +--------------------------------------+
             |     <H>  —  radioless hub node        |
             |  app_rpt, rxchannel = Local/pseudo    |
             +--------------------------------------+
```

Two repeaters on different bands, permanently cross-linked through a radioless hub node so
that a user on either band is heard on the other, full time. `<H>` issues `ilink 13`
(permanent transceive) to both, and app_rpt re-establishes automatically after any drop.

**The permanence is the point, not an accident.** These links are meant to be up forever —
if they drop, the two bands stop hearing each other. At capture time both legs had been
continuously established for ~126 hours, which is the intended steady state.

External nodes and app clients attach to the **repeater** nodes, not the hub. So client
churn happens at `<A>`/`<B>` while the spine legs never move. Every such attach transits
the hub, which is why the hub accumulates fastest — roughly **one stale entry per day**.

`RPT_LINKS` is surfaced to end users here (a status display, and users' own AllStar phone
apps show the node's connected-node list), so stale entries are directly user-visible:
people open an app and see seven or eight stations listed when one is connected.

### Observed

ASL3 3.10.0 / Asterisk 22.10.1 (`22.10.1+asl3-3.10.0-1.deb13`), stock packages.

`rpt xnode <B>`:

```
Conn:  <H>   <ip>   0   IN   126:26:54   ESTABLISHED         <- the ONE real link
RPT_NUMLINKS=8
RPT_LINKS=8,T<H>,T<A>,T<x1>,T<x2>,T<x3>,T<x4>,T<x5>,TK6OZY
RPT_NUMALINKS=1
RPT_ALINKS=1,<H>TU                                           <- correct
```

Eight claimed links against one actual connection. `rpt lstats <B>` agrees with the `Conn`
line — only `<H>`. The extra entries are nodes that had disconnected; my own event log
recorded their `disc` events **27 hours earlier**.

Two smaller observations that may or may not be related:

- `rpt nodes <B>` printed 6 entries while `RPT_LINKS` reported 8 — two peers appeared in
  the variable but not in the node list.
- The hub listed the same callsign client **twice** (`..., TK6OZY, TK6OZY`) after that
  client had attached and disconnected on more than one occasion.

`rpt cmd <B> ilink 11 <x1>` is a no-op against these entries, which makes sense — there is
no link struct to tear down.

### Minimal reproduction

I believe this needs only two nodes and a permanent link, no radios:

1. Two app_rpt instances, `<H>` and `<A>`. `<H>` can be radioless
   (`rxchannel = Local/pseudo`).
2. `<H>` issues `ilink 13 <A>` — permanent transceive. Leave it up.
3. From a third node or an app client, connect to `<A>`, then disconnect cleanly.
4. `<A>` reaps its own link correctly. But check `<H>`:
   `rpt xnode <H>` → the third party is still listed in `RPT_LINKS` / `rpt nodes`, and
   stays there indefinitely while the `<H>`↔`<A>` link remains up.
5. Repeat step 3 with different peers — `<H>`'s list grows monotonically.
6. Restarting Asterisk on `<H>` alone does **not** fix it: `<A>` re-advertises the stale
   composite on reconnect and `<H>` relearns it immediately.

The distinguishing check throughout is `RPT_ALINKS` (and `rpt lstats`), which stay correct
while `RPT_LINKS` diverges.

### Suggested triage

If it helps narrow it down, the things I would instrument:

- **`ast_debug(7, ...)` at `app_rpt.c:1945`** is already there —
  *"@@@@ node %s received node list %s from node %s"*. Turning that on shows the composite
  arriving and, I think, would show the same stale peers being re-sent every
  `linkpost_time` indefinitely.
- **`__mklinklist()` with `USE_FORMAT_RPT_LINK`** vs `USE_FORMAT_RPT_ALINK` on the same
  node — the divergence between the two outputs is the bug's signature, and ALINKS makes
  a convenient ground truth to diff against.
- **`rpt lstats` vs `rpt nodes`** on the same node is the cheapest user-level check: the
  former reflects real link structs, the latter the propagated composite. On an affected
  system they disagree, and the difference is exactly the stale set.

I am happy to run any of this on the setup above and report back — it reproduces on demand
and I can mint a fresh stale entry in about two minutes.

### Workaround, and why it is only a stopgap

Dropping every link in the mesh **simultaneously** and re-establishing frees all
`linklist` storage and rebuilds cleanly — 16 entries across the three nodes went to 6, all
stale ones gone, in about 15 seconds with no Asterisk restart. Restarting one node in
isolation does **not** work, for the reason in step 6 above.

I have that automated on a nightly timer for now, but it is not a real fix:

- It **severs every link in the mesh**, briefly, on a schedule. That is tolerable while the
  only permanent links are the two crossband legs. It stops being tolerable as soon as
  other always-up interconnects terminate on the hub — bridging digital modes in via
  `chan_usrp` is on my roadmap, and a periodic "drop everything" is fundamentally
  incompatible with an interconnect that is supposed to never drop.
- It treats the symptom on a timer rather than preventing accumulation, so the visible
  state is only ever as good as the time since the last purge.

So for anyone building on a permanent multi-node topology, there is currently no
maintenance action that clears this without disrupting the very links the topology exists
to provide.

### Possible directions

I did not want to presume the right fix, since the tradeoffs are yours:

- Stamp `linklistreceived` on receive and ignore or drop a list older than some multiple of
  `linkpost_time` — the field looks like it was intended for exactly this.
- Or reconcile received entries against something authoritative on a periodic sweep.

One caution from my own case: whatever the staleness policy, it needs to tolerate a link
that has legitimately been up for weeks. An expiry keyed to link age rather than to
last-received-list age would break permanent links, which is the topology that has the
problem in the first place.

Happy to test any patch, and happy to open a PR if you would like one — just say which
direction you would prefer.

---

### Appendix — raw sample data

All three nodes polled every 5 s with `rpt xnode <node>`, read-only. 258 samples.
`A` = present in `RPT_ALINKS`, digit = count in the node list, `C` = has a `Conn:` line.

**A live client is stable in ALINKS — it does not drop out between transmissions.**
An AllStar phone client attached to `<B>`, left idle and unkeyed:

```
ts(UTC)     <B>      <B> RPT_ALINKS          tx
08:28:15    A2C      2,K6OZYTU,<H>TU         0
08:28:37    A2C      2,K6OZYTU,<H>TU         0
08:29:00    A2C      2,K6OZYTU,<H>TU         0
08:29:23    A2C      2,K6OZYTU,<H>TU         0    <- 68s idle, 13/13 samples present
08:29:51    A2CK     2,K6OZYTK,<H>TU         1    <- keys: suffix U -> K only
08:29:57    A2CK     2,K6OZYTK,<H>TU         1
08:30:08    A2C      2,K6OZYTU,<H>TU         0    <- unkeys: suffix back to U
08:31:38    A2C      2,K6OZYTU,<H>TU         0    <- second idle window, 15/15 present
08:32:01    .1.      1,<H>TU                 0    <- clean disconnect, reaped in 1 sample
```

28 consecutive samples across two idle windows with the client continuously present.
Membership never changed; only the keyed suffix did. A force-killed client (no disconnect
frame) was likewise reaped in a single sample once the IAX timeout expired, ~45 s later.

Note the `2` in the middle column throughout: the node list held **two** entries for that
one callsign — the live session plus a stale copy from a session 27 hours earlier.

**Meanwhile the hub never reflected any of it.** Across the whole six-minute session —
attach, key, disconnect, re-attach, force-kill — the hub's node-list count for that
callsign stayed pinned at `2`, and it never appeared in the hub's `RPT_ALINKS`, even while
that client's audio was actively transiting the hub.

**Accumulated state before the workaround** (`LinkedNodes` per node):

```
<A>   NUMLINKS=3   T<H>, T<B>, TK6OZY
<B>   NUMLINKS=6   T<H>, T<x1>, T<x2>, T<A>, T<x3>, TK6OZY
<H>   NUMLINKS=7   T<x1>, T<x2>, T<A>, T<B>, T<x3>, TK6OZY, TK6OZY
                                                   ^^^^^^^^^^^^^^^^ duplicated
```

Real connections at that moment: `<A>`→`<H>`, `<B>`→`<H>`, and the hub's two legs. Every
other entry was stale.

**After dropping both legs and rejoining** (~15 s later, no restart):

```
<A>   T<H>, T<B>          ALINKS 1,<H>TU              conns [<H>]
<B>   T<H>, T<A>          ALINKS 1,<H>TU              conns [<H>]
<H>   T<A>, T<B>          ALINKS 2,<B>TU,<A>TU        conns [<A>, <B>]
```

16 entries to 6. Every remaining entry is a real link or a genuine one-hop-away node, and
`rpt lstats <H>` showed both legs `ESTABLISHED` immediately afterwards.

During the split, all three nodes momentarily reported `RPT_NUMLINKS=0` with empty
`RPT_ALINKS` and no `Conn` lines — consistent with `rpt_link_destroy()` freeing each
`l->linklist` as its link went away.

73, Ozy **K6OZY**

mkmer:
Fixes #1197 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when a link is disconnected by clearing stale link information before reconnection or cleanup actions are evaluated.
  * Helps ensure link state remains accurate after remote hangups.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4528,7 +4528,10 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 	 * This is done to prevent any stale links from being shared while the node
 	 * is attempting to reconnect (and is not "really" connected)
 	 */
-	ast_str_reset(l->linklist);
+	if (ast_str_len(l->linklist) > 0) {
+		ast_str_reset(l->linklist);
+		rpt_update_links(myrpt);
+	}
 
 	if (l->chan && !CHAN_TECH(l->chan, "echolink") && !CHAN_TECH(l->chan, "tlb")) {
 		/* If neither echolink nor tlb */
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4520,7 +4520,7 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 		 */
 		ast_autoservice_start(l->pchan);
 		link_process_textq(myrpt, l);
-		ast_safe_sleep(l->chan, MSWAIT * 10);  /* Allow the channel to send the text messages */
+		ast_safe_sleep(l->chan, MSWAIT * 10); /* Allow the channel to send the text messages */
 		ast_autoservice_stop(l->pchan);
 	}
 
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4529,7 +4529,9 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 	 * is attempting to reconnect (and is not "really" connected)
 	 */
 	if (ast_str_len(l->linklist) > 0) {
+		rpt_mutex_lock(&myrpt->lock);
 		ast_str_reset(l->linklist);
+		rpt_mutex_unlock(&myrpt->lock);
 		rpt_update_links(myrpt);
 	}
 
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4528,7 +4528,7 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 	 * This is done to prevent any stale links from being shared while the node
 	 * is attempting to reconnect (and is not "really" connected)
 	 */
-	if (ast_str_len(l->linklist) > 0) {
+	if (ast_str_strlen(l->linklist) > 0) {
 		rpt_mutex_lock(&myrpt->lock);
 		ast_str_reset(l->linklist);
 		rpt_mutex_unlock(&myrpt->lock);
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -4524,6 +4524,12 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 		ast_autoservice_stop(l->pchan);
 	}
 
+	/* When the node is disconnected we need to clear the list of links.
+	 * This is done to prevent any stale links from being shared while the node
+	 * is attempting to reconnect (and is not "really" connected)
+	 */
+	ast_str_reset(l->linklist);
+
 	if (l->chan && !CHAN_TECH(l->chan, "echolink") && !CHAN_TECH(l->chan, "tlb")) {
 		/* If neither echolink nor tlb */
 		if (l->disced == RPT_LINK_DISCONNECT_NONE && !ast_shutting_down()) {


mkmer:
@coderabbitai review

mkmer:
@coderabbitai review

Allan-N (approved):

PR #1201, opened 2026-08-18
app_rpt: autoservice channels on startup and shutdown 

mkmer:
Address areas where channels are on conferences but not being processed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved repeater channel startup and shutdown reliability.
  - Ensured channels are consistently stopped and cleaned up when setup or initialization encounters an error.
  - Improved recovery during repeater shutdown to prevent channels from remaining active unexpectedly.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3002,6 +3002,31 @@ static void rpt_stop_channel_autoservice(struct rpt *myrpt)
 	}
 }
 
+static void rpt_start_channel_autoservice(struct rpt *myrpt)
+{
+	if (myrpt->rxchannel) {
+		ast_autoservice_start(myrpt->rxchannel);
+	}
+	if (myrpt->txchannel && myrpt->txchannel != myrpt->rxchannel) {
+		ast_autoservice_start(myrpt->txchannel);
+	}
+	if (myrpt->pchannel) {
+		ast_autoservice_start(myrpt->pchannel);
+	}
+	if (myrpt->localtxchannel && myrpt->localtxchannel != myrpt->txchannel) {
+		ast_autoservice_start(myrpt->localtxchannel);
+	}
+	if (myrpt->monchannel) {
+		ast_autoservice_start(myrpt->monchannel);
+	}
+	if (myrpt->rxpchannel) {
+		ast_autoservice_start(myrpt->rxpchannel);
+	}
+	if (myrpt->txpchannel) {
+		ast_autoservice_start(myrpt->txpchannel);
+	}
+}
+
 static int rpt_setup_channels(struct rpt *myrpt, struct ast_format_cap *cap)
 {
 	int res = 0;
@@ -5915,6 +5940,7 @@ static void *rpt(void *this)
 	ast_debug(1, "%s disconnected, cleaning up...\n", myrpt->name);
 
 	myrpt->ready = 0;
+	rpt_start_channel_autoservice(myrpt);
 	usleep(100000);
 	while (myrpt->tele.next != &myrpt->tele) {
 		/* wait for telem to be done */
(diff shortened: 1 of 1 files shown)


mkmer:
@coderabbitai review

apps/app_rpt.c
```diff
@@ -2974,6 +2977,56 @@ static void _load_rpt_vars_by_rpt(struct rpt *myrpt, int force)
 #define IS_DAHDI_CHAN(c) (CHAN_TECH(c, "DAHDI"))
 #define IS_DAHDI_CHAN_NAME(s) (!strncasecmp(s, "DAHDI", 5))
 
+static void rpt_stop_channel_autoservice(struct rpt *myrpt)
```
Allan-N: `rpt_stop_channel_autoservice()` before `rpt_start_channel_autoservice()` ... OK

Allan-N: As alternate function names, maybe `rpt_autoservice_stop()` and `_start()` ?

apps/app_rpt.c
```diff
@@ -3002,20 +3055,25 @@ static int rpt_setup_channels(struct rpt *myrpt, struct ast_format_cap *cap)
...
+		}
 	} else {
 		myrpt->txchannel = myrpt->rxchannel;
 		/* If it is a DAHDI hardware channel (Not PSEUDO), use the configured txchannel. */
 		myrpt->localtxchannel = IS_DAHDI_CHAN_NAME(myrpt->rxchanname) && !IS_PSEUDO_NAME(myrpt->rxchanname) ? myrpt->txchannel : NULL;
 	}
 
 	if (rpt_request_local(myrpt, cap, RPT_PCHAN, "PChan")) {
+		rpt_stop_channel_autoservice(myrpt);
 		rpt_hangup_rx_tx(myrpt);
```
Allan-N, line 3069: Is it OK to stop autoservice on a channel multiple times?

apps/app_rpt.c
```diff
@@ -2974,6 +2977,56 @@ static void _load_rpt_vars_by_rpt(struct rpt *myrpt, int force)
 #define IS_DAHDI_CHAN(c) (CHAN_TECH(c, "DAHDI"))
 #define IS_DAHDI_CHAN_NAME(s) (!strncasecmp(s, "DAHDI", 5))
 
+static void rpt_stop_channel_autoservice(struct rpt *myrpt)
+{
+	if (myrpt->rxchannel) {
+		ast_autoservice_stop(myrpt->rxchannel);
```
Allan-N: Is it OK to stop autoservice on a channel that has not yet been started?

apps/app_rpt.c
mkmer: Yes, best I can tell, it looks up the channel in an internal list and only acts if it finds it.

apps/app_rpt.c
mkmer: Started?  IF there's a value in the variable, it's created.  Ideally we want to auto service BEFORE getting on the conference until we make it to the frame processing loop.

apps/app_rpt.c
Allan-N: OK.  Just wanted to be sure that it was OK to stop servicing (all) and then stop servicing RX (and TX) again.

apps/app_rpt.c
Allan-N: The question was here, and applied to all of the calls in this function to ast_autoservice_stop(), because I noticed many places where you are calling this function where the channel would exist but autoservice had not yet been started.  As long as we're OK stopping what hasn't been started then we should be good.

apps/app_rpt.c
mkmer: I suppose there's no reason to do that, even if it's ok.

apps/app_rpt.c
mkmer: As we create channels it starts autoservice individually, if there's a failure, we call the stop logic which only stops the populated pointers which should be started at that point (right?)
Maybe I've missed something?

Allan-N (approved):

PR #1210, opened 2026-08-24
app_rpt: treat an empty "L " linklist update as clearing the stored list

### LINKED ISSUE #1209: Stale RPT_LINKS entries never clear: empty "L " linklist update is discarded in handle_link_data() (verified one-line fix)
Follow-up to #1197, which was closed via #1198 (first released in 3.10.4). I re-tested on 3.10.4
and the behaviour I reported there is unchanged, so I dug further, found what I believe is the
actual root cause, and verified a one-line fix in both directions. Filing it as its own issue
because the mechanism is different from the `linklistreceived` aging I guessed at in #1197, and
because I cannot reopen that issue myself.

I think #1198 addresses a different case: it clears a link's stored list when *that link*
disconnects, but the stale entries here sit on a link that never disconnects.

### Setup

Four radioless nodes (`rxchannel = Local/pseudo`), private node numbers, no RF:

```
T1 --transient--> A <--permanent-- H
T2 --transient--> A
```

`A` is adjacent to the transient nodes. `H` is not. Same topology and same commands on both
versions. I watched the composite linklist that `rpt xnode` prints on `H`, not `RPT_LINKS`, since
`RPT_LINKS` is adjacent-only on a leaf node and looks clean even while a stale entry is held.

### Result on stock packages

| version | A (adjacent) after T1 disconnects | H (non-adjacent) after T1 disconnects |
|---|---|---|
| 3.10.0 | reaped within 10s | still lists T1 at t+240s |
| 3.10.4 | reaped within 10s | still lists T1 at t+240s |

### The part that localises it

With **T2 still connected**, disconnecting T1 behaves correctly:

```
t+10s  A: [H, T2]   H: [A, T1, T2]
t+40s  A: [H, T2]   H: [A, T2]      <- stale entry cleared, one linkpost cycle
```

So a received list is refreshed correctly whenever the sender still has at least one other link.
The entry only becomes permanent when the sender's list goes **empty**. That points at the receive
guard:

```c
/* apps/app_rpt.c:1945 on 3.10.4 */
if (*str == 'L') {
        if (strlen(str) < 3) {
                return;
        }
        rpt_mutex_lock(&myrpt->lock);
        ast_str_set(&mylink->linklist, 0, "%s", str + 2);
```

The sender is unconditional. It always builds `"L "` and writes it (`app_rpt.c:3550-3562`), with no
empty-list check. So when a peer's list is empty the payload is exactly the two character string
`"L "`, which fails `strlen(str) < 3`, and the receiver returns without touching
`mylink->linklist`. The last non-empty list it ever saw is kept indefinitely.

In other words "there is no update to apply" and "that peer now has no links" arrive as the same
thing on the wire, and the code takes the first reading.

This also explains why `RPT_ALINKS` was accurate throughout while `RPT_LINKS` was not:
`__mklinklist()` under `USE_FORMAT_RPT_ALINK` never reads `l->linklist`.

### The fix, built and verified

```diff
-               if (strlen(str) < 3) {
+               if (strlen(str) < 2) {
                        return;
                }
```

At length 2, `str[0]` is `'L'`, `str[1]` is `' '`, and `str + 2` points at the NUL terminator, so
`ast_str_set()` stores an empty list, which I believe is the intended meaning. A length 1 message
(`"L"`) is still rejected, so there is no overrun.

To check this is actually the mechanism and not a coincidence, I built `app_rpt.so` at tag 3.10.4
twice from the same overlaid asterisk 22.10.1 tree, once stock and once with only this change, and
swapped them into the same four containers running the packaged
`2:22.10.1+asl3-3.10.4-1.deb13` asterisk. Module identity was confirmed each run with
`module show like app_rpt` and a checksum of the file inside the container.

| module | H (non-adjacent) after T1 disconnects |
|---|---|
| patched (`< 2`) | stale entry dropped within 20s, still clean at t+240s |
| stock (`< 3`) | still lists T1 at t+240s |

Swapping the stock module back into the same containers brings the ghost back, so it is the guard
doing the clearing and not the asterisk restarts around the swap.

Is there a case where a length 2 `"L "` payload should genuinely be ignored rather than read as
"this peer now has no other links"? If not, happy to open a PR for this, and happy to run any
further test on the lab either way.

73, Ozy **K6OZY**

Ozy311:
Fixes #1209.

Two commits, separable if you want only one:

1. **`strlen(str) < 3` becomes `< 2` in `handle_link_data()`.** The sender always
   transmits `"L "` plus the list, so a peer with no other links sends exactly the two
   character payload `"L "`. The old guard discarded it, so the last non-empty list a
   permanent link ever sent was kept forever, and non-adjacent nodes showed phantom
   connections in `RPT_LINKS` and `rpt xnode` indefinitely. At length 2, `str + 2`
   points at the NUL terminator and `ast_str_set()` stores the empty list, which is
   what the sender meant. A length 1 payload is still rejected.

2. **Remove the unused `linklistreceived` member** from `struct rpt_link`, as
   requested in the issue discussion. It is declared and never read or written
   anywhere in the tree.

Testing: master is currently the same tree as tag 3.10.4, which is where the A/B in
#1209 ran. Built stock and patched `app_rpt.so` from one tree, swapped them through
the same four radioless lab nodes with module identity checked each arm. Patched: a
non-adjacent node drops the stale entry within ~20s of the transient node detaching,
still clean at t+240s. Stock: the entry is still there at t+240s, and swapping stock
back in brings the ghost back. The member removal in commit 2 was compile-verified
against the same asterisk 22.10.1 overlay build.

Happy to adjust either commit or run any further lab test you want.

73, Ozy **K6OZY**

## Summary by CodeRabbit

* **Bug Fixes**
  * Link-list messages with the minimum valid length are now accepted correctly.
  * Removed obsolete link-list timestamp data from the application’s link information.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -1942,7 +1942,7 @@ static void handle_link_data(struct rpt *myrpt, struct rpt_link *mylink, char *s
 		return;
 	}
 	if (*str == 'L') {
-		if (strlen(str) < 3) {
+		if (strlen(str) < 2) {
 			return;
 		}
 		rpt_mutex_lock(&myrpt->lock);
--- a/apps/app_rpt/app_rpt.h
+++ b/apps/app_rpt/app_rpt.h
@@ -600,7 +600,6 @@ struct rpt_link {
 	struct ast_channel *pchan;
 	struct ast_audiohook altaudio;
 	struct ast_str *linklist;
-	time_t linklistreceived;
 	int linklisttimer;
 	int linkunkeytocttimer;
 	struct timeval lastlinktv;


mkmer (approved):

mkmer:
The "outside of diff" comment makes some sense; except we just skip the next character no matter what.  If it is "Lx" we still just point at the `NULL`.

Allan-N (approved):

PR #1196, opened 2026-08-17
app_rpt: Don't put `pchan` on the conference until the calling channel is up

mkmer:
500ms delay on the incoming channel can cause `pchan` to have a long voice queue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved call handling by adding the link’s pseudo-channel after the incoming channel is answered.
  * Ensured phone-mode state and answer handling occur in the correct sequence for more reliable connections.

PR #1212, opened 2026-08-25
app_rpt: Don't put outbound link pchan on conference until channel is up

hardenedpenguin:
## Summary
- Extend the #1196 inbound link fix to outbound `rpt_link_connect()` so hub-initiated IAX links do not add `pchan` to the TX conference until the link channel is `AST_STATE_UP`.
- Mirror inbound behavior: answer if needed, optional 500 ms settle for non-numeric node names, then fail with CONNFAIL if the channel still is not up.
- Addresses the asymmetric link-setup path suspected in [forum reports of intermittent congestion/reorder tones](https://community.allstarlink.org/t/unwanted-reorder-or-fast-busy-tone/24944) when a voter hub links out to a SimpleUSB node (`ilink` / `linktolink`).

## Test plan
- [ ] Build/install package with this branch on a test node
- [ ] Hub-initiated link: `rpt cmd <hub> ilink 3 <remote>` — confirm link comes up cleanly and audio is normal
- [ ] Repeat several connect/disconnect cycles; verify no stuck 480/620 Hz congestion tone on the linked pair
- [ ] Confirm inbound links (remote initiates) still work as before (#1196 path unchanged)
- [ ] Echolink/non-numeric outbound link names: verify 500 ms settle path does not regress connect

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link connection reliability by verifying the channel is active before continuing.
  * Added recovery handling for channels that are not immediately ready.
  * Connections that remain unavailable now fail cleanly with appropriate reporting and teardown.

### DIFF
--- a/apps/app_rpt/rpt_link.c
+++ b/apps/app_rpt/rpt_link.c
@@ -718,7 +718,7 @@ static int rpt_wait_for_link_up(struct ast_channel *chan, int timeout_ms)
 		if (ast_channel_state(chan) == AST_STATE_UP) {
 			return 0;
 		}
-		if (ast_waitfor(chan, poll_ms) == chan) {
+		if (ast_waitfor(chan, poll_ms) > 0) {
 			f = ast_read(chan);
 			if (!f) {
 				return -1;
--- a/apps/app_rpt/rpt_link.c
+++ b/apps/app_rpt/rpt_link.c
@@ -701,7 +701,8 @@ void rpt_update_links(struct rpt *myrpt)
  * \brief Wait for an outbound link channel to reach AST_STATE_UP.
  *
  * Outbound Echolink and TLB calls return from ast_call while still ringing;
- * the remote answer arrives asynchronously via channel read processing.
+ * the remote answer arrives asynchronously via AST_CONTROL_ANSWER.  That
+ * frame is re-queued so process_link_channel() can perform link setup.
  *
  * \return 0 if the channel is up, -1 on hangup or timeout.
  */
@@ -710,6 +711,11 @@ static int rpt_wait_for_link_up(struct ast_channel *chan, int timeout_ms)
 	int elapsed = 0;
 	const int poll_ms = 50;
 	struct ast_frame *f;
+	struct ast_frame answer = {
+		.frametype = AST_FRAME_CONTROL,
+		.subclass.integer = AST_CONTROL_ANSWER,
+		.src = __PRETTY_FUNCTION__,
+	};
 
 	while (elapsed < timeout_ms) {
 		if (ast_check_hangup(chan)) {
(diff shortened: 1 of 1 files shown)


apps/app_rpt/rpt_link.c
```diff
@@ -697,6 +697,58 @@ void rpt_update_links(struct rpt *myrpt)
...
+			}
+			if (f->frametype == AST_FRAME_CONTROL && f->subclass.integer == AST_CONTROL_ANSWER) {
+				ast_frfree(f);
+				ast_queue_frame(chan, &answer);
+				return 0;
+			}
+			if (f->frametype == AST_FRAME_CONTROL && f->subclass.integer == AST_CONTROL_HANGUP) {
+				ast_frfree(f);
+				return -1;
+			}
```
mkmer, line 740: I think this could be "simplified" with an outer `if` for the frame type, then an inner if on the subclass using
`if (f->subclass.integer == AST_CONTROL_HANGUP) {...}
else if (f->subclass.integer == AST_CONTROL_ANSWER) {...}`
type of thing...

mkmer (commented):
I don't think we were doing this "ever", wonder why it is needed now (aka how did I break it?)... just sayin :)

hardenedpenguin:
@mkmer You're right — outbound never waited for UP before adding `pchan` to the conference. #1196 fixed an inbound race (`ast_answer` / 500 ms settle with `pchan` already on the mix). That path doesn't exist on outbound `rpt_link_connect()`, so this wait/re-queue helper isn't restoring something you broke; it was new machinery from mirroring #1196 / Greptile notes on Echolink/TLB.

For typical IAX outbound (`rpt_make_call` with a 2 s timeout), the channel is usually already UP, so the wait is a no-op and likely doesn't help the congestion-tone report that prompted this.

Cleaner approach: drop the wait helper and defer `rpt_conf_add(l->pchan)` until `process_link_channel` handles `AST_CONTROL_ANSWER`. Happy to rework the PR that way unless you'd rather park it.

mkmer:
We put `rpt_link_connect()` in it's own thread as part of the move to mutli-threading, but re-reviewing 3.6.0 code base, I don't see anything "different".  I'm more curious why calls are failing with fast busy with our current architecture that (maybe was not doing this) before.
 This PR looks like a good fix, but still not sure why behavior changed - would be good to understand why it is triggered. Maybe a bug with the move to threading `rpt_link_connect()` or a race, or missing lock?

apps/app_rpt/rpt_link.c
```diff
@@ -697,6 +697,58 @@ void rpt_update_links(struct rpt *myrpt)
 	ast_free(obuf);
 }
 
+/*!
+ * \brief Wait for an outbound link channel to reach AST_STATE_UP.
+ *
+ * Outbound Echolink and TLB calls return from ast_call while still ringing;
+ * the remote answer arrives asynchronously via AST_CONTROL_ANSWER.  That
+ * frame is re-queued so process_link_channel() can perform link setup.
```
Allan-N, line 705: This doesn't make sense.  Why would we receive an AST_CONTROL_ANSWER frame and then re-queue it back to ourselves?  When would this re-queuing stop?

PR #1214, opened 2026-08-28
chan_usbradio: drain TX backlog during unkeyed silence

Allan-N:
Asterisk can deliver bursts of audio frames, typically when a new connection is established, that increase the TX queue depth. While unkeyed, Asterisk also delivers frames of silence. Stop queueing these silent frames so the backlog can drain while unkeyed.

Track TX queue depth and log oversized queue episodes and their peak depth at debug level.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USB radio transmission queue handling during idle periods.
  * Prevented silent audio frames from unnecessarily delaying queued transmissions.
  * Ensured queued audio continues feeding output blocks consistently when ready.
  * Added queue-depth monitoring and logging to help identify transmission delays.

### DIFF
--- a/channels/chan_usbradio.c
+++ b/channels/chan_usbradio.c
@@ -2342,13 +2342,8 @@ static void *usbradio_audio_thread(void *arg)
 			}
 			ast_mutex_unlock(&o->txqlock);
 
-			/* Intentional keyed TX queue cushion */
-			if (num_frames <= 3 && (o->txkeyed || o->txtestkey)) {
-				last_frame_time = ast_radio_tvnow();
-			}
-
 			/* One queued frame per available output block */
-			if (tx_write_ready && num_frames && (num_frames > 3 || (!o->txkeyed && !o->txtestkey))) {
+			if (tx_write_ready) {
 				if (usbradio_feed_tx_queue(o, 1)) {
 					o->didpmrtx = 1;
 					last_frame_time = ast_radio_tvnow();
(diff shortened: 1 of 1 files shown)


channels/chan_usbradio.c
```diff
@@ -2070,13 +2090,24 @@ static int usbradio_write(struct ast_channel *c, struct ast_frame *f)
 		return 0;
 	}
 
+	/* PortAudio and XPMR are fed silence separately; do not let idle zero
+	 * frames prevent an accumulated TX queue from draining. */
+	if (!o->txkeyed && !o->txtestkey && usbradio_frame_is_silence(f)) {
```
mkmer, line 2098: Silence frames are valid frames.  We may still want to queue them.
That being said, we do add them back in on the other side of the `txq` if we don't have frames to send, but it's possible we pull a couple of silence frames out of the middle of audio frames causing some odd distortion?

IF we keep this + use the fill on the other side, we should add some docs around it so the next guy gets a head start :)

mkmer (approved):

channels/chan_usbradio.c
Allan-N: This comment has been updated.

But, to explain here, I agree that silence frames are valid but we don't need to queue/process each frame WHEN NOT KEYED (and we're only checking for silence frames ... and not queuing them ... when not keyed).  And, with the frames not being queued we don't need to dequeue / process the silence ... and if no frames are available we then add the silence back.

PR #1217, opened 2026-08-29
app_rpt: Stop permanent links from reconnecting after ilink 11

### LINKED ISSUE #1216: Perm Disconnect ilink 11 not working
We are currently running Allstar (Asterisk) 22.10.1+asl3-3.10.4-1.  We are having an issue on inbound perm connections to the server.  Using Allmon3 to disconnect the node, we can see the command from Allmon3: rpt cmd 28528 ilink 11 XXXXX to disconnect.  However the disconnected nodes is able to reconnect immediately.   Before we were able to issue the  ilink 11 command and disconnect the node.  After reading the man pages on Permanent Node Connection.  Should this stop the connection on the far end from trying to reconnect, until they issues another reconnect permanent to the system?

hardenedpenguin:
## Summary
- Fixes permanent outbound links immediately reconnecting after `ilink 11` / `!!DISCONNECT!!` (#1216): send DISCSTR before marking `disced`, demote `MAX_RETRIES_PERM`, and restore softhangup on receive.
- Count reconnect attempts only while the channel is down (avoids burning `MAX_RETRIES` while connected after #574).
- ilink 6 snapshots AO2 link/channel refs so disconnect I/O and silent stop-retries apply to the selected set, not a later re-traverse.

## Test plan
- [ ] Node A establishes permanent connect (`ilink 13`) to node B
- [ ] On B: `rpt cmd <B> ilink 11 <A>` (or Allmon3 Perm Disconnect)
- [ ] Confirm `!!DISCONNECT!!` is sent, A tears down, and A does **not** redial
- [ ] Unexpected network drop of a permanent link still reconnects
- [ ] `ilink 1` still refuses to drop a permanent link; `ilink 11` succeeds
- [ ] `ilink 6` disconnects all links including permalinks without immediate reconnect

Fixes #1216

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link disconnect and reconnect handling.
  * Prevented retry limits from being exhausted prematurely while a link is active.
  * Ensured disconnect notifications are delivered before channels close or reconnect attempts begin.
  * Improved “disconnect all links” behavior and handling of queued messages during shutdown.
  * Improved cleanup when nodes exit, including audio and conference resources.
  * Reduced unwanted redial attempts after intentional link disconnections.


apps/app_rpt/rpt_link.h
```diff
@@ -29,6 +29,21 @@ int altlink1(struct rpt *myrpt, struct rpt_link *mylink);
 
 void rpt_qwrite(struct rpt_link *l, struct ast_frame *f);
 
+/*!
+ * \brief Stop reconnect retries after an intentional disconnect.
+ *
```
Allan-N, line 36: add \param

apps/app_rpt/rpt_functions.c
```diff
@@ -182,21 +182,33 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
...
+		 */
+		{
+			struct ast_channel *chan = NULL;
+
+			if (l->chan && l->thisconnected) {
+				chan = ast_channel_ref(l->chan);
+			}
+			rpt_mutex_unlock(&myrpt->lock);
+			if (chan) {
 				struct ast_frame wf = {
```
Allan-N: Should we now be using `rpt_link_send_disconnect()` ?

apps/app_rpt/rpt_functions.c
```diff
@@ -182,21 +182,33 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
...
-		l->disced = RPT_LINK_DISCONNECT;
-		l->hasconnected = 1;
-		if (l->chan) {
-			if (l->thisconnected) {
+		/*
+		 * Send !!DISCONNECT!! before setting disced. If disced is set first,
+		 * process_link_channel can exit without flushing textq and the peer only
+		 * sees hangup — permanent links then immediately reconnect (#1216).
+		 */
+		{
```
Allan-N: Can we remove the `{}` block?

apps/app_rpt/rpt_functions.c
```diff
@@ -182,21 +182,33 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
...
 				struct ast_frame wf = {
 					.frametype = AST_FRAME_TEXT,
 					.src = __PRETTY_FUNCTION__,
 					.datalen = sizeof(DISCSTR),
 					.data.ptr = DISCSTR,
 				};
 
-				rpt_qwrite(l, &wf);
+				ast_write(chan, &wf);
+				ast_safe_sleep(chan, MSWAIT * 10);
```
Allan-N: I'm not wild about adding sleep's

apps/app_rpt/rpt_functions.c
```diff
@@ -311,15 +323,36 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
...
 		rpt_mutex_lock(&myrpt->lock);
-		myrpt->savednodes[0] = 0;
 
 		if (!myrpt->links) {
 			rpt_mutex_unlock(&myrpt->lock);
 			return DC_COMPLETE;
 		}
 
-		/* loop through all links */
+		maxchans = ao2_container_count(myrpt->links);
```
Allan-N: If you are trying to grab a copy of `myrpt->links` then I would suggest you look at `ao2_container_clone()`.

apps/app_rpt.c
```diff
@@ -4607,6 +4618,28 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
 
 	if (l->chan && !CHAN_TECH(l->chan, "echolink") && !CHAN_TECH(l->chan, "tlb")) {
 		/* If neither echolink nor tlb */
+		if (l->disced == RPT_LINK_DISCONNECT_NONE && !ast_shutting_down()) {
+			struct ast_frame *f;
+
+			/* !!DISCONNECT!! may still be queued ahead of hangup; honor it before redialing. */
```
Allan-N: I'm not sure about this block.  We send a `!!DISCONNECT!!`` to the other node.  Are we waiting for that node to hangup and then we process the remote hangup?  or are we sending the message and hanging up ourselves?  or ...?

apps/app_rpt.c
```diff
@@ -4607,6 +4618,28 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
...
+			while (ast_waitfor(l->chan, 0) > 0) {
```
Allan-N: Should we also check for `!ast_shutting_down()`?

apps/app_rpt.c
```diff
@@ -4607,6 +4618,28 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
...
+					break;
+				}
+				if (f->frametype == AST_FRAME_CONTROL && f->subclass.integer == AST_CONTROL_HANGUP) {
+					ast_frfree(f);
+					break;
+				}
+				ast_frfree(f);
+			}
+		}
 		if (l->disced == RPT_LINK_DISCONNECT_NONE && !ast_shutting_down()) {
```
Allan-N: This `if ()` looks to be testing the same conditions as above.  Do we need 2 `if ()`s?

apps/app_rpt.c
```diff
@@ -4983,6 +5016,13 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 		continue;
 	}
 	/* Link is done: Cleanup channels and link structure */
+	/* Flush !!DISCONNECT!! (or other textq) before hangup so the peer sees it. */
```
Allan-N: Why wouldn't we have been processing (and flushing) the textq?  Are we breaking out of our processing loop too soon?

Also, I'm not wild about adding sleep()s.

apps/app_rpt/rpt_link.h
hardenedpenguin: Done — added `\param l` on `rpt_link_stop_retries()`.

apps/app_rpt/rpt_functions.c
hardenedpenguin: Yes — ilink 1/11 now calls `rpt_link_send_disconnect()` instead of hand-rolling the frame write.

apps/app_rpt/rpt_functions.c
hardenedpenguin: Removed. `chan` is declared at function scope now.

apps/app_rpt/rpt_functions.c
hardenedpenguin: Agreed — dropped the `ast_safe_sleep`. Sending `!!DISCONNECT!!` before setting `disced` is the real fix; the sleep was only a hedge.

apps/app_rpt/rpt_functions.c
hardenedpenguin: Switched ilink 6 to `ao2_container_clone(myrpt->links, OBJ_NOLOCK)` (same pattern as CLI/manager), then collect channel refs from the clone for unlocked I/O.

apps/app_rpt.c
hardenedpenguin: This is the **receiver** path, not the sender.

When the peer hangs up after sending `!!DISCONNECT!!`, that text can still be queued ahead of HANGUP. We drain pending frames here so an intentional disconnect is honored (`rpt_link_stop_retries`) **before** we decide to redial a permanent outbound link.

We are not waiting for the far end after we ourselves sent DISCSTR — that side writes DISCSTR, then sets `disced`, then tears down.

apps/app_rpt.c
hardenedpenguin: Yes — the drain loop now also checks `!ast_shutting_down()`, and the whole reconnect block is gated on that once at the top.

apps/app_rpt.c
hardenedpenguin: Collapsed into one outer `if`. After the drain we only re-check `disced` (it may have been set by `!!DISCONNECT!!`) before redialing.

apps/app_rpt.c
hardenedpenguin: Yes — that was the original bug: setting `disced` first made `process_link_channel` exit the wait loop before `link_process_textq` ran, so hangup beat `!!DISCONNECT!!` on the wire.

Fix is send DISCSTR **before** `disced`. Removed the cleanup sleep; we still do a no-wait `link_process_textq` flush for anything else that might be queued.

apps/app_rpt/rpt_functions.c
```diff
@@ -182,21 +183,22 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
...
+		/*
+		 * Send !!DISCONNECT!! before setting disced. If disced is set first,
+		 * process_link_channel can exit without flushing textq and the peer only
+		 * sees hangup — permanent links then immediately reconnect (#1216).
+		 */
+		chan = NULL;
+		if (l->chan && l->thisconnected) {
+			chan = ast_channel_ref(l->chan);
+		}
+		rpt_mutex_unlock(&myrpt->lock);
```
Allan-N, line 194: Should the unlock/lock calls be moved inside the `if (chan)` block?

apps/app_rpt/rpt_functions.c
```diff
@@ -311,16 +313,38 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
...
-		myrpt->savednodes[0] = 0;
 
 		if (!myrpt->links) {
 			rpt_mutex_unlock(&myrpt->lock);
 			return DC_COMPLETE;
 		}
 
-		/* loop through all links */
-		RPT_LIST_TRAVERSE(myrpt->links, l, l_it) {
+		links_copy = ao2_container_clone(myrpt->links, OBJ_NOLOCK);
```
Allan-N, line 321: It looks like my last comments about using ao2_container_clone() were light on guidance.  What I was thinking about was providing a way to capture a copy of myrpt->links and have the clone keep any needed l->chan references.  In short, I was looking to get rid of the allocated/managed/free'd `chans`.

Also, do you need to re-acquire/hold the lock while iterating the cloned list?

In the end, this all relates to when we need to hold a lock (and when we can't hold the lock).

... and, if all you need is a list of channel references, maybe keep in mind that ast_channel_ref() returns an ao2 retained reference.  Would putting the channel references in an ao2 container help?

apps/app_rpt.c
```diff
@@ -3618,16 +3621,24 @@ static inline void periodic_process_link(struct rpt *myrpt, struct rpt_link *l,
...
-			} else {
-				/* We should not retry this node type */
-				l->retries = l->max_retries + 1;
+		/*
+		 * Reconnect only while the channel is down. Count attempts here — not on every
+		 * tick while connected (that exhausted MAX_RETRIES within ~100ms after ANSWER).
+		 * Permanent links never give up unless rpt_link_stop_retries() demoted them.
+		 */
+		if (!l->chan && !l->retrytimer && l->hasconnected) {
+			int give_up = (l->max_retries != MAX_RETRIES_PERM) && (l->retries >= l->max_retries);
```
Allan-N: Maybe flip from the negative `give_up` to a positive `try_reconnect` ?

apps/app_rpt.c
```diff
@@ -4605,9 +4616,35 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
...
+		while (!ast_shutting_down() && ast_waitfor(l->chan, 0) > 0) {
+			f = ast_read(l->chan);
+			if (!f) {
+				break;
+			}
+			if (f->frametype == AST_FRAME_TEXT && f->data.ptr && f->datalen >= (int) (sizeof(DISCSTR) - 1) &&
+				!strncmp(f->data.ptr, DISCSTR, sizeof(DISCSTR) - 1)) {
+				rpt_link_stop_retries(l);
+				ast_frfree(f);
+				break;
```
Allan-N: Do we want to `break` out of the loop?  or continue looking for any queued hangup frames?

apps/app_rpt.c
```diff
@@ -4605,9 +4616,35 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
...
+			}
+			if (f->frametype == AST_FRAME_CONTROL && f->subclass.integer == AST_CONTROL_HANGUP) {
+				ast_frfree(f);
+				break;
```
Allan-N: Do we want to `break` out of the loop?  or continue looking for any queued `!!DISCONNECT!!` frames?

apps/app_rpt.c
```diff
@@ -4983,6 +5020,14 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 		continue;
 	}
 	/* Link is done: Cleanup channels and link structure */
+	/*
+	 * Flush any remaining textq (e.g. if !!DISCONNECT!! was queued rather than
```
Allan-N: It does not look like we (no longer) ever put `!!DISCONNECT!!` on the textq.

apps/app_rpt.c
```diff
@@ -4605,9 +4616,35 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
...
+		 * Receiver path: peer may have sent !!DISCONNECT!! before HANGUP.
+		 * Drain pending text so we honor an intentional drop before redialing
+		 * a permanent outbound link. (Not waiting after we ourselves sent DISCSTR.)
+		 */
+		while (!ast_shutting_down() && ast_waitfor(l->chan, 0) > 0) {
+			f = ast_read(l->chan);
+			if (!f) {
+				break;
+			}
+			if (f->frametype == AST_FRAME_TEXT && f->data.ptr && f->datalen >= (int) (sizeof(DISCSTR) - 1) &&
```
Allan-N: The combination of `>=` and `strncmp()` means that we would match a `!!DISCONNECT!!ABC` text message.  Are the text messages expected to be NUL terminated strings?  If so, can we ignore the length and just use `strcmp()`?  If we need to check the length, would `> 0` be sufficient?  In the end, I think this code should be looking for an exact match.

apps/app_rpt/rpt_functions.c
hardenedpenguin: Dropped the calloc array. Clone still snapshots the links; live channels go into a private ao2 list via ao2_link (same refcount as ast_channel_ref). We still re-lock for stop_retries because that mutates shared link fields.

apps/app_rpt.c
hardenedpenguin: Done — flipped to a positive `try_reconnect`.

apps/app_rpt/rpt_functions.c
hardenedpenguin: Yes — unlock/lock now wrap only the channel write.

apps/app_rpt.c
hardenedpenguin: Same for HANGUP — drain until the queue is empty so we can still see DISCSTR after it.

apps/app_rpt.c
hardenedpenguin: Right — we never queue it now. Comment updated; the flush stays for leftover key/keepalive frames.

apps/app_rpt.c
hardenedpenguin: No longer break on DISCSTR — keep draining so a queued HANGUP is still consumed.

apps/app_rpt.c
hardenedpenguin: Exact match now: copy, NUL-terminate, `strcmp`. `!!DISCONNECT!!ABC` no longer matches.

Allan-N: Sorry to be a pain on this one but the text message is written with :
```
		.datalen = sizeof(DISCSTR),
		.data.ptr = DISCSTR,
```
Assuming the message is received with the same string/length, why not use :
```
			if (f->frametype == AST_FRAME_TEXT && f->data.ptr && (f->datalen == (int) sizeof(DISCSTR)) && !strncmp(f->data.ptr, DISCSTR, sizeof(DISCSTR))) {
				 rpt_link_stop_retries(l);
			}

```

hardenedpenguin: Good call — switched to your length + strncmp check so we mirror how rpt_link_send_disconnect() writes the frame.

mkmer:
The root cause is not "waiting" for the text queue to clear on the "normal" hangup path.
See #1218

Allan-N:
@mkmer and I have had a few side chats about this PR and #1218.  One bit that needs a bit more investigation is how `l->disctime` fits into the disconnect/hangup flow.

... and we're questioning the ASL2, ASL3 single-threading, and ASL3 multi-threading implementations.

hardenedpenguin:
Thanks Mike — agree the normal `process_link_channel` exit path not flushing textq is a real gap (and looks like the regression you called out in #1218 when links went multi-threaded).

That said, I don't think #1218 alone covers #1216:

- On master, ilink 1/11 sets `disced` before queuing DISCSTR, so the link thread exits the wait loop immediately; cleanup never had a flush/wait on that path.
- Even with a cleanup wait, perm outbound links still redial because `MAX_RETRIES_PERM` never exhausts (#574) unless we demote via `rpt_link_stop_retries()`.
- Retries were also being counted every periodic tick while connected, burning through `MAX_RETRIES` right after ANSWER.
- Receive path: `handle_link_data()` needs to honor `!!DISCONNECT!!` and softhangup so the peer actually stops; hangup can win the race otherwise.

#1217 addresses those directly: send DISCSTR before `disced` (direct `ast_write`, not textq), demote perm retries on intentional disconnect, fix retry counting, restore receive-side teardown, and drain inbound DISCSTR before redial.

Happy to take a normal-path textq flush from #1218 as complementary for other queued TEXT (keys, keepalive, etc.) — we already call `link_process_textq()` at cleanup without adding another sleep, per Allan's review.

mkmer:
I agree that this should need to be addressed. Something here is NOT working as we expect.  On my test client, using the PR #1218 , the !!DISCONNECT!! message comes through AND it stops trying to reconnect (perm or non perm).  
I suppose the question is WHY is it not doing what we think it should based your comments and code snips...
```
[2026-09-01 10:17:11.157] NOTICE[105072]: app_rpt.c:2577 attempt_reconnect: Reconnect Attempt to 287890 in progress
    -- Call accepted by 44.15.66.5:4569 (format ulaw)
    -- Format for call is (ulaw)
[2026-09-01 10:17:11.677] DEBUG[105072]: app_rpt.c:1913 handle_link_data: Received text over link: '!NEWKEY1!'
[2026-09-01 10:17:11.679] DEBUG[105072]: app_rpt.c:1913 handle_link_data: Received text over link: 'L '
[2026-09-01 10:17:11.690] DEBUG[75078]: app_rpt.c:1167 statpost: Making statpost to http://stats.allstarlink.org/uhandler?node=572921&time=1788272231&seqno=1601&nodes=T287890&apprptvers=3.10.5&apprptuptime=94348&totalkerchunks=1&totalkeyups=219&totaltxtime=4396&timeouts=0&totalexecdcommands=19&keyed=0&keytime=79992
[2026-09-01 10:17:11.816] DEBUG[105097]: app_rpt.c:1128 perform_statpost: Response: ok
[2026-09-01 10:17:16.965] DEBUG[105072]: app_rpt.c:1913 handle_link_data: Received text over link: '!!DISCONNECT!!'
[2026-09-01 10:17:16.965] DEBUG[105072]: app_rpt.c:4564 safe_hangup: Hard hanging up channel IAX2/44.15.66.5:4569-14601
    -- Hungup 'IAX2/44.15.66.5:4569-14601'
    -- Channel Announcer/IAXLink-00000179;2 left 'softmix' base-bridge <d1b88715-182d-440b-a784-cccf753d12cd>
```

Also of note, without the PR, the message does not come through.

mkmer:
The answer is:
https://github.com/AllStarLink/app_rpt/blob/32d268ad49aa44542704bce00873a0425bc71062/apps/app_rpt.c#L1916-L1917
When the client receives DISCSTR, it dumps the connection regardless of retry count, hanging up and destroying the link forever.

hardenedpenguin:
Your test lines up: without #1218, DISCSTR never leaves the sender — ilink sets `disced`, `rpt_qwrite()` queues DISCSTR, `process_link_channel` exits on `disced != NONE`, and normal cleanup doesn't flush textq. Delivery is the primary failure.

Once DISCSTR is received, master `handle_link_data()` (`disced` + `retries = max_retries + 1`) is enough in the happy path — your log shows that.

#1218 doesn't cover:

1. **Hangup-before-TEXT race** — if `AST_CONTROL_HANGUP` is processed before the TEXT frame, `handle_link_data()` never runs; `remote_hangup_helper()` still redials perm links (`retries++ < MAX_RETRIES_PERM`). Your trace is DISCSTR-then-hangup; #1216 is hangup-only when DISCSTR never sent or loses the race.

2. **Sender path** — #1217 uses `rpt_link_send_disconnect()` (`ast_write`) before `disced`, bypassing textq. Same delivery goal as #1218's cleanup flush, without `ast_safe_sleep`.

3. **Receive teardown** — master doesn't `softhangup` on DISCSTR; teardown waits for loop exit. #1217 restores `softhangup` + `rpt_link_stop_retries()` (demotes `MAX_RETRIES_PERM`).

4. **#574** — `retries` increments every periodic tick while connected; perm never hits give-up. Fixed separately in #1217.

#1218 fixes the textq flush regression; #1217 fixes delivery (direct write), the hangup/DISCSTR race (inbound drain), perm demotion, and retry counting. Open to combining a normal-path `link_process_textq()` from #1218 — we already flush at cleanup, without the sleep.

mkmer:
OK, here is where I broke things:
https://github.com/hardenedpenguin/app_rpt/blob/06bf55d2333e4ddc68629173f77630d485a60c47/apps/app_rpt.c#L4706
We need to remove `&& l->disced == RPT_LINK_DISCONNECT_NONE` from the `while` and only leave the loop on hangups.  Then in both places where we set `l->disced` to "SILENCE" or "DISCONNECT" perform `softhangup()`.  That will restore the original path calling the `remote_hangup_helper()`... then we need what you have done to stop retries (or whatever other way we want to inhibit the permanent check).

apps/app_rpt/rpt_functions.c
```diff
@@ -342,28 +358,39 @@ enum rpt_function_response function_ilink(struct rpt *myrpt, char *param, char *
...
+			if (l->chan && l->thisconnected) {
+				/* ao2_link holds the channel so hangup cannot free it before we write. */
+				ao2_link(chans, l->chan);
 			}
 		}
 		ao2_iterator_destroy(&l_it);
 		rpt_mutex_unlock(&myrpt->lock);
+
+		RPT_LIST_TRAVERSE(chans, chan, l_it) {
+			rpt_link_send_disconnect(chan);
```
mkmer: We can't write directly to channels in different threads.  This message needs to be entered into the text queue which is then processed in the channel thread.  When one tries to access a channel the belongs to another thread, it usually fails due to the channel being locked while waiting in the `ast_waitforn()`

apps/app_rpt.c
```diff
@@ -4605,9 +4616,29 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
...
+	if (l->chan && !CHAN_TECH(l->chan, "echolink") && !CHAN_TECH(l->chan, "tlb") && l->disced == RPT_LINK_DISCONNECT_NONE &&
+		!ast_shutting_down()) {
+		struct ast_frame *f;
+
+		/*
+		 * Receiver path: peer may have sent !!DISCONNECT!! before or after HANGUP.
+		 * Drain the queue so we honor an exact DISCSTR before redialing a permanent
+		 * outbound link. Do not stop at the first of DISCSTR or HANGUP.
+		 */
+		while (!ast_shutting_down() && ast_waitfor(l->chan, 0) > 0) {
```
mkmer: This is leaving `l->pchan` un-serviced - need to keep watching and processing `l->pchan` while waiting.

apps/app_rpt/rpt_functions.c
hardenedpenguin: Agreed — dropped cross-thread `ast_write`. DISCSTR goes through `rpt_link_queue_disconnect()` → `rpt_qwrite` again; the link thread flushes it in `remote_hangup_helper` after softhangup.

apps/app_rpt.c
hardenedpenguin: Fixed — inbound DISCSTR drain now runs while `l->pchan` is still on autoservice.

apps/app_rpt.c
```diff
@@ -4672,7 +4701,12 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 
 	looptimestart = rpt_tvnow();
 
-	while (ms >= 0 && l->disced == RPT_LINK_DISCONNECT_NONE) {
+	/*
+	 * Stay in the loop after an intentional disconnect while the channel is
+	 * still up so softhangup reaches remote_hangup_helper (textq flush).
+	 * Exit once disced is set and the channel is already gone (reconnect give-up).
+	 */
+	while (ms >= 0 && (l->disced == RPT_LINK_DISCONNECT_NONE || l->chan)) {
```
mkmer: (l->disced == RPT_LINK_DISCONNECT_NONE) should not be in the loop here.
We want to wait for a hangup to exit this loop.  `l->chan` should also not need to be here as it will cause an early exit.

Need to get to this code:
https://github.com/hardenedpenguin/app_rpt/blob/125632c777053227040514b777a1e09fd8b65132/apps/app_rpt.c#L3656-L3673

apps/app_rpt.c
```diff
@@ -5819,6 +5860,9 @@ static void *rpt(void *this)
 
 		RPT_LIST_TRAVERSE(myrpt->links, l, l_it) {
 			if (l->killme) {
+				if (l->chan) {
+					ast_softhangup(l->chan, AST_SOFTHANGUP_DEV);
+				}
 				l->disced = RPT_LINK_DISCONNECT;
```
mkmer: This is "really" close to the rpt_link_stop_retries_common(), should we make a helper and use that in rpt_link_stop_retries_common() and in these locations?

apps/app_rpt.c
```diff
@@ -5985,6 +6029,9 @@ static void *rpt(void *this)
 	rpt_mutex_lock(&myrpt->lock);
 	RPT_LIST_TRAVERSE(myrpt->links, l, l_it) {
 		/* hang-up any running links */
+		if (l->chan) {
+			ast_softhangup(l->chan, AST_SOFTHANGUP_DEV);
+		}
 		l->disced = RPT_LINK_DISCONNECT_SILENT;
```
mkmer: Same helper?

apps/app_rpt.c
```diff
@@ -7282,6 +7329,9 @@ static int rpt_exec(struct ast_channel *chan, const char *data)
 				l->killme = 1;
 				l->retries = l->max_retries + 1;
 				l->disced = RPT_LINK_DISCONNECT_SILENT;
+				if (l->chan) {
+					ast_softhangup(l->chan, AST_SOFTHANGUP_DEV);
+				}
```
mkmer: Same helper?

mkmer (commented):
Still need to deal with a
https://github.com/hardenedpenguin/app_rpt/blob/125632c777053227040514b777a1e09fd8b65132/apps/app_rpt.c#L3656-L3673
In the current PR state, the above code is not allowed to run.  Need to keep the loop alive and wait for a timeout, then exit the loop.  PR #1218 did this with a return value (rv), but it "could" be done by hanging up `l->pchan` and adding necessary tests, as current state expects `l->pchan` to always exist in the while loop

apps/app_rpt/rpt_link.c
```diff
@@ -210,6 +210,46 @@ void rpt_qwrite(struct rpt_link *l, struct ast_frame *f)
...
+static void rpt_link_demote_retries(struct rpt_link *l)
+{
+	l->perma = 0;
+	if (l->max_retries > MAX_RETRIES) {
+		l->max_retries = MAX_RETRIES;
+	}
+	l->retries = l->max_retries + 1;
+}
+
+static void rpt_link_stop_retries_common(struct rpt_link *l, enum rpt_link_disconnect disced)
```
Allan-N: Just a thought ... do we really need 2 functions (one normal, one silent)?  or is it OK to have one helper with an argument that specifies the (int/boolean or enum) disconnect flavor?

apps/app_rpt.c
hardenedpenguin: Done in 8b8de47.

`process_link_channel` now loops on `while (ms >= 0)` only — no `disced` / `l->chan` in the condition. Inbound hangup parks on `disctime` even if DISCSTR already set `disced`, so periodic ticks reach LINKDISC AA. That path (and outbound retry-exhausted) returns -1 and we leave. Same stay-alive-until-timeout shape as #1218's `rv`, without hanging up `pchan`.

apps/app_rpt.c
hardenedpenguin: Done — `killme` now goes through `rpt_link_stop_retries()`.

apps/app_rpt.c
hardenedpenguin: Done — node teardown uses `rpt_link_stop_retries_silent()`.

apps/app_rpt.c
hardenedpenguin: Done — replacing an existing connection uses `rpt_link_stop_retries_silent()`.

apps/app_rpt/rpt_link.c
hardenedpenguin: The flavor argument already lives in `rpt_link_stop_retries_common()`. I left the two wrappers so call sites stay `stop_retries` vs `stop_retries_silent` without repeating the enum. Happy to collapse them if you'd rather have one public helper.

Allan-N: Collapsing was just a thought.  Either way works.

apps/app_rpt/rpt_link.c
```diff
@@ -210,6 +210,46 @@ void rpt_qwrite(struct rpt_link *l, struct ast_frame *f)
...
+void rpt_link_stop_retries(struct rpt_link *l)
+{
+	rpt_link_stop_retries_common(l, RPT_LINK_DISCONNECT);
+}
+
+void rpt_link_stop_retries_silent(struct rpt_link *l)
+{
+	rpt_link_stop_retries_common(l, RPT_LINK_DISCONNECT_SILENT);
+}
+
```
mkmer: These could be macros?  I think in general one liner's have been macros (Maybe?)
Or not :)

mkmer (approved):

apps/app_rpt.c
```diff
@@ -4605,37 +4641,47 @@ static int remote_hangup_helper(struct rpt *myrpt, struct rpt_link *l)
...
-				rpt_mutex_unlock(&myrpt->lock);
-				return 1;
-			}
+	/*
+	 * Inbound links stay in process_link_channel after chan is gone so
+	 * periodic_process_link can run LINKDISC AA once disctime expires.
+	 * That must happen even when DISCSTR already set disced.
+	 */
+	if (!l->outbound) {
+		if ((l->name[0] <= '0') || (l->name[0] > '9') || l->isremote) {
```
Allan-N, line 4667: Only unexpected inbound hangups should linger for disctime to allow reconnection. If `l->disced` is already set, the disconnect was intentional, so complete cleanup immediately (matching ASL2 behavior).

To address this, can/should we :
```
	if (!l->outbound) {
...
		/* Intentional inbound disconnects bypass the unexpected-loss reconnect grace period. */
		if (l->disced != RPT_LINK_DISCONNECT_NONE) {
				hangup_link_chan(l);
				return 0;
		}
...
		if ((l->name[0] <= '0') || (l->name[0] > '9') || l->isremote) {
```

apps/app_rpt/rpt_link.c
hardenedpenguin: Done — wrappers are macros over `rpt_link_stop_retries_common()` now.

apps/app_rpt.c
hardenedpenguin: Agreed — intentional inbound (`disced` already set) now hangs up and runs `inbound_link_finished()` immediately (same LINKDISC AA side effects, no `DISC_TIME` wait). Unexpected inbound loss still parks on `disctime`.

Allan-N (approved):

PR #1218, opened 2026-08-31
app_rpt.c: Add hangup delay clearing text message queue

### LINKED ISSUE #1216: Perm Disconnect ilink 11 not working

mkmer:
Refactor IAX thread loop.  Use `softhangup()` to signal exit for the loop rather than `l->disced`.
Honor the `l->disctime` to expire before hanging up on the client.
 
Fixes #1216 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link hangup handling by allowing queued text frames to finish sending before disconnection.
  * Applied consistent hangup behavior across remote hangups and link cleanup.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3429,10 +3429,11 @@ static inline void link_process_textq(struct rpt *myrpt, struct rpt_link *l)
 	rpt_mutex_unlock(&myrpt->lock);
 }
 
-static inline void periodic_process_link(struct rpt *myrpt, struct rpt_link *l, const int elap)
+static inline int periodic_process_link(struct rpt *myrpt, struct rpt_link *l, const int elap)
 {
 	int newkeytimer_last, max_retries;
 	int myrx;
+	int rv = 0;
 
 	link_process_textq(myrpt, l);
 
@@ -3551,7 +3552,7 @@ static inline void periodic_process_link(struct rpt *myrpt, struct rpt_link *l, 
 	if ((!l->linklisttimer) && (l->name[0] != '0') && (!l->isremote)) {
 		struct ast_str *lstr = ast_str_create(RPT_AST_STR_INIT_SIZE);
 		if (!lstr) {
-			return;
+			return rv;
 		}
 		l->linklisttimer = myrpt->p.linkpost_time * 1000;
 		ast_str_set(&lstr, 0, "%s", "L ");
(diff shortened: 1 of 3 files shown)


apps/app_rpt.c
```diff
@@ -4575,24 +4575,33 @@ static inline void hangup_link_chan(struct rpt_link *l)
...
+	 */
 	if (l->chan) {
-		/* This will be a "long" delay, dump any audio in the l->pchan
-		 * as we are now about to close down the link channel.  If we don't
-		 * autoservice, l->pchan can report long voice queue and add unnecessary delay audio
-		 * on a reconnect
-		 */
 		ast_autoservice_start(l->pchan);
 		link_process_textq(myrpt, l);
 		ast_safe_sleep(l->chan, MSWAIT * 10); /* Allow the channel to send the text messages */
```
Allan-N, line 4601: Although using ast_safe_sleep() is a bit better because the API watches for hangups while waiting I will say that I'm not wild about arbitrary (?) sleeps.

Why `MSWAIT * 10`?  Asing because we've seen logs w/ `Exceptionally long voice queue length (97 voice / 97 total)` and while those may not be at all related I wonder if the long queues suggest that we should be using at least `MSWAIT * 100`?

mkmer: The channel is supposed to be servicing every 20ms, so it should be fine with 20 or 40ms JIC.  The original Jim code was 250ms.  In my testing 10x seemed fine although 40ms should do it.  Ofcourse I added the code to the while loop and missed the wait part on exiting the while loop, thus "re-breaking" the thing I had fixed when moving to the multi thread links.
Again, it would be much nicer if we could ask the channel if there was anything in it's queues and just wait till they empty, but alas that's not in the cards.

mkmer:
Folded into #1217

PR #1227, opened 2026-09-01
app_rpt.c: Optimize calculation of remrx, reducing locking requirements.

mkmer:
Reduce locking by determining `remrx` with `myrpt->remrx`.  Only search other links IF the link is receiving to find if some other link is also receiving.  

## Summary by CodeRabbit

* **Performance**
  * Improved link status processing by replacing repeated link-list scans with more efficient tracking.
  * Updated receive-state handling to provide faster, more consistent channel status updates.

* **Bug Fixes**
  * Improved cleanup when link connection setup cannot allocate required resources.

* **Diagnostics**
  * Added the tracked eligible receive-link count to diagnostic output.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -3422,6 +3422,23 @@ static inline void link_process_textq(struct rpt *myrpt, struct rpt_link *l)
 	rpt_mutex_unlock(&myrpt->lock);
 }
 
+static inline void rpt_update_link_rx_state(struct rpt *myrpt, struct rpt_link *l, int new_lastrx)
+{
+	if (l->lastrx == new_lastrx) {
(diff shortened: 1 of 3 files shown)


mkmer:
I believe this is syncronized as needed.  IF there is a mode change, it is now stored in `l->mode`.  When the channel analyzes frames, it first checks mode and `l->lastrx` and adjusts it based on the mode change.  Then it analyzes the rx state and again updates if mode < 2.  IF the mode changed while `lastrx` was set, we catch that first in `rpt_update_link_mode()`

mkmer:
This one needs a complete overhaul, will create a new PR with different design.

PR #1229, opened 2026-09-02
app_rpt.c: Address audiohook locking hang.  No need to detach an audiohook after hanging up

mkmer:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link teardown handling to safely detach audio processing before disconnecting channels.
  * Prevented cleanup of audio resources that were not successfully initialized, reducing the risk of shutdown errors.

### DIFF
--- a/apps/app_rpt.c
+++ b/apps/app_rpt.c
@@ -5017,17 +5017,16 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 		donodelog_fmt(myrpt, l->hasconnected ? "LINKDISC,%s" : "LINKFAIL,%s", l->name);
 	}
 	rpt_frame_queue_free(&l->frame_queue);
-
-	/* 1. Detach audiohook while l->chan is still valid */
-	ast_audiohook_lock(&l->altaudio);
-	ast_audiohook_detach(&l->altaudio);
-	ast_audiohook_unlock(&l->altaudio);
-
-	/* 2. Hang-up the channels */
 	hangup_link_chan(l);
+
 	if (l->pchan) {
 		ast_hangup(l->pchan);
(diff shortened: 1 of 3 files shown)


apps/app_rpt/app_rpt.h
```diff
@@ -595,6 +595,7 @@ struct rpt_link {
 	int retries;
 	int max_retries;
 	int reconnects;
+	int audiohook_init;
```
Allan-N: I question the value of this added variable.  From my read of the code, the value is set to the return value from the call to ast_audiohook_init().  Because that API **ALWAYS** returns 0 the variable will always have a value of 1.  The exception would be if we create a `struct rpt_link` without making a call to ast_audiohook_init().  Does that ever happen?  If not, there's no reason to check if the call was made.

apps/app_rpt.c
```diff
@@ -5017,25 +5017,23 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
...
-
-	/* 2. Hang-up the channels */
 	hangup_link_chan(l);
+
 	if (l->pchan) {
 		ast_hangup(l->pchan);
 		l->pchan = NULL;
 	}
 
+	if (!l->audiohook_init) {
```
Allan-N: Do we ever not call ast_audiohook_init() ?  If we always make the call then there's no point in checking l->audiohook_init.

Allan-N (approved):

apps/app_rpt.c
mkmer: Always create the hook on every link.  It appears we can change modes on the fly using COP 33,34,35.  It would be a bit more difficult to add the hook and remove the hook when the mode changes.

Issue #1069, opened 2026-06-18
chan_voter: thread blocking "problem"


mkmer:
I've noticed these for a while
Seems to be related to an RTCM reconnecting.  Maybe starting a new thread before the old one has "finished"?
Also appears to "recover" after a short time.

```
[2026-06-16 00:17:47.229] NOTICE[1081306] chan_voter.c: VOTER 2324: Client Ada connected.
[2026-06-16 00:17:47.229] NOTICE[1081306] chan_voter.c: Client: Ada is sending mix mode flag, setting client to mix mode
[2026-06-16 00:17:47.581] WARNING[1081294] chan_iax2.c: Silently dropping frame without existent call number: 10971
[2026-06-16 00:17:51.767] DEBUG[1808216][C-000003d2] channel.c: Thread LWP 1808216 is blocking 'voter/2324', already blocked by thread LWP 1081353 in procedure ast_waitfor_nandfds
[2026-06-16 00:17:51.914] DEBUG[1081353] channel.c: Thread LWP 1081353 is blocking 'voter/2324', already blocked by thread LWP 1808216 in procedure ast_waitfor_nandfds
[2026-06-16 00:17:52.034] DEBUG[1081353] channel.c: Thread LWP 1081353 is blocking 'voter/2324', already blocked by thread LWP 1808216 in procedure ast_waitfor_nandfds
```
AND

```
[2026-06-16 03:46:08.263] NOTICE[1081306] chan_voter.c: VOTER 2324: Client Ada connected.
[2026-06-16 03:46:08.263] NOTICE[1081306] chan_voter.c: Client: Ada is sending mix mode flag, setting client to mix mode
[2026-06-16 03:46:09.075] WARNING[1081289] chan_iax2.c: Silently dropping frame without existent call number: 7217
[2026-06-16 03:46:11.077] WARNING[1081294] chan_iax2.c: Silently dropping frame without existent call number: 7217
[2026-06-16 03:46:15.359] DEBUG[1831567][C-000003d5] channel.c: Thread LWP 1831567 is blocking 'voter/2324', already blocked by thread LWP 1081353 in procedure ast_waitfor_nandfds
[2026-06-16 03:46:15.374] DEBUG[1081353] channel.c: Thread LWP 1081353 is blocking 'voter/2324', already blocked by thread LWP 1831567 in procedure ast_waitfor_nandfds
[...]

PR #1232, opened 2026-09-09
app_rpt.c: wait on chan, not rxchannel in rpt_exec() thread

mkmer:
Fixes Issues #1069 

Appears to have been "wrong" for a long time?
https://github.com/AllStarLink/ASL-Asterisk/blob/d6a4b48d58288978bf2e5bf22e7e9cfcdea8b2f3/asterisk/apps/app_rpt.c#L23348-L23352

`rpt_exec()` is a separate thread that should NOT be trying to service channels managed by other threads - in this case rxchannel is managed by `rpt()`

## Summary by CodeRabbit

* **Bug Fixes**
  * Inbound rate-limit waits now monitor the caller’s connection, preventing unrelated radio-channel hangups from interrupting calls.
  * Rate-limit sleeping is now tied to the inbound call channel for more consistent call handling.


apps/app_rpt.c
Allan-N, line 7011: Since we are looking at `myrpt->rxchannel` usage, is this reference correct? or should we be using `chan`?

mkmer: I think that's "right" - we are keeping the repeater vars (set with a "V" option on the rpt call in the dialplan) on the `rxchannel`....
https://github.com/AllStarLink/app_rpt/blob/6966503d14cefb49a5bd269edb8524e549de0a85/apps/app_rpt/rpt_config.c#L1353-L1386
Maybe we need all of that protection here too?
Maybe we need to trigger the manager?
Maybe address this in a different PR if we should..

Allan-N (approved):

PR #1235, opened 2026-09-09
app_rpt.c: Create macro `IS_NODE_EXTEN()`

### LINKED ISSUE #1234: Create a macro for node checks, what is it really checking?
This chunk of code is "everywhere":
`(l->name[0] <= '0') || (l->name[0] > '9')`
or
`(l->name[0] > '0') && (l->name[0] <= '9')`

Maybe a macro...
`#define is_valid_ASLnode(l) (l[0] > '0' && l[0] <= '9')`

mkmer:
Fixes #1234

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recognition of valid AllStar Link nodes across link management workflows.
  - Corrected node classification during reconnect and disconnect handling.
  - Improved link control actions, including enabling, disabling, following, and demanding links.
  - Ensured valid node identification is applied consistently when selecting foreign links and tracking link activity.


apps/app_rpt/app_rpt.h
```diff
@@ -1224,3 +1224,5 @@ char *rpt_complete_function_list(const char *line, const char *word, int pos, in
 #define RPT_LIST_TRAVERSE(container, var, iterator) \
 	(iterator) = (container) ? ao2_iterator_init((container), 0) : (struct ao2_iterator) { 0 }; \
 	for (; ((var) = ao2_iterator_next(&(iterator))); ao2_ref((var), -1))
+
+#define rpt_valid_ASLNode(l) ((l[0] > '0') && (l[0] <= '9'))
```
Allan-N: Q? there are few references to ASL in the app_rpt/apps/... source code.  Is this the time/place to say names starting with 1-9 are ASL nodes?

Why `l` (that tends to be used as the variable or an rpt_link) ?  Why not `node`, 'name`, `nodename`, `nodenumber`, or `str` (to reflect that the argument is a string ?

mkmer: Not sure, I just picked a name that made sense to me at the time -> this is definitely a hard coded assumption about names (and I'd say those rules are based on AllStarLink rules right?) 

Should we go with name instead of `l`?  Make sense to me.

Allan-N: For discussion purposes, what names are expected?  Looking at the code surrounding this change I see one place with a `/* Ignore '0' nodes */` comment.  Is a `->name` value of "0" special?  what other names might we find that would not be "ASL" names?  and if the "0" name is the only one that's special then maybe we query/compare for that special name?

mkmer: I assume 0xxxxx is not a valid allstarlink node number?  The code forces it to be 1 -> 9, otherwise it's considered some other type of connection.
There is other code hiding in there with respect to the length of the node as well, must be 4 - 6 digits long.  I bet if we dig around, a 7 digit node would do something odd too.

hardenedpenguin:
## Review notes

Useful cleanup. Two things on tip `faca2cd`:

**Missing convert** — `function_ilink` case 6: first traverse uses the macro, second still has:
```c
if ((l->name[0] <= '0') || (l->name[0] > '9'))
```
Should be `!rpt_valid_ASLNode(l->name)` (or the final name).

**What it means / naming** — The check is only “first char is digit `1`–`9`”: AllStar-style node ids vs `'0'` IAXRPT/phone clients and non-numeric/foreign names. Not a full validator (no length/all-digits). Nearby `== '0'` / `> '9'` sites are different and fine to leave.

`rpt_valid_ASLNode` overclaims a bit vs nearby `IS_*` style. Prefer:
```c
#define IS_ALLSTAR_NODE(name) ((name)[0] > '0' && (name)[0] <= '9')
```
or more literal `IS_NUMERIC_NODE_NAME`. Parenthesize `(name)[0]`.

mkmer:
What's the consensus on the "correct" name?

Allan-N:
I like the `IS_xxx_NAME()` naming convention.  The comment about our not checking the full name is also valid.  Having a rpt_valid_ASLNode(). IS_ALLSTAR_NODE(), or IS_NUMERIC_NODE_NAME() that's only looking at the 1st character and not the full name is not necessarily accurate (unless we stop at the first non-digit).

>What's the consensus on the "correct" name?

I think we're all on the same page that a macro would help.  As for the name, I'd don't think any of us are putting a stake in the ground.  Toss a few coins, throw a few darts, ...

Q? is 3009999 a valid AllStar node name (3xxxxxx is EchoLink)?

... and I'm still curious what names we'll find other than "0" that we would be comparing against.  Will we really see a callsign here?  If we only care about "0" then maybe the test should be looking for "0".

mkmer:
There are tests for `=='0'`
https://github.com/AllStarLink/app_rpt/blob/def141082ab0d7bd8bd1c06ca952c9dad88f615a/apps/app_rpt.c#L1813-L1814
and `!='0'`
https://github.com/AllStarLink/app_rpt/blob/def141082ab0d7bd8bd1c06ca952c9dad88f615a/apps/app_rpt.c#L3442-L3444
This appear to determine the sending of text messages.

There is a test for `> '9'` (which is probably wrong?)
https://github.com/AllStarLink/app_rpt/blob/def141082ab0d7bd8bd1c06ca952c9dad88f615a/apps/app_rpt.c#L7409-L7410

Then there is this:
https://github.com/AllStarLink/app_rpt/blob/def141082ab0d7bd8bd1c06ca952c9dad88f615a/apps/app_rpt.c#L7405
Which would exclude the `'3'` as it looks for echolink....

Any of these could be "wrong" or questionable but probably not related to making a macro out of a bunch of repeated compares. 
If we don't like the validation procedure it probably should be a PR that changes that?
That leaves us with "what the hack was Jim testing here?"
  Essentially a name that starts with 1 -> 9 that's it, nothing more...
And what should we call it?

Allan-N (approved):

PR #1247, opened 2026-09-16
Reject invalid configuration during reload and channel startup

### LINKED ISSUE #1242: Reloading broken config crashes asterisk
To reproduce:
1 - start asterisk with a valid rpt.conf
2 - modify rpt.conf (ex. set up a context with an invalid or non-exisiting template)
3 - reload app rpt with `module reload app_rpt` on the cli.
4 - Asterisk reports bad config, disabling broken repeater, then Asterisk crashes.

hardenedpenguin:
## Summary

- abort `app_rpt` reload when `rpt.conf` cannot be loaded or parsed
- preserve running node state when either the shared or per-node reload read fails
- report shared reload failure through `AST_MODULE_RELOAD_ERROR`
- reject `CONFIG_STATUS_FILEINVALID` before using configuration results throughout the channel drivers
- keep active USB tuning values unchanged when tune-file parsing fails
- abort Voter channel creation through its cleanup path when `voter.conf` is missing or invalid
- keep initial-load failures distinct from rejected reloads in operator logs

Fixes #1242.

### Reload

`reload()` ignored the return value from `load_config(1)` and continued processing stale `reload_request` state. An invalid configuration could delete or restart live repeater nodes. With multiple active nodes this reproduced as a SIGSEGV and Asterisk restart.

Existing nodes subsequently read `rpt.conf` again in `load_rpt_vars()`. That function destroyed the current node configuration before validating its replacement, so a read/parse failure in that second phase could still terminate the node thread. The replacement is now loaded first; failure leaves the current node state intact.

### Initial startup

`chan_echolink` also reads `rpt.conf` during instance setup. It checked for `NULL` but not Asterisk's non-NULL `CONFIG_STATUS_FILEINVALID` sentinel. A bad inherited template therefore reached `ast_variable_retrieve()` with the sentinel and caused SIGSEGV before app_rpt could log that the repeater was disabled.

EchoLink now detects the sentinel, logs the invalid configuration, frees the staged instance/list allocations, and declines to load. Asterisk remains running; app_rpt's existing initial-load behavior disables repeaters until a valid configuration is supplied.

### Channel-driver configuration loads

A broader audit found the same unchecked sentinel in additional EchoLink, TLB, Voter, SimpleUSB, and USBRadio configuration paths. Every `ast_config_load()` under `channels/` now checks `CONFIG_STATUS_FILEINVALID` before using the result. SimpleUSB and USBRadio parse tuning values into temporary state and commit only after success, while Voter aborts channel creation through `ast_hangup()` so its existing cleanup removes the partially created instance.

## Verification

Tested on an isolated arm64 ASL3 node.

### Reload sequences

Built and deployed `22.10.1+asl3-3.10.5-11.deb13`, then repeated:

- invalid reload: CLI reported reload failure; existing node remained; PID unchanged
- valid reload followed by invalid reload: existing node remained; PID unchanged
- five-node crash reproduction: all five nodes remained; PID unchanged; no SIGSEGV

### Initial-startup/EchoLink path

With local-only EchoLink settings and a bad inherited template:

- before fix: `chan_echolink` load caused SIGSEGV; systemd restarted Asterisk (`39719` → `50647`)
- after fix: EchoLink logged `Config file rpt.conf is in an invalid format. Aborting.`; PID remained unchanged; no core dump; node remained running
- repeated after adding failure-path cleanup with the same successful result

All temporary configurations were restored. Telemetry and external links remained disabled during testing.

### Channel-driver audit

The updated EchoLink, TLB, Voter, SimpleUSB, and USBRadio sources all compiled and linked successfully in the ASL3 Asterisk build tree on mustang.

## Checks

- `clang-format-19` changed-line check clean
- `codespell` clean
- whitespace/diff checks clean
- IDE lints clean
- exact changed sources compiled and linked successfully


apps/app_rpt.c
```diff
@@ -6205,10 +6205,20 @@ static int load_config(int reload)
 
 	cfg = ast_config_load("rpt.conf", config_flags);
 	if (!cfg) {
-		ast_log(LOG_ERROR, "Unable to open radio repeater configuration rpt.conf.  Radio Repeater disabled.\n");
+		if (reload) {
+			ast_log(LOG_ERROR,
+				"Unable to open radio repeater configuration rpt.conf. Reload aborted; existing configuration remains active.\n");
```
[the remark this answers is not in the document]
hardenedpenguin: Fixed in 5280df2. Both reload-failure paths now say the existing state is preserved without asserting that an active configuration exists.

apps/app_rpt.c
```diff
@@ -8375,7 +8385,10 @@ static int reload(void)
 	int n;
 
 	ast_mutex_lock(&rpt_master_lock);
-	load_config(1);
+	if (load_config(1)) {
+		ast_mutex_unlock(&rpt_master_lock);
+		return AST_MODULE_RELOAD_ERROR;
+	}
```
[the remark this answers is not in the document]
hardenedpenguin: Fixed in ad18afc. load_rpt_vars() now loads and validates the replacement before destroying the current node configuration. A per-node reload read/parse failure logs the failure, retains the existing state, and returns without terminating the node thread; initial-load failures retain their existing disable/exit behavior.

mkmer:
I have not tested this PR yet, but I think there is a bit more to it.
If we startup with an invalid config (particularly with a "bad" template name on a context), it complains 1 time, then starts looping on core dumps:
```
[2026-09-16 09:02:02.511] Asterisk 22.10.1+asl3-3.10.5-1.deb13 built by builder @ allstarlink.org on a x86_64 running Linux on 2026-08-29 17:49:54 UTC
[2026-09-16 09:02:02.788] WARNING[388418] config.c: Inheritance requested, but category 'bad' does not exist, line 10 of /etc/asterisk/rpt.conf
[2026-09-16 09:02:09.765] Asterisk 22.10.1+asl3-3.10.5-1.deb13 built by builder @ allstarlink.org on a x86_64 running Linux on 2026-08-29 17:49:54 UTC
[2026-09-16 09:02:17.013] Asterisk 22.10.1+asl3-3.10.5-1.deb13 built by builder @ allstarlink.org on a x86_64 running Linux on 2026-08-29 17:49:54 UTC
[2026-09-16 09:02:24.247] Asterisk 22.10.1+asl3-3.10.5-1.deb13 built by builder @ allstarlink.org on a x86_64 running Linux on 2026-08-29 17:49:54 UTC
[2026-09-16 09:02:31.319] Asterisk 22.10.1+asl3-3.10.5-1.deb13 built by builder @ allstarlink.org on a x86_64 running Linux on 2026-08-29 17:49:54 UTC
[2026-09-16 09:02:38.516] Asterisk 22.10.1+asl3-3.10.5-1.deb13 built by builder @ allstarlink.org on a x86_64 running Linux on 2026-08-29 17:49:54 UTC
[2026-09-16 09:02:45.754] Asterisk 22.10.1+asl3-3.10.5-1.deb13 built by builder @ allstarlink.org on a x86_64 running Linux on 2026-08-29 17:49:54 UTC
[2026-09-16 09:02:52.800] Asterisk 22.10.1+asl3-3.10.5-1.deb13 built by builder @ allstarlink.org on a x86_64 running Linux on 2026-08-29 17:49:54 UTC
```

hardenedpenguin:
> I have not tested this PR yet, but I think there is a bit more to it. If we startup with an invalid config (particularly with a "bad" template name on a context), it complains 1 time, then starts looping on core dumps:
> 
That is a completely different issue, I can address it in the same PR if you would like.

mkmer:
That's the issue I was reporting - having a bad template in the config... I suppose it wasn't clear in the issue?

apps/app_rpt/rpt_config.c
```diff
@@ -693,19 +693,39 @@ void load_rpt_vars(int n, int init)
 	ast_verb(3, "%s config for repeater %s\n", (init) ? "Loading initial" : "Re-Loading", rpt_vars[n].name);
 	ast_mutex_lock(&rpt_vars[n].lock);
 
-	if (rpt_vars[n].cfg) {
-		ast_config_destroy(rpt_vars[n].cfg);
-	}
-
 	cfg = ast_config_load("rpt.conf", config_flags);
 	if (!cfg) {
```
Allan-N, line 697: Does the following help?
```
	if (!cfg || (cfg == CONFIG_STATUS_FILEINVALID)) {
```

hardenedpenguin:
@mkmer You were right—the bad template was the common trigger, and my earlier “completely different issue” characterization was incorrect. Reload and initial startup reached two different unsafe paths, but both belong to this report.

I reproduced your startup loop on crazytrain. The startup crash occurred in `chan_echolink`: `store_config()` loaded `rpt.conf` and checked only for `NULL`. A bad inherited template returns the non-NULL `CONFIG_STATUS_FILEINVALID` sentinel, which was then passed to `ast_variable_retrieve()`. That caused SIGSEGV before app_rpt's master thread could reach its “Radio Repeater disabled” log, explaining the single template warning followed by repeated systemd restarts.

Commit `24e4fb1` now checks for `CONFIG_STATUS_FILEINVALID` before dereferencing the result. It logs the error, frees the staged EchoLink instance and any permit/deny list backing allocations, and declines the EchoLink module load. This works because the invalid sentinel never reaches the config accessors, while app_rpt's existing initial-load path can safely disable repeater startup without taking down Asterisk.

Runtime verification with the same bad-template configuration:

- before: `SIGSEGV`, core dump, PID `39719` → `50647`
- after: `Config file rpt.conf is in an invalid format. Aborting.`, PID unchanged, no core dump, node/service preserved

The PR description and title now cover both the reload and initial-startup paths.

apps/app_rpt/rpt_config.c
```diff
@@ -693,19 +693,39 @@ void load_rpt_vars(int n, int init)
...
+		if (init) {
+			ast_log(LOG_ERROR, "Unable to open radio repeater configuration rpt.conf. Radio Repeater disabled.\n");
+		} else {
+			ast_log(LOG_ERROR,
+				"Unable to open radio repeater configuration rpt.conf. Reload for repeater %s aborted; "
+				"existing state preserved.\n",
+				rpt_vars[n].name);
```
Allan-N: If we have multiple nodes in a configuration will we end up with a blast of log messages, one for each node? ... all for the same "no config file" or "bad config file"?

hardenedpenguin: A normal invalid reload is rejected by `load_config(1)` before any nodes are scheduled, so it produces only the shared error. Per-node messages occur only if the file changes during the narrow window between shared validation and each node’s reload. Each affected node logs once because its reload flag is cleared first; the node name identifies which nodes retained their previous state.

apps/app_rpt/rpt_config.c
hardenedpenguin: It would combine the failure handling, but keeping the branches separate preserves the distinction between a missing and malformed configuration in the logs. I’d prefer to retain that operator-facing detail.

apps/app_rpt/rpt_config.c
Allan-N: OK

Allan-N (approved):

apps/app_rpt/rpt_config.c
```diff
@@ -693,19 +693,39 @@ void load_rpt_vars(int n, int init)
...
+			ast_log(LOG_ERROR,
+				"Errors detected in the radio repeater configuration rpt.conf. Reload for repeater %s aborted; "
+				"existing state preserved.\n",
+				rpt_vars[n].name);
+		}
 		ast_mutex_unlock(&rpt_vars[n].lock);
-		ast_log(LOG_ERROR, "Errors detected in the radio repeater configuration rpt.conf.  Radio Repeater disabled.\n");
-		pthread_exit(NULL);
+		if (init) {
+			pthread_exit(NULL);
```
mkmer: Could we just do this at the ln 712 `if (init) {`? 
We also have been replacing `pthread_exit()` with simple `return` in the general case.
That said, maybe we just don't need this chunk at all.

apps/app_rpt/rpt_config.c
```diff
@@ -693,19 +693,39 @@ void load_rpt_vars(int n, int init)
...
+			ast_log(LOG_ERROR,
+				"Unable to open radio repeater configuration rpt.conf. Reload for repeater %s aborted; "
+				"existing state preserved.\n",
+				rpt_vars[n].name);
+		}
 		ast_mutex_unlock(&rpt_vars[n].lock);
-		ast_log(LOG_ERROR, "Unable to open radio repeater configuration rpt.conf.  Radio Repeater disabled.\n");
-		pthread_exit(NULL);
+		if (init) {
+			pthread_exit(NULL);
```
mkmer: Same `pthread_exit()` comment

mkmer (commented):
~~I think we have the same config problem in `chan_tlb`~~ as it's mostly a 1:1 copy of `chan_echolink`

Looks like no. Sorry

mkmer:
Looking around at the other config load in `chan_tlb` and `chan_echolink` it appears the `load_module()` calls also are missing the `cfg == CONFIG_STATUS_FILEINVALID` tests for the driver config files :(

apps/app_rpt/rpt_config.c
Allan-N: This is not the thread function so we can't simply `return`.

hardenedpenguin: Allan is correct—`load_rpt_vars()` is a helper called by `rpt_master()`, not the thread entry function. Returning on an initial-load failure would let startup continue without valid node configuration. The `init` guard preserves the existing thread-exit behavior during startup, while reload failures return normally and keep the running node intact.

apps/app_rpt/rpt_config.c
hardenedpenguin: Same reasoning here: initial-load failure must stop `rpt_master()`, while a per-node reload failure must return without terminating the active node thread. Removing this guard would restore the reload failure we are fixing.

hardenedpenguin:
@mkmer Good catch. I audited every `ast_config_load()` under `channels/` and added missing `CONFIG_STATUS_FILEINVALID` handling in EchoLink, TLB, Voter, SimpleUSB, and USBRadio. The USB tuning loaders now stage values until parsing succeeds, and Voter aborts channel creation through its existing cleanup path. All five drivers compile successfully. This is included in `439d900`.

channels/chan_voter.c
```diff
@@ -4002,6 +4002,12 @@ static struct ast_channel *voter_request(const char *type, struct ast_format_cap
 	 */
 	if (!(cfg = ast_config_load(config, zeroflag))) {
 		ast_log(LOG_ERROR, "Unable to load config %s\n", config);
+		ast_hangup(chan);
```
Allan-N, line 4008: Before calling ast_hangup() do we also need to call ast_module_user_remove() ?

... and, if so, make the change in both places.

channels/chan_usbradio.c
```diff
@@ -799,32 +799,30 @@ static int load_tune_config(struct chan_usbradio_pvt *o, const struct ast_config
...
 	int configured = 0;
+	int rxmixerset = 500;
+	int txmixaset = 500;
+	int txmixbset = 500;
+	float rxvoiceadj = 0.5;
+	float rxctcssadj = 0.5;
+	int txctcssadj = 200;
+	int rxsquelchadj = 500;
+	int txslimsp = DEFAULT_TX_SOFT_LIMITER_SETPOINT;
+	int fever = o->fever;
```
Allan-N: Why default to `o->fever` ?

apps/app_rpt/rpt_config.c
mkmer: I guess a nit pick but I was thinking this:
```
		if (init) {
			ast_log(LOG_ERROR, "Unable to open radio repeater configuration rpt.conf. Radio Repeater disabled.\n");
                ast_mutex_unlock(&rpt_vars[n].lock);
		        pthread_exit(NULL); 

		}

		ast_log(LOG_ERROR,
			"Unable to open radio repeater configuration rpt.conf. Reload for repeater %s aborted; "
			"existing state preserved.\n",
			rpt_vars[n].name); 
		ast_mutex_unlock(&rpt_vars[n].lock);
		return;
```
Same for both locations....
Then when off the rails on `pthread_exit()`

channels/chan_voter.c
mkmer: If we hangup on one of these "critical" drivers, the repeater will continue to restart until it "locks out" on retries....
May be worth testing to see which way we want it to behave?

hardenedpenguin: Good catch. `p->u` is allocated before loading the configuration. I added `ast_module_user_remove()` centrally in `voter_hangup()` rather than duplicating it before both `ast_hangup()` calls. That covers both failure paths and normal Voter hangups. Fixed in `36f36b0`.

channels/chan_usbradio.c
hardenedpenguin: Initializing from `o->fever` is intentional. Unlike the other tuning values, the previous implementation did not reset `fever` before parsing, so omitting it preserved the current value. The temporary state retains that behavior; I added a comment in `36f36b0` to clarify it.

channels/chan_voter.c
hardenedpenguin: I traced the failed request through `app_rpt`: a NULL channel request makes `rpt_setup_channels()` fail, after which `rpt()` calls `disable_rpt()`. That sets `RPT_DELETED_PENDING` specifically to prevent continual thread restarts, so the affected repeater should be disabled rather than reaching the retry lockout. Continuing channel creation with an invalid configuration would be unsafe.

apps/app_rpt/rpt_config.c
hardenedpenguin: Yes, that structure is cleaner. I updated both failure paths in `36f36b0` so the initial-load branch logs, unlocks, and exits directly; the reload path then logs, unlocks, and returns normally. This preserves the behavior while simplifying the control flow.

channels/chan_voter.c
Allan-N: If I'm processing the comments and code changes I think you are saying that a call to `voter_request()` that allocates a channel, hangs up the channel, and returns NULL will be followed by a call to `voter_hangup()` ... and that's where the call would be made to `ast_module_user_remove()`?  What channel would be passed in the call to `voter_hangup()`?

Has this been tested?

hardenedpenguin: Yes. `ast_hangup(chan)` synchronously invokes the channel technology’s hangup callback with the same channel allocated by `voter_request()`.

I verified this on crazytrain without Voter hardware by loading the patched module with a valid configuration, replacing `voter.conf` with an invalid inherited template, and originating `Voter/990099`.

The log sequence was:

```
Asterisk channel created for Voter/990099
Config file voter.conf is in an invalid format
Channel ... 'voter/990099' hanging up
Channel voter/990099: Hangup
Channel ... 'voter/990099' destroying
```

The Asterisk PID remained unchanged, the module use count returned to zero, and `chan_voter.so` unloaded successfully. This confirms that `voter_hangup()` receives the allocated Voter channel and performs the centralized module-user cleanup.

The original module and configuration were restored after testing.

Allan-N (approved):

mkmer (approved):

PR #1250, opened 2026-09-17
Correct DNS node domain configuration handling

### LINKED ISSUE #1249: Incorrect variable used to assign dns_node_domain
### Problem:
The value of `node_lookup_method` is assigned to the variable for `dns_node_domain`.

### To reproduce:
rpt.conf:
```
[general]
node_lookup_method = testValue ; method used to lookup nodes
dns_node_domain = asllab.lan
```
Output in /var/log/asterisk/messages.log:
```
[2026-09-16 17:12:58.162] WARNING[1463] app_rpt.c: Configuration error: node_lookup_method, testValue, is not valid
[2026-09-16 17:12:58.162] NOTICE[1463] app_rpt.c: Domain used for DNS node lookup is: testValue
```

### Relevant code:
https://github.com/AllStarLink/app_rpt/blob/df049ea0e7976fe19201c765a981234ae8fef1f1/apps/app_rpt.c#L6243

### Suggested fix: 
The value of `cval` should be passed into the function `pt_is_valid_dns_name()` and if found to be valid should be assigned to `rpt_dns_node_domain` on the following line.  

Is there more? Perhaps.  Someone familiar with the design should review.

### Version Info
│ OS            : Debian GNU/Linux 13 (trixie)
│ OS Kernel     : 6.18.50+rpt-rpi-v8
│
│ Asterisk      : 22.10.1+asl3-3.10.5-1.deb13
│ ASL [app_rpt] : 3.10.5
│
│ Installed ASL packages :
│
│   Package                         Version
│   ==============================  ==============================
│   allmon3                         1.9.1-1.deb13
│   asl3                            3.18.2-2.deb13
│   asl3-appliance                  2.1.0-2.deb13
│   asl3-appliance-pi               2.1.0-2.deb13
│   asl3-asterisk                   2:22.10.1+asl3-3.10.5-1.deb13
│   asl3-asterisk-config            2:22.10.1+asl3-3.10.5-1.deb13
│   asl3-asterisk-modules           2:22.10.1+asl3-3.10.5-1.deb13
│   asl3-menu                       1.18-1.deb13
│   asl3-update-nodelist            2.1.0-1.deb13
│   asl-apt-repos                   2.0-1.deb13
│   cockpit                         337-1+deb13u2
│   cockpit-bridge                  337-1+deb13u2
│   cockpit-networkmanager          337-1+deb13u2
│   cockpit-packagekit              337-1+deb13u2
│   cockpit-sosreport               337-1+deb13u2
│   cockpit-storaged                337-1+deb13u2
│   cockpit-system                  337-1+deb13u2
│   cockpit-wifimanager             1.2.0-1.deb13
│   cockpit-ws                      337-1+deb13u2
│   dahdi                           1:3.1.0-2.1
│   dahdi-linux                     1:3.4.0-11.asl.deb13

hardenedpenguin:
## Summary

- validate and publish `dns_node_domain` using its own configuration value
- keep the configured domain in owned storage instead of retaining a pointer into a destroyed config object
- synchronize domain publication with DNS query construction
- reject node labels and complete DNS lookup names that exceed protocol limits

Fixes #1249.

## Cause

`load_config()` retrieved `dns_node_domain` into `cval`, but validated and assigned the previous `val` used for `node_lookup_method`. A valid custom domain was therefore replaced by the lookup-method text. Simply assigning `cval` would also leave `rpt_dns_node_domain` pointing into the configuration object after `ast_config_destroy(cfg)`.

The configured domain is now copied into private fixed-size storage through a locked setter. `node_lookup_bydns()` uses the same lock while building its local query name, then releases it before DNS resolution. The query buffer accounts for the SRV prefix, maximum node label, domain, separator, and terminator; oversized complete names are rejected rather than truncated.

## Verification

Built and linked the exact changed sources in the ASL3 Asterisk build tree on mustang.

Runtime verification on isolated crazytrain covered:

- reported reproduction: invalid `node_lookup_method = testValue` with valid `dns_node_domain = asllab.lan` logged `asllab.lan`
- malformed domain `bad_domain`: rejected and reset to `nodes.allstarlink.org`
- omitted domain: reset to `nodes.allstarlink.org`
- repeated module reloads: Asterisk PID unchanged and node `546050` remained active

The original module and configuration were restored after testing.

## Checks

- CodeRabbit Start Review: no actionable findings
- changed-line `clang-format-19`: clean
- `codespell`: clean
- whitespace/diff checks: clean
- IDE lints: clean
- compilation/link: successful

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DNS node domain configuration handling by reliably applying valid configured domains and retaining the default when invalid values are provided.
  * Added validation for DNS node and domain lengths, including support for valid fully qualified names with a trailing root dot.
  * Improved DNS lookup reliability by preventing malformed or oversized lookup names.
  * Improved consistency when updating and using the DNS node domain during runtime.


apps/app_rpt/rpt_config.c
```diff
@@ -408,8 +407,10 @@ static int node_lookup_bydns(const char *node, char *nodedata, size_t nodedatale
 
 		/* setup the domain to lookup */
 		memset(domain, 0, sizeof(domain));
+		ast_mutex_lock(&dns_node_domain_lock);
 		res = snprintf(domain, sizeof(domain), "_iax._udp.%s.%s", node, rpt_dns_node_domain);
-		if (res < 0) {
+		ast_mutex_unlock(&dns_node_domain_lock);
+		if (res < 0 || (size_t) res >= sizeof(domain) || (size_t) res > MAX_DNS_NODE_DOMAIN_LEN) {
```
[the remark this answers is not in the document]
hardenedpenguin: Fixed in `ad65395`. The runtime presentation-length check now permits 254 characters only when the completed lookup name ends in the root dot, while retaining the 253-character limit for relative names and the separate `snprintf()` truncation check.

apps/app_rpt.c
```diff
@@ -6254,16 +6254,14 @@ static int load_config(int reload)
 
 	cval = ast_variable_retrieve(cfg, "general", "dns_node_domain");
 	if (cval) {
-		if (rpt_is_valid_dns_name(val)) {
-			rpt_dns_node_domain = val;
+		if (rpt_is_valid_dns_name(cval)) {
```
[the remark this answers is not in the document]
hardenedpenguin: Fixed in `ad65395`. `is_valid_dns_node_domain()` now validates the configured domain against the complete worst-case `_iax._udp.<node>.<domain>` presentation length, reserving `MAX_DNS_NODE_LABEL_LEN` before publication. The node-specific runtime length check remains as defense in depth.

apps/app_rpt.c
```diff
@@ -6254,16 +6268,14 @@ static int load_config(int reload)
 
 	cval = ast_variable_retrieve(cfg, "general", "dns_node_domain");
 	if (cval) {
-		if (rpt_is_valid_dns_name(val)) {
-			rpt_dns_node_domain = val;
+		if (is_valid_dns_node_domain(cval)) {
+			dns_node_domain = cval;
```
mkmer, line 6274: This is the issue that needed fixing right?
We should that this in a stand alone PR linked to the issue...
The rest of the changes in here are more of a code quality fix not related (?)

apps/app_rpt/rpt_config.c
```diff
@@ -388,11 +386,12 @@ static int node_lookup_bydns(const char *node, char *nodedata, size_t nodedatale
 {
 	struct ast_dns_result *result;
 	const struct ast_dns_record *record;
-	char domain[256];
+	char domain[sizeof("_iax._udp.") - 1 + MAX_DNS_NODE_LABEL_LEN + 1 + MAX_DNS_NODE_DOMAIN_LEN + 1];
```
mkmer: We have been trying to make this more explicit as to what the magic number mean...
Something like ```MAX_DNS_NODE_LABEL_LEN + sizeof(".") + ...``` or whatever it means

apps/app_rpt/rpt_config.c
```diff
@@ -408,8 +407,13 @@ static int node_lookup_bydns(const char *node, char *nodedata, size_t nodedatale
...
 		/* setup the domain to lookup */
 		memset(domain, 0, sizeof(domain));
+		ast_mutex_lock(&dns_node_domain_lock);
 		res = snprintf(domain, sizeof(domain), "_iax._udp.%s.%s", node, rpt_dns_node_domain);
-		if (res < 0) {
+		ast_mutex_unlock(&dns_node_domain_lock);
+		if (res < 0 || (size_t) res >= sizeof(domain)) {
+			return -1;
+		}
+		if ((size_t) res > MAX_DNS_NODE_DOMAIN_LEN + (domain[res - 1] == '.')) {
```
mkmer, line 421: A comment about why we add 1 or not.
If I get it, we are saying it's OK to truncate a trailing "." ?

apps/app_rpt/app_rpt.h
```diff
@@ -454,6 +454,11 @@ enum patch_call_mode {
 
 /* for DNS resolution of node data */
 #define DEFAULT_DNS_NODE_DOMAIN "nodes.allstarlink.org"
+#define MAX_DNS_NODE_DOMAIN_LEN 253
+#define MAX_DNS_NODE_LABEL_LEN 63
```
mkmer: Since we are tweaking this stuff, should we eliminate the limits by using the `ast_str*` api ???

apps/app_rpt.c
hardenedpenguin: Yes — using `cval` instead of `val` is the actual #1249 bug.

The owned buffer + setter are part of that same fix, not a separate cleanup: once we assign the configured string correctly, keeping a pointer into `cfg` would dangle after `ast_config_destroy()`. The lock and length checks are there so publication and `_iax._udp.<node>.<domain>` construction stay consistent with that owned value.

Happy to split if you’d rather have a minimal “use `cval` + copy” PR and park the presentation-limit hardening, but I wouldn’t ship the assignment change without the owned storage.

apps/app_rpt/rpt_config.c
hardenedpenguin: Agreed. I’ll make the stack buffer composition explicit, e.g. SRV prefix + max node label + `'.'` + max domain + `NUL`, so the `+ 1`s aren’t opaque.

apps/app_rpt/rpt_config.c
hardenedpenguin: Not truncation — the opposite. Relative DNS presentation names max out at 253 characters; an absolute name may be 254 when it ends with the root `'.'`.

So this allows `res == 254` only when `domain` ends in `'.'`; anything longer, or 254 without a trailing root dot, is still rejected. I’ll add a short comment spelling that out.

apps/app_rpt/app_rpt.h
hardenedpenguin: I’d leave the fixed limits here. `253` / `63` (and `254` with a root dot) are DNS presentation/wire constraints, not just local buffer choices — `ast_str` wouldn’t remove the need to enforce them before we query.

Switching the domain storage to `ast_str` could be a follow-up if we want fewer fixed arrays, but it doesn’t change the protocol caps this PR is enforcing.

mkmer: makes sense and no need to change - maybe (for the next guy) add a note / comment about the protocol limit.

apps/app_rpt.c
mkmer: Makes sense and not as stand alone as I originally thought.
Let's keep it in one.

hardenedpenguin: Thank you Mike.

apps/app_rpt/app_rpt.h
hardenedpenguin: Added a short note above the `MAX_DNS_*` macros clarifying these are DNS presentation/wire protocol limits (so `ast_str` would not remove the need to enforce them).

jxmx (approved):

mkmer (approved):

apps/app_rpt/app_rpt.h
```diff
@@ -452,8 +452,20 @@ enum patch_call_mode {
...
+ * MAX_DNS_* values are DNS presentation/wire protocol limits (RFC 1035), not
+ * arbitrary local buffer choices. Absolute names may be 254 with a trailing root '.'.
+ * Using ast_str would not remove the need to enforce these caps before querying.
+ */
 #define DEFAULT_DNS_NODE_DOMAIN "nodes.allstarlink.org"
+#define MAX_DNS_NODE_DOMAIN_LEN 253 /* relative presentation max; absolute may be 254 with root '.' */
+#define MAX_DNS_NODE_LABEL_LEN 63
+/* "_iax._udp." + <node label> + "." + <domain> + NUL */
+#define DNS_SRV_LOOKUP_NAME_BUFSIZE \
+	(sizeof("_iax._udp.") - 1 + MAX_DNS_NODE_LABEL_LEN + sizeof(".") - 1 + MAX_DNS_NODE_DOMAIN_LEN + 1)
```
Allan-N: I'm fairly certain that the "_iax._udp." need to fit in MAX_DNS_NODE_DOMAIN_LEN sized buffer.  You don't get to have a really long name and then query for a name that's even longer.  I think what you want here is :
```
#define DNS_SRV_LABEL_LEN \
	(sizeof("_iax._udp.") - 1 + MAX_DNS_NODE_LABEL_LEN + sizeof(".") - 1)
#define DNS_SRV_LOOKUP_NAME_BUFSIZE \
	(MAX_DNS_NODE_DOMAIN_LEN - DNS_SRV_LABEL_LEN)
```

apps/app_rpt/app_rpt.h
```diff
@@ -452,8 +452,20 @@ enum patch_call_mode {
...
 
-/* for DNS resolution of node data */
+/* for DNS resolution of node data
+ * MAX_DNS_* values are DNS presentation/wire protocol limits (RFC 1035), not
+ * arbitrary local buffer choices. Absolute names may be 254 with a trailing root '.'.
+ * Using ast_str would not remove the need to enforce these caps before querying.
+ */
 #define DEFAULT_DNS_NODE_DOMAIN "nodes.allstarlink.org"
+#define MAX_DNS_NODE_DOMAIN_LEN 253 /* relative presentation max; absolute may be 254 with root '.' */
+#define MAX_DNS_NODE_LABEL_LEN 63
```
Allan-N: Instead of reserving space for a full DNS label where the node string would go I think we can safely assume that node #'s will be <= 10-digits (`sizeof("1234567890") - 1`)

apps/app_rpt/app_rpt.h
hardenedpenguin: Disagree — we’re keeping the RFC 1035 presentation limits here, which is what this code was already aiming at.

`MAX_DNS_NODE_DOMAIN_LEN` (253) / absolute 254 with a root `.` and `MAX_DNS_NODE_LABEL_LEN` (63) are the protocol caps for DNS names and labels. The lookup buffer is sized as the sum of those components (`_iax._udp.` + label + `.` + domain + `NUL`) so `snprintf` is not silently truncating, and we still reject any completed QNAME that exceeds the RFC presentation max at validation and at lookup time.

Shrinking the domain storage to `MAX_DNS_NODE_DOMAIN_LEN - DNS_SRV_LABEL_LEN` would invent an ASL-specific domain ceiling below the RFC max. That’s a different policy than “enforce RFC 1035,” and it’s not what we want this PR to do.

apps/app_rpt/app_rpt.h
hardenedpenguin: Disagree on special-casing a 10-digit node.

RFC 1035 allows a 63-octet label. The DNS path already validates the node as a DNS label up to `MAX_DNS_NODE_LABEL_LEN` (63). Hard-coding `sizeof("1234567890") - 1` would be an AllStar numbering assumption layered on top of the protocol limit, and it would disagree with that label check if a longer (still DNS-valid) label ever appeared.

We’ll keep the RFC 63-octet label reserve.

Allan-N: While I agree that a DNS label has a max size ... and we should most certainly check that the length of a node does not exceed that size ... I do want to call out that we effectively limit the length of a node name that can be queried.

Check out : https://github.com/AllStarLink/app_rpt/blob/b50efc52b87d493e82b7e13eee32562f00b8fee6/apps/app_rpt.c#L423

Also, if we're talking about the length of a DNS label we probably should rename `MAX_DNS_NODE_LABEL_LEN` to be `MAX_DNS_LABEL_LEN`

apps/app_rpt/rpt_config.c
```diff
@@ -388,11 +386,12 @@ static int node_lookup_bydns(const char *node, char *nodedata, size_t nodedatale
...
 	struct ast_dns_result *result;
 	const struct ast_dns_record *record;
-	char domain[256];
+	char domain[DNS_SRV_LOOKUP_NAME_BUFSIZE];
 	int res;
+	size_t node_length = strlen(node);
 
 	/* we require at least a node length of 4 digits */
-	if (strlen(node) < 4) {
+	if (node_length < 4 || node_length > MAX_DNS_NODE_LABEL_LEN) {
```
Allan-N: Instead of comparing the length against `MAX_DNS_NODE_LABEL_LEN`, should we use `rpt_max_dns_node_length`?

apps/app_rpt/app_rpt.h
Allan-N: Your comment says "Absolute names may be 254 with a trailing root '.'".  Don't the query names include the "_iax" and "_udp" labels?

and I made my initial comment because I saw our allocating a buffer larger than the size of an absolute name.

Allan-N (approved):
If my thoughts provoke changes, fine.  Else, what's on the branch is OK.

apps/app_rpt/app_rpt.h
hardenedpenguin: Done in `496b06a`. Domain storage is now `MAX_DNS_NODE_DOMAIN_LEN - DNS_SRV_LABEL_LEN`, and the lookup buffer is sized to the absolute presentation max (254 + NUL) so the full `_iax._udp.<node>.<domain>` QNAME always fits.

apps/app_rpt/app_rpt.h
hardenedpenguin: Done in `496b06a`. Reserve is now `MAX_DNS_NODE_DIGITS` (`sizeof("1234567890") - 1`), `MAX_DNS_NODE_LABEL_LEN` is renamed to `MAX_DNS_LABEL_LEN`, and `max_dns_node_length` is clamped to 10.

apps/app_rpt/rpt_config.c
hardenedpenguin: Done in `496b06a` — `node_lookup_bydns` now bounds against `rpt_max_dns_node_length`.

Allan-N (approved):

PR #1252, opened 2026-09-17
app_rpt.c: No need to remove the audiohook, hangup removes it automatically.

mkmer:
There is no need to remove or detach the audiohook when hangup is involved.
Simply removing the `l` from the list will prevent the `ast_audiohook_write()` from seeing the link.  Simply hanging up detaches the hook.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link cleanup during channel hangup while preserving audiohook resource release.


jxmx:
@mkmer - do we need this now with #1260?

mkmer:
NO, #1260 replaces the need for this.

PR #1257, opened 2026-09-19
chan_simpleusb/usbradio: Track TX queue depth instead of list walks

hardenedpenguin:
## Summary
- Add an O(1) `txq_depth` counter to `chan_simpleusb` (enqueue/dequeue/flush) so the audio and HID paths no longer walk `txq` every tick
- Read `txq_depth` under `txqlock` in the HID PTT path
- Drop the redundant `AST_LIST_TRAVERSE` count in `chan_usbradio` and use its existing `txq_depth`

## Test plan
- [ ] Build `chan_simpleusb.so` / `chan_usbradio.so`
- [ ] Confirm keyed/unkeyed TX audio is clean (no underrun chop)
- [ ] Confirm PTT still keys when frames are queued and drops when the queue drains
- [ ] Pager `?` status still reports `PAGES` / `NOPAGES` correctly

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of queued audio transmissions for more reliable processing.
  - Prevented invalid queued frames from being processed during audio transmission.
  - Improved stream-delay monitoring to maintain consistent behavior under queued transmission loads.
- **Performance**
  - Reduced overhead when checking pending outgoing audio frames, supporting more responsive channel operation.


channels/chan_simpleusb.c
```diff
@@ -2274,7 +2278,13 @@ static void *simpleusb_audio_thread(void *arg)
 					if (num_frames && (num_frames > 3 || (!o->txkeyed && !o->txtestkey))) {
 						ast_mutex_lock(&o->txqlock);
 						f1 = AST_LIST_REMOVE_HEAD(&o->txq, frame_list);
+						if (f1 && o->txq_depth) {
+							o->txq_depth--;
+						}
```
mkmer: We should check for <0 and set to 0 here covering any races on a full reset?
At a minimum we should log an error if it goes <0 helping us "know" things went wrong...

hardenedpenguin: Addressed in `2f9c703`.

On dequeue, if a frame comes off the queue while `txq_depth` is already 0 we now log `LOG_ERROR` and keep the counter at 0 (same on simpleusb and usbradio). That covers a queue/depth desync without letting an unsigned wrap leave a huge depth.

channels/chan_simpleusb.c
```diff
@@ -200,6 +200,7 @@ struct chan_simpleusb_pvt {
 	/* queue used to hold packets to transmit */
 	AST_LIST_HEAD_NOLOCK(, ast_frame) txq;
 	ast_mutex_t txqlock;
+	unsigned int txq_depth; /* frames queued; updated under txqlock */
```
mkmer: Structure here for all 3?  They mostly live together?
Then maybe `txqlock` = `lock` and `txq_depth` = `depth` in the structure?

hardenedpenguin: Done in `c362ec6`.

`txq` is now a struct owning `list` / `lock` / `depth` on both channels (usbradio also keeps `high_water` in there), so call sites use `o->txq.lock` and `o->txq.depth`.

channels/chan_usbradio.c
```diff
@@ -5311,7 +5313,7 @@ static struct chan_usbradio_pvt *store_config(struct ast_config *cfg, const char
 	ast_mutex_init(&o->eepromlock);
 	ast_mutex_init(&o->usblock);
 	ast_mutex_init(&o->device_lock);
-	ast_mutex_init(&o->txqlock);
+	ast_mutex_init(&o->txq.lock);
```
mkmer, line 5352: These are a bit interesting - we don't destroy them on unload. Probably should?
In a new PR.

mkmer (approved):

channels/chan_usbradio.c
hardenedpenguin: Agreed — worth cleaning up on unload, but I’ll leave mutex destroy for a follow-up PR.

channels/chan_simpleusb.c
```diff
@@ -1348,7 +1351,9 @@ static void *hidthread(void *arg)
 				o->rxhidctcss = ctcssed;
 			}
 
-			txreq = !(AST_LIST_EMPTY(&o->txq));
+			ast_mutex_lock(&o->txq.lock);
```
Allan-N: What's the point of locking/unlocking when the value can be stale right after we unlock?

channels/chan_simpleusb.c
```diff
@@ -2244,12 +2253,9 @@ static void *simpleusb_audio_thread(void *arg)
 				long frames_available;
 
 				num_frames = 0;
-				ast_mutex_lock(&o->txqlock);
-				AST_LIST_TRAVERSE(&o->txq, f1, frame_list) {
-					num_frames++;
-				}
-
-				ast_mutex_unlock(&o->txqlock);
+				ast_mutex_lock(&o->txq.lock);
```
Allan-N: What's the point of locking/unlocking when the value can be stale right after we unlock?

channels/chan_simpleusb.c
```diff
@@ -2464,12 +2481,9 @@ static void *simpleusb_audio_thread(void *arg)
...
 			 */
 			if (o->waspager) {
-				num_frames = 0;
-				ast_mutex_lock(&o->txqlock);
-				AST_LIST_TRAVERSE(&o->txq, f1, frame_list) {
-					num_frames++;
-				}
-				ast_mutex_unlock(&o->txqlock);
+				ast_mutex_lock(&o->txq.lock);
+				num_frames = (int) o->txq.depth;
```
Allan-N: What's the point of locking/unlocking when the value can be stale right after we unlock?

channels/chan_usbradio.c
```diff
@@ -2345,12 +2350,9 @@ static void *usbradio_audio_thread(void *arg)
 			 * claims OK but the hardware TX buffer is not draining, txq backs up
 			 * and MAX_FRAME_DELAY restarts the stream (same idea as simpleusb).
 			 */
-			num_frames = 0;
-			ast_mutex_lock(&o->txqlock);
-			AST_LIST_TRAVERSE(&o->txq, f1, frame_list) {
-				num_frames++;
-			}
-			ast_mutex_unlock(&o->txqlock);
+			ast_mutex_lock(&o->txq.lock);
```
Allan-N: What's the point of locking/unlocking when the value can be stale right after we unlock?

hardenedpenguin: Fair point — locking only long enough to sample `depth`, then unlocking before we act, doesn’t make the following decision race-free (same staleness we had with the old locked list walk).

Two ways we could go:

1. Hold `txq.lock` across the depth check and the decision that depends on it.
2. Drop the lock on these naked `depth` reads and treat them as best-effort, keeping the lock only on enqueue/dequeue that mutate the queue.

Which do you prefer?

Allan-N: I did not study the code following where we acquire the value.  If the behavior of the following code is dependent on the value then we should hold the lock across the entire operation.  Given that the lock looks to be for keeping the enqueue/dequeue/counts in sync and becuase we're not messing with the queue or count I think we'd be safe to follow option #2.

hardenedpenguin: Went with option #2 in `c62fe01` — dropped the lock on those advisory `txq.depth` snapshots; enqueue/dequeue still take `txq.lock`.

FYI: a follow-up CodeRabbit pass flagged the unlocked reads and suggested locking (or atomics) again. We’re intentionally keeping your option #2 here rather than re-adding a lock that doesn’t cover the decision that follows.

Allan-N (approved):

channels/chan_usbradio.c
mkmer: Isn't it possible to interrupt a non atomic add? Then see it partially added or subtracted?  I received those same warnings on my first swing at this and went with atomic vars....

channels/chan_simpleusb.c
```diff
@@ -1348,7 +1351,8 @@ static void *hidthread(void *arg)
 				o->rxhidctcss = ctcssed;
 			}
 
-			txreq = !(AST_LIST_EMPTY(&o->txq));
+			/* Best-effort snapshot; lock only protects enqueue/dequeue/depth updates. */
+			txreq = o->txq.depth != 0;
```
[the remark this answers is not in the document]
hardenedpenguin: Addressed in `523055d`.

`txq.depth` (and usbradio `high_water`) now go through `__atomic_*` helpers. Advisory reads stay lock-free; `txq.lock` still covers list enqueue/dequeue only. That keeps Allan’s option #2 and closes the non-atomic race Greptile flagged.

channels/chan_usbradio.c
hardenedpenguin: Follow-up in `523055d`: made `depth`/`high_water` atomic (`__atomic_*`), same concern Mike raised. Unlocked advisory snapshots stay best-effort for the *decision*, but the counter load/store itself is no longer a torn non-atomic race.

channels/chan_simpleusb.c
```diff
@@ -398,6 +401,34 @@ static int __attribute__((format(printf, 3, 4))) simpleusb_log_fault(struct chan
 	return 1;
 }
 
+/* txq.depth: atomic so advisory reads may omit txq.lock; list mutations still take the lock. */
+static unsigned int simpleusb_txq_depth_get(const struct chan_simpleusb_pvt *o)
+{
+	return __atomic_load_n(&o->txq.depth, __ATOMIC_RELAXED);
```
Allan-N: Shouldn't we be using asterisk's atomic functions (include/asterisk/lock.h) ?

channels/chan_simpleusb.c
```diff
@@ -398,6 +401,34 @@ static int __attribute__((format(printf, 3, 4))) simpleusb_log_fault(struct chan
...
+{
+	__atomic_add_fetch(&o->txq.depth, 1, __ATOMIC_RELAXED);
+}
+
+static void simpleusb_txq_depth_dec(struct chan_simpleusb_pvt *o)
+{
+	unsigned int depth = __atomic_load_n(&o->txq.depth, __ATOMIC_RELAXED);
+
+	if (depth) {
+		__atomic_store_n(&o->txq.depth, depth - 1, __ATOMIC_RELAXED);
```
Allan-N: Ummm... How can this be an atomic decrement when you are using 2 function calls?

channels/chan_simpleusb.c
mkmer: Ditto

mkmer: `ast_atomic_fetchadd_int()`: Atomically adds a value to an integer and returns the previous value.
`ast_atomic_dec_and_test()`: Atomically decrements a variable and checks if it reaches zero (frequently used for internal object reference counting).
`ast_atomic_sub_fetch()`: Atomically subtracts a value and returns the new value.

There are others in there that may fit your needs

hardenedpenguin: Good catch — switched to `ast_atomic_*` from `lock.h` in `3be90fa` (`fetch_add` / `add_fetch` / `fetch_sub` / `and_fetch`). Also dropped `const` on the getters in `9126f4b` so Clang accepts the mutable pointer those macros need.

channels/chan_simpleusb.c
hardenedpenguin: You’re right — that load/store pair wasn’t a single atomic op. Fixed in `3be90fa` to use `ast_atomic_fetch_sub` for the decrement (underflow still clamps back to 0 with a `LOG_ERROR`).

channels/chan_simpleusb.c
```diff
@@ -398,6 +401,32 @@ static int __attribute__((format(printf, 3, 4))) simpleusb_log_fault(struct chan
...
+}
+
+static void simpleusb_txq_depth_inc(struct chan_simpleusb_pvt *o)
+{
+	ast_atomic_add_fetch(&o->txq.depth, 1, __ATOMIC_RELAXED);
+}
+
+static void simpleusb_txq_depth_dec(struct chan_simpleusb_pvt *o)
+{
+	unsigned int prev = ast_atomic_fetch_sub(&o->txq.depth, 1, __ATOMIC_RELAXED);
```
mkmer, line 417: `ast_atomic_dec_and_test()` ?

channels/chan_simpleusb.c
```diff
@@ -398,6 +401,32 @@ static int __attribute__((format(printf, 3, 4))) simpleusb_log_fault(struct chan
...
+
+	if (!prev) {
+		ast_log(LOG_ERROR, "Channel %s: txq_depth underflow (queue/depth desync)\n", o->name);
+		ast_atomic_and_fetch(&o->txq.depth, 0, __ATOMIC_RELAXED);
```
Allan-N: Here, you are setting the value to zero ... but the value may have changed since it was decremented.  I'm thinking that the fix is to just increment the value by 1 (restoring the previous value of 0).

channels/chan_usbradio.c
```diff
@@ -511,6 +513,48 @@ static int __attribute__((format(printf, 3, 4))) usbradio_log_fault(struct chan_
...
+
+	if (!prev) {
+		ast_log(LOG_ERROR, "Channel %s: txq_depth underflow (queue/depth desync)\n", o->name);
+		ast_atomic_and_fetch(&o->txq.depth, 0, __ATOMIC_RELAXED);
+	}
+}
+
+static void usbradio_txq_counters_clear(struct chan_usbradio_pvt *o)
+{
+	ast_atomic_and_fetch(&o->txq.depth, 0, __ATOMIC_RELAXED);
```
Allan-N: Two separate atomic operations are not, together, atomic.  Here, I think you need to lock/unlock.

mkmer: Boy this is more complicated than it first appears.
Maybe we just add a mutex for math, keep the mutex for lists (`lock_list` and `lock_depth`? or something along that line)
Maybe?

channels/chan_simpleusb.c
hardenedpenguin: Good catch — forcing 0 after a wrapping `fetch_sub` could clear a concurrent enqueue. Fixed in `8a02457`: on underflow we undo with `ast_atomic_add_fetch(..., 1)` instead (same change on usbradio).

channels/chan_usbradio.c
hardenedpenguin: Agreed — those two clears aren’t one atomic pair on their own. `flush_tx_queue` (and the high-water reset path) already hold `txq.lock`; in `8a02457` the helper now does plain stores of both counters under that caller-held lock so they clear as one critical section. Same idea for the high-water peak update on enqueue (already under `txq.lock`).

channels/chan_simpleusb.c
```diff
@@ -198,8 +198,11 @@ struct chan_simpleusb_pvt {
 	struct ast_frame read_f;											 /* returned by simpleusb_read */
 
 	/* queue used to hold packets to transmit */
-	AST_LIST_HEAD_NOLOCK(, ast_frame) txq;
-	ast_mutex_t txqlock;
+	struct {
+		AST_LIST_HEAD_NOLOCK(, ast_frame) list;
```
mkmer, line 202: Maybe we should use `AST_LIST_HEAD(, ast_frame) list;`
And use the locks created  with:
```
AST_LIST_LOCK(&list);
...
AST_LIST_UNLOCK(&list);
``` 
???

And a separate lock for the depth?

Allan-N (approved):
Looks good!

channels/chan_simpleusb.c
hardenedpenguin: Good thought. For this PR I’d rather keep the current shape: one lock for list mutations (`AST_LIST_HEAD_NOLOCK` + `txq.lock`) and `ast_atomic_*` for depth so advisory unlocked reads stay cheap without a second mutex. A dedicated depth lock (or switching to `AST_LIST_HEAD` + `AST_LIST_LOCK`) would mostly reshuffle the same serialization and risks lock-order complexity on enqueue/dequeue. Happy to revisit a list-lock idiom cleanup in a follow-up if you still want that separately.

mkmer: That's fine, I don't think we want to mess with the lock version as we don't use it anywhere else at this point.

mkmer (approved):

mkmer:
Well, I think we may have made this WAY more complicated that it needs to be (probably partly due to some of my comments).

Let's look at this one more time:
When we load things into the queue, we lock and add to the depth. (Safe)
https://github.com/hardenedpenguin/app_rpt/blob/8a02457b8f2c97de6d573457603f8ac44a526446/channels/chan_simpleusb.c#L1947-L1950
and
https://github.com/hardenedpenguin/app_rpt/blob/8a02457b8f2c97de6d573457603f8ac44a526446/channels/chan_simpleusb.c#L2092-L2095

Then when we remove from the queue, we lock and subtract from depth (safe):
https://github.com/hardenedpenguin/app_rpt/blob/8a02457b8f2c97de6d573457603f8ac44a526446/channels/chan_simpleusb.c#L2305-L2310

The only "unsafe" part was looking at the value of depth, which should be copied locked (I think as you had it) so that we don't snag a partially operated on variable, then unlock and use the copy.  Yes it's "stale" but it was right for the moment we snagged it and it will be updated next time around.

Only possible suggestion around it would be to make a function for accessing the `depth` that locks, copies, unlocks and returns the value?

Maybe make a insert / remove function that has the ++ and -- and locking integrated? (Maybe)

I don't think we really need any atomic magic here do we?

hardenedpenguin:
Fair point — looking at it again, enqueue/dequeue already serialize depth under `txq.lock`, so the atomics are doing more than we need. I’m fine dropping them and going back to a plain depth counter, with a small helper that locks, copies, unlocks, and returns the snapshot for the advisory reads. Optional insert/remove helpers that own the lock + `++`/`--` are a nice cleanup if we want them in the same pass; otherwise we can keep those call sites as they are.

Allan-N:
Another titbit to ponder : on my last pass I noticed paired calls to insert/remove from the list and increment/decrement the counters.  One might think about having helpers perform both operations, maybe even including the lock/unlock.

and, of course, we can iterate more and more and more.  When do we stop and say "enough" :)

PR #1258, opened 2026-09-19
app_rpt: Track remrx with edge counters instead of link scans

hardenedpenguin:
## Summary
- Maintain `remrx_links` / `remrx_links_txable` / `voteremrx_links` on `lastrx` and mode changes via `rpt_link_set_lastrx` / `rpt_link_set_mode`
- Replace per-link and main-loop remrx link walks with O(1) counter checks
- Clear lastrx from aggregates before unlink; keep EchoLink/TLB mode changes on the counter path

## Test plan
- [ ] Build `app_rpt.so`
- [ ] Multi-link node: key one remote and confirm local TX/COS and telemetry still reflect remrx
- [ ] Key a second MONITOR/TRANSCEIVE link and confirm the first link still sees "other remrx"
- [ ] Disconnect a keyed link and confirm remrx clears
- [ ] EchoLink/TLB mode change while keyed still updates txable remrx behavior

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved accuracy when tracking linked radio receive activity and voter status.
  - Corrected transmit-capability tracking when a linked connection changes mode.
  - Ensured activity counters remain accurate as links connect, disconnect, or are removed.
  - Improved handling of link activity during reconnects and periodic processing.

- **Performance**
  - Reduced repeated scanning during link activity processing, improving efficiency in configurations with multiple links.


hardenedpenguin:
@coderabbitai review

apps/app_rpt.c
```diff
@@ -3565,7 +3565,11 @@ static inline int periodic_process_link(struct rpt *myrpt, struct rpt_link *l, c
 		if (!l->voxtostate)
 			myrx = myrx || l->wasvox;
 	}
-	l->lastrx = myrx;
+	if (l->lastrx != !!myrx) {
```
mkmer: `!!myrx` ?

apps/app_rpt.c
```diff
@@ -5137,6 +5135,7 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 	}
 	rpt_mutex_lock(&myrpt->lock);
 	ao2_ref(l, +1);					  /* prevent freeing while we finish up */
+	rpt_link_set_lastrx(myrpt, l, 0); /* drop from remrx aggregates before unlink */
 	rpt_link_remove(myrpt->links, l); /* remove from queue */
```
mkmer, line 5139: Should we handle this cleanup in the remove from list? `rpt_link_remove()` should always clear the lastrx right?

apps/app_rpt.c
```diff
@@ -5586,18 +5584,8 @@ static void *rpt(void *this)
...
-				if (l->voterlink)
-					myrpt->voteremrx = 1;
-				if (IS_NODE_EXTEN(l->name)) { /* Ignore '0' nodes */
-					ast_copy_string(myrpt->lastnodewhichkeyedusup, l->name, sizeof(myrpt->lastnodewhichkeyedusup));
-				}
-			}
-		}
-		ao2_iterator_destroy(&l_it);
+		myrpt->remrx = myrpt->remrx_links != 0;
+		myrpt->voteremrx = myrpt->voteremrx_links != 0;
```
mkmer: Should we test for >0 so that <0 problems will not keep us keyed?

apps/app_rpt/rpt_link.c
```diff
@@ -433,6 +433,65 @@ void rpt_link_remove(struct ao2_container *links, struct rpt_link *l)
...
+	/* Caller must hold myrpt->lock. */
+	if (!myrpt || !l || l->mode == mode) {
+		return;
+	}
+
+	was_txable = l->mode < MODE_LOCAL_MONITOR;
+	now_txable = mode < MODE_LOCAL_MONITOR;
+	if (l->lastrx && was_txable != now_txable) {
+		if (was_txable && myrpt->remrx_links_txable) {
+			myrpt->remrx_links_txable--;
```
mkmer: Test this does not go <0 and at a minimum log that it happened, bound it to 0?

apps/app_rpt/rpt_link.c
```diff
@@ -433,6 +433,65 @@ void rpt_link_remove(struct ao2_container *links, struct rpt_link *l)
...
+		}
+		if (l->voterlink) {
+			myrpt->voteremrx_links++;
+		}
+		if (IS_NODE_EXTEN(l->name)) {
+			ast_copy_string(myrpt->lastnodewhichkeyedusup, l->name, sizeof(myrpt->lastnodewhichkeyedusup));
+		}
+	}
+	myrpt->remrx = myrpt->remrx_links != 0;
+	myrpt->voteremrx = myrpt->voteremrx_links != 0;
```
mkmer: >0 so <0 errors don't signal `remrx` and `voterremrx`

apps/app_rpt/rpt_link.c
```diff
@@ -433,6 +433,65 @@ void rpt_link_remove(struct ao2_container *links, struct rpt_link *l)
...
+
+	if (l->lastrx) {
+		if (myrpt->remrx_links) {
+			myrpt->remrx_links--;
+		}
+		if (l->mode < MODE_LOCAL_MONITOR && myrpt->remrx_links_txable) {
+			myrpt->remrx_links_txable--;
+		}
+		if (l->voterlink && myrpt->voteremrx_links) {
+			myrpt->voteremrx_links--;
```
mkmer: Test this does not go <0 and at a minimum log that it happened, bound it to 0?

apps/app_rpt/rpt_link.c
```diff
@@ -433,6 +433,65 @@ void rpt_link_remove(struct ao2_container *links, struct rpt_link *l)
...
+	if (!myrpt || !l || l->lastrx == keyed) {
+		return;
+	}
+
+	if (l->lastrx) {
+		if (myrpt->remrx_links) {
+			myrpt->remrx_links--;
+		}
+		if (l->mode < MODE_LOCAL_MONITOR && myrpt->remrx_links_txable) {
+			myrpt->remrx_links_txable--;
```
mkmer: Test this does not go <0 and at a minimum log that it happened, bound it to 0?

apps/app_rpt/rpt_link.c
```diff
@@ -433,6 +433,65 @@ void rpt_link_remove(struct ao2_container *links, struct rpt_link *l)
...
+	int keyed = !!rx;
+
+	/* Caller must hold myrpt->lock. */
+	if (!myrpt || !l || l->lastrx == keyed) {
+		return;
+	}
+
+	if (l->lastrx) {
+		if (myrpt->remrx_links) {
+			myrpt->remrx_links--;
```
mkmer: Test this does not go <0 and at a minimum log that it happened, bound it to 0?

apps/app_rpt/app_rpt.h
```diff
@@ -949,6 +949,12 @@ struct rpt {
 	rpt_bool rxchankeyed:1; /*!< \brief Receiver RxChan Key State */
 	rpt_bool localtx:1;
 	rpt_bool remrx:1;
+	/*! \brief Count of links with lastrx currently true (drives remrx). */
+	unsigned int remrx_links;
+	/*! \brief Count of links with lastrx and mode MONITOR/TRANSCEIVE (mode < LOCAL_MONITOR). */
+	unsigned int remrx_links_txable;
+	/*! \brief Count of voter links with lastrx (drives voteremrx). */
+	unsigned int voteremrx_links;
```
mkmer, line 957: These should be signed, or we need to make sure >0 when we to -- so that we don't rollover to a huge number?

apps/app_rpt/app_rpt.h
```diff
@@ -949,6 +949,12 @@ struct rpt {
 	rpt_bool rxchankeyed:1; /*!< \brief Receiver RxChan Key State */
 	rpt_bool localtx:1;
 	rpt_bool remrx:1;
+	/*! \brief Count of links with lastrx currently true (drives remrx). */
+	unsigned int remrx_links;
```
mkmer, line 953: signed, or check for >0 on all -- operations?

mkmer (commented):
Some overlapping comments about counters <0 or subtracting while at 0 - while it shouldn't happen, we need to log that it happened and test to be sure we don't roll over if unsigned or go negative is signed (depending on the resolution).

apps/app_rpt.c
hardenedpenguin: Addressed in `218fcba`.

`rpt_link_remove()` now takes `myrpt` and calls `rpt_link_set_lastrx(myrpt, l, 0)` before unlink, so keyed links are always dropped from the remrx aggregates at remove time.

apps/app_rpt.c
hardenedpenguin: Addressed in `218fcba`.

Both places that publish the flags now use `remrx_links > 0` / `voteremrx_links > 0`, so a prior underflow (after clamp-to-zero) cannot leave remrx latched.

apps/app_rpt/rpt_link.c
hardenedpenguin: Addressed in `218fcba`.

All remrx aggregate decrements go through `remrx_count_dec()`, which logs `LOG_ERROR` and clamps to 0 if the counter is already zero (covers this mode-change path too).

apps/app_rpt/rpt_link.c
hardenedpenguin: Addressed in `218fcba`.

`remrx` / `voteremrx` are now set from `count > 0` in `rpt_link_set_lastrx()` as well as the periodic refresh in `rpt()`.

apps/app_rpt/rpt_link.c
hardenedpenguin: Addressed in `218fcba` via `remrx_count_dec()` — log + clamp to 0 on underflow.

apps/app_rpt/rpt_link.c
hardenedpenguin: Addressed in `218fcba` via `remrx_count_dec()` — log + clamp to 0 on underflow.

apps/app_rpt/rpt_link.c
hardenedpenguin: Addressed in `218fcba` via `remrx_count_dec()` — log + clamp to 0 on underflow.

apps/app_rpt/app_rpt.h
hardenedpenguin: Kept them unsigned (they are counts) and guarded every `--` with `remrx_count_dec()` in `218fcba`, so we cannot wrap to a huge value. Happy to switch to signed if you’d rather see negatives in the log before clamp.

apps/app_rpt/app_rpt.h
hardenedpenguin: Same as above — unsigned counts with `> 0` checks on all decrements in `218fcba`, plus `LOG_ERROR` if a decrement is attempted at zero.

apps/app_rpt.c
mkmer: Maybe that was a bad call on my part? 
OR now that we pass `myrpt` why pass `myrpt->links`?.  I'm leaning toward your original separate code, but it's not clear to me which way is "right"...
I think we would want to keep `rpt_link_add()` and `rpt_link_remove()` the "same"....

hardenedpenguin: Agreed — restored in `f0ede2d`.

`rpt_link_remove()` is back to a plain container unlink matching `rpt_link_add()`, and the remrx clear stays at the call site:

```c
rpt_link_set_lastrx(myrpt, l, 0);
rpt_link_remove(myrpt->links, l);
```

mkmer (approved):

apps/app_rpt/rpt_link.c
```diff
@@ -427,12 +427,81 @@ void rpt_link_add(struct ao2_container *links, struct rpt_link *l)
 	ao2_link(links, l);
 }
 
+/*! Decrement an remrx aggregate; log and clamp if already zero (unsigned wrap guard). */
```
Allan-N: It seems unbalanced to use a function to decrement but a direct operation to increment.  Maybe add `remrx_count_inc()`?

... or have a single `remrx_ref()` function (think `ao2_ref(xxx, +1)`, `ao2_ref(xxx, -1)`)

Allan-N (approved):

apps/app_rpt/rpt_link.c
hardenedpenguin: Done in `2541bcf` — increments and decrements now both go through `remrx_count_ref(..., ±1, ...)`.

apps/app_rpt.c
Allan-N: `!!myrx` ??

PR #1259, opened 2026-09-19
app_rpt: Track simplex delay queue depth instead of list walks

hardenedpenguin:
## Summary
- Add shared `rpt_framelist_*` helpers with O(1) depth counters for simplex delay FIFOs
- Typedef `rpt_framelist_t` in `app_rpt.h` and use it for `rpt`/`rpt_link` delay queues
- Replace per-frame list walks in `localtxchannel_read`, link phonevox, and `exec_chan_read`
- Free the input frame and continue (return 0) if silent padding allocation fails

## Test plan
- [ ] Build `app_rpt.so`
- [ ] Simplex duplex &lt; 2: autopatch delay still sounds correct when keyed
- [ ] Phone/vox link path: simplex phone delay still applies; no hang on alloc stress
- [ ] Remote phone vox path continues after a pad failure (no spurious hangup)
- [ ] Full-duplex path still flushes `txq` cleanly

## Summary by CodeRabbit

* **Performance**
  * Improved voice frame queue processing for simplex and phone delay audio paths, reducing overhead when managing queued frames.

* **Reliability**
  * Improved handling of queue padding and allocation failures to prevent stale audio frames from being retained.
  * Voice queues now track their depth consistently, improving empty-queue and non-empty-queue handling.


apps/app_rpt.c
```diff
@@ -4027,6 +4027,67 @@ static inline void rpt_frame_queue_free(struct rpt_frame_queue *frame_queue)
...
+}
+
+/*!
+ * \brief Pad FIFO with muted copies of template until depth reaches target.
+ * \retval 0 success (or already deep enough)
+ * \retval -1 allocation failure (partial pad may remain)
+ */
+static inline int rpt_framelist_pad_silent(rpt_framelist_t *q, unsigned int *depth, struct ast_frame *template, unsigned int target)
+{
+	while (*depth < target) {
```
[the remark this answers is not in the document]
hardenedpenguin: Addressed in `57cfbfb`.

`simplexpatchdelay` and `simplexphonedelay` now use `RPT_CONFIG_VAR_INT_DEFAULT_MIN_MAX(..., 0, 1000)` so negative values cannot become huge unsigned pad targets in `rpt_framelist_pad_silent`.

apps/app_rpt/app_rpt.h
```diff
@@ -550,6 +550,9 @@ struct rpt_frame_queue {
 	struct ast_frame *lastf1, *lastf2;
 };
 
+/*! Voice-frame FIFO used for simplex delay queues (with parallel depth counters). */
+typedef AST_LIST_HEAD_NOLOCK(, ast_frame) rpt_framelist_t;
```
mkmer: Is this necessary, when it's the same list in other locations?  
Not sure it's helpful, other's input welcome.

apps/app_rpt.c
```diff
@@ -4027,6 +4027,67 @@ static inline void rpt_frame_queue_free(struct rpt_frame_queue *frame_queue)
...
+	AST_LIST_INSERT_TAIL(q, f, frame_list);
+	(*depth)++;
+}
+
+static inline struct ast_frame *rpt_framelist_dequeue(rpt_framelist_t *q, unsigned int *depth)
+{
+	struct ast_frame *f = AST_LIST_REMOVE_HEAD(q, frame_list);
+
+	if (f && *depth) {
+		(*depth)--;
```
mkmer: We should check <0, at a minimum log it happened.

apps/app_rpt.c
```diff
@@ -4027,6 +4027,67 @@ static inline void rpt_frame_queue_free(struct rpt_frame_queue *frame_queue)
...
+}
+
+static inline void rpt_framelist_flush(rpt_framelist_t *q, unsigned int *depth)
+{
+	struct ast_frame *f;
+
+	while ((f = AST_LIST_REMOVE_HEAD(q, frame_list))) {
+		ast_frfree(f);
+	}
+	*depth = 0;
```
mkmer: should we make sure the pointer is not null? (at all uses of `depth`)

apps/app_rpt/app_rpt.h
hardenedpenguin: Kept it in `2e8e71e`, with a short comment clarifying why.

The typedef is only for the counted delay FIFOs so the shared helpers stay readable (`rpt_framelist_t *q`). Lists that do not carry a depth counter (e.g. `textq`) still use the expanded `AST_LIST_HEAD_NOLOCK` form. Happy to drop the alias if you’d rather see the expanded type everywhere.

apps/app_rpt.c
hardenedpenguin: Addressed in `2e8e71e`.

Depth is unsigned, so the guard is on dequeue: if a frame comes off the list while `*depth` is already 0 we log `LOG_ERROR` and clamp to 0 (same idea as the USB `txq_depth` fix).

apps/app_rpt.c
hardenedpenguin: Addressed in `2e8e71e`.

All shared helpers (`enqueue` / `dequeue` / `flush` / `pad_silent` / `take_or_mute`) now null-check `q`/`depth` (and the frame/template where applicable), log, and no-op or return failure instead of dereferencing.

apps/app_rpt/app_rpt.h
mkmer: Should we make structures out of the frame variable that include the depth?  Then we would just pass in the structure and reference the depth in there?  Seems like these "require" the depth to use them.

hardenedpenguin: Good call — done in `1f5ec7e`.

`rpt_framelist_t` is now a struct owning both the list and depth, helpers take a single pointer, and the empty checks use the O(1) `.depth` field instead of walking the list.

apps/app_rpt.c
```diff
@@ -4028,70 +4028,70 @@ static inline void rpt_frame_queue_free(struct rpt_frame_queue *frame_queue)
...
-	while (*depth < target) {
+	while (fl->depth < target) {
 		struct ast_frame *f1 = ast_frdup(template);
 
 		if (!f1) {
 			return -1;
 		}
 		RPT_MUTE_FRAME(f1);
-		rpt_framelist_enqueue(q, depth, f1);
+		rpt_framelist_enqueue(fl, f1);
```
mkmer: Not sure I like `fl` and `f1` in the same world like this.  It's sure to confuse someone (mostly me)
Probably ok just to use `f` in there?

hardenedpenguin: Done in `3160de2` — local is now `f`. Also rebased onto current `master` so the PR merges cleanly with `mix_altaudio`.

mkmer (approved):

Allan-N (approved):

Issue #1240, opened 2026-09-14
Crash / Runaway CPU with 491d9c0fcc236d54b75124917445e0d1209879c9


jxmx:
Trying to test #1229 and built packages at commit 491d9c0fcc236d54b75124917445e0d1209879c9. At start of asterisk I get the below and the CPU balloons to 200% (full) usage.

```
[2026-09-13 19:49:22.032] Asterisk 22.11.0+asl3-3.10.5.1-j1.deb13 built by builder @ allstarlink.org on a x86_64 running Linux on 2026-09-13 23:29:13 UTC
[2026-09-13 19:49:22.042] NOTICE[337105] loader.c: 70 modules will be loaded.
[2026-09-13 19:49:22.062] NOTICE[337105] cdr.c: CDR simple logging enabled.
[2026-09-13 19:49:22.062] NOTICE[337105] dnsmgr.c: Managed DNS entries will be refreshed every 300 seconds.
[2026-09-13 19:49:22.066] NOTICE[337105] indications.c: Default country for indication tones: us
[2026-09-13 19:49:22.066] NOTICE[337105] indications.c: Setting default indication country to 'us'
[2026-09-13 19:49:22.088] NOTICE[337105] res_smdi.c: No SMDI interfaces are available to listen on, not starting SMDI listener.
[2026-09-13 19:49:22.131] WARNING[337105] loader.c: Module 'res_adsi' has been loaded but may be removed in a future release.
[2026-09-13 19:49:22.146] WARNING[337108] pbx_config.c: users.conf is deprecated and will be removed in a future version of Asterisk
[2026-09-13 19:49:22.349] NOTICE[337146] app_rpt.c: Domain used for DNS node lookup is: nodes.allstarlink.org
[2026-09-13 19:49:22.350] NOTICE[337146] app_rpt.c: Normal Repeater Init  48496
[2026-09-13 19:49:22.350] NOTICE[337146] app_rpt.c: Normal Repeater Init  430780
[2026-09-13 19:49:22.351] NOTICE[337146] app_rpt.c: Normal Repeater Init  430781
[2026-09-13 19:49:22.353] NOTICE[337146] app_rpt.c: Normal Repeater Init  430782
[2026-09-13 19:49:22.354] NOTICE[337146] app_rpt.c: Normal Repeater Init  430783
[2026-09-13 19:49:22.358] NOTICE[337146] app_rpt.c: Normal Repeater Init  430784
[...]

PR #1260, opened 2026-09-19
rpt_link: Fix/link whisper framehook spin

jxmx:
Attaching a whisper audiohook now installs Asterisk's whisper framehook, which opens a timer and adds its fd to the hooked channel (upstream 29f769ef). l->altaudio on l->pchan is our only whisper hook, so every ilink connection gives process_link_channel() a third, level triggered fd that only the framehook acknowledges, and only when ast_channel_fdno() matches its own slot. ast_waitfor_nandfds() records one fd per channel per wakeup and lets later channels win, so with pchan first in cs[] the tick went unacked, the fd stayed readable and the link thread free-ran at 100% CPU. DAHDI and pthread timing make it permanent by clearing one expiration per ack, hence the system dependence.

Poll l->pchan last so it wins the tie, and skip the audiohook entirely on links that can never be an altlink() - for those the hook is attached, never fed and never mixed. Also treat a NULL ast_waitfor_n() return as a timeout the way rpt() already does.

The framehook's own frame leak, which trips the ao2 excessive refcount assertion in __ast_frdup(), still needs an Asterisk side fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary CPU usage during linked-channel processing.
  * Prevented audio processing overhead and resource leakage for links that cannot support alternate-link behavior.
  * Improved timeout handling to prevent processing loops from spinning on stale timing values.


jxmx:
Original prompt:

There is a bug in the current code within app_rpt. When Asterisk starts, the app_rpt.so module starts fine. When an ilink connection is made on certain systems the CPU spikes to 100%. It eventually crashes with: {omitted}. I can see when we make a connection it's the newest thread that goes to 99.7% -> so I think it points to our process_link_channel() thread doing something crazy. Something wierd is happening in the link thread causing it to use so much CPU.  It should be waiting at who = ast_waitfor_n(cs, n, &ms); looping around 2 times per 20ms. 1 frame from each channel. the waitfor_n function is supposed to trigger on a new frame OR 20ms (max) -> so if we are spinning, either the 20ms is "wrong" or we are not clearing the fd it's supposed to trigger on. We only look at 2: l->pchan and l->chan , if there is activity from the specific one, we read the channel and "do stuff" -> the read is supposed to clear the activity indicator. How does process_link_channel() could loop without waiting utilizing 100% cpu. Or is there something else you spot?

jxmx:
Opus Response #1:

hort answer: the reorder fixes the spin, not the crash. It turns a crash-in-seconds into a crash-in-tens-of-minutes. And the who == NULL guard does nothing for this bug at all — in the spin state who is always non-NULL (it's l->chan). I flagged it as hygiene and should have been clearer that it's unrelated.

Here's the arithmetic, after checking two things I'd left open.

What the reorder actually buys you:
ast_unreal_read() unconditionally returns &ast_null_frame (core_unreal.c:313), so the dropped-frame leak only bites when a real frame was sitting in pchan's readq at the moment the timer won the race. The dup leak, though, fires on every injection — and the injection is skipped only if a non-timer voice write hit pchan within the last 40ms (last_external_write_tv, audiohook.c:194-199). app_rpt writes to pchan only when the link is delivering audio, so a quiet link leaks at the full timer rate, ~50 frames/sec; a link passing audio leaks ~0.

Spinning (today): leaks at loop rate, ~10⁴–10⁵/sec → 100,000 refs in seconds. Matches your report exactly.
After the reorder: leaks at 50/sec per quiet link. 100,000 ÷ 50 ≈ 33 minutes with one quiet link, ~8 minutes with four. Same FRACK, same log storm, just later. (~50 MB of leaked frames by then, so the refcount ceiling trips before memory becomes the issue on most boxes.)
So the reorder is worth doing — a pegged core is its own outage — but it is not a workaround for the crash.

The app_rpt change that does eliminate it for ilinks
altlink() (rpt_link.c:81) short-circuits on values that are all fixed for the life of the link:

```c
if (!mylink->phonemode && IS_NODE_EXTEN(mylink->name)
    && !CHAN_TECH(mylink->chan, "echolink") && !CHAN_TECH(mylink->chan, "tlb")) {
	return 0;
}
```

phonemode and name are set at link setup, the channel tech never changes, and IS_NODE_EXTEN is just name[0] > '0' && name[0] <= '9'. So for a plain node-to-node ilink, altlink() returns 0 for the entire life of the link — in both directions ([rpt_link.c:940](vscode-webview://0uku21eql3e9rhubt0b8tcqnq6plk9s16lj9fb43ve8iqr4a4aat/apps/app_rpt/rpt_link.c#L940) outbound, [app_rpt.c:7558](vscode-webview://0uku21eql3e9rhubt0b8tcqnq6plk9s16lj9fb43ve8iqr4a4aat/apps/app_rpt.c#L7558) inbound).

And [monchannel_read()](vscode-webview://0uku21eql3e9rhubt0b8tcqnq6plk9s16lj9fb43ve8iqr4a4aat/apps/app_rpt.c#L5219-L5228) is the only thing that ever writes into that hook, gated on altlink(myrpt, l). So on a normal ilink the whisper audiohook is attached, never fed, never mixed — it exists purely to drag a framehook, a timer, and a third fd onto l->pchan.

Skipping the attach when altlink is structurally impossible is a behavioral no-op that removes the framehook, the timer fd, the spin, and both leaks:


ast_audiohook_init(&l->altaudio, AST_AUDIOHOOK_TYPE_WHISPER, "Broadcast", 0);
if (link_may_altlink(l)) {   /* mirrors altlink()'s immutable early-out */
	ast_audiohook_attach(l->pchan, &l->altaudio);
}
Keep the init unconditional so the rest of the code is untouched: status stays NEW, so monchannel_read's status == AST_AUDIOHOOK_STATUS_RUNNING test skips it; the cleanup at [app_rpt.c:5170-5174](vscode-webview://0uku21eql3e9rhubt0b8tcqnq6plk9s16lj9fb43ve8iqr4a4aat/apps/app_rpt.c#L5170-L5174) is already guarded by ast_channel_audiohooks(l->pchan), which stays NULL; and ast_audiohook_destroy() still pairs with the init.

What's left exposed
Phone links, EchoLink, TLB, and IAXRPT/web-transceiver links (name starting with '0' isn't a node exten) legitimately need the hook, so they keep the framehook and keep leaking at ~50/sec when quiet. Those still require the Sangoma fix — which is the one to file:

audiohook.c:238-242 — ast_write() never takes ownership (if (f && f != fr) ast_frfree(f); in channel.c), so dup must be freed unconditionally, not only when write_res < 0.
audiohook.c:245 — returning &ast_null_frame drops the caller's frame; framehook_list_push_event() does not free the original, the callback must.
audiohook.c:145 — ast_timer_ack(timer, 1) clears one tick, and only when fdno matches; on DAHDI/pthread timing a single missed ack makes the fd permanently hot.
So: do the reorder and the conditional attach, and you're clean for the case you're reproducing.

jxmx:
Opus Response #2 when told to fix the issues:

app_rpt — branch fix/link-whisper-framehook-spin, commit 8e20045
Three changes, 69 insertions across [app_rpt.c](vscode-webview://0uku21eql3e9rhubt0b8tcqnq6plk9s16lj9fb43ve8iqr4a4aat/apps/app_rpt.c), [rpt_link.c](vscode-webview://0uku21eql3e9rhubt0b8tcqnq6plk9s16lj9fb43ve8iqr4a4aat/apps/app_rpt/rpt_link.c), [rpt_link.h](vscode-webview://0uku21eql3e9rhubt0b8tcqnq6plk9s16lj9fb43ve8iqr4a4aat/apps/app_rpt/rpt_link.h):

* app_rpt.c:4784-4800: chan now goes into cs[0] and l->pchan into cs[1], so pchan wins the tie in ast_waitfor_nandfds() and its timer fd actually gets acked. Comment explains the override-previous-winners mechanic so nobody "tidies" the order back.

* rpt_link.c:67-83]: new link_may_altlink() mirroring altlink()'s immutable early-out; both attach sites 
 
* rpt_link.c:958-967, app_rpt.c:7574-7584:  now gate ast_audiohook_attach() on it. ast_audiohook_init() stays unconditional — I verified it's what sets up both slinfactories and that ast_audiohook_destroy() is its exact inverse, so init-without-attach is balanced, status stays NEW so monchannel_read() skips it, and the cleanup path is already guarded by ast_channel_audiohooks(l->pchan).

* app_rpt.c:4801-4805: the who == NULL guard, for parity with rpt(). Hygiene, not part of the fix.
I confirmed l->chan, l->phonemode and l->name are all populated at app_rpt.c:7491-7497 well before the inbound attach, so the guard sees a complete link. Every source that includes the changed header compiles clean against the real Asterisk headers (mdc_decode.c, mdc_encode.c, pocsag.c fail, but they fail identically without my change — missing autoconfig network defines — and include neither header).

jxmx:
Fixes #1240

jxmx:
In investigating the upstream further, this is new behavior as of Asterisk 22.11. Also, https://github.com/AllStarLink/asl3-asterisk/pull/92 is needed to build until upstream is fixed.

jxmx:
https://github.com/asterisk/asterisk/pull/2171 also upstream

apps/app_rpt.c
```diff
@@ -4781,16 +4781,32 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
...
+		 * to its own slot. Poll l->pchan last so it wins the tie and the timer gets
+		 * acked; otherwise the timer fd stays readable, ast_waitfor_n() stops blocking
+		 * and this loop free-runs at 100% CPU.
+		 */
 		if (l->chan) {
 			cs[n++] = l->chan;
 		}
+		if (l->pchan) {
+			cs[n++] = l->pchan;
+		}
```
mkmer: This ordering should be the "fix" to our 100% cpu problem (right?)

apps/app_rpt/rpt_link.c
```diff
@@ -64,6 +64,23 @@ void set_linkmode(struct rpt_link *mylink, enum rpt_linkmode linkmode)
...
+	 * are fixed for the life of the link: phonemode and name are set when the link
+	 * is created and a channel never changes technology. A plain node to node link
+	 * therefore can never be an altlink, so it never needs the whisper audiohook.
+	 */
+	if (!mylink->phonemode && IS_NODE_EXTEN(mylink->name) && !CHAN_TECH(mylink->chan, "echolink") && !CHAN_TECH(mylink->chan, "tlb")) {
+		return 0;
+	}
+	return 1;
+}
+
```
mkmer, line 83: There are several alt link "reasons" - I don't think we've captured all of the "reasons" it will "never" altlink here have we?
https://github.com/AllStarLink/app_rpt/blob/7b134dcd2e36ec6b9371ca22b5604ef3b0bec726/apps/app_rpt/rpt_link.c#L67-L88
https://github.com/AllStarLink/app_rpt/blob/7b134dcd2e36ec6b9371ca22b5604ef3b0bec726/apps/app_rpt/rpt_link.c#L149-L184

Seems like we should create a helper for the "common" code between these 3 functions so that they "can't" disagree?

We can also set the link mode on the fly -> so as long as the first "stuff" is possible we could change modes after a call is established:
https://github.com/AllStarLink/app_rpt/blob/7b134dcd2e36ec6b9371ca22b5604ef3b0bec726/apps/app_rpt/rpt_link.c#L45-L65

mkmer (commented):
That is a nasty little nuanced problem...

apps/app_rpt/rpt_link.c
mkmer: On this one:
https://github.com/AllStarLink/app_rpt/blob/7b134dcd2e36ec6b9371ca22b5604ef3b0bec726/apps/app_rpt/rpt_link.c#L67-L88
Maybe we have captured them all -> might make sense to just move ln81 above line 78, then all of the "static" reasons exit first and we leave the dynamic reasons at the end...

mkmer: ON third thought, maybe just use `link_may_altlink()` in `altlink()` for the tests...

apps/app_rpt.c
mkmer: Ordering seems to make no difference to the problem here.

mkmer:
Adding a bit of testing docs:
This is a 22.11.x asterisk issue - 22.10.x appears to work fine.
 attaching the audiohook to `l->pchan` :2 side using `AST_AUDIOHOOK_WRITE` still exhibits the issue.
 attaching the audiohook to `l->chan' still exhibits the issue.

Moving forward, the "roll your own" mixer is the only path available.  It maybe a better path either way?

apps/app_rpt.c
mkmer: Turns out this order change does nothing for our problem...

apps/app_rpt.c
```diff
@@ -4781,16 +4811,32 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
...
+		 * Poll l->pchan last. ast_waitfor_nandfds() builds its pollfd array channel by
+		 * channel and lets later channels "override previous winners", so when both
+		 * channels are ready only the last one gets its fd recorded via
+		 * ast_channel_fdno_set(). A channel carrying an fd that only its own reader can
+		 * clear - a timer fd installed by a framehook, say - then has that event
+		 * swallowed whenever the other channel wins the tie, and because such fds are
+		 * level triggered ast_waitfor_n() stops blocking and this loop free-runs at
+		 * 100% CPU. Nothing puts such an fd on l->pchan today (see the altaudio comment
+		 * in struct rpt_link), and this ordering keeps it harmless if anything does.
+		 */
```
mkmer: I'm not sure this comment is correct in any way.  I did try just swapping these and it did not appear to change the behavior at all.
Since we removed the framehook, probably should just remove the comment and put it back as it's not doing anything for us?

apps/app_rpt.c
```diff
@@ -4781,16 +4811,32 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
...
+		}
 		if (!n) {
 			break;
 		}
 		who = ast_waitfor_n(cs, n, &ms);
+		if (!who) {
+			/* No winner: ast_waitfor_n() can return NULL without touching ms
+			 * (e.g. interrupted poll), so do not spin on a stale timeout. */
+			ms = 0;
+		}
```
mkmer: If we need this bit of code, we should add it over in the other ast_waitfor_n() loops as well.

mkmer (commented):
Just a couple of nits - it is working well on my nodes.
Also wonder if it's worth adding back the "can't altlink" logic and not start an slinfactory if a link can't be altlink?  I thought it was a nice bit of optimization.

apps/app_rpt.c
mkmer: After a bit of review of other core asterisk code, I don't think we need this.

apps/app_rpt.c
jxmx: Agreed.

jxmx:
> Just a couple of nits - it is working well on my nodes. Also wonder if it's worth adding back the "can't altlink" logic and not start an slinfactory if a link can't be altlink? I thought it was a nice bit of optimization.

I'll leave that up to you?

apps/app_rpt/app_rpt.h
```diff
@@ -3,6 +3,7 @@
 #define VERSION_PATCH 5
 
 #include "asterisk/audiohook.h"
```
Allan-N, line 5: Do we still need audiohook.h ?

apps/app_rpt.c
```diff
@@ -5168,10 +5206,6 @@ void process_link_channel(struct rpt *myrpt, struct rpt_link *l)
 	/* 1. Hang-up the channels */
```
Allan-N: Since we have removed `/* 2. Destroy audiohook resources */`, can we remove the `1. ` ?

Allan-N (approved):

apps/app_rpt.c
jxmx: Fixed.

apps/app_rpt/app_rpt.h
jxmx: Yes. Used in app_rpt.c and rpt_telemetry.c

